PR #1, opened 2022-10-19
Multiple Enhancements

marioevz:
### Test Functions as Generators
Each test function should now be modeled as a generator in order to yield multiple test vectors which result in multiple similar test cases with minimal changes.
I think this could reduce the amount of code necessary in the fillers when multiple test cases are very similar to each other and, instead of rewriting the same filler function again, one can simply change the `pre`, `post` or `txs` objects and yield a new test.

### Test Functions Now Take Fork as the Single Parameter
Some test cases require behavior changes depending on the fork they run. E.g. when a fork changes the opcode gas cost, the resulting balances might differ depending on the fork.
The test can now take a single parameter, which is the fork name, to modify its behavior accordingly.

Another potentially better option could be for the test functions to take `**kwargs`, and then read the fork from there.
This could make the test cases a bit more future-proof, and even the decorators could include more keyword parameters to be read by the test functions to modify its behavior.

### Yul Compiler
The `ethereum_test` package now includes a `Yul` class which is a compiler of Yul source code.
This results in the filler now requiring the `solc` binary to be installed to properly fill all the tests, which I think is an acceptable tradeoff to make the tests more readable.

### Alloc and Result are now checked
The returned `alloc` mapping returned by the `evm t8n` is now checked against the `post` object proposed by the test vector.
`result` mapping is also checked to verify that, if any transaction failed to execute, the failure is expected by the test vector.

### Expanded README
Added the section `Writing Tests` to the README document to include a small guide on how to write a test.
I think this section needs a lot of cleanup and further info, but this could be only a start.
Also, we could host a more detailed guide somewhere else if necessary.

### New Arguments for `ethereum_test_filler` 
Added new arguments `--test-module` and `--test-case` to limit test filling of a single test module or test case.

### ToDos (Can be separate PRs)

Add more internal tests for each of the newly introduced objects.

Wanted to add a proof of concept of a decorator that takes a test case and adds multiple variations of it.
E.g. an EOF (EVM Object Format) decorator that could generate both combinations of a test case with legacy contracts or EOF contracts.

Add info to the generated fixture such as the `evm` or `b11r` versions used to generate it.

Add automatic fixture generation to this repo.


src/ethereum_tests/example/example.py
```diff
@@ -0,0 +1,61 @@
...
+def test_yul(fork):
+    """
+    Test YUL compiled bytecode.
+    """
+    env = Environment()
+
+    pre = {
+        "0x1000000000000000000000000000000000000000": Account(
+            balance=0x0ba1a9ce0ba1a9ce,
+            code=Yul("""
```
lightclient, line 25: so cool

lightclient (approved):
Awesome work!

winsvega:
no coverage on dup tests ?

danceratopz:
> no coverage on dup tests ?

DUP tests were added in this first EEST PR #1 :) ...a long time before you added the automated coverage checker.

PR #2, opened 2022-10-28
Add Support for BlockchainTests

marioevz:
## Changes Included
- Adds support for multi-block tests, which are necessary to verify behavior that transcends a single block, such as basefee calculation, fork transitions, withdrawals, blobspace transaction fees calculation, etc.
- Adds requirement for `rlp` module import, because `t8n` tool currently returns transactions' RLP that includes rejected ones, and these should be removed before producing the final block.
- Introduces new types to better make a distinction between the types that are used to describe a test and the ones used to format the inputs to the transition or the block tool.

### ToDos
- [x] Requires https://github.com/ethereum/go-ethereum/pull/26051 to properly work, since this PR introduces `currentBasefee` as output of the execution result.
- [x] `mypy` is currently complaining about the usage of `kw_only` property of `dataclass`.
- [x] Use `kw_only` on all `dataclass` classes to force test specs to specify keywords when declaring test objects. The argument to this is that it allows the objects to be updated in the future and not having to refactor all tests due to positional argument conflicts.


src/ethereum_test/types.py
```diff
@@ -265,11 +286,23 @@ def __post_init__(self) -> None:
         if self.signature is None and self.secret_key is None:
             self.secret_key = TestPrivateKey
 
+        if self.ty is None:
+            # Try to deduce transaction type from included fields
+            if (
+                self.max_fee_per_gas is not None
```
lightclient, line 292: If only one is set, will the other be set automatically? I think yes?

src/ethereum_test/types.py
```diff
@@ -290,6 +323,49 @@ class Header:
     base_fee: Optional[int]
     hash: Optional[str] = None
 
+    @staticmethod
+    def from_dict(source: Dict[str, str]) -> "FixtureHeader":
+        """
+        Creates a FixedHeader object from a Dict.
+        """
+        return FixtureHeader(
```
lightclient, line 331: Is there a cleaner way of defining this? Sort of difficult to parse visually in this format.

src/ethereum_test/types.py
```diff
@@ -318,14 +394,137 @@ def to_geth_dict(self) -> Mapping[str, Any]:
...
+    """
+    exception: Optional[str] = None
+    """
+    If set, the block is expected to be rejected by the client.
+    """
+    txs: Optional[List[Transaction]] = None
+    """
+    List of transactions included in the block.
+    """
+    uncles: Optional[List[Header]] = None
```
lightclient, line 449: Pedantic, but would be great to switch to `ommers` if it doesn't complicate things too much.

src/ethereum_test/types.py
```diff
@@ -318,14 +394,137 @@ def to_geth_dict(self) -> Mapping[str, Any]:
...
+        """
+        These values are required, but they depend on the previous environment,
+        so they can be calculated here.
+        """
+        if self.number is not None:
+            new_env.number = self.number
+        else:
+            # calculate the next block number for the environment
+            if len(new_env.block_hashes) == 0:
+                new_env.number = 1
```
lightclient, line 488: This assumes the genesis block isn't in the env block hashes?

src/ethereum_test/types.py
```diff
@@ -398,22 +597,44 @@ def default(self, obj):
...
+            env = {
                 "currentCoinbase": obj.coinbase,
-                "currentDifficulty": hex(obj.difficulty),
-                "parentDifficulty": None,
                 "currentGasLimit": str(obj.gas_limit),
                 "currentNumber": str(obj.number),
                 "currentTimestamp": str(obj.timestamp),
-                "parentTimstamp": None,
-                "blockHashes": {},
+                "currentRandom": str(obj.prev_randao)
```
lightclient, line 605: Could these be replaced with a function like `valueOrNone(x)` ?

lightclient (commented):
Some small comments and nits, overall looks great!

src/ethereum_test/base_test.py
```diff
@@ -0,0 +1,124 @@
...
+                raise Exception("tx expected to fail succeeded")
+            elif not tx.error and error:
+                raise Exception(f"tx unexpectedly failed: {error}")
+
+            # TODO: Also we need a way to check we actually got the
+            # correct error
+    return list(rejected_txs.keys())
+
+
+def remove_transactions_from_rlp(txs: str, rejected_txs: List[int]) -> str:
```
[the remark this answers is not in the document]
marioevz: That's a good question. I checked the previous tests and there are indeed some that include invalid transactions, not within the block/rlp, but within a `transactionSequence` field.

For example this fixture contains some invalid transactions in some of the blocks:
https://github.com/ethereum/tests/blob/9d91961e98e97ba319e089f31388d4685da9b362/BlockchainTests/ValidBlocks/bcEIP1559/baseFee.json#L157

Which come from this source (base fee is invalid):
https://github.com/ethereum/tests/blob/9d91961e98e97ba319e089f31388d4685da9b362/src/BlockchainTestsFiller/ValidBlocks/bcEIP1559/baseFeeFiller.yml#L150

And in the filler they are marked as `expectException`.

However, I could not find the tag `transactionSequence` anywhere in geth's code, so I am not sure if we are actually testing anything by adding this field.

I think the correct approach might be, as you mention, to not remove these from the RLP and expect the entire block to fail.

src/ethereum_test/blockchain_test.py
```diff
@@ -0,0 +1,227 @@
...
+                    header.to_geth_dict(), txs, [], None
+                )
+
+                if block.exception is None:
+                    head = header.hash
+
+                    """
+                    Update environment for the following block
+                    """
+                    env.parent_base_fee = header.base_fee
```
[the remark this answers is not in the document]
marioevz: Added `apply_new_parent` function.

src/ethereum_test/types.py
marioevz: I think it will fail, I checked geth code and it requires both, since there is no way to deduce the other value if only one is set I think.

src/ethereum_test/types.py
marioevz: I cleaned up a bit, let me know if this looks better. Another option could be to add a method that iterates over the fields.

src/ethereum_test/types.py
marioevz: Done, at least only in the fields of the types defined in this package.
The json fields for geth could not be changed for example.

src/ethereum_test/types.py
marioevz: Fixed this error. Thanks!

src/ethereum_test/types.py
marioevz: Created a couple of helpers for this.

marioevz:
The only remaining issue seems to be whether to include or remove the invalid transactions from blocks.

I can follow up on this in a subsequent PR if that's ok.

src/ethereum_test/base_test.py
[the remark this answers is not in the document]
MariusVanDerWijden: Yeah I think it makes sense to not remove invalid transactions, and just let the block fail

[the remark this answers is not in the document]
marioevz: I submitted #3 to tackle this issue separately, which also includes the possible solutions for a problem that involves the current `StateTest` mode of operation.

lightclient:
I might be missing something, but it feels like this PR is making an active choice to remove invalid transactions from the RLP and so we should remove that logic from this PR and decide the best way to deal with invalid txs separately. Is there a reason that work is necessary for this specific PR?

src/ethereum_test/types.py
```diff
@@ -257,24 +342,58 @@ def __post_init__(self) -> None:
...
+@dataclass(kw_only=True)
+class FixtureHeader:
+    """
+    Representation of an Ethereum header within a test Fixture.
     """
 
+    coinbase: str
     parent_hash: str
     ommers_hash: str
-    coinbase: str
```
lightclient, line 277: What is the rationale for moving this? It's kind of nice that it is currently in the RLP list order.

src/ethereum_test/types.py
lightclient: this is much more readable, thx!

marioevz:
> I might be missing something, but it feels like this PR is making an active choice to remove invalid transactions from the RLP and so we should remove that logic from this PR and decide the best way to deal with invalid txs separately. Is there a reason that work is necessary for this specific PR?

No it was just simply to keep adding more and more modifications to the PR, but yes now that you mention it, we should not be adding a new import that we are supposed to remove in another PR. Let me add the changes in this PR :+1:

src/ethereum_test/types.py
marioevz: This was a mistake, I moved temporarily for mental order while I was refactoring but forgot to put it back. Changing it bacl now.

src/ethereum_test/blockchain_test.py
```diff
@@ -22,6 +22,11 @@
     JSONEncoder,
 )
 
+default_base_fee = 1
```
lightclient, line 25: This should be 7 because it is impossible for the base fee to go below that value.

lightclient (approved):

lightclient:
Looks good to merge once CI goes green!

PR #7, opened 2022-11-08
Type Fixes

marioevz:
## Changes Included
- ~Adds `UNSPECIFIED` type in order to better discern between an unspecified test value and when the test writer explicitly wants to verify against `None`~
- Adds `NONEXISTENT` as a sentinel value of class `Account` that can be used to specify when an account must not exist in the resulting `alloc`
- `None` is now interpreted as a parameter that does not need to be verified.
- `Union` typing is removed from almost everywhere in favor of the python 3.10+ nomenclature `|`.
  Only place where it couldn't be removed is on the `add` and `radd` methods of `Code` class (pytest fails when `Union[str, bytes, "Code"]` is replaced by `str | bytes | "Code"` because of the quoted self-class).
- Tests within fixture outputs are now enumerated in the format `{index}[_{test_name}]_{fork}` instead of only `{fork}`, where `{test_name}` is an optional name that can be set by the test spec writer.
- Adds helper `to_hash`
- Adds more unit tests for types
- Adds the post state to the output fixture, so in case there is `stateRoot` mismatch it will be easier to debug.


src/ethereum_test/types.py
```diff
@@ -13,7 +13,33 @@
 from .common import AddrAA, TestPrivateKey
 
 
-def hex_or_none(input: Union[int, None], default=None) -> Union[str, None]:
+class UNSPECIFIED(object):
```
lightclient, line 16: Could you expand on why this class is necessary? I _think_ we should be able to generally work around this by having different methods for creating objects that set the defaults to non-`None` values and if a test writer desires `None` actually being specified they can use a more general initializer that accepts `None`?

lightclient: Also, this may just be due to my ignorance, but I would be curious to know if this a pattern you've seen before. I'm not as familiar with larger-scale python projects, maybe this is common.

src/ethereum_test/types.py
```diff
@@ -13,7 +13,33 @@
...
+    """
+
+    def __new__(cls):
+        """Create the singleton UNSPECIFIED object"""
+        if not hasattr(cls, "instance"):
+            cls.instance = super(UNSPECIFIED, cls).__new__(cls)
+        return cls.instance
+
+
+def hex_or_none_unspecified(
```
lightclient, line 31: Couldn't we just expand `hex_or_none` with an "unspecified" default?

lightclient (commented):
Overall looks good - just a couple questions about `UNSPECIFIED`.

src/ethereum_test/types.py
marioevz: I actually copied the idea from the `dataclasses` library `MISSING` sentinel object over here: https://docs.python.org/3/library/dataclasses.html#dataclasses.field.

I think it's definitely possible to skip this pattern completely, so I think it's just a matter of deciding how we want the test definitions to look like, and here are some examples of how a tester could specify an Account's code in post and specifically define if they want it verified or not:

Without `UNSPECIFIED`:
```
Account()
- code is not checked

Account(code=None)
- code is not checked

Account(code="0x")
- code is expected to be empty
```
With `UNSPECIFIED`:
```
Account()
- code is not checked

Account(code=None)
- code is expected to be empty

Account(code="0x")
- code is expected to be empty
```
 Where `code is not checked` means that the `tf` tool will not check what was received in the transition tool and simply pass it along onto the fixture.

src/ethereum_test/types.py
marioevz: Indeed, check out the proposed solution 😉 (spoiler alert: it uses `UNSPECIFIED`)

README.md
```diff
@@ -213,13 +207,70 @@ Currently supported built-in compilable objects are:
 
 `Code` objects can be concatenated together by using the `+` operator.
 
+### Verifying the accounts' Post State
```
lightclient, line 210:
```suggestion
### Verifying the Accounts' Post State
```

README.md
```diff
@@ -213,13 +207,70 @@ Currently supported built-in compilable objects are:
...
+  created is indeed empty.
+
+### The `Account` object
+
+The `Account` object is used to specify the properties of an account to be
+verified in the post state.
+
+It has the following Python representation.
+```
+class Account:
```
lightclient, line 239: I wish there were a nice way to provide a ref in markdown to a code object and have it fetch it so that you don't need to worry about keeping the two in sync.

src/ethereum_test/base_test.py
```diff
@@ -59,16 +57,15 @@ def verify_post_alloc(post: Mapping[str, Account], alloc: Mapping[str, Any]):
...
-            # If an account is None in post, it must not exist in the
-            # alloc.
-            if address in alloc:
-                raise Exception(f"found unexpected account: {address}")
-        else:
-            if address in alloc:
-                account.check_alloc(address, alloc[address])
+        if account is not None:
+            if account == Account.NONEXISTENT:
+                if address in alloc:
```
lightclient, line 62: Minor nit: I think `post` and `alloc` could be more descriptive. I'm not actually sure which value is the expected and which was computed. I _think_ `post` is expected from test and `alloc` is the computed value? In geth we often use `got`/`want` or `got`/`exp`.

src/ethereum_test/fill.py
```diff
@@ -46,14 +46,20 @@ def fill_test(
                 head=head,
                 fork=fork,
                 pre_state=test.pre,
-                post_state=None,
+                post_state=alloc,
                 seal_engine=engine,
+                name=test.name,
+                index=index + 1,
```
lightclient, line 52: Is this to not 0 index them?

src/ethereum_test/types.py
```diff
@@ -181,19 +186,69 @@ def must_contain(self, other: "Storage"):
...
     """
 
-    nonce: Optional[int] = None
-    balance: Optional[int] = None
-    code: Optional[Union[bytes, str, Code]] = None
-    storage: Optional[
-        Union[Storage, Dict[Union[str, int], Union[str, int]]]
-    ] = None
+    nonce: int | None = None
+    """
```
lightclient, line 229: Are python comments really below the actual code definition?

src/ethereum_test/types.py
lightclient: Thanks for all this extra context!! I do like the new `NONEXISTENT` approach.

README.md
marioevz: Yeah, I think it might be better to remove it altogether and point to the file containing the definition.

src/ethereum_test/base_test.py
marioevz: Fixed!

src/ethereum_test/fill.py
marioevz: Yeah, I don't know why I did that, but I set it to zero now :+1:

src/ethereum_test/types.py
marioevz: Yes! check it out: https://peps.python.org/pep-0224/

lightclient (approved):
Sorry for all the back and forth. PR looks great; thanks for all the hard work!

PR #8, opened 2022-11-09
Fix `--evm-bin` option

petertdavies:


src/ethereum_test_filler/main.py
```diff
@@ -37,6 +38,7 @@ def parse_arguments() -> argparse.Namespace:
             help="path to evm executable that provides `t8n` and `b11r` "
             + "subcommands",
             default="evm",
```
marioevz, line 40:
```suggestion
            default=None,
```
Default should be `None`, otherwise `which(evm)` never kicks into action when needed.

marioevz (approved):

PR #9, opened 2022-11-14
Bump golang to 1.18

lightclient:
Geth bumped the minimum golang version to 1.18 [a few days ago](https://github.com/ethereum/go-ethereum/pull/26160). Since we're building directly from master we need to do the same.


lightclient:
Going to go ahead and just merge since CI is broken on main.

PR #10, opened 2022-11-14
Folder Structure Refactor

marioevz:
## Changes Included
- Refactors the project's folder structure with the following changes:
  - `/src` now only contains source code pertaining to the actual execution specs library packages and no fillers
  - Test fillers are now located at ~`/execution_tests_fillers`~ `/fillers`
  - Modules' unit tests are now included within folders called `tests` inside of each package.
- Rewrite of the module-importing routine responsible for searching the test fillers
- Fix decorators to use generic `TestSpec` type instead of `StateTestSpec`


marioevz:
@lightclient do you like more `execution_tests_fillers` or `execution_test_fillers` for the name of the folder containing the fillers?
I included `execution_tests_fillers` in this PR but now I'm leaning more towards `execution_test_fillers`.
Other suggestions are welcome of course (maybe just `fillers`?).

lightclient (commented):
I would say at minimum I would like to remove the `_library` suffix on `execution_tests_library`. 

Some additional thoughts on the matter:

Naming is hard. I'm ambivalent on the right approach. One idea is to namespace these packages under something like `ethereum_test_tools` and then have the actual package names super short like `filler` and `tests`. Then expose the `tests` subpackage under the root `ethereum_test_tools`.

I think this kinda looks nice.
```python
from ethereum_test_tools import StateTest
```

`ethereum` vs. `execution` isn't a huge deal, but I do like that `ethereum` is much more descriptive than `execution`. This is all mostly bikeshedding. I don't think we should get hung up too much beyond what I mentioned above. I just think we should contemplate how to keep package names i) short and ii) descriptive.

edit: not sure if we could release under the `ethereum` package somehow eventually, but this also looks nice:

```python
from ethereum.testtools import StateTest
```

marioevz:
@lightclient I like the `ethereum.testtools` too, I think this would require coordination to check if there is a package name collision?

For the time being, I renamed `execution_tests_library` to `ethereum_test_tools`.

lightclient (approved):

PR #12, opened 2022-12-01
Transition Tool EIPs Support, Print Traces, Helper Methods

marioevz:
## Changes Included
- Adds EIPs support for a test to specify extra EIPs used to call the transition tool
- When a test fails during the filling phase, now the full traces are printed
- Include a gas measuring helper, to easily create tests that expect a certain amount of gas usage on a code excerpt
- Add custom exception types, to better catch specific types of errors

### DIFF
--- a/src/ethereum_test_tools/common/helpers.py
+++ b/src/ethereum_test_tools/common/helpers.py
@@ -1,6 +1,10 @@
 """
-Helper functions for generating Ethereum tests.
+Helper functions/classes used to generate Ethereum tests.
 """
+
+from dataclasses import dataclass
+
+from ..code import Code, code_to_bytes
 
 
 def to_address(input: int | str) -> str:
@@ -25,3 +29,71 @@ def to_hash(input: int | str) -> str:
     if type(input) is int:
         return "0x" + input.to_bytes(32, "big").hex()
     raise Exception("invalid type to convert to hash")
+
+
+@dataclass(kw_only=True)
+class CodeGasMeasure(Code):
+    """
+    Helper class used to generate bytecode that measures gas usage of a
+    bytecode, taking into account and subtracting any extra overhead gas costs
+    required to execute.
+    """
+
+    code: bytes | str | Code
+    """
+    Bytecode to be executed to measure the gas usage.
+    """
+    overhead_cost: int
+    """
+    Extra gas cost to be subtracted from extra operations.
+    """
+    extra_stack_items: int = 0
+    """
+    Extra stack items that remain at the end of the execution.
+    To be considered when subtracting the value of the previous GAS operation,
+    and to be popped at the end of the execution.
+    """
+    sstore_key: int = 0
+    """
+    Storage key to save the gas used.
+    """
+
+    def assemble(self) -> bytes:
+        res = bytes()
+        res += bytes(
+            [
+                0x5A,  # GAS
+            ]
+        )
+        res += code_to_bytes(self.code)  # Execute code to measure its gas cost
+        res += bytes(
+            [
+                0x5A,  # GAS
+            ]
+        )
+        # We need to swap and pop for each extra stack item that remained from
+        # the execution of the code
+        res += (
+            bytes(
+                [
+                    0x90,  # SWAP1
+                    0x50,  # POP
+                ]
+            )
+            * self.extra_stack_items
+        )
+        res += bytes(
+            [
+                0x90,  # SWAP1
+                0x03,  # SUB
+                0x60,  # PUSH1
+                self.overhead_cost + 2,  # Overhead cost + GAS opcode price
+                0x90,  # SWAP1
+                0x03,  # SUB
+                0x60,  # PUSH1
+                self.sstore_key,  # -> SSTORE key
+                0x55,  # SSTORE
+                0x00,  # STOP
+            ]
+        )
+        return res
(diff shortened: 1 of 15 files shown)


src/ethereum_test_tools/common/helpers.py
```diff
@@ -25,3 +29,74 @@ def to_hash(input: int | str) -> str:
     if type(input) is int:
         return "0x" + input.to_bytes(32, "big").hex()
     raise Exception("invalid type to convert to hash")
+
+
+@dataclass(kw_only=True)
+class CodeGasMeasure(Code):
```
lightclient, line 35: Would you mind expanding upon the motivation for this vs. using `gasUsed` in a tx receipt?

lightclient: I'm also curious how the caller will use this. For instance: does the caller need to manually calculate the overhead costs? Or determine if their code leaves the stack dirty?

src/ethereum_test_tools/common/types.py
```diff
@@ -72,6 +72,62 @@ class Storage:
...
+        def __str__(self):
+            """Print exception string"""
+            return f"invalid type for key/value: {self.v}"
+
+    class MissingKey(Exception):
+        """
+        Test expected to find a storage key set but key was missing.
+        """
+
+        k: int
```
lightclient, line 95:
```suggestion
        key: int
```
seems short enough and improve readability

src/ethereum_test_tools/common/types.py
```diff
@@ -72,6 +72,62 @@ class Storage:
...
+                Storage.key_value_to_string(self.k)
+            )
+
+    class KeyValueMismatch(Exception):
+        """
+        Test expected a certain value in a storage key but value found
+        was different.
+        """
+
+        k: int
```
lightclient, line 113:
```suggestion
        key: int
```
same here

src/ethereum_test_tools/common/types.py
```diff
@@ -250,6 +282,81 @@ class Account:
...
+                + f"want {self.want}, got {self.got}"
+            )
+
+    class CodeMismatch(Exception):
+        """
+        Test expected a certain bytecode for an account but a different
+        one was found.
+        """
+
+        account: str
```
lightclient, line 341:
```suggestion
        addr: str
```
or `address`

src/ethereum_test_tools/common/types.py
```diff
@@ -250,6 +282,81 @@ class Account:
...
+                + f"want {self.want}, got {self.got}"
+            )
+
+    class BalanceMismatch(Exception):
+        """
+        Test expected a certain balance for an account but a different
+        value was found.
+        """
+
+        account: str
```
lightclient, line 316:
```suggestion
        addr: str
```

src/ethereum_test_tools/common/types.py
```diff
@@ -250,6 +282,81 @@ class Account:
     state.
     """
 
+    class NonceMismatch(Exception):
+        """
+        Test expected a certain nonce value for an account but a different
+        value was found.
+        """
+
+        account: str
```
lightclient, line 291:
```suggestion
        addr: str
```

lightclient (commented):
Lots of good stuff here, mainly just a few nits. Also curious if you know how much recording traces affects the performance of filler a large amount of tests? Would it be worth having a `--no-traces` flag or something?

src/ethereum_test_tools/common/helpers.py
marioevz: Yes, right now it is expected that the user provides this analysis manually. Though I would really like for this to be automatic, but I'm not sure if it's possible without having a list of all opcodes and their stack modifications.

This is an example of how I am using it right now in the Warm COINBASE tests:
```
"BALANCE": CodeGasMeasure(
            code=bytes(
                [
                    0x41,  # addr: COINBASE
                    0x31,  # BALANCE
                ]
            ),
            overhead_cost=2,
            extra_stack_items=1,
        ),
```

Here, `CodeGasMeasure` is meant to automate the computation of the gas cost of the `BALANCE` opcode.
`overhead_cost` is the gas cost of the `COINBASE` opcode, which we need to subtract to obtain the expected result in the storage.

`extra_stack_items` is the dangling stack item left by the `BALANCE` opcode, which we need to swap+pop before we do the gas measurement, otherwise the subtract operation ends up subtracting the first gas measurement minus the dangling stack element.

marioevz:
> Lots of good stuff here, mainly just a few nits. Also curious if you know how much recording traces affects the performance of filler a large amount of tests? Would it be worth having a `--no-traces` flag or something?

Yes I think performance will be affected, maybe it could be the other way around and require the user to do `--traces` to fetch the information. And then, only if there's an error, and the flag was not used, print a warning in the output to suggest the user to add the flag to get the traces.

marioevz:
Hey @lightclient, changes are now implemented, and I liked how the `TransitionTool` class is the one that stores the traces only if configured to do so, then the specs retrieve the traces in case of error.
Let me know what you think.

src/ethereum_test_tools/common/helpers.py
lightclient: Okay I understand, thank you.

lightclient (approved):
LGTM - thx Mario

PR #28, opened 2023-01-11
fillers/eips/eip3860: Update EIP-3860 Due to Small Spec Change

spencer-tb:
Currently if there is an initcode length violation for 'CREATE'/'CREATE2' the gas for the initcode execution is not deducted but stored. In the new spec change here -> [link](https://github.com/ethereum/EIPs/pull/6249) the evm exceptionally aborts as if it runs out of gas. A small fix is added to achieve the latter.

Another small change is added to match the salt addition here -> ([link](https://github.com/ethereum/execution-spec-tests/pull/20/commits/8cd2b0356a6959881c1a433b8fc0fb86152c1efa))

### DIFF
--- a/fillers/eips/eip3860.py
+++ b/fillers/eips/eip3860.py
@@ -460,8 +460,7 @@ def generate_create_opcode_initcode_test_cases(
     call_code = Yul(
         """
         {
-            let initdata := calldataload(0)
-            mstore(0, initdata)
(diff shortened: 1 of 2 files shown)


spencer-tb:
Still to add extra checks to catch OOG error.

spencer-tb:
I have made a mess.

PR #29, opened 2023-01-13
fillers/eips/eip3860: Update EIP-3860 Due to Spec Change

spencer-tb:
Currently if there is an initcode length violation for 'CREATE'/'CREATE2' the gas for the initcode execution is not deducted but stored. In the new spec change here -> https://github.com/ethereum/EIPs/pull/6249 the evm exceptionally aborts as if it runs out of gas. A small fix is added to achieve the latter.

### DIFF
--- a/fillers/eips/eip3860.py
+++ b/fillers/eips/eip3860.py
@@ -457,6 +457,19 @@ def generate_create_opcode_initcode_test_cases(
     """
     env = Environment()
 
+    call_code = Yul(
+        """
+        {
+            let initdata := calldataload(0)
+            mstore(0, initdata)
+            let call_result := call(10000000, \
+                0x0000000000000000000000000000000000000100, \
+                0, 0, calldatasize(), 0, 0)
+            sstore(0, call_result)
+        }
+        """
+    )
+
     if opcode == "create":
         code = Yul(
             """
@@ -499,6 +512,7 @@ def generate_create_opcode_initcode_test_cases(
             salt=0xDEADBEEF,
             initcode=initcode.assemble(),
         )
+
     else:
         raise Exception("invalid opcode for generator")
 
@@ -508,13 +522,17 @@ def generate_create_opcode_initcode_test_cases(
             code=code,
             nonce=1,
         ),
+        to_address(0x200): Account(
+            code=call_code,
+            nonce=1,
+        ),
     }
 
     post: Dict[Any, Any] = {}
 
     tx = Transaction(
         nonce=0,
-        to=to_address(0x100),
+        to=to_address(0x200),
         data=initcode,
         gas_limit=10000000,
         gas_price=10,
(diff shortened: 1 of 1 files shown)


fillers/eips/eip3860.py
```diff
@@ -457,6 +457,18 @@ def generate_create_opcode_initcode_test_cases(
...
     env = Environment()
 
+    call_code = Yul(
+        """
+        {
+            calldatacopy(0, 0, calldatasize())
+            let call_result := call(10000000, \
+                0x0000000000000000000000000000000000000100, \
+                0, 0, calldatasize(), 0, 0)
+            sstore(0, call_result)
```
marioevz, line 467: Could we do `sstore(call_result, 1)` to store a value regardless of the output here?

We would then need to check `s[0]==1, s[1]==0` for failing test cases, and `s[0]==0, s[1]==1` for passing ones.

We should also perform the call with a value less than `10000000` (e.g. `5000000`), because `10000000` is the total gas of the transaction, and luckily we are only able to keep executing because the actual gas sent follows this formula:
```
min(call_gas_value, remaining_gas * 63 / 64)
```
And since we are OOG in the call, we are wasting the entire `remaining_gas * 63 / 64`, so we are left with very little gas after `CALL`, and `SSTORE` consumes `20000` so we are safe, but if this changes in the future for any reason, the transaction could fail unnecessarily.

marioevz (changes requested):

fillers/eips/eip3860.py
```diff
@@ -457,6 +457,19 @@ def generate_create_opcode_initcode_test_cases(
...
 
+    call_code = Yul(
+        """
+        {
+            calldatacopy(0, 0, calldatasize())
+            let call_result := call(5000000, \
+                0x0000000000000000000000000000000000000100, \
+                0, 0, calldatasize(), 0, 0)
+            sstore(0, call_result)
+            sstore(call_result, 1)
```
marioevz, line 468:
```suggestion
            sstore(call_result, 1)
```
Only line 468 should remain.

fillers/eips/eip3860.py
```diff
@@ -525,14 +543,23 @@ def generate_create_opcode_initcode_test_cases(
         expected_gas_usage += PUSH_DUP_OPCODE_GAS
 
     if len(initcode.assemble()) > MAX_INITCODE_SIZE and eip_3860_active:
+        # Call returns 0 as out of gas s[0]==1
+        post[to_address(0x200)] = Account(
+            nonce=1,
+            storage={
+                0: 1,
```
marioevz, line 550:
```suggestion
                0: 1,
                1: 0,
```
Also check for `s[1]`

fillers/eips/eip3860.py
```diff
@@ -552,6 +579,15 @@ def generate_create_opcode_initcode_test_cases(
                 len(initcode.assemble())
             )
 
+        # Call returns 1 as valid initcode length s[0]==1 && s[1]==1
+        post[to_address(0x200)] = Account(
+            nonce=1,
+            storage={
+                0: 1,
+                1: 1,
```
marioevz, line 587:
```suggestion
                0: 0,
                1: 1,
```
Change due to the first suggestion.

marioevz (approved):

Issue #30, opened 2023-01-13
Update withdrawals fillers to use gwei amounts instead of wei


marioevz:
PR https://github.com/ethereum/EIPs/pull/6325 updates the amounts in a withdrawal to match the units used by the CL, gwei.

This change must be accounted for in the balance increments for all tests in `fillers/withdrawals/withdrawals.py`.

PR #32, opened 2023-01-13
fillers/withdrawals/withdrawals: Update to use Gwei instead of Wei.

spencer-tb:
PR for #30.

### DIFF
--- a/fillers/withdrawals/withdrawals.py
+++ b/fillers/withdrawals/withdrawals.py
@@ -564,6 +564,10 @@ def test_withdrawals_zero_amount(_):
             code="0x00",
             balance=0,
         ),
+        to_address(0x300): Account(
+            code="0x00",
+            balance=0,
+        ),
     }
 
     withdrawal_1 = Withdrawal(
@@ -586,6 +590,10 @@ def test_withdrawals_zero_amount(_):
             code="0x00",
             balance=0,
         ),
+        to_address(0x300): Account(
+            code="0x00",
+            balance=0,
+        ),
     }
 
     yield BlockchainTest(pre=pre, post=post, blocks=[block])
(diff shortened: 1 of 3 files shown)


fillers/withdrawals/withdrawals.py
```diff
@@ -77,7 +79,7 @@ def test_withdrawals_use_value_in_tx(_):
         index=0,
         validator=0,
         address=TestAddress,
-        amount=tx.gas_price * tx.gas_limit + 1,
+        amount=(tx.gas_price * tx.gas_limit + ONE_GWEI),
```
marioevz, line 82: There is a problem when specifying `(tx.gas_price * tx.gas_limit)` here, because their unit is wei, but `withdrawal.amount` expects gwei, and this multiplication actually translates to `(tx.gas_price * tx.gas_limit * 1e9)` in post's balance, so the test is failing.

One trick could be to set the gas_price to `ONE_GWEI` and remove `tx.gas_price` from this formula.

Another thing to consider is that, when we use `ONE_GWEI` in the withdrawal amount, it translates to `ONE_GWEI * ONE_GWEI` in the post balance, so we need to use a `1` here and check for `ONE_GWEI` in post.

marioevz (changes requested):

marioevz:
I think some of the tests are failing because of the issue described in my review.

The up-to-date go-ethereum branch with gwei withdrawals is https://github.com/lightclient/go-ethereum/tree/withdrawals-timestamp, so I think we can use this branch to double check that the changes are correct.

fillers/withdrawals/withdrawals.py
spencer-tb: Apologies for this, a simple error on my part, after fixing the errors I realised I misunderstood the conversions. I've updated the PR with the commits below

fillers/withdrawals/withdrawals.py
```diff
@@ -597,7 +599,7 @@ def test_withdrawals_zero_amount(_):
         amount=1,
     )
     block.withdrawals.append(withdrawal_2)
-    post[to_address(0x200)].balance = 1
+    post[to_address(0x200)].balance = ONE_GWEI
     yield BlockchainTest(pre=pre, post=post, blocks=[block])
```
marioevz, line 603: Could we add a new test case that uses the maximum value for the amount field in the withdrawal? I.e. use 2**64 -1 in amount. It should pass without issues.

marioevz (commented):
Two minor comments, but overall looks very good! :D

fillers/withdrawals/withdrawals.py
spencer-tb: Done! I've added to the commit below

setup.cfg
```diff
@@ -72,6 +72,7 @@ extend-ignore =
     D414    # Ignore: Section has no content
     D415    # Ignore: First line should end with a period, question mark, or exclamation point
     D416    # Ignore: Section name should end with a colon
+    SC200   # Ignore: Spelling error in name (e.g. variable, function, class)
```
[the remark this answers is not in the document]
spencer-tb: Ahh perfect! I will do that :)

marioevz (approved):

PR #33, opened 2023-01-13
fillers/vm/vm_arithmetic: Port vmArithmetic fillers

spencer-tb:
PR for the ported test of https://github.com/ethereum/tests/tree/develop/src/GeneralStateTestsFiller/VMTests/vmArithmeticTest


spencer-tb:
Currently the test filler: `tf` does not fill tests within sub-directories, as the `iter_modules` function only searches 1 depth from the given root. For example if I have a test `add.py` within the sub-directory `vm/vm_arithmetic/vm_add` it will not be found. 

The above 2 commits have updated `tf` to account for tests within sub-directories. Instead of using `iter_modules` solely a new recursive function was created: `recursive_iter_modules`. This continuously calls `iter_modules` if a module is not found, appending the sub-directory onto the root each time.

fillers/vm/vm_arithmetic/add.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    total_tests = 5
+    solutions = {
+        to_address(0x100): to_hash(-2),
+        to_address(0x101): 0x03,
+        to_address(0x102): 0x00,
+        to_address(0x103): 0x00,
+        to_address(0x104): 0x00,
+    }
```
marioevz, line 77: I think using multiple accounts accounts here is redundant if we are already using a switch statement in the code itself.

We can either use multiple accounts and different code for each without a switch, or a single account (perhaps only `to_address(0x100)` and the switch statement, but both is redundant and a bit confusing.

fillers/vm/vm_arithmetic/arith.py
```diff
@@ -0,0 +1,86 @@
...
+        # noqa: 24 PUSH1 9      6009                      9,0,4
+        # noqa: 26 PUSH1 0x11   6011                      17,9,0,4
+        # noqa: 28 EXP          0A                        17^9,0,4
+        # noqa: 29 PUSH1 0      6000                      0,17^9,0,4
+        # noqa: 2B SSTORE       55 The original was MSTORE, but that's not testable
+        # noqa: 2C PUSH1 8      6008                      8,0,4
+        # noqa: 2E PUSH1 0      6000                      0,8,0,4
+        # noqa: 30 RETURN       F3
+        "0x60016001900160070260050160029004600490602190056017016\
+            0030260059007600303600960110A60005560086000F3"
```
marioevz, line 65: I think we can represent this code by using the Opcodes enum list, like some of the examples in this file:
https://github.com/ethereum/execution-spec-tests/blob/4dd97c93515857a52f6d0876f9cad65d749f6563/fillers/eof1/code_validation_jump.py#L31

We could keep the stack as a comments though (last column), but however you prefer is ok.

fillers/vm/vm_arithmetic/add.py
spencer-tb: Thanks for pointing this out, I'll go for multiple accounts!

fillers/vm/vm_arithmetic/arith.py
spencer-tb: Sounds good to me!

spencer-tb:
Deleting PR as its a bit of a mess. Fresh PR created.

PR #34, opened 2023-01-19
fillers/example: Remove unnecessary Yul output bytecode verification

marioevz:
Removes `post[0x1000000000000000000000000000000000000000].code` verification, because the output from `solc` can change from version to version.


spencer-tb (commented):
LGTM!

spencer-tb (approved):

PR #35, opened 2023-01-19
spec,filler: Fix `difficulty>0` on post merge fixtures, add `genesisRLP`

### LINKED ISSUE #27: Release v0.1.0 chain format is wrong
Issue 1: 
- test format missing `genesisRLP` field

Issue 2:
- genesis block has difficulty != 0, but also has withdrawalsRoot defined which makes it Merge. Merge block has difficulty = 0.

marioevz:
Fixes #27


spencer-tb (approved):
LGTM

lightclient:
FYI @marioevz @spencer-tb, when I was looking into this the other day I found something kind of weird. I think there is an issue when building block tests. In `set_environment` I _think_ it takes the difficulty from the block and writes it into the env difficulty, when sometimes it should write it into prevRandao. But maybe it is worked out somewhere else, because I think the post merge tests would be failing to fill if they didn't supply prevRandao. Anyways, just wanted to mention.

Issue #36, opened 2023-01-20
RLP of Shanghai blocks should always have withdrawals list


yperbasis:
Even when they don't contain withdrawals, RLP of such blocks should contain an empty list for withdrawals post-Shanghai. For example, block 2 in `withdrawals_use_value_in_contract.json` in Release [v0.2.1](https://github.com/ethereum/execution-spec-tests/releases/tag/v0.2.1) fixtures has RLP 
```0xf90282f9021aa08444423fbde249cfcd3c41bdf009368a8b77763ed9fe344db4a32b0ecdc8d7b3a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347942adc25665018aa1fe0e6bc666dac8fc2697ff9baa06ab526318fc1842c06d5440bf245b03112e15241ea5d1f9f013ff6cf81296376a08c500edaa09a8dce41c0dcce8ee2876f5f3150b8101d09faa9ba61db523507eaa0cda4364d9e43f15119746d96fb9d842997acbfa55f65b6d3f7a3b5f06cd250d1b9010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800288016345785d8a000083012e731880a0000000000000000000000000000000000000000000000000000000000000000088000000000000000007a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421f862f860010a830186a0940000000000000000000000000000000000000100808026a025b1502ca1862804cfb6dca49ffb3cca3d8225a66df185ca857aab6f8a1f97a3a05d9178ca2d7a23b22e07ff2fe9a01ceb23a4acf041a0fcdf3d31752de802b74ec0```
but it should end with `c0c0` (one empty list for uncles and one empty list for withdrawals).

Also blocks in `initcode_limit_contract_creating_tx_gas_usage.json` suffer from this issue.

spencer-tb:
https://github.com/ethereum/execution-spec-tests/pull/37 is a WIP but should resolve the issue. Thanks for the find :happy:

gurukamath:
[...]

PR #37, opened 2023-01-20
Fix rlp for blocks with no withdrawals.

spencer-tb:
Regarding https://github.com/ethereum/execution-spec-tests/issues/36

It looks like before RLP serialization for blocks:

image

the `Block` class (block) doesn't set withdrawals to the empty list `[ ]` for Shanghai, and instead this remains as a None type. This 

This PR contains one solution to the latter. As withdrawals are within the Environment class, and it is updated according to the fork type before RLP serialization:

image

we can use it instead, as the withdrawals None type here is set to the empty list `[ ]` within `set_fork_requirements`.

### DIFF
--- a/src/ethereum_test_tools/spec/blockchain_test.py
+++ b/src/ethereum_test_tools/spec/blockchain_test.py
@@ -184,7 +184,7 @@ class BlockchainTest(BaseTest):
                 header=header.to_geth_dict(),
                 txs=txs_rlp,
                 ommers=[],
-                withdrawals=to_json_or_none(block.withdrawals),
+                withdrawals=to_json_or_none(env.withdrawals),
             )
 
             if block.exception is None:


spencer-tb:
@marioevz 

**Some other alternative options:**
1) To keep the withdrawals field consistent between the Environment class and the Block class we could add `block` as another input within `set_fork_requirements(env, block, fork)`, and update the withdrawals for block to `[ ]` if it has a none type for Shanghai alongside `env`
2) Achieve a similar outcome with a new function called `set_block_requirements(block, fork)`.
3) Make the default type for withdrawals in the Block class the empty list as opposed to the None type.

src/ethereum_test_tools/spec/blockchain_test.py
```diff
@@ -184,7 +184,7 @@ def make_block(
                 header=header.to_geth_dict(),
                 txs=txs_rlp,
                 ommers=[],
-                withdrawals=to_json_or_none(block.withdrawals),
+                withdrawals=to_json_or_none(env.withdrawals),
```
marioevz, line 187: I think this is good, but I think we should also copy the withdrawals from the block into the environment on `block.set_environment` (inside the `set_environment` method), because I think if the block indeed had withdrawals, they would be disregarded at this point.

marioevz (changes requested):

marioevz:
A couple of things:
- Looking at my code on `make_block` in BlockchainTests it seems like was really easy to make this mistake in the first place because of how the `block` and `env` are just mixed all over the place. I think it would be nice to refactor this a bit on a different PR to make a clear distinction that `block` structure is the source of the test information, but `env` is the structure that should contain all the correct information at some point.
- I think this same mistake could be present in the `StateTests`, it's just that we don't have any withdrawals tests written in the `StateTests` format.

marioevz (approved):
LGTM

PR #44, opened 2023-01-25
fillers/withdrawals: Add another withdrawals test with large amounts

marioevz:
Adds a withdrawals test that checks uint64 values which, when multiplied by 1e9, would overflow uint64 but not uint256.


spencer-tb (approved):
LGTM!

Issue #39, opened 2023-01-21
Exported test format not following the existing tests schema: Value format


winsvega:
Pyscpecs export 

```
"nonce": "0x0",
```

its a t8n tool format. we have different schema in tests, requested long time ago by teams
 
```
"0x00" for 0  "0x" for empty code
"0x01"  for 1.  so the data always padded to 1 byte
```

there was a discussion long time ago. I don't know if it is relevant to the testing team who initially asked about it. but we stick to this format

```
"balance": "0x432ae"    ->                   "balance": "0x0432ae"
```

but in storage we save the space 

```
                    "0x0000000000000000000000000000000000000000000000000000000000000001": "0x000000000000000000000000000000000000000000000000000000003b9aca00",
                    "0x0000000000000000000000000000000000000000000000000000000000000002": "0x0000000000000000000000000000000000000000000000000000000077359400"
```

would be 
```
                    "0x01": "0x3b9aca00",
                    "0x02": "0x77359400"
```

spencer-tb:
Here is an example of the updated format for the accounts:

![image](https://user-images.githubusercontent.com/60348173/214648983-8d9e5ace-a02f-4f24-a850-916ac2d836bf.png)



Compared to the old format:

![image](https://user-images.githubusercontent.com/60348173/214648893-315c0e4b-3cd2-465a-b407-04a75f3f09b8.png)


Is the fixture formatting in these issues only relevant to the pre and postState accounts?

winsvega:
if we speak about accounts yes. but the format remains the same for all fields.

 such as in transaction: value,data,gasPrice and so on.
 in Blockheaders. btw. nonce and mixhash has fixed length. as well as other hashes. in GenesisHeader. 

the only exception I see is                 `"blocknumber" : "2",` because that one is a debug field for human to read it is in dec format. also optional.

spencer-tb:
Closing after merging: #47

Issue #38, opened 2023-01-21
Exported test format not following the existing tests schema: Short Accounts


winsvega:
**Issue 1**
Short accounts:

```
Error: Expected field 'nonce' not found in config: Account 0x0000000000000000000000000000000000000100
"0x0000000000000000000000000000000000000100" : {
    "balance" : "0x0",
    "code" : "0x6000358031435550",
    "storage" : {
        "0x0000000000000000000000000000000000000000000000000000000000000001" : "0x000000000000000000000000000000000000000000000000000000003b9aca00",
        "0x0000000000000000000000000000000000000000000000000000000000000002" : "0x0000000000000000000000000000000000000000000000000000000077359400"
    }
} (EIPTests/bc4895-pyspecs020/000_shanghai, step: BlockchainTest)
```

In the tests all account fields are defined so no ambiguity

spencer-tb:
Closing after merging: #47.

PR #47, opened 2023-01-25
Update existing fixture test format to export correct schema

spencer-tb:
This PR addresses the 2 issues regarding the exported fixture json format: #39 #38


spencer-tb:
Here is an example of the updated format for the accounts:

image



Compared to the old format:

image

spencer-tb:
`even_padding()` can re-format any `Dict`. Waiting to find out what fields need to be excluded from re-formating.

src/ethereum_test_tools/common/types.py
```diff
@@ -425,6 +448,24 @@ def with_code(cls: Type, code: bytes | str | Code) -> "Account":
...
+def alloc_to_accounts(got_alloc: Dict[str, Any]) -> Mapping[str, Account]:
+    """
+    Converts the post state alloc returned from t8n to a mapping of accounts.
+    """
+    accounts = {}
+    for address, value in got_alloc.items():
+        account = Account(
+            nonce=int(value.get("nonce", None), 16)
+            if "nonce" in value
+            else None,
```
marioevz, line 460: I think we could use `int_or_none` here, but it depends on whether the value returned here from t8n has the `0x` at the beginning or not.

src/ethereum_test_tools/common/types.py
```diff
@@ -982,6 +1041,12 @@ def default(self, obj):
                 b["blocknumber"] = str(obj.block_number)
             if obj.chain_name is not None:
                 b["chainname"] = obj.chain_name
+            if obj.txs is not None:
+                b["transactions"] = obj.txs
+            if obj.ommers is not None:
+                b["uncleHeaders"] = obj.ommers
+            if obj.withdrawals is not None:
+                b["withdrawals"] = obj.withdrawals
```
marioevz, line 1049: Are we actually adding these fields during filling to the block object ?

src/ethereum_test_tools/common/types.py
```diff
@@ -1005,3 +1070,19 @@ def default(self, obj):
...
+    """
+    for key, value in input.items():
+        if key not in excluded:
+            if isinstance(value, dict):
+                even_padding(value, excluded)
+            elif value != "0x":
+                strip = value.lstrip("0x")
+                new_value = "0" if strip == "" else strip
+                if len(new_value) % 2:
+                    input[key] = "0x0" + new_value
```
marioevz, line 1087: I think this statement should have an else (adding only "0x"), unless I misunderstood something.

marioevz (changes requested):
Looks good overall but I have a few comments!

src/ethereum_test_tools/common/types.py
spencer-tb: Sounds good - I'll add the else for sanity!

src/ethereum_test_tools/common/types.py
spencer-tb: Sounds good!

src/ethereum_test_tools/common/types.py
spencer-tb: I have only added them within `make_blocks`  for `blockchain_test.py` so far, but plan to add them to the latter within `state_test.py` as well.

src/ethereum_test_tools/common/types.py
```diff
@@ -455,10 +455,10 @@ def alloc_to_accounts(got_alloc: Dict[str, Any]) -> Mapping[str, Account]:
     accounts = {}
     for address, value in got_alloc.items():
         account = Account(
-            nonce=int(value.get("nonce", None), 16)
-            if "nonce" in value
+            nonce=int_or_none(value["nonce"]) if "nonce" in value else None,
```
marioevz, line 458: I think `int_or_none(value.get("nonce", None))` would work here.

Same for balance below.

src/ethereum_test_tools/common/types.py
```diff
@@ -1028,6 +1027,25 @@ def default(self, obj):
                 ],
             )
         elif isinstance(obj, FixtureBlock):
+            b_txs = (
```
marioevz, line 1030: I think this part can be completely removed by using something like:
```
to_json_or_none(obj.txs)
```
Which in theory should simply reuse the code of the `elif isinstance(obj, Transaction):` branch in this same function.

The `even_padding` could also be added in that part of the code.

marioevz (commented):
Some comments on the new code.

src/ethereum_test_tools/common/types.py
spencer-tb: That was my original plan but the fields expected by `t8n.evaluate` within `blockchain_test.py/make_block()`, via `block.txs`:

image

are different to those required for the fixture format. For example `t8n` wants `"gas"` but the fixture format requires `"gasPrice"`, hence it was not possible to use the exact same `Transaction` type. As a workaround, the `Transaction` type is pre-processed as `b_txs` during the json formatting of a `FixtureBlock` hence fixture format of blocks are correct.

image

It would be useful if the format was standardised

src/ethereum_test_tools/common/types.py
```diff
@@ -260,6 +260,29 @@ def must_be_equal(self, other: "Storage"):
...
+        new_value = value.lstrip("0x").lstrip("0")
+        new_value = "00" if new_value == "" else new_value
+        if len(new_value) % 2 == 1:
+            new_value = "0" + new_value
+
+        addr = "0x" + new_addr
+        value = "0x" + new_value
+        new_storage[addr] = value
+
+    return new_storage
```
marioevz, line 283: I think this section could be improved by introducing another extra function to perform the padding like `storage_key_value_padding`, which would do the stripping and the padding on a single key/value string.

Then we can use dictionary comprehension to make this a one-liner like so:
```
new_storage = { storage_key_value_padding(k):storage_key_value_padding(v) for (k,v) in storage}
```

Also it would be nice to have a small test for this specific function in `src/ethereum_test_tools/tests/test_types.py`.

PR #48, opened 2023-01-26
feat(README.md): Remove prompts from commands

qbzzt:
Without those prompts users can just cut and paste all the commands, instead of doing it one by one.


spencer-tb (approved):
LGTM.

Issue #49, opened 2023-01-26
How do I get this working?


qbzzt:
I followed the directions to install, and now when I try to run a test I get:

```
(venv) qbzzt1@spec-tests:~/execution-spec-tests$ tf --output="fixtures" --test-case dup
DEBUG:ethereum_test_filling_tool.main:searching example.example for fillers
DEBUG:ethereum_test_filling_tool.main:searching vm.chain_id for fillers
DEBUG:ethereum_test_filling_tool.main:searching vm.dup for fillers
DEBUG:ethereum_test_filling_tool.main:searching withdrawals.withdrawals for fillers
DEBUG:ethereum_test_filling_tool.main:searching eips.eip3651 for fillers
DEBUG:ethereum_test_filling_tool.main:searching eips.eip3855 for fillers
DEBUG:ethereum_test_filling_tool.main:searching eips.eip3860 for fillers
INFO:ethereum_test_filling_tool.main:collected 1 fillers
DEBUG:ethereum_test_filling_tool.main:filling dup
Exception during test ''
Traceback (most recent call last):
  File "/home/qbzzt1/execution-spec-tests/venv/bin/tf", line 8, in <module>
    sys.exit(main())
  File "/home/qbzzt1/execution-spec-tests/src/ethereum_test_filling_tool/main.py", line 180, in main
    filler.fill()
  File "/home/qbzzt1/execution-spec-tests/src/ethereum_test_filling_tool/main.py", line 140, in fill
    fixture = filler(t8n, b11r, "NoProof")
  File "/home/qbzzt1/execution-spec-tests/src/ethereum_test_tools/filling/decorators.py", line 67, in inner
    return fill_test(
  File "/home/qbzzt1/execution-spec-tests/src/ethereum_test_tools/filling/fill.py", line 55, in fill_test
    raise e
  File "/home/qbzzt1/execution-spec-tests/src/ethereum_test_tools/filling/fill.py", line 45, in fill_test
    (blocks, head, alloc) = test.make_blocks(
  File "/home/qbzzt1/execution-spec-tests/src/ethereum_test_tools/spec/state_test.py", line 111, in make_blocks
    (alloc, result, txs_rlp) = t8n.evaluate(
  File "/home/qbzzt1/execution-spec-tests/src/evm_transition_tool/__init__.py", line 209, in evaluate
    raise Exception("failed to evaluate: " + result.stderr.decode())
[...]

PR #50, opened 2023-01-26
Update Readme: Most stable geth not compatible

spencer-tb:
As the most stable geth version:
```
evm version 1.10.26-stable-e5eb32ac
```
is not compatible with the current test filling procedure #49 , update the readme accordingly.


spencer-tb:
Closing as a duplicate.

PR #40, opened 2023-01-23
docs: Migrate repo documentation to a readthedocs.io wiki

spencer-tb:
This PR will start the process of creating a wiki/documentation page for this repo. 

Much like https://ethereum-tests.readthedocs.io/en/latest/#.

Upon approval of this PR, the current README.md documentation will be available here. 
https://execution-spec-tests.readthedocs.io/en/latest/

Extra documentation will be made in a separate PR.

PR #55, opened 2023-01-29
Tutorial for state transition tests in execution-spec-tests

qbzzt:


spencer-tb:
Nice job, I definitely learned a bit going through this. 

I've added some small tweaks to what you have wrote in the above commit, such that it will be compatible when this PR #40 is merged. If you are happy we can remove your original file and merge

qbzzt:
> I've added some small tweaks to what you have wrote in the above commit, such that it will be compatible when this PR #40 is merged. If you are happy we can remove your original file and merge

Great. Almost ready. I turned `theCamel` into `a_snake` in the test. I'll update the tutorial and then it'll be ready.

spencer-tb (approved):

PR #59, opened 2023-02-02
Additional Tox Checks: pep8-naming and fname8 for correct python naming conventions

spencer-tb:
Some small additions to the tox pipeline to check that files are in accordance with [PEP8](https://peps.python.org/pep-0008/).

1) To make sure file names are correctly named I have added a custom name checking package `fname8`: this checks that python file names are of the type `[a-z0-9_]`, for example `test_python.py` or `eip1897.py` would pass but `testPython.py` or `eip-1897.py` would fail.

      image

2) As `flake8` doesn't report mixed-case function names, I have added `pep8-naming` which integrates directly with `flake8` as an extension. This also asserts that class/variable names are in accordance with PEP8. 

      image

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,7 @@ package_dir =
 python_requires = >=3.10
 
 install_requires =
-    ethereum@git+https://github.com/ethereum/execution-specs.git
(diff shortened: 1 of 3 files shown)


marioevz (approved):
Looks good, thanks!

Issue #65, opened 2023-02-09
single test execution command


winsvega:
`tf --output="fixtures" --testfile ./fillers/withdrawals/withdrawals.py --outputfolder ./filled/withdrawals`

for single test generation or debug.

spencer-tb:
`tf --output="./filled/withdrawals" --test-module withdrawals` should work here.

PR #66, opened 2023-02-09
Feature: Update `tf` to fill tests within subdirectories with depth > 1, change path defaults, add --no-structure option.

spencer-tb:
Currently when adding a new test file, we are limited to having the tests within one sub-directory from `fillers` e.g `fillers/vm`. Tests within sub-directories deeper than that will not be found by the test filler `tf`.

This PR updates `tf` to search recursively when running from a top level directory i.e:
```console
tf --output="fixtures" --test-categories vm
```
hence tests within sub-directories of `vm` will be found, e.g `vm/vm_tests`.

Similarly it adds the extra option to only fill tests from a specific sub-directory i.e:
```console
tf --output="fixtures" --test-categories vm/vm_tests/vm_chain
```

### DIFF
--- a/src/ethereum_test_filling_tool/main.py
+++ b/src/ethereum_test_filling_tool/main.py
@@ -157,17 +157,36 @@ def find_modules(root, include_pkg, include_modules):
         root,
         include=include_pkg if include_pkg is not None else ("*",),
     ):
-        for info in iter_modules([os.path.join(root, package)]):
-            if not info.ispkg:
-                module_full_name = package + "." + info.name
-                if module_full_name not in modules:
-                    if not include_modules or include_modules in info.name:
-                        yield (
-                            package,
-                            info.name,
-                            info.module_finder.find_module(module_full_name),
-                        )
-                    modules.add(module_full_name)
+        package = package.replace(
+            ".", "/"
+        )  # sub_package tests i.e 'vm.vm_tests'
+        for info, package_path in recursive_iter_modules(root, package):
+            module_full_name = package_path + "." + info.name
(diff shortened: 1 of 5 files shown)


spencer-tb:
Another small change added in the above 3 commits. Default values for `--filler-path` and `--output` clearly set in `ethereum_test_filling_tool/main.py`. Can run the tool solely with `tf`.

`README.md` and `docs` updated to match this to make the tool more intuitive to use at first. (linking #65).

spencer-tb:
Add --no-structure option for output required in https://github.com/ethereum/execution-spec-tests/issues/70.

src/ethereum_test_filling_tool/main.py
```diff
@@ -157,17 +166,36 @@ def find_modules(root, include_pkg, include_modules):
...
+    Helper function for find_packages.
+    Iterates through all sub-packages (packages within a package).
+    Recursively navigates down the package tree until a new module is found.
+    """
+    for info in iter_modules([os.path.join(root, package)]):
+        if info.ispkg:
+            for sub_info in recursive_iter_modules(
+                root, os.path.join(package, info.name)
+            ):
+                yield sub_info
```
marioevz, line 195:
```suggestion
            yield from recursive_iter_modules(root, os.path.join(package, info.name))
```

marioevz (approved):
LGTM

(Only a small suggestion in one of the files)

src/ethereum_test_filling_tool/main.py
spencer-tb: Nice I will add the change! :)

PR #72, opened 2023-03-07
Feature & Refactor: Add Concurrency and Filler Skips to Improve `tf` Effienciency. Additional CLI arguments.`

spencer-tb:
This PR improves our run time of `tf` from ~15s to ~5s.

So far I've added an additional function called `parallel_fill` that is equivalent to `fill` but with concurrency, as well as a `--benchmark` CLI argument that logs the run time of `tf`.

To test for yourself you will need to edit `filler.fill()` in main and switch between `filler.parallel_fill()`. Using `tf --benchmark` to log the times.

`fill()`:
image

`parallel_fill()`:
image

### DIFF
--- a/src/ethereum_test_filling_tool/filler.py
+++ b/src/ethereum_test_filling_tool/filler.py
@@ -113,7 +113,8 @@ class Filler:
 
         # Only skip if the fixture file already exists, the module
         # has not been modified since the last test filler run, and
-        # the user does not want to overwrite the fixtures (--overwrite).
+        # the user doesn't want to force a refill the
+        # fixtures (--force-refill).
         if (
             os.path.exists(path)
             and not is_module_modified(
--- a/src/ethereum_test_filling_tool/filler.py
+++ b/src/ethereum_test_filling_tool/filler.py
@@ -119,7 +119,7 @@ class Filler:
             and not is_module_modified(
                 path, self.options.filler_path, module_path
             )
-            and not self.options.overwrite
+            and not self.options.force_refill
         ):
             self.log.debug(f"skipping - {full_name}")
             return
--- a/src/ethereum_test_filling_tool/filler.py
+++ b/src/ethereum_test_filling_tool/filler.py
@@ -1,0 +1,132 @@
+"""
+Provides the Filler Class:
+
+Fillers are python functions that, given an `EvmTransitionTool` and
+`EvmBlockBuilder`, return a JSON object representing an Ethereum test case.
+
+This tool will traverse a package of filler python modules, fill each test
+case within it, and write them to a file in a given output directory.
+"""
+import argparse
+import concurrent.futures
+import json
+import logging
+import os
+import time
+
+from ethereum_test_tools import JSONEncoder
+from evm_block_builder import EvmBlockBuilder
+from evm_transition_tool import EvmTransitionTool
+
+from .modules import find_modules, is_module_modified
+
+
+class Filler:
+    """
+    A command line tool to process test fillers into full hydrated tests.
+    """
+
+    log: logging.Logger
+
+    def __init__(self, options: argparse.Namespace) -> None:
+        self.log = logging.getLogger(__name__)
+        self.options = options
+
+    def fill(self) -> None:
+        """
+        Fill test fixtures.
+        """
+        if self.options.benchmark:
+            start_time = time.time()
+
+        fillers = self.get_fillers()
+        self.log.info(f"collected {len(fillers)} fillers")
+
+        os.makedirs(self.options.output, exist_ok=True)
+
+        t8n = EvmTransitionTool(
+            binary=self.options.evm_bin, trace=self.options.traces
+        )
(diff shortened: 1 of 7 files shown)


src/ethereum_test_filling_tool/main.py
```diff
@@ -154,6 +165,92 @@ def fill(self) -> None:
...
+                    self.options.output,
+                    *(filler.__filler_metadata__["module_path"])
+                    if not self.options.no_output_structure
+                    else "",
+                )
+                os.makedirs(output_dir, exist_ok=True)
+                path = os.path.join(output_dir, f"{name}.json")
+
+                name = path[9 : len(path) - 5].replace("/", ".")
+                self.log.debug(f"filling - {name}")
```
marioevz, line 232:
```suggestion
                module_path = filler.__filler_metadata__["module_path"]
                output_dir = os.path.join(
                    self.options.output,
                    *module_path
                    if not self.options.no_output_structure
                    else "",
                )
                os.makedirs(output_dir, exist_ok=True)
                path = os.path.join(output_dir, f"{name}.json")

                full_name = ".".join(module_path + [name])
                self.log.debug(f"filling - {full_name}")
```

src/ethereum_test_filling_tool/main.py
```diff
@@ -154,6 +165,92 @@ def fill(self) -> None:
...
+        self.log.info(f"collected {len(fillers)} fillers")
+
+        os.makedirs(self.options.output, exist_ok=True)
+
+        t8n = EvmTransitionTool(
+            binary=self.options.evm_bin, trace=self.options.traces
+        )
+        b11r = EvmBlockBuilder(binary=self.options.evm_bin)
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
marioevz, line 218: I think this method can be made the default one and perhaps we can include an argument to specify `max_workers` here, which could be set to `1` if the user wants to fill them serially.

marioevz (commented):
Looks amazing!

Only a few comments.

Thanks!

src/ethereum_test_filling_tool/main.py
```diff
@@ -85,6 +87,25 @@ def parse_arguments() -> argparse.Namespace:
...
+            type=int,
+            help="specifies the max number of workers for the test filler \
+                  set to 1 for serial execution",
+        )
+
+        parser.add_argument(
+            "--no-skip",
+            action="store_true",
+            help="fill all test fillers and don't skip any tests",
+        )
```
marioevz, line 107: I think something like `"--force-refill"` would be a bit clearer.

Also adding a explanation of when the tests are skipped would be helpful.

spencer-tb: Changed to `--force-refill`. Comments below hopefully address the latter.

```python
# Only skip if the fixture file already exists, the module
# has not been modified since the last test filler run, and
# the user does not want to overwrite the fixtures (--overwrite).
if (os.path.exists(path) 
    and not is_module_modified(path, self.options.filler_path, module_path)
    and not self.options.force_refill):
          self.log.debug(f"skipping - {full_name}")
          return
```

spencer-tb:
**To summarise, we now have a refactor of `ethereum_test_filling_tool`:**
- `__init__.py`
- `filler.py` 
- `main.py`
- `modules.py`

**With the following features:**
- Default `tf` execution is concurrent/parallel. So far ~65% increase in speed. This should scale when more tests are added in the future.
- If a filler has already been filled into a fixture (.json file), it is skipped. Unless the filler file (the entire module not the test case) has been modified or the fixture file that is generated by the module does not exist in the output path (./fixtures by default).
  - This improves speed significantly. For example, now if I have ran `tf` for every filler, and then modify the withdrawals module `withdrawals.py`, `tf` will only fill this module. This takes a total time of ~1.76s. 
- New CLI arguments:
  - `tf --benchmark`. Adds timing of filling to the output e.g: `INFO:ethereum_test_filling_tool.filler:Filled test fixtures in 7.34 seconds.`
  - `tf --max-workers`. Allows the user to set the number of workers used for concurrent `tf` execution. For serial execution (the previous way) use `tf --max-workers 1`.
  - `tf --force-refill`. Force tests to be filled. Works with `--test-module` and `--test-categories`.

src/ethereum_test_filling_tool/filler.py
```diff
@@ -0,0 +1,132 @@
...
+            self.options.output,
+            *(module_path if not self.options.no_output_structure else ""),
+        )
+        os.makedirs(output_dir, exist_ok=True)
+        path = os.path.join(output_dir, f"{name}.json")
+        full_name = ".".join(module_path + [name])
+
+        # Only skip if the fixture file already exists, the module
+        # has not been modified since the last test filler run, and
+        # the user does not want to overwrite the fixtures (--overwrite).
```
marioevz, line 116: Force refill instead of overwrite.

spencer-tb: Cheers! Updated

marioevz (approved):
LGTM. Thanks!

PR #78, opened 2023-03-15
Update readme/docs for python versioning.

spencer-tb:
To avoid confusion and errors when new users are getting started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Ethereum tests. [Further documentation](https://execution-spec-tests.readthedocs
 
 The following are required to either generate or develop tests:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,7 +9,7 @@ Ethereum tests. [Further documentation](https://execution-spec-tests.readthedocs
 
 The following are required to either generate or develop tests:
 
-1. Python >=`3.10.0`.
+1. Python ==`3.10.0`.
```
marioevz, line 12: Should be something like ">= 3.10.0 < 3.11", since 3.10.8 still works.

marioevz (approved):
LGTM

Issue #74, opened 2023-03-08
output fixtures has empty pre or postState


winsvega:
in  withdrawals_use_value_in_tx.json
in     "000_shanghai" : {

the pre and postState are empty. 
Current test base has no empty states whatsoever.
It good to define the sender account as empty or having 0 balance for explicit checks

spencer-tb:
**For `use_value_in_tx/000/shanghai` want:** (Expect an error)
```json
        "pre": {
            "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
                "nonce": "0x00",
                "balance": "0x00",
                "code": "0x",
                "storage": {}
            }
        },
        "postState": {
            "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
                "nonce": "0x00",
                "balance": "0x00",
                "code": "0x",
                "storage": {}
            }
        },
```

**For `use_value_in_tx/001/shanghai` want:**
```json
        "pre": {
            "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
                "nonce": "0x00",
                "balance": "0x00",
                "code": "0x",
                "storage": {}
            }
        },
        "postState": {
             "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba": {
                "nonce": "0x00",
                "balance": "0x1319718811c8",
                "code": "0x",
                "storage": {}
            },
            "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
                "nonce": "0x00",
                "balance": "0x00",
                "code": "0x",
                "storage": {}
            }
        },
```

PR #79, opened 2023-03-15
Fixture Format: Add default pre-state to withdrawals/use_value_in_tx so pre/post in fixture is not empty.

spencer-tb:
Addresses: #74


marioevz (approved):
LGTM, thanks!

marioevz:
I think there are some extra changes that might need to be removed before merging :)

PR #81, opened 2023-03-17
Add reference spec information to fixtures and fillers

marioevz:
This PR introduces functionality to add reference specification information to all fillers, including a path to a file and a version.

For example, one can reference a specific EIP markdown file in the ethereum/EIPs repository, and its SHA digest, and the filler will automatically check if this is the latest version of the spec, then issue a warning if the spec has been updated since (see example provided for file `fillers/eips/eip3651.py`).

The reference spec information is also written to the generated fixture.


src/ethereum_test_tools/common/reference_spec.py
```diff
@@ -0,0 +1,162 @@
...
+            f"https://api.github.com/repos/{self.RepoOwner}/"
+            + f"{self.RepoName}/git/blobs/{self.SpecVersion}"
+        )
+        if response.status_code != 200:
+            return None
+        content = json.loads(response.content)
+        content["content"] = decode_base64_content(content["content"])
+        return content
+
+    def latest_spec(self) -> Dict:
```
spencer-tb, line 114: Return of `latest_spec` should be `Dict | None`.

src/ethereum_test_tools/common/reference_spec.py
```diff
@@ -0,0 +1,162 @@
...
+    def write_info(self, info: Dict[str, str]):
+        """
+        Writes info about the reference specification used into the output
+        fixture.
+        """
+        info["reference-spec"] = self.name()
+        info["reference-spec-version"] = self.SpecVersion
+
+
+_ = GitReferenceSpec(SpecPath="")
```
spencer-tb, line 162: Maybe this line is not needed?

spencer-tb (approved):
Really nice! This will be super useful. Only a couple of small things caught my attention. 

Maybe we could add another argument to `tf` like `--check-spec` in the future.

src/ethereum_test_tools/common/reference_spec.py
marioevz: Done, thanks!

spencer-tb:
Nice! Yeah I prefer the 2nd option for the reasons you mentioned:
> It looks cleaner and simpler for the end-user to me

As long as the extra changes required aren't too complex on the filler :D

src/ethereum_test_tools/common/reference_spec.py
marioevz: I'm not sure how to do this properly in python.

Basically what I want to do is for mypy to immediately warn if there was an abstract method from the base class, `ReferenceSpec`, that was not implemented.

spencer-tb (approved):
LGTM, docs are concise and easily understandable!

PR #83, opened 2023-03-18
add access list txn

gurukamath:
Add `AccessList` type and provide an example of a case with access list transactions.

### DIFF
--- a/fillers/example/acl_example.py
+++ b/fillers/example/acl_example.py
@@ -2,15 +2,9 @@
 Test ACL Transaction Source Code Examples
 """
 
-from ethereum_test_tools import (
-    AccessList,
-    Account,
-    Environment,
-    Opcodes as Op,
-    StateTest,
-    Transaction,
-    test_from_until,
-)
+from ethereum_test_tools import AccessList, Account, Environment
+from ethereum_test_tools import Opcodes as Op
+from ethereum_test_tools import StateTest, Transaction, test_from_until
 
 
 @test_from_until("berlin", "london")
--- a/fillers/example/acl_example.py
+++ b/fillers/example/acl_example.py
@@ -6,6 +6,7 @@ from ethereum_test_tools import (
     AccessList,
     Account,
     Environment,
+    Opcodes as Op,
     StateTest,
     Transaction,
     test_from_until,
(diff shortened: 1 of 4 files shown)


fillers/example/acl_example.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Test type 1 transaction.
+    """
+    env = Environment(
+        coinbase="0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
+        difficulty=0x20000,
+        gas_limit=68719476736,
+        number=16777216,
+        timestamp=4,
+    )
```
marioevz, line 26: It seems like the values `difficulty=0x20000` and `timestamp=4` are impossible for block 1 (`env` describes Block 1 in state tests), since the genesis block already has the minimum values of `difficulty=0x20000` and `timestamp=0`, the minimum expected difficulty for block 1 (assuming a timestamp increase of only 4) should be `0x20040`.

Also geth complains about the block number not being 0 for genesis and 1 for the first block (in the case of the resulting block test for this file, the numbers are 16777215 for genesis and 16777216 for the first block).

I filled an issue for this here: https://github.com/ethereum/execution-spec-tests/issues/87

We could discuss the best course of action in the issue, but for the time being I think we should use the default values for `env` in this test.

fillers/example/acl_example.py
```diff
@@ -0,0 +1,75 @@
...
+
+    pre = {
+        "0x000000000000000000000000000000000000aaaa": Account(
+            balance=0x03,
+            code="0x5854505854",
```
marioevz, line 31:
```suggestion
            code=Op.PC + Op.SLOAD + Op.POP + Op.PC + Op.SLOAD,
```
It's easier to read, and simply requires this import:
```
from ethereum_test_tools import Opcodes as Op
```

marioevz (commented):
Thanks so much for the changes! Just a couple of comments.

fillers/example/acl_example.py
gurukamath: Sure. I just picked these values from another test without much thought to that. I'll make the updates

fillers/example/acl_example.py
gurukamath: done

fillers/example/acl_example.py
gurukamath: done

marioevz (approved):
Thanks!

PR #91, opened 2023-03-30
Fork Definitions as Classes

marioevz:
## Changes Included
Forks are now specified as a fork object instead of strings, which allows for a more coherent use of forks across the repo.


src/ethereum_test_forks/__init__.py
```diff
@@ -0,0 +1,279 @@
...
+    """
+    Constantinople Fix fork
+    """
+
+    pass
+
+
+class Istanbul(ConstantinopleFix):
+    """
+    Istambul fork
```
spencer-tb, line 100: Minor spelling: Istambul -> Istanbul

spencer-tb (approved):
Really nice and clean!! LGTM

Issue #86, opened 2023-03-20
Transaction data export in .json is wrong


winsvega:
here my check fails: 
in withdrawals test in self_destructing_account.json  the transaction json is 
```
 "transactions" : [
                    {
                        "data" : "0x0200",
                        "gasLimit" : "0x0186a0",
                        "gasPrice" : "0x0a",
                        "nonce" : "0x00",
                        "secretKey" : "0x45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8",
                        "to" : "0x0000000000000000000000000000000000000100",
                        "value" : "0x00"
                    }
                ],
```

but the one enocded in RLP is 

```
RLP transaction number: 1
{
    "data" : "0x0000000000000000000000000000000000000000000000000000000000000200",
    "gasLimit" : "0x0186a0",
    "gasPrice" : "0x0a",
    "nonce" : "0x00",
    "to" : "0x0000000000000000000000000000000000000100",
    "value" : "0x00",
    "v" : "0x26",
    "r" : "0xfd75940662272e5daac655a822b10f101866916a2975c3821111ad4ab481d48b",
    "s" : "0x67a11d0fcd31933ad4b62b4d6fd4f1540f93406d501e41b82ff59223303f60c5"
}
```

pyspec removed leading zeroes in data field when writing rlp data to json.

spencer-tb:
#105

PR #92, opened 2023-04-03
Fixture Format: Remove zero padding for data field in txs fixture format.

spencer-tb:
Zero padding shouldn't be in data field for fixture format -> https://github.com/ethereum/execution-spec-tests/issues/86


marioevz (approved):
LGTM

PR #95, opened 2023-04-04
Update Datahash Opcode Module to Test Spec

spencer-tb:
I've added a couple more tests: `test_datahash_blob_versioned_hash()`, `test_datahash_invalid_blob_index()`, and updated `test_datahash_gas_cost()`. Mostly to match the test spec in the hack.md for the 4844 tests (more below).

**Edit: now I have also added `test_datahash_multiple_txs_in_block()`, and refactored `test_datahash_opcode_contexts()`.**

---

📖 Opcode DATAHASH

Test Module - eip4844/datahash_opcode.py

File Doc String: Verifies that the DATAHASH opcode, works as intended for a variety of contexts, retrieves the versioned hash correctly, handles out-of-range indices, and consumes the correct amount of gas.

1) 🔴 test_datahash_opcode_contexts(): :heavy_check_mark:

Tests that the DATAHASH opcode functions correctly when called in different contexts, including:

    DATAHASH opcode on the top level of the call stack.
    DATAHASH opcode on the max value.
    DATAHASH opcode on CALL, DELEGATECALL, STATICCALL, and CALLCODE.
    DATAHASH opcode on Initcode.
    DATAHASH opcode on CREATE and CREATE2.
    DATAHASH opcode on transaction types 0, 1 and 2.

2) 🔴 test_datahash_blob_versioned_hash(): :heavy_check_mark:

Tests that the DATAHASHopcode returns the correct versioned hash for various valid indexes. This test should cover various scenarios with random data versions within the valid range.

3) 🔴 test_datahash_invalid_blob_index(): :heavy_check_mark:

Tests that the DATAHASH opcode returns a zeroed bytes32 value for invalid indexes. This test should include cases where the index is negative or exceeds the maximum number of versions stored, where index >= len(tx.message.blob_versioned_hashes). It should confirm that the returned value is a zeroed bytes32 value for these cases.

4) 🟠 test_datahash_gas_cost(): :heavy_check_mark:

Asserts the gas consumption of the DATAHASH opcode and ensures it matches HASH_OPCODE_GAS = 3. Includes a variety of random index sizes, including min=0 and max=2^256-1, for tx types 2 and 5.

5) :yellow_circle:  test_datahash_multiple_txs_in_block(): :heavy_check_mark:

Tests that the `DATAHASH` opcode returns the appropriate values when there is more than one blob tx type within a block (for tx types 2 and 5). Scenarios involve tx type 5 followed by tx type 2 running the same code. In this case `DATAHASH` returns 0, but for the opposite scenario `DATAHASH` should return the correct blob hash value.


fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -538,118 +545,238 @@ def test_datahash_gas(_: str):
...
+    def create_tx(ty: int, to: str, nonce: int):
+        return Transaction(
+            ty=ty,
+            nonce=nonce,
+            data=to_hash_bytes(0),
+            to=to,
+            gas_limit=3000000,
+            max_fee_per_gas=10,
+            max_priority_fee_per_gas=10,
+            max_fee_per_data_gas=10,
```
marioevz, line 579: I'm not sure if this could cause issues (`max_fee_per_data_gas != None`) for type < 5.

We could set it to `None` for those cases just to be safe.

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -538,118 +545,238 @@ def test_datahash_gas(_: str):
...
-        gas_limit=3000000,
-        max_fee_per_gas=10,
-        max_priority_fee_per_gas=10,
-        max_fee_per_data_gas=10,
-        access_list=[],
-        blob_versioned_hashes=[to_hash_bytes(2**256 - 1)],
+
+
+@test_from(fork=ShardingFork)
+def test_datahash_blob_versioned_hash(_: str):
```
marioevz, line 622: We could add another test case where we request `DATAHASH` multiple times, to check if the result is the same.
E.g.:
```
- DATAHASH(0x00), DATAHASH(0x00)
- DATAHASH(0x00), DATAHASH(0xff..ff), DATAHASH(0x00)
- DATAHASH(0x00), DATAHASH(0x01), DATAHASH(0x00)
```
Where blobs 0x00 and 0x01 exist, just to name a few.

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -538,118 +545,238 @@ def test_datahash_gas(_: str):
...
+                    hex"{Op.DATAHASH.hex()}",pos)
+                sstore(pos, datahash)
+            }}
+        }}
+        """
+    )
+
+    # Create a list of random blob hashes
+    blob_ver_hashes = [
+        to_hash_bytes(random.randint(MAX_BLOB_PER_BLOCK, 2**256 - 1))
```
marioevz, line 656: I think we should use hard-coded values here too.

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -538,118 +545,238 @@ def test_datahash_gas(_: str):
...
     )
 
-    post = {
-        datahash_zero_gas_measure_code_address: Account(
-            storage={
-                0: DATAHASH_GAS_COST,
-            }
-        ),
-        datahash_max_gas_measure_code_address: Account(
+    datahash_code = Yul(
```
marioevz, line 639: I think we should generate this bytecode using `Op` and a python loop instead of relying on Yul.

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -538,118 +545,238 @@ def test_datahash_gas(_: str):
...
-        access_list=[],
+
+@test_from(fork=ShardingFork)
+def test_datahash_invalid_blob_index(_: str):
+    """
+    Tests that the `DATAHASH` opcode returns a zeroed bytes32 value
+    for invalid indexes.
+    """
+
+    INVALID_DEPTH = 3
```
marioevz, line 705: 3 is technically a valid depth, if a tx is using the max amount of blobs.
Maybe `MAX_BLOB_INDEX` ?

Also it should be `MAX_BLOB_PER_BLOCK - 1`.

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -538,118 +545,238 @@ def test_datahash_gas(_: str):
...
-        nonce=0,
-        data=to_hash_bytes(0),
-        to=datahash_zero_gas_measure_code_address,
-        gas_limit=3000000,
-        max_fee_per_gas=10,
-        max_priority_fee_per_gas=10,
-        access_list=[],
+
+@test_from(fork=ShardingFork)
+def test_datahash_invalid_blob_index(_: str):
```
marioevz, line 699: Also add a test where a block contains a blob transaction and then a type 2 tx, and this subsequent tx uses `DATAHASH`, it should return zero for all indexes where the first type 5 tx had blobs.

Similar test with two blob txs in the same block, and the second tx has less blobs, e.g., tx 1 has 2 blobs, tx 2 has 1 blob, DATAHASH(0x01) on tx 2 must return zero.

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -538,118 +545,238 @@ def test_datahash_gas(_: str):
...
 
-    yield StateTest(
-        env=env,
+    yield BlockchainTest(
         pre=pre,
         post=post,
-        txs=[tx1, tx2],
-        tag="cost_tx_type_2",
+        blocks=blocks,
     )
```
marioevz, line 782: We should also test opcodes that access tx information, since type 5 tx is now SSZ, there could be issues there. E.g.
- ORIGIN
- GASPRICE
- CHAINID
- CALLER (On the top level call stack)
- CALLVALUE (On the top level call stack)

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -538,118 +545,238 @@ def test_datahash_gas(_: str):
...
+    for i in range(NUM_BLOCKS):
+        address = to_address(0x100 + i * 0x100)
+        rnd_blob_size = random.randint(0, MAX_BLOB_PER_BLOCK)
+        pre[address] = Account(
+            code=Yul(
+                datahash_invalid_index_code.substitute(num_blobs=rnd_blob_size)
+            )
+        )
+        blob_ver_hashes = [
+            to_hash_bytes(random.randint(MAX_BLOB_PER_BLOCK, 2**256 - 1))
```
marioevz, line 746: Also remove randomness here.

marioevz (commented):
Thanks! Some comments :)

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -243,413 +277,497 @@ def test_datahash_opcode(_: Fork):
...
-    )
-    datahash_zero_gas_measure_code_address = to_address(0x100)
-
-    datahash_max_gas_measure_code = CodeGasMeasure(
-        code=(Op.PUSH32(2**256 - 1) + Op.DATAHASH),
-        overhead_cost=3,
-        extra_stack_items=1,
+    datahash_valid_call = b"".join(
+        [
+            Op.PUSH1(i) + Op.DATAHASH + Op.PUSH1(i) + Op.SSTORE()
```
marioevz, line 718: Same as comments above.

marioevz (commented):
Looks great so far! Just a couple more suggestions.
Thanks!

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -243,413 +277,497 @@ def test_datahash_opcode(_: Fork):
...
+                        access_list=[],
+                        blob_versioned_hashes=blob_hashes,
+                    )
+                ]
+            )
+        )
+        post[address] = Account(
             storage={
-                0: 0,
+                index: 0
```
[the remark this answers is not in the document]
spencer-tb: For the fixtures I believe it shouldn't show up in storage, as the values written to any negative key (index) should be zero. For example if we are storing `Datahash(-1)` this should be zero.

And currently I think a zero'd value in storage is not generated in the fixture format.

fillers/eips/eip4844/datahash_opcode.py
```diff
@@ -243,413 +289,493 @@ def test_datahash_opcode(_: Fork):
...
-        ),
-    }
+    post = {}
+
+    # Declare datahash indexes: zero, max & random values
+    datahash_index_measures: List[int] = [
+        0x00,
+        0x01,
+        0x02,
+        0x03,
```
marioevz, line 454:
```suggestion
        0x03,
        0x04,
```

marioevz (approved):
LGTM! Just some minor suggestions :)

PR #97, opened 2023-04-05
ethereum_test_forks: Refactor

marioevz:
## Changes Included
`LatestFork` can now be dynamically specified when filling tests by using flag `--latest-fork`, which can be used to specify filling tests for future forks.

By default, tests are filled up until the latest fork that most clients have already implemented (as of today, `Shanghai`).

This allows to merge tests for future forks into the main branch, but these tests will not be filled by default.


marioevz:
Changing to draft since I tried refactoring the 4844 branch and there were errors during filling.

src/ethereum_test_forks/base_fork.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class BaseFork:
+    """
+    An abstract class representing an Ethereum fork.
+
+    Must contain all the methods used by every fork.
+    """
+
+    @classmethod
```
spencer-tb, line 14: Super nice and clean! I am being nit-picky as we don't yet do this anywhere else but we could add the following to further specify its an abstract class method:

```py
from abc import ABC, abstractmethod

class BaseFork(ABC):
    ...

    @abstractmethod
    def header_base_fee_required(...)
    ...
```

src/ethereum_test_forks/helpers.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def set_latest_fork(fork: Fork):
+    """
+    Sets the latest fork
+    """
+    latest_fork_resolver.latest_fork = fork
+
+
+def set_latest_fork_by_name(fork_name: str):
```
spencer-tb, line 31: We could add type hints for these 2 functions: `-> None:` for both I think

spencer-tb (approved):
Super nice! I liked it a lot before - I didn't think I could like it any more :D Very easy to read

Only added a couple small thing but feel free to merge without!!

src/ethereum_test_forks/base_fork.py
marioevz: Yea, strictly speaking, BaseFork should be abstract, and the definition of the starting behavior should happen at the first fork (Frontier), with the only caveat being that when we add a new resolver (like base_fee or excess_data_gas), it needs to be defined in BaseFork as an abstract function, and at Frontier with the starting implementation.

PR #100, opened 2023-04-06
 Add Functions to copy Transaction/Block Objects

spencer-tb:
Currently we modify only some yielded Transaction objects correctly. This PR adds generic functions to allow us to modify all yielded Transaction/Block object correctly: using `tx.with_fields()` or `block.with_fields()`. 

These create a copy of the object such that when you have a Transaction/Block object yielded within a test case, you can create a simply copy/modification of that object for another test case to yielded later within the test.

### DIFF
--- a/fillers/withdrawals/withdrawals.py
+++ b/fillers/withdrawals/withdrawals.py
@@ -88,7 +88,11 @@ def test_use_value_in_tx(_):
 
     blocks = [
         Block(
-            txs=[tx.with_error("intrinsic gas too low: have 0, want 21000")],
+            txs=[
+                tx.with_fields(
+                    error="intrinsic gas too low: have 0, want 21000"
+                )
+            ],
             withdrawals=[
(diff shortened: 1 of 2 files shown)


fillers/withdrawals/withdrawals.py
```diff
@@ -492,7 +498,7 @@ def test_newly_created_contract(_):
 
     # Same test but include value in the contract creating transaction
 
-    tx.value = ONE_GWEI
+    block = block.with_fields(txs=[tx.with_fields(value=ONE_GWEI)])
```
marioevz, line 501: I feel like this should not be necessary. I see the use of `with_fields` when creating multiple similar txs that are going to be yielded to the same test, but once the yield has happened, it's like a clean slate, and the changes should never affect an already yielded test.

spencer-tb:
Will create a separate PR into main.

PR #104, opened 2023-04-11
fillers: 4844 Precompile Tests

marioevz:
Tests for the Point Evaluation Precompile included in EIP-4844:

- Verify correct output on correct kzg verifications, using all CALL* opcodes
- Verify behavior on incorrect length inputs to the precompile
- Verify behavior on incorrect versioned hash inputs to the precompile
- Verify correct gas usage on both correct and incorrect kzg verifications, using all CALL* opcodes
- File `fillers/eips/eip4844/point_evaluation_vectors/go_kzg_4844_verify_kzg_proof.json` uses test vectors from https://github.com/crate-crypto/go-kzg-4844/tree/master/tests/verify_kzg_proof/small as basis to call the precompile and verify outputs
- Verify calling the precompile before Cancun fork


danceratopz:
An amazing collection of tests. I couldn't come up with any missing cases yet :) 

Most of my suggestions are cosmetic, but there was one mistake in the gas usage tests (the `call_type` was not specified which led to all test cases being generated with Op.CALL (instead of CALLCODE, DELEGATECALL, STATICCALL, as intended).

I worked and ran the code locally, so I added my suggestions to https://github.com/marioevz/execution-spec-tests/pull/6; I hope you don't mind that I didn't add the suggestions in a review here.

Btw, as the test case generation is wonderfully abstracted to the respective `KZGPointEvaluation` methods, I think this will be one of the easier modules to convert to pytest format.

spencer-tb (approved):
LGTM! :rocket:

PR #105, opened 2023-04-11
tools/types: Convert to JSON on fixture init

marioevz:
Fixes some tests where there was a modification to the transactions/withdrawals after yield, by immediately converting to JSON on fixture initialization.


spencer-tb (approved):
Great!! LGTM

PR #106, opened 2023-04-12
tools/types: Correctly handle negative numbers in storage

marioevz:
Allow negative values in the storage, which are automatically converted to the EVM's two's complement of 32-byte.

Also add a check for an ambiguous storage check, i.e., the tester defined the equivalent EVM value as negative and positive numbers, and throw an exception.


spencer-tb (approved):
Nice will use this within the datahash module!!

PR #107, opened 2023-04-12
fillers/withdrawals: Add untouched accounts on zero-amount withdrawals

marioevz:
Adds a simple check to verify behavior on untouched accounts when a zero-amount withdraw happens.

This behavior is not specified on https://eips.ethereum.org/EIPS/eip-4895


spencer-tb (approved):
LGTM!

PR #109, opened 2023-04-21
tools/types: Assert tx[to] always within fixtures.

spencer-tb:
Previous PR merged changes the fixture format of transactions in the `newly_created_contract` withdrawals test case. The `tx["to"]` field should be present in fixture even set to `None`. The PR makes sure that `tx["to"] = ""` if its set to 'None' in a test case.


marioevz (approved):
LGTM

PR #112, opened 2023-04-25
fillers/security: Add test for tx selfdestuct balance security bug

spencer-tb:
Adds a test for an older bug:  https://github.com/ethereum/go-ethereum/security/advisories/GHSA-xw37-57qp-9mm4

### DIFF
--- a/fillers/security/__init__.py
+++ b/fillers/security/__init__.py
@@ -1,0 +1,3 @@
+"""
+Cross-client Ethereum Virtual Machine Tests
+"""
--- a/fillers/security/__init__.py
+++ b/fillers/security/__init__.py
@@ -1,3 +1,3 @@
 """
-Cross-client Ethereum Virtual Machine Tests
+Cross-client Ethereum Security Tests
 """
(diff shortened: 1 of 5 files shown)


fillers/security/selfdestruct_balance_bug.py
```diff
@@ -0,0 +1,135 @@
+"""
+Tests the following consensus-vulnerability.
+
+https://github.com/ethereum/go-ethereum/security/advisories/GHSA-xw37-57qp-9mm4
+
+Requires an geth evm version between >= v1.9.4 and v1.9.20 to fill the test.
```
marioevz, line 6: This is only true if we need the test to fail I think, otherwise we should always be able to fill a test with the latest geth version (for a fixed issue).

marioevz (approved):
Looks great! I think we need to rebase first before merging.

fillers/security/selfdestruct_balance_bug.py
spencer-tb: Yeah you are right here! I'll update the comment.

fillers/security/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
+Cross-client Ethereum Virtual Machine Tests
```
marioevz, line 2:
```suggestion
Cross-client Ethereum Virtual Machine Security Tests
```

PR #114, opened 2023-04-25
feature: Improve Logging of `tf`

spencer-tb:
This PR adds the minimal changes from #82 for now before we upgrade the repo to pytest.

Reference specs are updated for relevant Shanghai eips as they have been moved final. Commit example: https://github.com/ethereum/EIPs/commit/5ccb61c9d28348c07a095b0b478cf52c38ab107d

Reference spec checks are now ignored for specific modules: `["vm", "example", "security"]`.

Logging is set to `INFO` level in `logger.cfg`. We now have a very clean output so if we reference spec warnings it will be very easy to see.


danceratopz:
Really like the format/colorisation! Works for me with the change above. We can still use this when we move to pytest; pytest will write the logs directed to stdout in the console if it's ran with the `-s` flag.

tox.ini
```diff
@@ -2,9 +2,14 @@
 envlist = py3
 
 [testenv]
+deps =
+    pytest
+    colorlog
```
marioevz, line 7: I think we need this dependency also in the `setup.cfg`.

I'm getting the following error locally:

```
ModuleNotFoundError: No module named 'colorlog'
```

marioevz (changes requested):
Just one comment regarding the dependencies.

setup.cfg
```diff
@@ -48,6 +48,7 @@ test =
     pytest>=7.2
     pytest-cov>=2.12,<3
     pytest-xdist>=2.3.0,<3
+    colorlog>=6.7.0
```
marioevz, line 51: It's still failing in my local setup :/

I moved it to the `install_requires` category and it seems to be working fine.

marioevz (approved):
Thanks!

PR #119, opened 2023-04-28
fillers: update gas usage for subcalls in warm coinbase tests fixes #118

### LINKED ISSUE #118: Update exact gas specified for warm coinbase contract calls in eip3651 fillers
Commit https://github.com/ethereum/execution-spec-tests/commit/0670954459d3744db021fa4aed91f041a188149f refactored the call functions under test in `test_warm_coinbase_call_out_of_gas` to use the opcode format instead of specifying the calls as Yul. The exact gas usage specified for each call function now appears to be too generous as the opcode format's bytecode is more optimized than that generated by solc from the Yul format (I assume).

Although this is unlikely to effect the test outcome (as the difference between calls to cold and warm accounts is much larger than the discrepancy described above; 2600 vs 100 gas) it seems worthwhile to update the exact gas usage in this test.

FYI: The pytest port of this test introduced a new test that additionally tests that these calls indeed fail if `exact_gas_usage - 1` is specified as the call gas for the subcall, see https://github.com/ethereum/execution-spec-tests/pull/116/commits/79ed2a2fe6a20360a8c9a401ad6718038dcfb605

danceratopz:
Fixes #118.

### DIFF
--- a/fillers/eips/eip3651.py
+++ b/fillers/eips/eip3651.py
@@ -53,20 +53,20 @@ def test_warm_coinbase_call_out_of_gas(fork):
 
            switch calladdr
            case 0x100 {
-             // Extra: COINBASE + 6xPUSH1 + DUP6 + 2xPOP
-             callgas := add(callgas, 27)
+             // Extra: COINBASE + 4xPUSH1 + 2xDUP1 + POP
+             callgas := add(callgas, 22)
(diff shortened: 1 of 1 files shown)


marioevz (approved):
Looks good! Thanks! :)

PR #122, opened 2023-05-03
EIP 6780 Tests

jwasinger:
I'm unsure how to fill these.  I have tried:
```
>  tf --filler-path="fillers" --output="fixtures" --test-categories eips.eip6780  --force-refill

DEBUG:ethereum_test_filling_tool.filler:searching eips.eip6780.eip6780 for fillers
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): api.github.com:443
DEBUG:urllib3.connectionpool:https://api.github.com:443 "GET /repos/ethereum/EIPs/contents/EIPS/eip-6780.md HTTP/1.1" 200 3449
```

After a bit of digging, I figured out that the filler is not detected because it is configured for an upcoming fork (have also tried using `Cancun` with the same result).


fillers/eips/eip6780/eip6780.py
```diff
@@ -0,0 +1,138 @@
...
+                number=1,
+                timestamp=1000,
+        )
+        post: Dict[Any, Any] = {}
+
+        test_ops =[
+                "6032",
+                "601c",
+                "6000",
+                "39", # codecopy
```
marioevz, line 40: This bytecode can be replaced with the `Op` library to make it more readable.

For example, these lines can be replaced with `Op.CODECOPY(0, 28, 50)`, which will automatically place the numbers in the stack using PUSH operations.

danceratopz:
> I'm unsure how to fill these. I have tried:
> 
> ```
> >  tf --filler-path="fillers" --output="fixtures" --test-categories eips.eip6780  --force-refill
> 
> DEBUG:ethereum_test_filling_tool.filler:searching eips.eip6780.eip6780 for fillers
> DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): api.github.com:443
> DEBUG:urllib3.connectionpool:https://api.github.com:443 "GET /repos/ethereum/EIPs/contents/EIPS/eip-6780.md HTTP/1.1" 200 3449
> ```
> 
> After a bit of digging, I figured out that the filler is not detected because it is configured for an upcoming fork (have also tried using `Cancun` with the same result).

Hi @jwasinger, I just struggled to fill these myself. It turns out that you need to specify `--latest-fork Cancun` (I found this info hidden in the [v0.2.5 release notes](https://github.com/ethereum/execution-spec-tests/releases/tag/v0.2.5))
```console
tf --latest-fork Cancun --test-categories eips.eip6780
```
If I run this using the `evm` tool from https://github.com/jwasinger/go-ethereum/commit/06c795dc4a384554a381d781e6287887a9249211 branch I get:
```
➜ tf --latest-fork Cancun --test-categories eips.eip6780 --traces
DEBUG:ethereum_test_filling_tool.filler:searching eips.eip6780.eip6780 for fillers
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): api.github.com:443
DEBUG:urllib3.connectionpool:https://api.github.com:443 "GET /repos/ethereum/EIPs/contents/EIPS/eip-6780.md HTTP/1.1" 200 None
INFO:ethereum_test_filling_tool.filler:collected 2 fillers
DEBUG:ethereum_test_filling_tool.filler:skipping - eips/eip6780/eip6780/eip6780_create_selfdestruct_same_tx
Traceback (most recent call last):
  File "/home/dtopz/code/github/jwasinger/execution-spec-tests/venv/bin/tf", line 8, in <module>
<- snip ->
Exception: failed to evaluate: ERROR(3): failed constructing chain configuration: unsupported fork "ShanghaiToCancunAtTime15k"
```
I get further, but the filler's post condition fails, if I run it against the version of go-ethereum (https://github.com/mdehoog/go-ethereum/commit/ac64c4400e35cf8c7e2d02f99e257534096d2ec9) that was used for the execution-spec-tests [v0.2.5 release](https://github.com/ethereum/execution-spec-tests/releases/tag/v0.2.5)), but https://github.com/mdehoog/go-ethereum/commit/ac64c4400e35cf8c7e2d02f99e257534096d2ec9 is focussed on 4844 and hasn't implemented 6780.

@marioevz or @spencer-tb, perhaps you've already cleared this up? If not, could you help further?

spencer-tb:
One thing I just realized is that the new selfdestruct behaviour will change the outcome of this test: https://github.com/ethereum/execution-spec-tests/blob/main/fillers/security/selfdestruct_balance_bug.py

I am thinking we could run the security tests in a different scope, or have a method of specifying a specific `evm` version for these tests.

fillers/eips/eip6780/eip6780.py
jwasinger: yes, that would be a good improvement.

jwasinger:
@spencer-tb @danceratopz thanks for taking a further look into this.

> It looked like the account was being marked as deleted if it called selfdestruct, regardless of whether it was created in the same transaction.

This is the existing behavior for `selfdestruct` that we retain for pre-6780 blocks.  With 6780, the `suicided` flag is no longer set and now `sendalled` is set when an account is the originator of `selfdestruct`.  The only case where we delete a non-empty state object is when the state object created and selfdestructed in the same tx (`(obj.created && obj.sendalled)`).

The reason the test was failing to fill is that the Cancun instruction set wasn't being enabled.  I've added that to https://github.com/jwasinger/go-ethereum/commit/0aebb62768ea416c899910b86f39168f5c4bb965 and  removed https://github.com/spencer-tb/go-ethereum/commit/5a1923dcbb0028e4c3dd1d2767982fa7173f6ce4 .  The tests now fill for me.

> One thing I just realized is that the new selfdestruct behaviour will change the outcome of this test: https://github.com/ethereum/execution-spec-tests/blob/main/fillers/security/selfdestruct_balance_bug.py
> I am thinking we could run the security tests in a different scope, or have a method of specifying a specific evm version for these tests.

IMO it should be sufficient to just disable filling this test for post-cancun/6780 forks.

jwasinger:
IMO, a way to make it easier to fill tests for feature branches (potentially) included with upcoming hard-forks would be to allow `tf` to take a custom genesis configuration.

marioevz:
Hey @jwasinger, we recently made a lot of changes to the repo with the inclusion of the pytest framework, if you need any help running your tests with the new changes, just let me, @danceratopz or @spencer-tb  and we can help you out :)

danceratopz:
Hi @jwasinger, I just rebased this on main and refactored the tests to run with the new pytest-based execution-spec-tests framework - we can pick up work on these tests again!

You can run the tests with:
```console
fill tests/cancun/eip6780_selfdestruct/ --fork=Shanghai --evm-bin=/path/to/jwasinger/go-ethereum/build/bin/evm  -v --runxfail 
```
The test ` test_selfdestruct_prev_created` is currently marked to xfail, `--runxfail` is a pytest flag which runs this test as though it's not marked to xfail.


Couple of comments:
1. I tried to refactor the bytecode to use the Opcodes class: https://github.com/ethereum/execution-spec-tests/blob/a14258b19ff35ebefb25d33628dc8b34c9f76cb4/src/ethereum_test_tools/vm/opcode.py. Perhaps it helps you, I think the bytecode for `test_create_selfdestruct_same_tx` is still a work in progress? Either way feel free to use your original code instead.
2. The fork (Shanghai) in the above command-line call is currently wrong, as is the validity marker in the test:
    https://github.com/ethereum/execution-spec-tests/blob/7dc5a63ff52715eeed8815263f200f57eba7753d/tests/cancun/eip6780_selfdestruct/test_selfdestruct.py#L19
    This is a work-around to run the test with the `evm t8n` tool from your branch - the framework would otherwise pass "Cancun" as the fork, which is currently not implemented in your branch (rightly so 🙂). We both need to have a look how to enable 6780 by specifying the EIP to `evm t8n`: We need to make a minor improvement on our side to (re-)enable this functionality (I think it was working, but relatively untested before). And I think you need to add 6780 to the list of "avaliable extra eips", I refer to this list from `evm t8n --help`:
    ```
              Shanghai
            Available extra eips:
                1153, 1344, 1884, 2200, 2929, 3198, 3529,
          3855, 3860
            Syntax <forkname>(+ExtraEip)
    ```

There's no rush on 2., as it currently stands we can resume work on these tests.

jwasinger:
Thanks for updating this @danceratopz .  I have pushed a change to the 6780 geth branch to make the EIP available as an extra eip via `t8n`.

marioevz:
> Thanks for updating this @danceratopz . I have pushed a change to the 6780 geth branch to make the EIP available as an extra eip via `t8n`.

@jwasinger I added the fixture so the tests run both with `--state.fork=Shanghai` and `--state.fork=Shanghai+6780`, and did the appropriate expected storage changes each time.

I'll be adding some more tests to this branch in the next couple of hours if you don't mind :)

marioevz:
Some test ideas:
- Destroy and re-create multiple times in the same tx
- Create and destroy multiple contracts in the same tx
- Create multiple contracts in a tx and destroy only one of them, but attempt to destroy the other one in a subsequent tx
- Create contract, attempt to destroy it in a subsequent tx in the same block
- Create a contract and try to destroy using another calltype (e.g. Delegatecall then destroy)

cc @winsvega

tests/cancun/eip6780_selfdestruct/test_selfdestruct.py
```diff
@@ -0,0 +1,176 @@
...
+    # copy inner payload to memory
+    code += Op.CODECOPY(0, 0x0C, 0x08)
+    # return payload
+    # inner payload:
+    code += Op.RETURN(0x00, 0x08)
+    code += Op.SSTORE(0, 1)
+    code += Op.PUSH0
+    code += Op.SELFDESTRUCT
+
+    post["0x5fef11c6545be552c986e9eaac3144ecf2258fd3"] = Account.NONEXISTENT
```
shemnon, line 107: The offsets for the codecopy operations are off, an intrinsic risk in bytecode authoring.  I think a Yul version would be easier to read and more resiliant to changes.

```suggestion
def test_create_selfdestruct_same_tx(state_test: StateTestFiller, yul: YulCompiler):
    """
    TODO test that if a contract is created and then selfdestructs in the same
    transaction the contract should not be created.
    """
    env = Environment(
        coinbase="0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
        difficulty=0x20000,
        gas_limit=10000000000,
        number=1,
        timestamp=1000,
    )
    post: Dict[Any, Any] = {}

    pre = {
        TestAddress: Account(balance=1000000000000000000000),
    }
    code = yul(
        """
        object "Outer" {
            code {
                let size := datasize("Middle")
                datacopy(0, dataoffset("Middle"), size)
                let contract := create(0, 0, size)
                pop(call(gas(), contract, 0, 0, 0, 0, 0))
                selfdestruct(0)
            }
            object "Middle" {
                code {
                    let size := datasize("Inner")
                    datacopy(0, dataoffset("Inner"), size)
                    return(0, size)
                }
                object "Inner" {
                    code { selfdestruct(0) }
                }
            }
        }
        """
    )

    post["0x6295ee1b4f6dd65047762f924ecd367c17eabf8f"] = Account.NONEXISTENT
    post["0x64e2ebd6405af8cb348aec519084d3fff42ebba6"] = Account.NONEXISTENT
```

marioevz: Hey @shemnon, thanks for the suggestion :)  I was finishing up the refactor when I saw your comment.

This is how it ended up:
https://github.com/jwasinger/execution-spec-tests/blob/8d1f4ff795a6cc2d47f89f0e29e29fd6495c260a/tests/cancun/eip6780_selfdestruct/test_selfdestruct.py#L74-L90
and
https://github.com/jwasinger/execution-spec-tests/blob/8d1f4ff795a6cc2d47f89f0e29e29fd6495c260a/tests/cancun/eip6780_selfdestruct/test_selfdestruct.py#L140-L153

The first code extract is using an `Initcode` construct that's written in python and wraps up any bytecode in the appropriate initcode, taking into consideration the bytecode length and other things.

The second one, I ended up using extcodecopy to make it easier to copy the initcode to the self-destructing contract.

I've also added several other test cases, but it's not by any means exhaustive yet, so any feedback is greatly appreciated!

shemnon:
New test idea, since sending ether to a non-existant account creates it:
single tx - 
* call into non-existent account, sending value
* create that account with code that will self destruct
* call the code, self destruct.
* assert the account doesn't persist.

Variation of the above, but self destruct is in the initcode, skipping the call test but keeping the assert.

marioevz:
Added more tests and packaged them all in here:

[eip-6780-blockchain-tests.tar.gz](https://github.com/ethereum/execution-spec-tests/files/11986967/eip-6780-blockchain-tests.tar.gz)

More tests are still missing but these should provide a nice start point for implementations.

tests/cancun/eip6780_selfdestruct/test_selfdestruct.py
```diff
@@ -0,0 +1,784 @@
...
+        # Address where the contract is created already had some balance,
+        # which must be included in the send-all operation
+        sendall_amount += selfdestruct_contract_initial_balance
+
+    post: Dict[str, Account] = {
+        entry_code_address: Account(
+            code="0x00",
+            storage=entry_code_storage,
+        ),
+        selfdestruct_contract_address: Account.NONEXISTENT,  # type: ignore
```
shemnon, line 295: For some cases of cancun this wiill fail.  Here's what I have (I have Cancun wired in instead of plus notation).  If the account is pre-funded it will sweep out balance, but storage and nonce remains.

```suggestion
        selfdestruct_contract_address: Account(balance=0, storage={0:1}, nonce=1) if is_fork(fork=fork, which=Cancun) & prefunded_selfdestruct_address else  Account.NONEXISTENT,
```

shemnon: Other than this, my besu impl passes all test (https://github.com/hyperledger/besu/pull/5430)

shemnon: This case is fairly tricky.  Account exists prior to transaction start.  Create adds the new code to the existing code.  Because the contract creates a new contract a nonce increment occurs.  

Is this a contract create when done into an account with a balance?

- So we need semantic clarification if the phrase `contract calling SELFDESTRUCT was created` - is it created when the CREATE operation returns or is it created when ether is deposited in the account? Or are account creation and contract creation two distinct events within the spec? Either is fine, but both can be argued from the spec with no clear winner.

marioevz: Yes, I think that was the point of the tests where there's a previous balance in the address.

I think it would be a huge problem if someone could inhibit the ability to self-destruct by simply sending 1 wei before the transaction creating the contract.

What's besu's current implementation? These tests should pass if the account can self-destruct even when there was balance beforehand.

shemnon: Besu's initial implementation was to track account creation.  I just checked in code to only consider the results of CREATE/CREATE2 and the address of a transaction in create mode.  So if you self-destruct in your root-level init code there is no footprint left.

shemnon: PR to clarify the two cases I felt were unclear

https://github.com/ethereum/EIPs/pull/7308

holiman:
Great that this is coming along. I'll try to check it out in more detail, but some  of the things I'd like to ensure have coverage are: 

- pre: Account `X` has `1` wei balance. tx: sender calls `Y`, which creates `X` as a contract with code, and then calls a method on it which calls `selfdestruct(y)`. The selfdestruct should work, and `Y` should be gone. 
- Tx: Contract `X` is created with code+balance `1`. Does self-destruct to self. `X` should be gone and `1` wei ceased to exist. 
- Pre: Contract `x` exists with code + balance `1`. Tx: does self-destruct to self, which becomes a send-all. `X` should still exist, with `1` wei balance.

marioevz:
Current implementation assumes all clarifications indicated by https://github.com/ethereum/EIPs/pull/7308 are accepted.

shemnon:
Besu passes all the tests for self destruct.

One question, is there a way to flag this warning as "expected" and not output it?

```
=============================== warnings summary ===============================
tests/conftest.py:22
  /Users/dannoferrin/git/github.com/ethereum/execution-spec-tests/tests/conftest.py:22: UserWarning: Compiling Yul source with Shanghai, not Cancun.
    warnings.warn("Compiling Yul source with Shanghai, not Cancun.")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================ 148 passed, 1 deselected, 1 warning in 11.43s =================
```

danceratopz (approved):
:+1:

danceratopz:
> Besu passes all the tests for self destruct.
> 
> One question, is there a way to flag this warning as "expected" and not output it?
> 
> ```
> =============================== warnings summary ===============================
> tests/conftest.py:22
>   /Users/dannoferrin/git/github.com/ethereum/execution-spec-tests/tests/conftest.py:22: UserWarning: Compiling Yul source with Shanghai, not Cancun.
>     warnings.warn("Compiling Yul source with Shanghai, not Cancun.")
> 
> -- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
> ================ 148 passed, 1 deselected, 1 warning in 11.43s =================
> ```

@shemnon, you can suppress this warning on the command line using the following flag:
```
fill  -W "ignore::UserWarning:tests.conftest"
```
We could certainly consider removing this warning, because, as you mention, it's expected (if any tests use Yul) for forks under development (as solc typically won't add support for the fork until after it's deployed).

tehran19r (approved):

PR #124, opened 2023-05-05
src/eth_test_tools/common/types: Remove even padding on tx data field

spencer-tb:
Adds a tiny change that addresses the inconsistency between the block RLP and tx field data field.

Flagged by retesteth: http://retesteth.ethdevops.io/results/log//2023-04-27-1682591043-dretesteth.txt


marioevz:
Looks good! I think we should add a test for the formatting of the tx in the output to prevent this issue in the future.

marioevz (approved):

PR #125, opened 2023-05-06
tools: Types fixes

marioevz:
## Changes Included
- The dict type needed as input to instantiate a the Storage class is now properly defined and can be used in tests
- Forks and opcodes display proper names when printed, which will be useful when using pytest parametrize


spencer-tb (approved):
LGTM!

PR #126, opened 2023-05-11
fillers/eips/eip4844: Add Blob commitment version kzg to blob versioned hashes.

spencer-tb:
Adds the KZG Version to each blob hash in the current set of 4844 tests. 

To give an example, in the fixtures the blob hash will go from:
```
"0x0000000000000000000000000000000000000000000000000000000000000000",
```
to 
```
"0x0100000000000000000000000000000000000000000000000000000000000000",
```
appending the kzg version to the front.

Re: https://github.com/ethereum/EIPs/issues/7009

It's not the most elegant solution but when we move towards a full pytest suite the elegance can be improved!


spencer-tb:
WIP until geth evm implements invalidity for blob hashes without kzg versioning.

src/ethereum_test_tools/common/helpers.py
```diff
@@ -104,3 +104,26 @@ def to_hash(input: int | str) -> str:
     Converts an int or str into proper 32-byte hash hex string.
     """
     return "0x" + to_hash_bytes(input).hex()
+
+
+def add_kzg_version(b_hashes, kzg_version):
+    """
+    Adds the Kzg Version to each blob hash.
+    """
+    kzg_version_hex = format(kzg_version, "02x")
```
marioevz, line 113: We could do `bytes([kzg_version])` to avoid working with strings.

spencer-tb: I was wondering if we could combine it with the function you added to for the precompile tests? What do you think? So we have a general function that can be used in all the 4884 tests?

With `pytest` we could even have a `common.py`/`helpers.py` (`spec.py`)  within the `eips/eip4884/` folder for all the globally used variables and helper functions related to 4844?

spencer-tb:
Tagging the following task: https://github.com/ethereum/execution-spec-tests/issues/130

> Add explicit requirement of correct version byte for every versioned hash in blob txs

This PR will not be merged until both the spec change https://github.com/ethereum/EIPs/pull/7014 & `geth t8n-tool` change https://github.com/mdehoog/go-ethereum/pull/130 are merged.

marioevz (approved):
LGTM!

PR #134, opened 2023-05-29
keep `protected` field in transactions

gurukamath:
Keep the `protected` field in the transactions. This field might be important for signing transactions with the right signing hash.


marioevz (approved):
LGTM :)

spencer-tb:
Is this field the same field as "v" for the transaction signing? 

In this case we should maybe add that to the transaction field and omit the "protected" field. CC'ing this issue:
https://github.com/ethereum/execution-spec-tests/issues/168

PR #139, opened 2023-06-08
tests/4844: `excessDataGas` and `dataGasUsed` spec update

marioevz:
Addresses the following spec changes:
- [x] https://github.com/ethereum/EIPs/pull/6985
- [x] https://github.com/ethereum/EIPs/pull/7062
- [x] https://github.com/ethereum/EIPs/pull/7095
- [x] https://github.com/ethereum/EIPs/pull/7100
- [x] https://github.com/ethereum/EIPs/pull/7154
- [x] Fix minimum balance required calculation on blob transactions where the `max_fee_per_data_gas > get_data_gasprice(header.excess_data_gas)`

### DIFF
--- a/tests/eips/eip4844/test_blobhash_opcode_contexts.py
+++ b/tests/eips/eip4844/test_blobhash_opcode_contexts.py
@@ -16,7 +16,7 @@ from ethereum_test_tools import (
     to_hash_bytes,
 )
 
-from .utils import BlobhashContext, simple_blob_hashes
+from .utils import BlobhashContext, MAX_BLOBS_PER_BLOCK, simple_blob_hashes
 
 pytestmark = pytest.mark.valid_from("Cancun")
 
@@ -136,7 +136,7 @@ def opcode_context(yul: YulCompiler, request):
                 ),
             },
             tx_type_3.with_fields(
-                data=to_hash_bytes(0) + to_hash_bytes(3),
+                data=to_hash_bytes(0) + to_hash_bytes(MAX_BLOBS_PER_BLOCK - 1),
                 to=BlobhashContext.address("delegatecall"),
             ),
             {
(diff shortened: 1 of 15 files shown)


tests/eips/eip4844/test_blobhash_opcode_contexts.py
```diff
@@ -191,21 +191,6 @@ def opcode_context(yul: YulCompiler, request):
                 ),
             },
         )
-    elif test_case == "on_INITCODE":
```
spencer-tb, line 194: Why was this test case removed? I'm curious :D

marioevz: It was basically made impossible by https://github.com/ethereum/EIPs/pull/7100.

I think we can still add an invalidity test where verify that a block that includes a blob transaction with nil "to" is properly rejected.

tests/eips/eip4844/test_excess_data_gas.py
```diff
@@ -204,11 +147,18 @@ def tx_max_fee(  # noqa: D103
 
 
 @pytest.fixture
-def tx_data_cost(  # noqa: D103
+def tx_max_fee_per_data(  # noqa: D103
```
spencer-tb, line 150: I noticed that this might have been duplicated. There is an equivalent fixture `tx_max_fee_per_data_gas` that I think can be deleted.

To keep things consistent with `blob_txs.py` I think we should rename the `tx_max_fee_per_data` fixture to `tx_max_fee_per_data_gas`. Delete the old `tx_max_fee_per_data_gas` fixture. And rename the `tx_max_fee` fixture to `tx_max_fee_per_gas`

spencer-tb (changes requested):
Utils file is well used :D 

I will re-review after you have re-based onto Dan changes

tests/eips/eip4844/test_blobhash_opcode_contexts.py
spencer-tb: Add it here so we don't forget - https://github.com/ethereum/execution-spec-tests/issues/144

tests/eips/eip4844/test_excess_data_gas.py
marioevz: Great catch! I'll push a fix :)

marioevz:
@spencer-tb rebased and fixed the comment :+1:

spencer-tb (approved):

PR #145, opened 2023-06-13
DataGasUsed EIP-4844 Genesis Header Workaround

spencer-tb:
**Description :book:**
Provides an alternative way to generate test fixtures that originally contained the `dataGasUsed` field in the Genesis header.

**Key File Changes :key:**
- Add TestAddress/TestPrivateKey2 to `common/constants.py`.
- `test_blob_txs.py` uses intermediate block for single case.
- `test_excess_data_gas.py` uses intermediate block for all cases.
- @danceratopz added a new `conftest.py` file for common 4844 fixtures.

---

**Motivation :chart_with_upwards_trend:**
Existing client teams such as Nethermind, noted that having a non-nil value for the `dataGasUsed` field in the Genesis header, would imply that the Genesis block should have valid blobs. On mainnet this should never happen, so it did not make sense to force client teams to add this field. Similarly, the latter could result in adding an extra vulnerability for the clients.

**Solution :bulb:**
This PR addresses the issue by adding an extra intermediate block between the Genesis block (Block 0) and the test blocks (Block 1 -> N) for any test that initially contained a non-nil `dataGasUsed` field. This intermediate block replicates the original Genesis block, but with the original values for `excessDataGas` and `dataGasUsed`.

To implement this for the test cases where a non-nil `dataGasUsed` field is used, the `excessDataGas` in the Genesis block is set to `excessDataGas + TARGET_DATA_GAS_PER_BLOCK`, anticipating the reduction of `TARGET_DATA_GAS_PER_BLOCK` from the Genesis block to the intermediate block. The `dataGasUsed` of the intermediate block is determined by the number of blobs initially hard coded in the original the Genesis block.


danceratopz (commented):
Hey, I added a local conftest in 38d33a11c6a9cbf2c1f5ee0f5411437454f6eb10 - see comment below. Perhaps there are other 4844-specific fixtures we could add there?

The approach with the intermediate blocks seems reasonable, but I think I'd need some help to understand it better. Happy to discuss.

tests/eips/eip4844/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+from .utils import (
+    BLOB_COMMITMENT_VERSION_KZG,
+    TARGET_DATA_GAS_PER_BLOCK,
+    calc_excess_data_gas,
+    get_data_gasprice,
+)
+
+
+@pytest.fixture
+def block_intermediate(
```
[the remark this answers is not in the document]
spencer-tb: Big fan of this! Thanks for the addition :D

spencer-tb:
I think we can add a lot in `conftest.py` - so maybe for another PR?

tests/eips/eip4844/test_blob_txs.py
```diff
@@ -447,13 +449,19 @@ def test_invalid_tx_max_fee_per_data_gas(
...
     blocks: List[Block],
+    block_intermediate: Block,
 ):
     """
     Reject blocks with invalid blob txs due to:
 
     - tx max_fee_per_data_gas is barely not enough
     - tx max_fee_per_data_gas is zero
     """
+    if parent_blobs is not None:
```
marioevz, line 460: `parent_blobs` needs to be directly added as a parameter to this function, otherwise we are referring to the fixture function directly, which will never be `None`.

tests/eips/eip4844/test_excess_data_gas.py
```diff
@@ -507,8 +513,8 @@ def test_invalid_static_excess_data_gas(
 
     Test is parametrized to `MAX_BLOBS_PER_BLOCK` and `TARGET_BLOBS_PER_BLOCK`.
     """
-    blocks[0].rlp_modifier = Header(excess_data_gas=parent_excess_data_gas)
-    blocks[0].exception = "invalid excessDataGas"
+    blocks[1].rlp_modifier = Header(excess_data_gas=parent_excess_data_gas)
+    blocks[1].exception = "invalid excessDataGas"
```
marioevz, line 517:
```suggestion
    blocks[-1].rlp_modifier = Header(excess_data_gas=parent_excess_data_gas)
    blocks[-1].exception = "invalid excessDataGas"
```
To always use the last block.

tests/eips/eip4844/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+    parent_blobs: int,
```
marioevz, line 25: We could potentially check this value for zero:
- if it's greater than zero, return the block
- if it's zero, return `None`

We can then conditionally append the block to the blocks list, and make tests shorter (1 block tests for those that do not rely on the parent having blobs).

tests/eips/eip4844/test_excess_data_gas.py
```diff
@@ -124,12 +128,10 @@ def block_base_fee() -> int:  # noqa: D103
 @pytest.fixture
 def env(  # noqa: D103
     parent_excess_data_gas: int,
-    parent_blobs: int,
     block_base_fee: int,
 ) -> Environment:
     return Environment(
-        excess_data_gas=parent_excess_data_gas,
-        data_gas_used=parent_blobs * DATA_GAS_PER_BLOB,
+        excess_data_gas=parent_excess_data_gas + TARGET_DATA_GAS_PER_BLOCK,
```
marioevz, line 134: Directly related to the comment on `common.py`, we can check if the `parent_blobs` parameter:
- If zero, `excess_data_gas=parent_excess_data_gas`
- else, `excess_data_gas=parent_excess_data_gas + TARGET_DATA_GAS_PER_BLOCK`

tests/eips/eip4844/test_excess_data_gas.py
```diff
@@ -233,28 +236,31 @@ def blocks(  # noqa: D103
     tx: Transaction,
     header_excess_data_gas: Optional[int],
     header_data_gas_used: Optional[int],
+    block_intermediate: Block,
```
marioevz, line 239: Directly related to the comment on `common.py`, we can check if the `block_intermediate` is `None`, and only add it to the start of the block list if it contains a value.

marioevz (commented):
Looks good! Just a couple of comments :)

tests/eips/eip4844/test_blob_txs.py
spencer-tb: Nice catch added here: https://github.com/ethereum/execution-spec-tests/pull/145/commits/5f05c1016fd9b2bd76a8c197f9257457f12e75c9

tests/eips/eip4844/conftest.py
spencer-tb: I think we do this on line 69:
```python
        if parent_blobs != 0
        else [],
```
If parent blobs is 0 in this case we return []

spencer-tb: I've changed the above essentially:
```python
if parent_blobs != 0
        else None,
 ```
So then used this None logic in your last comment!!

File: tests/eips/eip4844/test_blob_txs.py [hunk 7]
```diff
@@ -447,13 +449,20 @@ def test_invalid_tx_max_fee_per_data_gas(
...
+    parent_blobs: int,
+    block_intermediate: Block,
 ):
     """
     Reject blocks with invalid blob txs due to:
 
     - tx max_fee_per_data_gas is barely not enough
     - tx max_fee_per_data_gas is zero
     """
+    if parent_blobs is not None:
```
marioevz, line 461: We could reduce it to:
```suggestion
    if parent_blobs:
```
to check both `not None` and `> 0`.

tests/eips/eip4844/conftest.py
```diff
@@ -0,0 +1,73 @@
...
+    parent_excess_data_gas: int,
+    tx_max_fee_per_gas: int,
+):
```
marioevz, line 28:
```suggestion
) -> Block | None:
```

marioevz (approved):

PR #147, opened 2023-06-14
feature: specify the target evm version to solc for yul compilation

### LINKED ISSUE #135: evm version not passed to solc when compiling yul
Currently, several fillers fail (for forks older than Shanghai) if ran with solc 0.8.20, which added Shanghai support.

This is due to the resulting bytecode containing an opcode that isn't recognized by the `evm` tool for older forks. For example, the bytecode from `fillers/example/example_yul.py` compiled with solc 0.8.20 contains PUSH0 which is not recognized by forks prior to Shanghai. 

The contract code specified via the `Yul` helper class is compiled with solc using the flags `--assemble -`  (the yul code is provided on stdin). This can be fixed by the evm version to compile via `--evm-version=FORK`:
```
  --evm-version version (=shanghai)
                       Select desired EVM version. Either homestead, 
                       tangerineWhistle, spuriousDragon, byzantium, 
                       constantinople, petersburg, istanbul, berlin, london, 
                       paris or shanghai.
```
This could be added as a parameter to Yul's `assemble()` method, but this call is tucked away. Might need a bit of clean-up to propagate the fork value to where it is needed.

To check: Whether some fillers call assemble() directly.

danceratopz:
This PR:
- [x] Adds an optional keyword argument to the Yul constructor that allows the user to specify the fork for which the code should be compiled for. This fixes #135.
- [x] Adds a `--solc-bin` command-line flag analogous to `--evm-bin`.
- [x] ~~Checks whether any forks (configured the command-line) don't supported by the configured solc binary.~~ This may be done in the forks plugin, but perhaps undesirable as it would potentially force users to use a development version of `solc` to run `fill` with dev forks.

Note, after changing the behavior so that Yul gets compiled with the test's current fork parameter value: `tests/security/test_selfdestruct_balance_bug.py` fails if compiled with Constantinople or Shanghai. Shanghai can be explained: This is just because `solc` refuses to compile contracts with SELFDESTRUCT. I didn't understand why the code doesn't compile for Constantinople yet. We don't need to fix this in the scope of this PR, imo, but we should make a ticket to not forget. This is currently avoided by applying `@pytest.mark.compile_yul_with("Merge")`.

### DIFF
--- a/src/ethereum_test_tools/code/yul.py
+++ b/src/ethereum_test_tools/code/yul.py
@@ -4,16 +4,34 @@ Yul frontend
 
 from pathlib import Path
 from subprocess import PIPE, run
-from typing import Optional
+from typing import Mapping, Optional, Tuple, Type
+
+from ethereum_test_forks import Fork
 
 from .code import Code
 
 SOLC: Path = Path("solc")
-SOLC_ARGS = (
-    SOLC,
-    "--assemble",
-    "-",
-)
+DEFAULT_SOLC_ARGS = ("--assemble", "-")
+
+
+def get_evm_version_from_fork(fork: Fork | None):
+    """
+    Get the solc evm version corresponding to `fork`.
+
+    Args
+    ----
+        fork (Fork): The fork to retrieve the corresponding evm version for.
+
+    Returns
+    -------
+        str: The name of evm version as required by solc's --evm-version.
+    """
+    if not fork:
+        return None
+    fork_to_evm_version_map: Mapping[str, str] = {"Merge": "paris"}
+    if fork.name() in fork_to_evm_version_map:
+        return fork_to_evm_version_map[fork.name()]
+    return fork.name().lower()
 
 
 class Yul(Code):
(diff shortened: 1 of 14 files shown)


src/ethereum_test_tools/tests/test_code.py
```diff
@@ -96,7 +98,8 @@ def test_yul():
         expected_bytecode += bytes.fromhex("80")
         expected_bytecode += bytes.fromhex("55")
 
-    assert Yul(long_code).assemble() == expected_bytecode
+    # TODO(dan): workaround until it's understood why Shanghai takes so long to compile
+    assert Yul(long_code, fork=Merge).assemble() == expected_bytecode
```
marioevz, line 102: Maybe we found an issue in the solc ?

marioevz (approved):
Looks good! :)

Only two minor comments that would not affect the tests.

src/ethereum_test_tools/code/yul.py
```diff
@@ -4,16 +4,34 @@
...
+    ----
+        fork (Fork): The fork to retrieve the corresponding evm version for.
+
+    Returns
+    -------
+        str: The name of evm version as required by solc's --evm-version.
+    """
+    if not fork:
+        return None
+    fork_to_evm_version_map: Mapping[str, str] = {"Merge": "paris"}
```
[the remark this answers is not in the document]
danceratopz: Happy to do that (in a separate PR) :) we'd have to patch it on geth's evm tool though, because in the help it's known as "Merge" :upside_down_face: :

```
evm t8n --help
```

grafik

src/ethereum_test_tools/tests/test_code.py
danceratopz: Thanks for bringing this up. Could be, need to look into this! Made a ticket so not to forget #148.

PR #152, opened 2023-06-16
tools: fix solc compilation for the ConstantinopleFix fork

danceratopz:
We have to additionally map the `ConstantinopleFix` fork to "constantinople" when setting the target evm version for solc (`solc --evm-version FORK`.

Additionally, update the yul example to run from Homestead. Frontier is not supported by solc.


marioevz (approved):
LGTM!

In theory we could make pass the fork fallback to the next fork supported by solc, e.g. Frontier -> Homestead. But I really think it's not worth the complexity.

PR #153, opened 2023-06-19
github: update golang version

marioevz:
Updates the golang version required in the github actions file, so the `evm` tool compilation succeeds.


danceratopz:
LGTM! 🙂

PR #155, opened 2023-06-20
Pytest plugins: Implement EIP reference version checking as dynamic tests

danceratopz:
This PR:
- Gets the `spec_version_checker` plugin ready for the test folder restructure.
- Implements the version checks as tests instead of issuing warnings. A new test is dynamically added by the plugin for each test module whose path matches `eip\d{1,4}` that checks that the reference spec defined in the module is up-to-date with ethereum/EIPs.

These tests as-is are currently always ran. We might have to be careful that our (especially our CI/CD) Github API access doesn't get rate limited :slightly_smiling_face: 

Only run the EIP version check tests:
```console
pytest -m eip_version_check
```
Run all tests except the version check tests:
```console
pytest -m "not eip_version_check"
```

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -7,3 +7,4 @@ addopts =
     -p pytest_plugins.test_filler.test_filler
     -p pytest_plugins.forks.forks
     -p pytest_plugins.spec_version_checker.spec_version_checker
+    -m "not eip_version_check"
--- a/src/ethereum_test_tools/reference_spec/git_reference_spec.py
+++ b/src/ethereum_test_tools/reference_spec/git_reference_spec.py
@@ -44,11 +44,17 @@ class GitReferenceSpec(ReferenceSpec):
         """
         return self.SpecVersion
 
-    def _get_latest_known_spec(self) -> Dict | None:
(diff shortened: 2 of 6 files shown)


marioevz (approved):
LGTM!

I'm a bit worried about the rate limiting we could incur when enabling this tests, should we disable them by default if they cause too many issues during the rename PR?

danceratopz:
> LGTM!
> 
> I'm a bit worried about the rate limiting we could incur when enabling this tests, should we disable them by default if they cause too many issues during the rename PR?

Sounds good to me! I'll disable for tox and ci/cd via pytest.ini.

Longer term we can perform conditional requests:
https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#conditional-requests
and potentially get credentials for the Github API which would increase the rate limit:
https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#increasing-the-unauthenticated-rate-limit-for-oauth-apps
But that's a bit tight for our current schedule!

PR #156, opened 2023-06-21
tests/docs: reorganize test module locations by fork

danceratopz:
Implements @spencer-tb's idea to organize the test modules in the `tests/` folder by fork and updates the doc flow to handle this. 

It additionally adds two improvements to the Test Case Reference section:
1. Adds a page under each test module which lists the tests cases as generated with `fill --collect-only -q`. This appears in the site nav panel as "Test Cases".
2. Links each test path to its corresponding Github page for the version at which the doc was generated.

Preview available at: https://danceratopz.github.io/execution-spec-tests

Example of 1:
grafik

Example of 2:
grafik


marioevz (approved):
Looks amazing!!! Thanks for this :)

PR #157, opened 2023-06-22
tests: Add EIP-5656, MCOPY, Tests

marioevz:
Adds tests for the MCOPY opcode.

### DIFF
--- a/tests/cancun/eip5656_mcopy/test_mcopy.py
+++ b/tests/cancun/eip5656_mcopy/test_mcopy.py
@@ -57,7 +57,7 @@ def bytecode_storage(
 
     # Fill memory with initial values
     for i in range(0, len(initial_memory), 0x20):
-        bytecode += Op.MSTORE(i, initial_memory[i : i + 0x20])
+        bytecode += Op.MSTORE(i, Op.PUSH32(initial_memory[i : i + 0x20]))
 
     # Perform the MCOPY according to calldata values
     bytecode += Op.MCOPY(
@@ -76,7 +76,9 @@ def bytecode_storage(
     if len(final_memory) > len(initial_memory):
         last_word = ceiling_division(len(final_memory), 0x20) - 1
         bytecode += Op.SSTORE(last_word, Op.MLOAD(last_word * 0x20))
-        storage[last_word] = final_memory[last_word * 0x20 : (last_word + 1) * 0x20]
+        storage[last_word] = final_memory[last_word * 0x20 : (last_word + 1) * 0x20].ljust(
+            32, b"\x00"
+        )
 
     # We could also keccak the memory and store the hash in storage for good measure ?
 
@@ -127,6 +129,9 @@ def post(
         (0x00, 0x10, 0x40),
         (0x0F, 0x10, 0x40),
         (0x100, 0x01, 0x01),
+        (0x100, 0x01, 0x20),
+        (0x100, 0x01, 0x1F),
+        (0x100, 0x01, 0x21),
         (0x00, 0x00, 0x100),
         (0x100, 0x00, 0x100),
         (0x200, 0x00, 0x100),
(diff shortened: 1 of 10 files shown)


marioevz:
This is ready for review but we should merge #142 first.

tests/cancun/eip5656_mcopy/test_mcopy_memory_expansion.py
```diff
@@ -0,0 +1,286 @@
+"""
+abstract: Tests [EIP-5656: MCOPY - Memory copying instruction](https://eips.ethereum.org/EIPS/eip-5656)
+
+    Test copy operations of [EIP-5656: MCOPY - Memory copying instruction](https://eips.ethereum.org/EIPS/eip-5656)
```
spencer-tb, line 4: Maybe we should have a unique title docstring here for the docs

tests/cancun/eip5656_mcopy/test_mcopy.py
```diff
@@ -0,0 +1,218 @@
...
+        "single_word_plus_one_byte_memory_extension",
+        "full_memory_rewrite",
+        "full_memory_copy",
+        "full_memory_copy_offset",
+        "full_memory_clean",
+        "out_of_bounds_memory_extension",
+    ],
+)
+@pytest.mark.valid_from("Cancun")
+def test_valid_mcopy_operations(
```
spencer-tb, line 177: We could maybe add a separate test case similar to this for invalid parameters - `test_mcopy_invalid_params` only if we are able to generate them or if you think they are useful:
```python
(2**256, 2**256, 0x00),  # src_and_dest_out_of_bounds
(2**256, 0x00, 0x00),  # src_out_of_bounds
...
(0x00, 0x00, -0x01),  # negative_length
```

spencer-tb (approved):
LGTM! Only suggestions below - happy to merge :D

tests/cancun/eip5656_mcopy/test_mcopy.py
```diff
@@ -0,0 +1,218 @@
...
+    to_address,
+    to_hash_bytes,
+)
+
+from .common import REFERENCE_SPEC_GIT_PATH, REFERENCE_SPEC_VERSION, mcopy
+
+code_address = to_address(0x100)
+"""
+Code address used to call the test bytecode on every test case.
+"""
```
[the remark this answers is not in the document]
marioevz: I just made it into a comment because there was nothing really to explain what was happening with these variables.

tests/cancun/eip5656_mcopy/test_mcopy.py
```diff
@@ -0,0 +1,218 @@
...
+    # If the memory was extended beyond the initial range, store the last word of the resulting
+    # memory into storage too
+    if len(final_memory) > len(initial_memory):
+        last_word = ceiling_division(len(final_memory), 0x20) - 1
+        bytecode += Op.SSTORE(last_word, Op.MLOAD(last_word * 0x20))
+        storage[last_word] = final_memory[last_word * 0x20 : (last_word + 1) * 0x20].ljust(
+            32, b"\x00"
+        )
+
+    # We could also keccak the memory and store the hash in storage for good measure ?
```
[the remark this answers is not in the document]
marioevz: Done! I implemented the check, thanks!

tests/cancun/eip5656_mcopy/test_mcopy.py
marioevz: I don't think there's a way to cover the first two since the EVM stack only reaches `2**256 - 1`, so even using these as inputs for the transaction data, by the time it gets converted into the EVM, we will lose the range.

The last one is covered by `test_mcopy_huge_memory_expansion` (`-1` is equivalent to `2**256 - 1`)

tests/cancun/eip5656_mcopy/test_mcopy_memory_expansion.py
marioevz: Done!

PR #158, opened 2023-06-22
pytest: modify --help to only print option help from ./src/pytest_plugins

danceratopz:
"fill" is defined as an entry point in execution-spec-tests and is a wrapper for pytest.

This PR modifies the behavior of "fill --help":
- 'fill --help' only prints the options defined by ./src/pytest_plugins/.
- 'fill --help -v' prints the full help (equivalent to 'pytest --help')

The approach is brittle, but pragmatic. There isn't a direct way to trace back a command-line option to the plugin that registered it because when a plugin registers a command-line option, it doesn't tag that option with its own name or ID.

There's an open feature request for this:
https://github.com/pytest-dev/pytest/issues/9452

Now:
grafik

### DIFF
--- a/src/entry_points/fill.py
+++ b/src/entry_points/fill.py
@@ -1,14 +1,56 @@
 """
 Define an entry point wrapper for pytest.
+
+Essentially an alias for pytest but modifies the behavior of --help:
+- 'fill --help' only prints the options defined by ./src/pytest_plugins/.
+- 'fill --help -v' prints the full help (equivalent to 'pytest --help')
+
+The approach is brittle, but pragmatic. There isn't a direct way to trace
+back a command-line option to the plugin that registered it because when
+a plugin registers a command-line option, it doesn't tag that option with
+its own name or ID.
+
+There's an open feature request for this:
+https://github.com/pytest-dev/pytest/issues/9452
 """
 
+import subprocess
 import sys
 
 import pytest
 
 
 def main():  # noqa: D103
-    pytest.main(sys.argv[1:])
+    if "--help" in sys.argv and "-v" not in sys.argv:
+        sys.argv.remove("--help")
+
+        result = subprocess.run(["pytest", "--help"], capture_output=True, text=True)
+        output = result.stdout
+
+        start_string = "Arguments defining evm executable behavior:"
+        end_string = "distributed and subprocess testing:"
+
+        start = output.find(start_string)
+        end = output.find(end_string)
+
+        if start != -1 and end != -1:
+            required_help_text = output[start : end - 2]
+            print(
+                "The 'fill' command is an alias for 'pytest'. Run 'fill --help -v' to\n"
+                "see the complete pytest cli help.\n\n"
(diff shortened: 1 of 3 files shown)


marioevz:
It already broke for me :(

I'm thinking we could somehow mark all `parser.getgroup` with a keyword and use it to get the options during the parsing of the help?

```python
fork_group = parser.getgroup("Forks", "[Filling] Specify the fork range to generate fixtures for")
```

danceratopz:
> It already broke for me :(
Silly question, but did you run:
```
pip install --force-reinstall -e .
```
to re-install the endpoint?

> I'm thinking we could somehow mark all `parser.getgroup` with a keyword and use it to get the options during the parsing of the help?
> 
> ```python
> fork_group = parser.getgroup("Forks", "[Filling] Specify the fork range to generate fixtures for")
> ```

I did consider this, but I didn't try. Might work in combination with another plugin whose hooks get loaded last to retrieve all the other hook's option groups. But if the approach in this PR works for you, I'd probably go with this simplicity.

src/entry_points/fill.py
```diff
@@ -1,14 +1,56 @@
...
 def main():  # noqa: D103
-    pytest.main(sys.argv[1:])
+    if "--help" in sys.argv and "-v" not in sys.argv:
+        sys.argv.remove("--help")
+
+        result = subprocess.run(["pytest", "--help"], capture_output=True, text=True)
+        output = result.stdout
+
+        start_string = "Arguments defining evm executable behavior:"
+        end_string = "distributed and subprocess testing:"
```
spencer-tb, line 31:
```suggestion
        end_string = "Fill tests until and including the specified fork."
```

src/entry_points/fill.py
```diff
@@ -1,14 +1,56 @@
...
+            )
+            print(required_help_text)
+        else:
+            print(
+                "Unable to extract required help text. Run `fill --help` instead and "
+                "go and shout at the maintainers."
+            )
+        return
+    else:
+        sys.argv.remove("-v")
```
spencer-tb, line 52:
```suggestion
```
I get the following error with this line when running only with `fill`:

```
 line 52, in main sys.argv.remove("-v")
ValueError: list.remove(x): x not in list
```

spencer-tb (commented):
Nice one! I had to make the following changes to get it work for me, even after:
`pip install --force-reinstall -e .`

danceratopz:
> It already broke for me :(
> 
> I'm thinking we could somehow mark all `parser.getgroup` with a keyword and use it to get the options during the parsing of the help?
> 
> ```python
> fork_group = parser.getgroup("Forks", "[Filling] Specify the fork range to generate fixtures for")
> ```

Let's do it properly :laughing: - implemented this approach in https://github.com/ethereum/execution-spec-tests/pull/172

src/entry_points/fill.py
danceratopz: Thanks for the feedback! Helped me decide to go for a more robust approach, see https://github.com/ethereum/execution-spec-tests/pull/172

PR #163, opened 2023-06-22
all: Sign and Serialize Transactions Internally

marioevz:
Introduce methods to be able to sign and serialize (to rlp only at the moment) transactions of all types.


src/ethereum_test_tools/common/types.py
```diff
@@ -785,6 +872,316 @@ def with_fields(self, **kwargs) -> "Transaction":
...
+                        Uint(self.value),
+                        code_to_bytes(self.data),
+                        [a.to_list() for a in self.access_list]
+                        if self.access_list is not None
+                        else [],
+                        Uint(self.max_fee_per_data_gas),
+                        [hash_to_bytes(h) for h in self.blob_versioned_hashes],
+                        Uint(self.signature[0]),
+                        Uint(self.signature[1]),
+                        Uint(self.signature[2]),
```
spencer-tb, line 926: We could create a Payload class that we use throughout, with optional fields depending on the transaction type! 

Very happy with how it is now though, its very read-able :)

spencer-tb (commented):
This is amazing!! :fire: 

I never thought this would have been possible - I just filled all the tests, with your other PRs merged (##164 & #167)  including `test_blob_txs_full.py` just now :D 

I noticed the block RLP was very large in the new blob test... :eyes: This is how big it is with one blob using the network block type?

marioevz:
> I noticed the block RLP was very large in the new blob test... eyes This is how big it is with one blob using the network block type?

Yes! That block contains a full blob, which is 4096 times 32 bytes, but I think this is necessary to verify the behavior that brought down the previous devnet.

src/ethereum_test_tools/common/types.py
marioevz: Definitely some room for improvement, yes I like the idea of separating this into its own object.

I think in the future this call should somehow be fork-aware, because some of the transaction types will be updated to SSZ after a given fork.

marioevz:
Superseded by https://github.com/ethereum/execution-spec-tests/pull/167

PR #164, opened 2023-06-23
tests/4844: Add blob network type in block test

marioevz:
Adds tests where one or more blobs in one or more transactions contained in the serialized version of the block is in its full blob (network) version.

Adds three tests:
- A block with a single transaction with a single blob in full blob form
- A block with multiple transactions with the first transaction having full blob form
- A block with multiple transactions with the last transaction having full blob form


spencer-tb (approved):
LGTM! Just needs the docstring added to L128 for the blob class in `common.py` for tox

marioevz:
Thanks for the review @spencer-tb ! I will fix the docstring issue and merge :)

PR #165, opened 2023-06-23
fix: update requests

danceratopz:
Update requests in setup.cfg and remove read the docs requirements file as it's not needed and also contains an out-dated requests version.


marioevz (approved):
LGTM

PR #167, opened 2023-06-24
all: Completely remove external block building tool, build blocks internally, sign transactions internally

marioevz:
Completely removes all references to `b11r` tool and builds the blocks internally.

Also all transaction types are now signed internally.


src/ethereum_test_tools/common/types.py
```diff
@@ -1115,38 +1503,31 @@ def default(self, obj):
                 "gas": hex(obj.gas_limit),
                 "value": hex(obj.value),
                 "input": code_to_hex(obj.data),
-                "to": "0x" + int.to_bytes(obj.to, length=20, byteorder="big").hex()
-                if obj.to is int
-                else obj.to,
+                "to": "0x" + address_to_bytes(obj.to).hex() if obj.to is not None else None,
```
spencer-tb, line 1506: I'm assuming that the `address_to_hex(...)` function uses `hex_or_none(...)` within it to return None for the cases where `obj.to` is None here
```suggestion
                "to": address_to_hex(obj.to),
```

src/ethereum_test_tools/common/conversions.py
```diff
@@ -0,0 +1,174 @@
+"""
+Common conversion methods.
+"""
```
spencer-tb, line 3: Really like these in a separate file!! I think at some point we should consolidate on these conversions, and maybe create class types for `Address` & `Hash` to make conversions easier

I added some suggested changes/ideas at the end of `types.py` but I think in the long term we should try to make these as concise as possible or change some of the naming to avoid errors - i.e `hex_or_none` returns a hex string in one case so we could call it `hex_string_or_none` etc

src/ethereum_test_tools/common/types.py
```diff
@@ -1115,38 +1503,31 @@ def default(self, obj):
...
-                        hashes.append(h)
-                    elif type(h) is bytes:
-                        if len(h) != 32:
-                            raise TypeError("improper byte size for blob_versioned_hashes")
-                        hashes.append("0x" + h.hex())
-                    else:
-                        raise TypeError("improper type for blob_versioned_hashes")
-                tx["blobVersionedHashes"] = hashes
+                tx["blobVersionedHashes"] = [
+                    "0x" + hash_to_bytes(h).hex() for h in obj.blob_versioned_hashes
```
spencer-tb, line 1515: We could also have a hash_to_hex(...) function in conversions.py, but I think we could do refactor for these in the future

```suggestion
                    hash_to_hex(h) for h in obj.blob_versioned_hashes
```

spencer-tb (approved):
Super impressed with this and glad to be rid of `bllr` - happy to merge as is :100: :rocket: , all comments relate to how we can improve and refactor `types.py` in the future - I think we should move the json encoder to separate file 

I'd read my comments from the bottom upwards :D

src/ethereum_test_tools/common/types.py
```diff
@@ -785,6 +796,324 @@ def with_fields(self, **kwargs) -> "Transaction":
...
+                        Uint(self.max_fee_per_data_gas),
+                        [hash_to_bytes(h) for h in self.blob_versioned_hashes],
+                        Uint(self.signature[0]),
+                        Uint(self.signature[1]),
+                        Uint(self.signature[2]),
+                    ],
+                    self.blobs,
+                    self.blob_kzg_commitments,
+                    self.blob_kzg_proofs,
+                ]
```
spencer-tb, line 855: We could create separate class types for each Transaction at some point in the future, i.e TransactionType0, ..., TransactionType3 that are members of Transaction to avoid duplication

This could be useful when writing tests later down the line to improve readability, maintenance, etc

src/ethereum_test_tools/common/types.py
```diff
@@ -1304,11 +1685,11 @@ def key_value_padding(value: str) -> str:
     """
     Adds even padding to a dictionary key or value string.
     """
```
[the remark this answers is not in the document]
marioevz: We are converting the format ;)

src/ethereum_test_tools/common/types.py
```diff
@@ -1115,38 +1503,31 @@ def default(self, obj):
...
+                tx["v"] = hex(obj.signature[0])
+                tx["r"] = hex(obj.signature[1])
+                tx["s"] = hex(obj.signature[2])
             else:
                 tx["v"] = ""
                 tx["r"] = ""
                 tx["s"] = ""
+
+            if obj.sender is not None:
+                tx["sender"] = "0x" + address_to_bytes(obj.sender).hex()
```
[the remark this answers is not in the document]
marioevz: The problem is that `address_to_bytes` returns empty `bytes()` if a `None` is passed, and `hex_or_none` would return `"0x"`. 

I can modify `address_string` to return a default in case of `None`.

src/ethereum_test_tools/common/types.py
marioevz: Ah we have `hash_string`. We really need a better way to organize these.

src/ethereum_test_tools/common/types.py
marioevz: Fixed :)

src/ethereum_test_tools/common/types.py
marioevz: Definitely, I thought of it but I could not figure out a way to do it in a clean manner.
It would be good to have a follow up PR focused only in this part.

marioevz: Created #181

src/ethereum_test_tools/common/conversions.py
marioevz: Created #182

danceratopz:
Btw, this gives almost a 2x speed-up on my laptop, running `fill` (until Shanghai):
* d34385b2b150af8e802d9cbef9fe1cac165db1b1: ~17s
* 86d8a818a229e32e33bf388de89a12af891a34e6: ~9s

PR #169, opened 2023-06-26
chore: move xdist to the main install requires

marioevz:
This PR proposes to move `xdist` to the main requires list in the setup.

Reasoning being that the `--dist` option is now included by default in the `pystest.ini` and usually users (in my opinion) would like to do:
```bash
pip install -e .
```
and then run `fill` command, which is no longer possible.

Granted that the `README.md` correctly states that the correct install command is `pip install -e .[docs,lint,test]`, but even so, I think all the tools necessary to fill the tests should be installed with `pip install .`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,7 @@ install_requires =
     requests>=2.31.0
     colorlog>=6.7.0
     pytest==7.3.2
+    pytest-xdist>=3.3.1,<4
 
 [options.package_data]
 ethereum_test_tools =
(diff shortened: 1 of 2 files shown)


spencer-tb (approved):
LGTM :rocket:

danceratopz:
:100: Thanks for fixing this! I additionally cleaned up the install tox `docs` environment by removing `test` extras.. As we shouldn't have to include the packages defined in the `test` extras for this. 

We could consider deleting the `test` extras section in the future (currently only contains pytest-cov, which I occasionally use locally).

danceratopz (approved):

PR #174, opened 2023-06-27
pytest: compile Yul with --target-version shanghai when fork is cancun

danceratopz:
This fixes test fixture generation when the fork fixture's parameter value is Cancun. The root cause is that we pass the fork parameter value to solc's `--target-version` option and, clearly, the last stable version of solc doesn't offer Cancun support :-)

Additionally, warnings are issued in the pytest summary:
-  If verbosity <= 1 (no verbosity flag or `-v` ): "UserWarning: Compiling Yul source with Shanghai, not Cancun.".
-  If verbosity is >=2, then each warning also mentions the test node id.

All Yul tests can be executed using:
```
fill --until Cancun -m yul_test
```


marioevz (approved):
LGTM! Thanks for this :)

PR #177, opened 2023-06-28
genesis reward is -1

winsvega:


marioevz:
Just to add the Geth code as reference to this behavior:
https://github.com/ethereum/go-ethereum/blob/8bbb16b70ef4a9e732745aac104525a6f16306d6/cmd/evm/internal/t8ntool/execution.go#L239

And it does not affect forks after Homestead.

Hey @gurukamath, I'm not sure if this would affect how you calculate the results in your version of t8n.

marioevz (approved):
LGTM!

PR #183, opened 2023-06-30
feature: Add engine API execution payload information in fixtures

marioevz:
Allows tests to specify the metadata that describes how each block in a blockchain test should be delivered through the Engine API, which is used in hive testing.

The forks now also have information about the versions of the Engine API endpoints that should be used with all the payloads generated for them.

Currently only `engine_newPayloadVX` is supported.


src/ethereum_test_tools/common/types.py
```diff
@@ -1116,6 +1116,22 @@ def serialize_transactions(input_txs: List[Transaction] | None) -> bytes:
...
+    """
+    Gets a list of ordered blob versioned hashes from a list of transactions.
+    """
+    versioned_hashes: List[bytes] = []
+
+    if input_txs is None:
+        return versioned_hashes
+
+    for tx in input_txs:
+        if tx.blob_versioned_hashes is not None:
```
spencer-tb, line 1129: I think we need to include the tx type check, maybe we could add it here:
```suggestion
        if tx.blob_versioned_hashes is not None and tx.type >= 3:
```

spencer-tb (approved):
Thanks for adding this! I was able to remove so many lines in the `pyspec` simulator as a result :100: 

I think we could add a flag for this in the future. If at some point by default we generate the original StateTest format when running `fill` in the future we could add the `--hive` or `--pyspec` flag for generating these fixtures with the payloads

src/ethereum_test_tools/common/types.py
marioevz: Thanks! I modified the field name from `type` to `ty` and used `== 3` instead of `>= 3` because we don't know yet if all new types of transactions could contain blobs.

PR #186, opened 2023-07-04
docs: add markdown linting and spellcheck

danceratopz:
This PR:
1. Adds markdown spellchecking and linting to tox and Github Actions.
2. Fixes existing linting problems and spelling in `docs/**/*.md` & `README.md`.

**Motivation**

We now have CD of our docs from the main branch and we should try and maintain a certainly level of quality:
https://ethereum.github.io/execution-spec-tests/main

**External Tooling Required**

- **spelling:** Uses the [pyspelling](https://github.com/facelessuser/pyspelling) Python cli tool which requires the external aspell command. This is a standard package and can be installed on Debian-based distros via:
    ```
    sudo apt-get install aspell aspell-en
    ```
- **lint:** Uses the [markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2) command-line utility. This is node.js program, so introduces quite a hefty dependency. If you don't already use node.js (and require multiple versions), then you can install markdownlint with:
    ```
    sudo apt install nodejs
    sudo npm install markdownlint-cli2 --global
    ```
    Experienced users would probably use `nvm` to manage and install different node versions, see https://www.freecodecamp.org/news/node-version-manager-nvm-install-guide/.

**Tooling and Integration with VS Code:**

- Markdown spellchecking is already covered in VS Code via the [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) that we already recommend in our `extensions.json`. 

- Dictionaries: Now "whitelist.txt" is split into 3 separate wordlist files for maintainability. One reason: If you do "Quick Fix" -> "Add word to dictionary 'project-words'", VS Code sorts the wordlist alphabetically, which is not so optimal for our list of opcodes. Pyspelling supports multiple wordlists. flake8-spellcheck works only supports one wordlist file, which must be called `whitelist.txt`. This is generated using the `src/entry_points/create_whitelist.py` in tox prior to running flake8.

- Btw, I did check whether we could spellcheck the markdown via the flake8-spellcheck plugin for simplicity, but it was not possible.

- I used the [vscode-markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) for a couple of weeks - it works well. This extension uses the `markdownlint-cli2` and there's a corresponding Github Action that uses this tool.

If you use these two plugins and clean-up any errors, then tox should pass.

**Soft Fail Mode Rational**

It's not ideal to require external (non-native Python) tooling and these tools are only required for people who work heavily on the docs. As such, if these tools are not available, tox prints a hint that they're missing, but will still pass. Most devs will only be working on the tests and should not break the docs or need to change them (docstrings in tests will get spell-checked from flake8; linting the output is more difficult, perhaps there's a flake8 plugin extension?). 

**Keeping tox cross-platform** (hopefully)

Additionally, the tools are executed via Python scripts in `src/entry_points` to avoid using external commands (e.g. bash) in tox to ensure it still runs cross-platform.

**Github Actions**

These checks will always be ran in Github actions:
- Spellcheck is ran as part of tox (it's easy to install the aspell dependencies).
- Lint is ran using [DavidAnson/markdownlint-cli2-action@v11](https://github.com/marketplace/actions/markdownlint-cli2-action) to avoid dealing with node.js dependencies in the workflow.


danceratopz:
Would be good to get some feedback, and if you like it, I'll add some documentation before merge.

src/entry_points/markdownlintcli2_soft_fail.py
```diff
@@ -0,0 +1,34 @@
...
+
+def main():
+    """
+    Run markdownlint-cli2 as a subprocess if it's installed, if not fail silently.
+    """
+    markdownlint = shutil.which("markdownlint-cli2")
+    if not markdownlint:
+        # Note: There's an additional step in test.yaml to run markdownlint-cli2 in github actions
+        print("*********  Install 'markdownlint-cli2' to enable markdown linting *********")
+        sys.exit(0)
```
marioevz, line 27: Sorry if I missed something obvious here but, don't we need `GITHUB_ACTIONS` check to fail here too?

marioevz (commented):
Looks great, I think this will certainly help us maintain the quality of the docs :)

src/entry_points/markdownlintcli2_soft_fail.py
danceratopz: Ooh, this is a little inconsistent, Markdown lint is handled separately by the [markdownlint-cli2-action](https://github.com/marketplace/actions/markdownlint-cli2-action) in Github Actions:
https://github.com/ethereum/execution-spec-tests/blob/a77f3939df6f676abf0a09d96a3dc6e6aadbbc36/.github/workflows/test.yaml#L49

danceratopz:
Rebase due to:
- #193 
- various wordlist/whitelist.txt additions.

spencer-tb (approved):
LGTM! I really like the dictionaries - I feel like we are bulletproof now :D

PR #187, opened 2023-07-04
do not export protected field into filled tests as vrs is already exp…

winsvega:
…orted

### DIFF
--- a/src/ethereum_test_tools/common/types.py
+++ b/src/ethereum_test_tools/common/types.py
@@ -1506,7 +1506,6 @@ class JSONEncoder(json.JSONEncoder):
                 "input": code_to_hex(obj.data),
                 "to": address_or_none(obj.to),
                 "accessList": obj.access_list,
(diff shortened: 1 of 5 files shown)


marioevz:
Hey @gurukamath, since #167 we are signing and including the `(v, r, s)` signature values for all transactions before we call `t8n`, so the removal of `protected` field makes sense since the `t8n` will not be signing any transactions from us anymore.

Let me know if that sounds ok for you! :)

marioevz (approved):

gurukamath:
> Hey @gurukamath, since #167 we are signing and including the `(v, r, s)` signature values for all transactions before we call `t8n`, so the removal of `protected` field makes sense since the `t8n` will not be signing any transactions from us anymore.
> 
> Let me know if that sounds ok for you! :)

If we don't need to sign the transactions, I think it is ok if we don't have the protected field.

PR #192, opened 2023-07-04
fix -v --version t8n args for different t8ns

winsvega:


spencer-tb (approved):
LGTM!

src/evm_transition_tool/__init__.py
```diff
@@ -174,6 +175,10 @@ def __init__(
             raise Exception("evm process unexpectedly returned a non-zero status code: " f"{e}.")
         except Exception as e:
             raise Exception(f"Unexpected exception calling evm tool: {e}.")
+        if not result.stdout.strip():
```
winsvega, line 178: I don't like this check it does not seem reliable

danceratopz:
Hi @winsvega, I think this is now handled via the changes made to enable `evm-one`, see #142.

Also, this adds support for `ethereum-spec-evm` from execution-specs: #200.

PR #193, opened 2023-07-05
chore: enable python 311

danceratopz:
Allow usage of execution-spec-tests with Python 3.11.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: ['3.10']
+        python: ['3.10', '3.11']
         golang: ['1.20.5']
         solc: ['0.8.20']
     steps:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -56,7 +56,7 @@ The generated test fixtures can be used:
 
 ### Prerequisites
 
-The following requires a Python 3.10 installation.
+The following requires a Python 3.10 or Python 3.11 installation.
```
spencer-tb, line 59: I was able to `fill` using the all versions between these ranges, so we could add something like the following: 
```suggestion
The following requires an installation of Python between versions 3.10.3 and 3.11.4.
```

spencer-tb (approved):
LGTM!

README.md
danceratopz: Thanks for the very thorough check - good to know! I would prefer not to mention the patch version in order to avoid having to update the docs upon every patch release. In the unlikely case that a new patch release would break our code, we'll find out in Github Actions:
https://github.com/ethereum/execution-spec-tests/blob/4b43545b23f7f7585839efd206b387d1a14e9e2a/.github/workflows/test.yaml#L10
As we run only specify version as major-minor, the latest patch release will be used (this is also quicker apparently):
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input

PR #194, opened 2023-07-05
chore: add a pre-commit config

danceratopz:
Adds a [pre-commit](https://pre-commit.com/) config file in the repository root (`.pre-commit-config.yaml`) that can be used to enable automatic checks upon commit - the commit will not go through if the checks don't pass.

To enable pre-commit, the following must be ran once:

```console
pip install pre-commit
pre-commit install
```
Enabling of pre-commit checks is not mandatory (it cannot be enforced) and even if it is enabled, it can always be bypassed with:

```console
git commit --no-verify
```

I'm sure we can tweak the hooks quite a bit as we get more experience and ideas, but we can leave it as this for now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,15 @@
+repos:
+  - repo: local
+    hooks:
+      - id: tox
+        name: tox
+        entry: tox
+        language: system
+        types: [python]
+        pass_filenames: false
+      - id: tox-docs
+        name: tox-docs
+        entry: tox -e docs
+        language: system
+        files: \.md$
+        pass_filenames: false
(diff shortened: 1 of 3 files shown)


spencer-tb (approved):
LGTM! I'm a big fan of this workflow :100: 

We should consider making it a requirement for committing in the future, when we have more people contributing to the repo!

danceratopz:
> LGTM! I'm a big fan of this workflow 100
> 
> We should consider making it a requirement for committing in the future, when we have more people contributing to the repo!

Nice, it is very helpful! We can't make it a requirement, as it can't be enforced, but we can certainly encourage it :smile_cat:

PR #195, opened 2023-07-05
Add correct fields to tx types

shemnon:
There are a number of tests that vary across transaction types.  Besu's t8n tool is sensitive to unsupported fields in some tx types and requires a value for other fields in those types.

This PR makes sure all required and no unsupported fields are in each TX type for cancun blob tx tests.


marioevz (approved):
LGTM. Thanks!

PR #200, opened 2023-07-11
evm_transition_tool: Add execution-specs, `ethereum-spec-evm`

marioevz:
Brings back support of using `ethereum-spec-evm` transition tool to fill tests.

### DIFF
--- a/src/evm_transition_tool/execution_specs.py
+++ b/src/evm_transition_tool/execution_specs.py
@@ -7,9 +7,11 @@ https://github.com/ethereum/execution-specs
 from pathlib import Path
 from re import compile
 
-from ethereum_test_forks import Fork
+from ethereum_test_forks import ConstantinopleFix, Fork
 
 from .geth import GethTransitionTool
+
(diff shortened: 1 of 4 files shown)


danceratopz:
Seems to work nice in combo with #204.

danceratopz:
Added doc. It's a bit hidden under "execution-spec-tests/library/evm_transition_tool/#evm_transition_tool.ExecSpecsTransitionTool", but at least we can point people to it, if need be.

grafik

src/evm_transition_tool/execution_specs.py
```diff
@@ -0,0 +1,67 @@
+"""
+Ethereum Specs EVM Transition tool frontend.
```
danceratopz, line 2: Perhaps more of an "interface" than a "frontend" :smile:

danceratopz (approved):
Amazing that we can add the execution-specs tool almost for free! :partying_face: 

I think I would prefer `ExecutionSpecsTransitionTool` for clarity, but don't have a very strong feeling about it.

PR #204, opened 2023-07-12
pytest: remove excessively restrictive unsupported fork check

### LINKED ISSUE #201: Any unsupported forks in a transition tool results in the test not running for any forks
The problem might've to do with this line:
https://github.com/ethereum/execution-spec-tests/blob/08876580c1660380362efb10b40034af8e330f67/src/pytest_plugins/forks/forks.py#L166

Where we collect all the unsupported forks for a given tool, but this might be too restrictive as we might still want to run the test for the remaining tests.

Found during development of #200 

@danceratopz

danceratopz:
1. Fixes #201.
2. Improves reporting for the case when the intersection between a test's validity range and the specified fork range (on the command-line or default) is empty:
    1. If `-vv` is specified (`config.getoption("verbose") >= 2`) then report each test with the reason why it was skipped in the pytest execution output.
    2. If `config.getoption("verbose") >= 2`, use the previous behaviour, report that tests were skipped in the collection summary at the beginning of the test run, but don't report each test explicitly in the execution output.

### DIFF
--- a/src/pytest_plugins/forks/forks.py
+++ b/src/pytest_plugins/forks/forks.py
@@ -305,9 +305,27 @@ def pytest_generate_tests(metafunc):
 
     if "fork" in metafunc.fixturenames:
         if not intersection_range:
-            pytest.skip(  # this reason is not reported on the command-line
-                f"{test_name} is not valid for any any of forks specified on the command-line."
-            )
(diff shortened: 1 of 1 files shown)


marioevz (approved):
LGTM!

PR #23, opened 2023-01-05
fillers: EVM Object Format Reference Tests

marioevz:
Introduces reference tests for EOF EIPs, including:
- https://eips.ethereum.org/EIPS/eip-3540
- https://eips.ethereum.org/EIPS/eip-3670
- https://eips.ethereum.org/EIPS/eip-4200
- https://eips.ethereum.org/EIPS/eip-4750
- https://eips.ethereum.org/EIPS/eip-5450

PR #205, opened 2023-07-12
WIP: EVM Object Format Reference Tests

marioevz:
Supersedes #23 

Works:
- Prague fork
- Evmone running tests

Doesn't work:
- CREATE3 procedure to deploy the EOF contracts


src/ethereum_test_tools/eof/v1/__init__.py
```diff
@@ -0,0 +1,492 @@
...
+        The initcode can be padded to a specified length for testing purposes.
+        """
+        self.name = name
+        self.assembled_output = deploy_container.assemble()
+        self.init_container = Container(
+            sections=[
+                Section(
+                    kind=SectionKind.CODE,
+                    data=Op.CALLDATACOPY(0, 0, len(self.assembled_output))
+                    + Op.RETURN(0, len(self.assembled_output)),
```
gumb0, line 382: Some clarification how this should work, maybe it will help:

So in the simplest case `init_container` can be something like just
```
code: RETURNCONTRACT{0}(0, 0)
container: assembled_output
data: empty
```
and then `initcode` can be
```
code: CREATE3{0}(0, 0, 0, 0) STOP
container: init_container
data: empty
```


More complicated case can append data to `init_container`, by for example getting it from calldata:
`init_container`:
```
code: CALLDATACOPY(0, CALLDATASIZE) RETURNCONTRACT{0}(0, CALLDATASIZE)
container: assembled_output
data: empty
```
`initcode`:
```
code: CALLDATACOPY(0, CALLDATASIZE) CREATE3{0}(0, 0, 0, CALLDATASIZE) STOP
container: init_container
data: empty
```

marioevz: Thanks for the insights! @spencer-tb is helping me to rebase to main right now, which should fix the issues you mentioned on the discord, and I think it would be very helpful if we start collaborating on the same branch to start improving these python tests.

marioevz:
closed in favor of #512

PR #206, opened 2023-07-12
Add support for Besu

shemnon:
Adds support for Besu using the `t8n-server` command.  This command avoids the Java startup penalties and provides similar performance to other clients.  Support for "cleanup" tasks is added, in this case to terminate the t8n-server process for Besu.

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -10,6 +10,7 @@ balance
 base64
 basefee
 bb
+besu
 big0
 big1
 blockchain
--- a/.wordlist_python_pytest.txt
+++ b/.wordlist_python_pytest.txt
@@ -30,6 +30,7 @@ parametrize
 popen
 pytester
 pytestmark
+readline
 regexes
 reportinfo
 ret
--- a/src/evm_transition_tool/__init__.py
+++ b/src/evm_transition_tool/__init__.py
@@ -2,6 +2,7 @@
 Library of Python wrappers for the different implementations of transition tools.
 """
 
+from .besu import BesuTransitionTool
 from .evmone import EvmOneTransitionTool
 from .execution_specs import ExecutionSpecsTransitionTool
 from .geth import GethTransitionTool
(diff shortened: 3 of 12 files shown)


danceratopz (approved):
Had to rebase due to changes on main. Sorry that you lost authorship of the first commit, @shemnon, I'm not sure how I managed that.

Anyway, this is working very nicely. The Besu `t8n-server` doesn't work well with the xdist plugin, which should speed up test execution by concurrently executing tests on different workers. If a few amount of workers (<6ish) are used, it does work, but is slower than normal execution (`-n=0`) so the `test_filler` plugin raises a usage error if xdist is used (i.e., `-n j, if j >0`) in combo with Besu.

I did experiment with running with `-n=auto` using a single instance of the `t8n-server`, but it wasn't fast or reliable (occasional timeouts). Perhaps we can revisit this later if need be, but I don't think it's worth it for now; `t8n-server` runs quickly enough for the existing test case set as is!

shemnon:
I'd have to learn about about pytest to provide suggestions.  Is the session scope shared across runners or is it one session per runner?

Also, the timeouts may be besu not handling bad input fast and with finality.  Looks like we need a "server error" response that I catch for all errors.

danceratopz:
> I'd have to learn about about pytest to provide suggestions. Is the session scope shared across runners or is it one session per runner?

Thanks for your comments. It's one session per runner. 

> Also, the timeouts may be besu not handling bad input fast and with finality. Looks like we need a "server error" response that I catch for all errors.

For now, I consider running `fill` with pytest-xdist a nice-to-have, at least from my side. It will definitely become more relevant as the number of test cases/forks grow, though. 

If a speed-up would be relevant to you and your workflows sooner rather than later, then I'd be more than happy to explore a solution now. I'd keep it out of the scope of this PR though.

danceratopz:
> LGTM - is there anything more needed from me?

Hey @shemnon, nope looks good, excuse the delay!

PR #210, opened 2023-07-18
Add `disable-hive` flag

marioevz:
Adds a flag that removes all hive related properties from the output fixtures.

At the moment it only removes the Engine API directive instructions.

~Requires #207, hence draft.~ Rebased and removed dependency.

### DIFF
--- a/src/pytest_plugins/test_filler/test_filler.py
+++ b/src/pytest_plugins/test_filler/test_filler.py
@@ -16,6 +16,7 @@ import pytest
 from ethereum_test_forks import Fork
 from ethereum_test_tools import (
     BaseTest,
+    BaseTestConfig,
     BlockchainTest,
     BlockchainTestFiller,
     Fixture,
@@ -84,6 +85,13 @@ def pytest_addoption(parser):
         dest="flat_output",
         default=False,
         help="Output each test case in the directory without the folder structure.",
+    )
+    test_group.addoption(
+        "--disable-hive",
+        action="store_true",
+        dest="disable_hive",
+        default=False,
+        help="Output tests skipping hive-related properties.",
     )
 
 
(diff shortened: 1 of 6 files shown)


src/pytest_plugins/test_filler/test_filler.py
```diff
@@ -315,7 +333,7 @@ def __init__(self, *args, **kwargs):
 
 @pytest.fixture(scope="function")
 def state_test(
-    request, t8n, fork, engine, reference_spec, eips, fixture_collector
+    request, t8n, fork, engine, reference_spec, eips, fixture_collector, base_test_config
```
danceratopz, line 336: I think this should also be passed to the `blockchain_test` fixture? Currently, filled `blockchain_test` tests do not differ and still contain the `engineNewPayload` field.

src/pytest_plugins/test_filler/test_filler.py
```diff
@@ -331,6 +349,7 @@ def state_test(
 
     class StateTestWrapper(StateTest):
         def __init__(self, *args, **kwargs):
+            kwargs["base_test_config"] = base_test_config
```
danceratopz, line 352: Here, too :)

danceratopz (commented):
Looks good! But, I get identical fixtures for the following 2 commands:
```console
fill -n auto  -m blockchain_test --output=fixtures-hive-disabled-blockchain-tests-only --disable-hive
﻿
fill -n auto  -m blockchain_test --output=fixtures-hive-enabled-blockchain-tests-only
```
See comments below.

src/pytest_plugins/test_filler/test_filler.py
marioevz: Yes, I missed this completely! I checked only one state test when I was verifying the output.

spencer-tb (approved):
LGTM! Runs as expected :)

danceratopz (approved):
LGTM!

danceratopz:
Rebased due to a minor conflict in `src/ethereum_test_tools/__init__.py` from #157; no other changes.

PR #211, opened 2023-07-19
Chore: Fix `fill --test-help` formatting for store true/false arguments

danceratopz:
Whilst reviewing #210 I noticed that the output of `fill --test-help` was misleading for `--disable-hive` (and all other `store_true`/`store_false` arguments) as it suggested that it took an argument:
grafik

This PR fixes that:
grafik


spencer-tb (approved):
Nice :D

PR #212, opened 2023-07-19
chore: add helpful vscode launch configs

danceratopz:
Add configs to `.vscode/launch.recommended.json` that assist debugging test cases with an interactive choice of an evm binary and test path.


spencer-tb (approved):
Like this a lot! Will be useful for using multiple t8n branches :)

PR #213, opened 2023-07-19
readme: fix pip install command

chfast:
zsh does not like the `[` character so wrap it with `''`.


danceratopz (approved):
Thanks for this!

I added the fix to the online version of the docs, too. The readme content is unfortunately duplicated there. I fixed up the formatting while I was at it, too. It'll be updated here after merge:
https://ethereum.github.io/execution-spec-tests/main/getting_started/quick_start/

PR #222, opened 2023-07-21
feature: Pass Correct Fork Name to T8N on Transition Forks

### LINKED ISSUE #208: pyspec pass transition fork to t8n
how ShanghaiToCancunAtTime  work? geth t8n rejects it
actually t8n does not have to care at all. we just pass it Shanghai if condition is fakse and Cancun otherwise

marioevz:
## Changes Included
- Correct fork is now passed to transition tools on transition forks:
  - E.g. on `BerlinToLondonAt5`, only `Berlin` and `London` forks are now passed to the transition on the appropriate block number
- Transition fork classes now automatically derive information from their base classes

Fixes #208


spencer-tb (approved):
LGTM!

src/ethereum_test_forks/forks/forks.py
```diff
@@ -12,6 +12,13 @@ class Frontier(BaseFork):
     Frontier fork
     """
 
+    @classmethod
+    def fork(cls, block_number: int, timestamp: int) -> str:
```
danceratopz, line 16: Point 1. How about specifying keyword arguments and providing defaults here, instead of defining them in `transition_tool.py`? I appreciate they belong more to the t8n tool, but as parameters required by the framework, having them here would keep the transition tool code simpler?
```suggestion
    def fork(cls, block_number: int = ALWAYS_TRANSITIONED_BLOCK_NUMBER, timestamp: int = ALWAYS_TRANSITIONED_BLOCK_TIMESTAMP) -> str:
```

src/ethereum_test_tools/spec/blockchain_test.py
```diff
@@ -155,6 +155,8 @@ def make_block(
                 txs=to_json(txs),
                 env=to_json(env),
                 fork=fork,
+                block_number=Number(env.number),
+                block_timestamp=Number(env.timestamp),
```
danceratopz, line 159: Point 2. How about doing this instead? Or do you think we'll need the Fork object in evaluate at some point in the future (instead of just the name)?
```suggestion
                fork_name=fork.fork(block_number=Number(env.number), block_timestamp=Number(env.timestamp))
```

src/ethereum_test_tools/spec/state_test.py
```diff
@@ -122,6 +122,8 @@ def make_blocks(
             txs=to_json(txs),
             env=to_json(env),
             fork=fork,
+            block_number=Number(env.number),
+            block_timestamp=Number(env.timestamp),
```
danceratopz, line 126:
```suggestion
            fork_name=fork.fork(block_number=Number(env.number), block_timestamp=Number(env.timestamp))
```

src/evm_transition_tool/besu.py
```diff
@@ -82,6 +82,8 @@ def evaluate(
         txs: Any,
         env: Any,
         fork: Fork,
+        block_number: int,
+        block_timestamp: int,
```
danceratopz, line 86: fork_name: str
```suggestion
        fork_name: str
```

src/evm_transition_tool/besu.py
```diff
@@ -92,7 +94,7 @@ def evaluate(
         if not self.process:
             self.start_server()
 
-        fork_name = fork.name()
+        fork_name = fork.fork(block_number, block_timestamp)
```
danceratopz, line 97:
```suggestion

```

danceratopz (commented):
Rebased to get besu support that was added in #206.

Looks good! I did wonder about two changes to keep rather framework-specific code out of the transition tool interface implementation:
1. Change base fork's `fork()` signature to take keyword arguments with default values in order to get "reasonable" default behaviour that return the merged fork. 
2. Avoid adding additional function arguments to transition tools' `evaulate()` methods.

More clarification below.

src/ethereum_test_forks/forks/forks.py
marioevz: Awesome! Let me change this throughout the changes :)

marioevz: Changed the default values to `ALWAYS_TRANSITIONED_BLOCK_*` but only on transition forks, the default for the rest is zero. It worked out pretty nice because I also changed the way the transition decorator creates the methods.

src/ethereum_test_tools/spec/blockchain_test.py
marioevz: Yes, I think I was saving it to open up for the possibility of some transition tool implementing a different name at some point, but I don't think this will ever happen now that we have them all supported and we can just raise a PR on the specific tool to align its naming of the fork.

src/ethereum_test_forks/transition_base_fork.py
```diff
@@ -26,24 +26,43 @@ def transitions_from(cls) -> Fork:
...
-            def name(cls) -> str:
-                """
-                Returns the name of the transition fork.
-                """
-                return transition_name
+            setattr(
+                NewTransitionClass,
+                method_name,
+                classmethod(_method),
+            )
```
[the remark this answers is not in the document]
marioevz: The need for a dynamic method was because the methods were neither from one fork or the other, but it performs the block/timestamp check before calling either of them.

I think it should be possible to override the methods automatically when the subclass is created, and we could definitely explore if `__init_subclass__` could suit our needs for this.

Issue #209, opened 2023-07-15
make a custom t8n (mostlikely pyt8n) to generate invalid states


winsvega:
How do you like the idea of a super t8n tool (potentially pyt8n) that can construct states with ethereum rules violated (bigint balances, withdrawals, gaslimit, formulas violated)

I think it is too much to ask from dev teams to support it so I am thinking about pyt8n as a candidate for this role. 

https://github.com/ethereum/tests/issues/1244

danceratopz:
@spencer-tb Perhaps a complimentary approach to #242?

spencer-tb:
Closing in favor of #242 and the WIP PR https://github.com/ethereum/execution-spec-tests/pull/272

PR #226, opened 2023-07-25
evm_transition_tool: Fix besu debug dump

marioevz:
Adds support for besu to support `--t8n-dump-dir` to dump debug information, cf #219.


danceratopz (approved):
This is very cool! 

Tested locally with besu.

I missed #209, so I took the chance to add the command to the output `--test-help` and to the docs here :slightly_smiling_face:

Issue #148, opened 2023-06-14
Investigate long compilation time of long byteocode with `solc --evm-version shanghai`


danceratopz:
Upon using a newer version of solc and compiling the "long bytecode" test with `--evm-version=shanghai`, the test took much longer to complete - it actually looked like it had completely hanged. But it does terminate, cf #147.

danceratopz:
This had nothing to do with solc or its compilation time. This long delay was triggered when `test_yul[large]` failed due to specifying  `--evm-version Shanghai` to solc. This test case was failed due to the solc compiling "6000" to "5f" (PUSH0), so that the produced bytecode differed from the (hardcoded) expected bytecode.

The delay appears to rather occur in pytest and is probably due to the large and complicated diff that pytest performs on the two very long byte arrays. Indeed, the delay was only visible when pytest was ran with `-v` (without `-v` presumably only the start of the byte arrays were diffed).

PR #228, opened 2023-07-26
Tools: Update framework tests to run with solc 0.8.21

danceratopz:
The latest solc release ([0.8.21](https://github.com/ethereum/solidity/releases/tag/v0.8.21)) generates different bytecode for the Yul we use in our framework tests. This PR:
- Makes the framework tests compatible with 0.8.20 and 0.8.21.
- Fixes `test_yul['large']` when compiled with `--evm-version=Shanghai`, cf #148.
- Updates the docs to say which versions of solc are supported.
- Adds 0.8.21 to the solc version matrix under test in Github Actions.


marioevz (approved):
LGTM!

PR #229, opened 2023-07-27
tools: add nimbus t8n support

danceratopz:
Adds nimbus' t8n tool support for filling tests, see https://github.com/status-im/nimbus-eth1/blob/master/tools/t8n/readme.md.

Regarding 264e5ccde54b24e47a91f315ac8cc876528b5cb6, Nimbus is the first tool added that doesn't support `-v`, which is no problem (it uses `--version` instead) - this change catches stderr in order to prevent the console output being polluted with:
```
Unrecognized option 'v'
Try t8n --help for more information.
```
during `fill` startup.

Todo:
- [x] Fix tracing.
- [x] Potentially cleanup duplicate code in `src/evm_transition_tool/nimbus.py` and `src/evm_transition_tool/geth.py`.


danceratopz:
The Nimbus t8n tool integrates well in execution-test-specs and only requires minor changes to the geth interface. I moved geth's implementation of `evaluate` to the base class to be used as a default as it can also be used by nimbus.

Only issue with test cases until Shanghai is https://github.com/status-im/nimbus-eth1/issues/1693.

Currently untested with Cancun test cases.

marioevz (approved):
Looks great!

PR #232, opened 2023-07-31
tests/cancun: 4788-proof all cancun tests

marioevz:
Adds 1 wei to the stateful pre-compile introduced by 4788 in all Cancun tests, in order for it to not be deleted from the state at the end of the block execution due to EIP-161 rules.

This must be a temporary workaround and must be removed when a proper solution to this issue reaches consensus.


spencer-tb:
Nice one! :D

spencer-tb (approved):
Awesome job - I'm going to merge these into master! Then sort out the `tests-4788` branch :D

PR #239, opened 2023-08-02
tools/code: Add conditional bytecode generator

marioevz:
Implements a conditional code generator, to skip using Yul in these cases.

It uses a combination of the `PC`, `JUMP` and `JUMPI` opcodes, but in the future these could be replaced by `RJUMP` and `RJUMPI`.


spencer-tb (approved):
LGTM! 

This is very smart, took me a while to understand how it works!

It gives me hope for the for loop case - I think I have an idea for it

PR #241, opened 2023-08-02
filler: allow genesis state root calc to modify pre-alloc

marioevz:
After Cancun fork, due to [EIP-4788](https://eips.ethereum.org/EIPS/eip-4788), `t8n` automatically updates the storage when calculating the genesis state root to perform the nonce increment of the beacon root stateful pre-compile:
```json
"0x000000000000000000000000000000000000000b": {
                "nonce": "0x01",
                "balance": "0x00",
                "code": "0x",
                "storage": {}
            }
```

Instead of doing this manually for every test, with this PR we fetch the updated `pre` that is the result of calling the `t8n` to calculate the genesis state root, and use that as input `pre` for each test.

This has the added benefit that it will not matter if the client reading the genesis applies or not the pre-compile nonce update because it will be contained in the `pre` of the test.

### DIFF
--- a/src/ethereum_test_tools/common/types.py
+++ b/src/ethereum_test_tools/common/types.py
@@ -684,6 +684,15 @@ class Account:
             expected_storage.must_be_equal(address=address, other=actual_storage)
 
     @classmethod
+    def from_dict(cls: Type, data: "Dict | Account") -> "Account":
+        """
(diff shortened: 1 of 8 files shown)


spencer-tb (approved):
Very clean! :D

PR #245, opened 2023-08-07
tests/4844: Fix Block Expected Error on Full Blob Tests

marioevz:
Fixes the expected error on the full blob block tests.

It was missing because none of the transactions actually fail when sent to the transition tool: they are all being sent in the correct format and only modified when put in the final block.

Thanks @jochem-brouwer for finding and raising this issue!

### DIFF
--- a/tests/cancun/eip4844_blobs/test_blob_txs_full.py
+++ b/tests/cancun/eip4844_blobs/test_blob_txs_full.py
@@ -162,10 +162,10 @@ def tx_max_fee_per_blob_gas(  # noqa: D103
 @pytest.fixture
 def tx_error() -> Optional[str]:
     """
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
Nice catch guys! LGTM

jochem-brouwer:
Heya, I think the release 1.0.1 got updated, but now in `fixtures.tar.gz` there are weird underscore JSON files which are incorrect JSON, seems to be something generated by Mac OS?

jochem-brouwer:
Also, on the new version of these fixtures we are now suddenly failing 1/3 of the tests?

spencer-tb:
Sorry this was me! I used the incorrect branch to generate the tests! New fixtures should work now

PR #246, opened 2023-08-08
tests/cancun/eip4788: Update to use pre-deployed contract

marioevz:
Updates EIP-4788 tests to use a pre-deployed bytecode instead of the regular pre-compile.

## Framework Considerations
A new `pre_allocation` method is added to the base fork abstract class, which specifies a pre-allocation that must be performed for a given fork.

This method also has a new decorator `prefer_transition_to_method`, which marks the method so the `to-fork` implementation of the method takes priority when the method is evaluated.

This is necessary because the pre-allocation cannot be modified at fork transition, so it needs to be present in the genesis if we are transitioning to the given fork that requires this allocation. E.g. Shanghai->Cancun transition fork needs the Cancun pre-allocation of the beacon root bytecode pre-deploy.

## EVM Considerations
- `DELEGATECALL` and `CODECALL` no longer work as previously because the target address is not a precompile
- Using address `0x000000000000000000000000000000000000000b` until the mainnet address reaches consensus
- Gas tests no longer work in current implementation due to gas not being a constant

Fill using: https://github.com/marioevz/go-ethereum/tree/4844-devnet-6-t8n

### DIFF
--- a/src/ethereum_test_forks/forks/forks.py
+++ b/src/ethereum_test_forks/forks/forks.py
@@ -1,7 +1,7 @@
 """
 All Ethereum fork class definitions.
 """
-from typing import Optional
+from typing import Mapping, Optional
 
 from ..base_fork import BaseFork
 
@@ -98,6 +98,15 @@ class Frontier(BaseFork):
         5_000_000_000_000_000_000
         """
         return 5_000_000_000_000_000_000
+
+    @classmethod
+    def pre_allocation(cls, block_number: int = 0, timestamp: int = 0) -> Mapping:
+        """
+        Returns whether the fork expects pre-allocation of accounts
+
+        Frontier does not require pre-allocated accounts
+        """
+        return {}
 
 
 class Homestead(Frontier):
@@ -290,6 +299,21 @@ class Cancun(Shanghai):
         Parent beacon block root is required starting from Cancun.
         """
         return True
+
+    @classmethod
+    def pre_allocation(cls, block_number: int = 0, timestamp: int = 0) -> Mapping:
+        """
+        Cancun requires pre-allocation of the beacon root contract for EIP-4788
+        """
+        new_allocation = {
+            0x0B: {
+                "nonce": 1,
+                "code": "0x3373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5f"
+                "fd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b426201800042"
+                "06555f3562018000420662018000015500",
+            }
+        }
+        return new_allocation | super(Cancun, cls).pre_allocation(block_number, timestamp)
 
     @classmethod
     def engine_new_payload_version(
--- a/src/ethereum_test_tools/common/types.py
+++ b/src/ethereum_test_tools/common/types.py
@@ -699,11 +699,53 @@ class Account:
         """
         return Account(nonce=1, code=code)
 
+    @classmethod
+    def merge(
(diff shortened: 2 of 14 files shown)


src/ethereum_test_forks/forks/forks.py
```diff
@@ -291,6 +300,21 @@ def header_beacon_root_required(cls, block_number: int = 0, timestamp: int = 0)
...
+        new_allocation = {
+            0x0B: {
+                "nonce": 1,
+                "code": "0x3373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5f"
+                "fd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b426201800042"
+                "06555f3562018000420662018000015500",
+            }
+        }
+        return new_allocation | super(Cancun, cls).pre_allocation(block_number, timestamp)
+
```
spencer-tb, line 317: I'm thinking that if we are deploying the beacon root contract on mainnet before the Cancun hardfork then we could add this to Shanghai?

Using the previous method, some of the `ShanghaiToCancunAt15k` transition tests were failing because we didn't have the beacon root in the pre allocation at genesis. As our first block started at Shanghai the pre allocation didn't include the beacon root, so when we got to Cancun block we were missing it in storage and hence got a state root mismatch

marioevz: I added an explanation of this to the description, and I'll work in a bit on adding transition tests so this is tested :raised_hands:

src/ethereum_test_tools/common/types.py
```diff
@@ -699,12 +699,60 @@ def with_code(cls: Type, code: BytesConvertible) -> "Account":
...
+                kwargs |= account_2
+            elif isinstance(account_2, cls):
+                for f in fields(cls):
+                    if (v := getattr(account_2, f.name)) is not None:
+                        kwargs[f.name] = v
+            else:
+                raise TypeError(f"Unexpected type for account_2: {type(account_2)}")
 
-class Alloc(dict, Mapping[FixedSizeBytesConvertible, Account | Dict], SupportsJSON):
+        return cls(**kwargs)
```
spencer-tb, line 730: We could add an extra function here to reduce repeated logic, something like:

```python
@classmethod
def merge(
    cls: Type, account_1: "Union[Dict, 'Account', None]", account_2: "Union[Dict, 'Account', None]"
) -> "Account":
    """
    Create a merged account from two sources.
    """
    def process_account(account: "Union[Dict, 'Account', None]", account_name: str) -> Dict:
        data: Dict = {}
        if account is not None:
            if isinstance(account, dict):
                return account
            if isinstance(account, cls):
                for f in fields(cls):
                    if (v := getattr(account, f.name)) is not None:
                        data[f.name] = v
                return data
        else:
            raise TypeError(f"Unexpected type for {account_name}: {type(account)}")

    kwargs: Dict = {}
    kwargs.update(process_account(account_1, "account_1"))
    kwargs.update(process_account(account_2, "account_2"))

    return cls(**kwargs)
 ```

spencer-tb (approved):
`test_multi_block_beacon_root_timestamp_calls` is super nice :D 

Only one comment but nothing essential, needs a rebase I think as well

spencer-tb (approved):

src/ethereum_test_tools/common/types.py
marioevz: Yes, absolutely :D let me do this :)

PR #249, opened 2023-08-09
filler: Fix dump t8n on blockchain tests

marioevz:
Fixes an oversight that BlockchainTest did not dump the t8n output as StateTest do.


spencer-tb (approved):
Ah nice, I wondered why it wasn't working before - I thought it was because I switched laptops 💻 

Works for me now :D

PR #250, opened 2023-08-10
entrypoints: Add new line at the end of each file when merging word lists

marioevz:
Currently, the resulting flake8 wordlist contains the last word and the first word of the next file merged together because some of the files do not have a new line at the end.

This fix appends a new line at the end of each file.


spencer-tb (approved):
Awesome! I was using:
```python
[flake8]
extend-ignore = SC100, SC200 
```
in `setup.py` as workaround  :D

LGTM!

PR #257, opened 2023-08-16
evm_transition_tool: Fix t8n dump on `str` types

marioevz:
Small fix to properly write the T8N dumps if the content is simply a `str` type.


danceratopz (approved):
lgtm!

PR #263, opened 2023-08-23
Tools: Add tracing to evmone t8n

danceratopz:
Adds tracing for `evmone-t8n` which was recently implemented in https://github.com/ethereum/evmone/pull/616.


spencer-tb (approved):
Yay nice one! LGTM

Issue #268, opened 2023-08-23
Add output.body/txs rlp support for `ethereum-spec-evm` 


danceratopz:
This is likely an issue with the `ethereum-spec-evm` t8n tool; this is a reminder to double-check this and re-enable when solved.

Specifying `--output.body=stdout` doesn't write the body to stdout, but rather to a file called "stdout" in the directory specified by `--output.basedir`.

As of #269, a temporary output directory for `--output.basedir` is no longer specified if tracing is not enabled (it's not required for geth/nimbus); previously this would have worked as the "stdout" file would have been written here and subsequently copied.

danceratopz:
Fixed by https://github.com/ethereum/execution-specs/pull/823.

PR #261, opened 2023-08-18
Tools: add a t8n script to t8n dump dir to easily reproduce executed t8n tool commands

danceratopz:
Requires https://github.com/ethereum/execution-spec-tests/pull/229.

Adds an executable script to each subdir in the `--t8n-dump-dir` in order to reasily rerun t8n tool executions performed during the test run in isolation.

E.g.
```
fill --fork=Shanghai --traces --t8n-dump-dir=/tmp/geth-t8n
```
results in:
grafik
and then:
```
/tmp/geth-t8n/test_yul_fork_Shanghai/2/t8n.sh
```
reruns the t8n command ran for the test and outputs:
```

INFO [08-18|15:52:41.232] Trie dumping started                     root=91d0a9..9fd44b
INFO [08-18|15:52:41.232] Trie dumping complete                    accounts=3 elapsed="162.897µs"
INFO [08-18|15:52:41.232] Wrote file                               file=/tmp/tmp32lqi24l/txs.rlp
{
  "alloc": {
    "0x1000000000000000000000000000000000000000": {
      "code": "0x600f565b5f828201905092915050565b6019600260016003565b5f5560205ff3",
      "storage": {
        "0x0000000000000000000000000000000000000000000000000000000000000000": "0x0000000000000000000000000000000000000000000000000000000000000003"
      },
      "balance": "0xba1a9ce0ba1a9ce"
    },
    "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba": {
...
```

PR #269, opened 2023-08-23
tools: add `t8n.sh` to the t8n-dump-dir for evmone, besu & ethereum-specs-evm (and various t8n tool clean-up)

danceratopz:
Implements the `t8n.sh` script added in #261 for the remaining supported t8n tools (evmone-t8n, besu t8n-server, ethereum-specs-evm t8n). The script gets created to the `--t8n-dump-dir` if specified on the command-line, e.g.,
```
fill --fork=Shanghai --traces --t8n-dump-dir=/tmp/geth-t8n
```

This PR also contains quite a lot of other small improvements and clean-up across the transition tools. Here's a quick overview:
1. Adds file extensions to all the files created in the t8n dump dir.
2. Writes input and output files to input/output subdirectories.
3. The result of`output.body` is now also written to `txs.rlp` in the dump dir. Except ethereum-spec-evm cf #268.
4. The default `evaluate()` method (used by geth, nimbus, ethereum-specs-evm) no longer creates a temporary directory (unless tracing is enabled). All input/output is handled by stdin/stdout, so the directory is not required.
5. A default `collect_traces()` method has been added.
6. Besu t8n-server:
    i. Print the request's `text.response` in the test failure (by raising an exception in evaluate()) if the response is invalid.
    ii. Add more information about the request to the t8n dump dir (response.text, time elapsed, status code).
7. ethereum-spec-evm t8n: Required a few changes to get it running, see associated commit messages which reference the issues found; most of these are very small problems that should be easy to solve, but are out of the scope of this PR.
8. Add a small section to the docs describing `--t8n-dump-dir` and `t8n.sh`, preview: 
    https://danceratopz.github.io/execution-spec-tests/main/getting_started/debugging_t8n_tools/


danceratopz:
Force pushed to include #271.

danceratopz:
Marked as draft: It's probably worth waiting for https://github.com/ethereum/execution-specs/pull/823 as this will remove some additional workaround logic added when the transition tool is `ethereum-specs-evm`.

docs/getting_started/debugging_t8n_tools.md
```diff
@@ -0,0 +1,101 @@
...
+
+where the directories `0` and `1` correspond to the different calls made to the `t8n` tool executed during the test, in this case (pre-Shanghai state):
+
+- `0` corresponds to the call used to calculate the state root of the test's initial alloc (which is why it has an empty transaction list).
+- `1` corresponds to the call used to execute the first transaction from the test.
+
+For post-Shanghai forks:
+
+- `0` corresponds to the call used to calculate the `stateRoot` for the test's initial alloc.
+- `1` (optional) corresponds to the call used to calculate the initial `withdrawalsRoot` for the test's initial alloc if required.
```
marioevz, line 59: Let's just merge https://github.com/ethereum/execution-spec-tests/pull/273 first and get rid of this once and for all :)

docs/getting_started/debugging_t8n_tools.md
```diff
@@ -0,0 +1,101 @@
...
+- `2` corresponds to the call used to execute the first transaction from the test.
+
+Note, there may be more directories present `3`, `4`,... if the test executes more transactions.
```
marioevz, line 62:
```suggestion
Note, there may be more directories present `3`, `4`,... if the test executes more transactions/blocks.
```

docs/getting_started/debugging_t8n_tools.md
```diff
@@ -0,0 +1,101 @@
...
+where the directories `0` and `1` correspond to the different calls made to the `t8n` tool executed during the test, in this case (pre-Shanghai state):
+
+- `0` corresponds to the call used to calculate the state root of the test's initial alloc (which is why it has an empty transaction list).
+- `1` corresponds to the call used to execute the first transaction from the test.
+
+For post-Shanghai forks:
+
+- `0` corresponds to the call used to calculate the `stateRoot` for the test's initial alloc.
+- `1` (optional) corresponds to the call used to calculate the initial `withdrawalsRoot` for the test's initial alloc if required.
+- `2` corresponds to the call used to execute the first transaction from the test.
```
marioevz, line 60:
```suggestion
- `2` corresponds to the call used to execute the first transaction/block from the test.
```

src/evm_transition_tool/transition_tool.py
```diff
@@ -261,15 +289,25 @@ def evaluate(
...
+        if str(self.default_binary) != "ethereum-spec-evm":
+            # See https://github.com/ethereum/execution-spec-tests/issues/268
+            # if enabled for ethereum-spec-evm, it will create a file called stdout in the cwd
+            args.append("--output.body=stdout")
 
         if self.trace:
+            if str(self.default_binary) == "ethereum-spec-evm":
+                raise Exception(
+                    "`ethereum-spec-evm` tracing is not currently implemented in "
+                    "execution-test-specs, see "
```
marioevz, line 305:
```suggestion
                    "execution-spec-tests, see "
```

src/evm_transition_tool/transition_tool.py
```diff
@@ -286,23 +324,30 @@ def evaluate(
...
+            t8n_call = " ".join(args)
+            t8n_output_base_dir = os.path.join(debug_output_path, "t8n.sh.out")
+            if self.trace:
+                t8n_call = t8n_call.replace(temp_dir.name, t8n_output_base_dir)
             t8n_script = textwrap.dedent(
                 f"""\
                 #!/bin/bash
-                mkdir {temp_dir.name}
-                {' '.join(args)} < {debug_output_path}/stdin
+                rm -rf {t8n_output_base_dir}
```
marioevz, line 334: I feel kinda uneasy of adding this line to the script, it could be really bad if a typo would result in `t8n_output_base_dir=='/'`.

src/evm_transition_tool/evmone.py
```diff
@@ -110,63 +119,52 @@ def evaluate(
...
+                    os.path.dirname(file_path), os.path.join(debug_output_path, "input")
+                )
+            t8n_call = t8n_call.replace(  # use a new output path for basedir and outputs
+                temp_dir.name,
+                t8n_output_base_dir,
+            )
+            t8n_script = textwrap.dedent(
+                f"""\
+                #!/bin/bash
+                output_directory={os.path.join(t8n_output_base_dir, "output")}
```
marioevz, line 136: Similar comment to the last one.
Perhaps it would be safer to assume `t8n_output_base_dir` is the current working directory?

marioevz (changes requested):
Looks good overall, just a few comments :+1:

src/evm_transition_tool/evmone.py
```diff
@@ -110,63 +119,52 @@ def evaluate(
...
+            t8n_call = " ".join(args)
+            for file_path in input_paths.values():  # update input paths
+                t8n_call = t8n_call.replace(
+                    os.path.dirname(file_path), os.path.join(debug_output_path, "input")
+                )
+            t8n_call = t8n_call.replace(  # use a new output path for basedir and outputs
+                temp_dir.name,
+                t8n_output_base_dir,
+            )
+            t8n_script = textwrap.dedent(
```
spencer-tb, line 133: Since there is no output for evmone when running the `t8n.sh` script, we could echo the contents written to `$output_directory` - something like:
```python
t8n_script = textwrap.dedent(
    f"""\
    #!/bin/bash
    output_directory={os.path.join(t8n_output_base_dir, "output")}
    rm -rf $output_directory
    mkdir -p $output_directory
    {t8n_call}
    for file in $output_directory/*; do
    echo \-e "\\n>>> $(basename $file) <<<"
    cat $file
    done
    """
)
```

spencer-tb (approved):
Super awesome addition!

Tried with all the evm t8n tools (minus nimbus):
- Besu works great (took me a while to figure out the server part.. :P)
- Evmone: note the comment on adding output via `t8n.sh`, and needs https://github.com/danceratopz/execution-spec-tests/pull/23 merged into your branch.
- Execution specs works well with your changes :D 

Happy for you to merge when the extra changes for evmone/execution-specs are in!

src/evm_transition_tool/evmone.py
danceratopz: As it stands:

https://github.com/ethereum/execution-spec-tests/blob/c7cc487cd3374c9222d8b64270773d370135abc4/src/evm_transition_tool/evmone.py#L137-L138

the `t8n.sh` script will delete the "output" sub-directory (where ever that might be), but doesn't seem to be dangerous. Only disadvantage is that files in the base directory aren't deleted, but I think we can live with that.

src/evm_transition_tool/transition_tool.py
danceratopz: Thanks for spotting this! See 3070d94b. This hardcodes the "t8n.sh.out" directory (instead of f-string variable substitution), so that we are only at risk of deleting these directories, no matter where they're located (or what the value of `debug_output_path` is).

danceratopz: Update: See 3070d94b and comment above on same issue with transition_tool.py.

src/evm_transition_tool/evmone.py
danceratopz: We could do, but I'd leave as is for now; rodiazet already seems happy with it, at least with the addition of his changes in https://github.com/ethereum/execution-spec-tests/pull/273

danceratopz:
@spencer-tb @marioevz thanks a lot for your fixes and suggestions. 

Regarding `execution-specs-evm`: I reverted 32ac3bf879936341fa356386e22ca8e722f2dae1(to reduce complexity) in [504b7e4](https://github.com/ethereum/execution-spec-tests/pull/269/commits/504b7e4991b0a00b401c6378d298b37ec3ce624c), I wouldn't wait for https://github.com/ethereum/execution-specs/pull/823 in order to merge this PR, I'm sure 823 (or a different fix) will get merged in due course.

@marioevz if you're happy with the changes regarding `rm -rf` in the `t8n.sh` script, see https://github.com/ethereum/execution-spec-tests/commit/3070d94bbf1206a8dafcd374f806add5e62d3f44 (and my comments on your review), I think we can merge.

marioevz (approved):
LGTM :)

danceratopz:
@rodiazet this is now merged into main, thanks for your contribution!

PR #271, opened 2023-08-24
chore: simplify spellcheck setup by combining wordlists

danceratopz:
This change combines the previous 3 wordlists into one file, so that it 'just works' with flake8 (without running an additional script). 

The only disadvantages are that we will:
1. Apply python specific words in whitelist.txt to the markdown spellcheck.
2. Make it harder to keep track of types of words within the wordlist, e.g., opcodes. As adding a word via the VS Code cSpell interface reorders the dictionary alphabetically.

There aren't a big deal though.


spencer-tb (approved):
LGTM, thanks for this :D

PR #273, opened 2023-08-25
all: Replace t8n call with internal withdrawals root calculation

marioevz:
Removes the unnecessary call to t8n to calculate the withdrawals root with an internal method.

Imports `trie` for this, which did not have type stubs, so a file with only the necessary stubs was generated.


danceratopz (approved):
Looks great, thanks! Nice trick with the mypy stubs!

src/ethereum_test_tools/spec/state_test.py
```diff
@@ -94,13 +94,7 @@ def make_genesis(
...
             withdrawals_root=Hash.or_none(
-                t8n.calc_withdrawals_root(
-                    withdrawals=env.withdrawals,
-                    fork=fork,
-                    debug_output_path=self.get_next_transition_tool_output_path(),
-                )
-                if env.withdrawals is not None
-                else None
+                withdrawals_root(env.withdrawals) if env.withdrawals is not None else None
             ),
```
danceratopz, line 98: Here we silently set the withdrawals and the beacon_root to `None`:
https://github.com/ethereum/execution-spec-tests/blob/6e3fed66bdcd1b9ca5253ad73a562bb142bf19a6/src/ethereum_test_tools/spec/state_test.py#L64-L66

For transparency, how about failing the test if they're set? Something like:
```python
if env.withdrawals is not None:
    raise Exception("'withdrawals' can't be set in a StateTest's genesis 'env'.")
...
```


In the state test case, we could even, instead of calculating the withdrawals root, write:
```python
withdrawals_root=Hash.or_none(default_trie_root if env.withdrawals is not None else None),
```
where we've previously defined:
```python
default_trie_root = "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"
```

danceratopz (commented):
I just spotted this, just a small comment/question.

danceratopz:
See also https://github.com/marioevz/execution-spec-tests/pull/9

spencer-tb (approved):
Nice one! Personally not the biggest fan of using`stubs/trie` but can see the use of `stubs/` for other things in the future :D

marioevz:
> Nice one! Personally not the biggest fan of using`stubs/trie` but can see the use of `stubs/` for other things in the future :D

Yeah, it's not ideal :/ Hopefully we don't need it often.

src/ethereum_test_tools/spec/state_test.py
marioevz: So `env.withdrawals` might be used on the actual block creation. It's the problem with how state tests are implemented at the moment, we reuse a lot of stuff from the `env` in genesis and in the block 1.
This is definitely due a proper refactoring.

marioevz:
@danceratopz Made the changes of your comment in the latest commit :+1: 

Next stop, get rid of `calc_state_root`!

Issue #274, opened 2023-08-25
Update coincurve from version 17 to 18


danceratopz:
execution-spec-tests and [execution-specs](https://github.com/ethereum/execution-specs) has [coincurve](https://github.com/ofek/coincurve) as a dependency. As of ec18f1e5be520e6d0bc10bf183e8497379c312e2 / 25 Aug 2023 both repos use v17, which doesn't "properly" support Python 3.11. According to the [v18 releases](https://github.com/ofek/coincurve/releases/tag/v18.0.0) adds Python 3.11 support.

Whilst we haven't observed any problems using v17 with Python 3.11, extra dependencies are required to install it on macOS:
```
brew install autoconf automake libtool
```
Upgrading to v18 removes this extra dependency requirement; confirmed by @spencer-tb.

Update: As we install execution-specs as a dependency, we won't be able to update to coincurve 18 unless execution-specs also upgrades to v18.

Here's the output of install execution-specs-tests as `pip install -e .[docs,lint,test]` if we upgrade to 18:
```
INFO: pip is looking at multiple versions of ethereum to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install test-filler and test-filler[docs,lint,test]==1.0.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    test-filler[docs,lint,test] 1.0.0 depends on coincurve==18.0.0
    ethereum 0.1.0 depends on coincurve<18 and >=17

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

PR #275, opened 2023-08-25
docs: add extra installation step for macOS when using Python >=3.11

spencer-tb:
Current work around for the following issue: #274 related to installing coincurve v17 on macOS with Python 3.11.

image


danceratopz (approved):
LGTM, thanks!

PR #276, opened 2023-08-25
chore: run github actions on macos

danceratopz:
Additionally run github actions on macos-latest. Previously, it was only ubuntu-latest.

To help detect issues such as #274.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -38,7 +38,7 @@ jobs:
         run: |
           if [ "$RUNNER_OS" == "Linux" ]; then PLATFORM="linux-amd64"; else PLATFORM="macosx-amd64"; fi
           RELEASE_NAME=$(curl https://binaries.soliditylang.org/${PLATFORM}/list.json | jq -r --arg SOLC_VERSION "${{ matrix.solc }}" '.releases[$SOLC_VERSION]')
-          wget -O $GITHUB_WORKSPACE/bin/solc https://binaries.soliditylang.org/linux-amd64/$RELEASE_NAME
+          wget -O $GITHUB_WORKSPACE/bin/solc https://binaries.soliditylang.org/${PLATFORM}/$RELEASE_NAME
           chmod a+x $GITHUB_WORKSPACE/bin/solc
           echo $GITHUB_WORKSPACE/bin >> $GITHUB_PATH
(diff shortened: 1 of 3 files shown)


marioevz:
I'm not sure why the CI failed but it seems to have failed on fetching the latest spec for one of the EIPs.
Maybe github it's rate limiting itself when trying to get the files from the EIP repo?
We also probably have too many combinations now, I'd say that we remove at least one, maybe the previous version of solc?

danceratopz:
> I'm not sure why the CI failed but it seems to have failed on fetching the latest spec for one of the EIPs. Maybe github it's rate limiting itself when trying to get the files from the EIP repo?

I could confirm that we were indeed getting rate limited with the help of 4440bde5d132749fdaae3c35411bcb377d6a58b9 in https://github.com/ethereum/execution-spec-tests/actions/runs/6016591968/job/16320904696:
```
git_reference_spec.py:70: UserWarning: Unable to get latest version, status code: 
403 - text: {"message":"API rate limit exceeded for 199.19.85.226. .....
```
Very strange that we should get rate limited on this runner so easily though, but it could even be due to other workflows that run there and use the github api. Either way, I mocked the test to remove this source of flakiness: cce0ac57fe532e8378d0890e15856d1cc01a8761.

The network does seem to be very slow on the macos runner, which even appears to be why the job takes so long to execute. It takes twice as long and is 5 minutes slower than the ubuntu runner (it takes almost 2 minutes to create the framework tox environment, for example; seemingly the time it takes to download python packages?). 

> We also probably have too many combinations now, I'd say that we remove at least one, maybe the previous version of solc?

I reduced it to three jobs that cover the most important params.

danceratopz:
The "framework" and "tests" test runs are about 50% (12.4 vs 5.4 seconds and 16.4 vs 10.9 seconds). But it's the setup of the tox venvs that consume most of the time, perhaps it's not predominantly network, see https://github.com/actions/runner-images/issues/6147#issuecomment-1228450302. 

macos-latest currently runs macOS 12 Monterey.

marioevz (approved):
LGTM! Thanks for the changes :)

danceratopz:
#274 was reproduced in run https://github.com/ethereum/execution-spec-tests/actions/runs/5977100121/job/16216331893
image

PR #284, opened 2023-09-01
tests/cancun: Add test scenario where the beacon root contract is missing

marioevz:
## Changes included

Adds a simple test where the expected beacon root contract is not present in the state before the Cancun fork, and verify that the storage for this address is not updated, nor that the calls performed to it return any value, nor that the client reaches an exception or error.

~~Requires #283~~ Merged.


spencer-tb (approved):
LGTM - nice test case  :)

PR #289, opened 2023-09-06
feat(evm): enable tracing support for ethereum-spec-evm

danceratopz:
https://github.com/ethereum/execution-specs/pull/828 implements full tracing for `ethereum-spec-evm`. This PR simply removes the exception so that tracing with the execution-specs t8n tool can be enabled within execution-spec-tests.

We should hold off merging this PR until https://github.com/ethereum/execution-specs/pull/828 is merged.


spencer-tb (approved):
LGTM - we really need a `t8n` spec don't we? Will be really nice once everything is in the same format etc

spencer-tb:
Looks like we can merge now :D

PR #293, opened 2023-09-08
framework: add switch-case construct for conditional bytecode evaluation

danceratopz:
Adds a `Switch` bytecode class that implements switch case behaviour. It should help make the transient storage reentrancy tests more readable; see the refactored test case in 0045bf80ec446ceaf14ff12f41edfbf9d91a8f56 which performs two reentrant calls. It might be clearer to use `Switch` instead of `Conditional` in the other reentrancy test cases, too.

Example from test code:
https://github.com/ethereum/execution-spec-tests/blob/253eb361ac786cadb85b0658bd880b6d6769babd/src/ethereum_test_tools/tests/test_code.py#L318-L331

### DIFF
--- a/src/ethereum_test_tools/code/generators.py
+++ b/src/ethereum_test_tools/code/generators.py
@@ -304,7 +304,28 @@ class Switch(Code):
         # The length required to jump over the condition and action of the next case
         condition_jump_length = len(self.bytecode) + 3
 
-        # Reversed: first case in list has priority; it will become the outer most onion layer
+        # Reversed: first case in list has priority; it will become the outer-most onion layer.
+        # We build up layers around the default_action, after 1 iteration of the loop, the bytecode
+        # is laid out as:
+        #
+        #  [ case[n-1].condition + JUMPI
+        #    default_action + JUMP,
+        #    JUMPDEST + case[n-1].action,
+        #    JUMPDEST]
+        #
+        # and after n=len(cases) iterations:
+        #
+        #  [ case[0].condition + JUMPI
+        #    case[1].condition + JUMPI
+        #    ...
+        #    case[n-1].condition + JUMPI
+        #    default_action + JUMP,
+        #    JUMPDEST + case[n-1].action,
+        #    ...
+        #    JUMPDEST + case[1].action,
+        #    JUMPDEST + case[0].action,
+        #    JUMPDEST]
+        #
         for i, case in enumerate(reversed(self.cases)):
             action_jump_length = (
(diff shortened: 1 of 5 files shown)


danceratopz:
:onion:

src/ethereum_test_tools/code/generators.py
```diff
@@ -237,3 +237,99 @@ def __post_init__(self):
         # Finally we append the true and false branches, and the condition, plus the jumpdest at
         # the very end
         self.bytecode = condition_bytes + if_false_bytes + if_true_bytes + Op.JUMPDEST
+
+
+@dataclass
+class BytecodeCase:
```
marioevz, line 243: I feel like we could almost get away with calling this `Case`. What do you think?

src/ethereum_test_tools/tests/test_code.py
```diff
@@ -308,3 +312,312 @@ def test_opcodes_if(conditional_bytecode: bytes, expected: bytes):
...
+                default_action=b"",
+            ),
+            {0: 1},
+            id="no-default-action-condition-met",
+        ),
+        pytest.param(
+            to_hash_bytes(0),
+            Switch(
+                cases=[
+                    BytecodeCase(condition=Op.EQ(Op.CALLDATALOAD(0), 1), action=Op.SSTORE(0, 1)),
```
marioevz, line 336: I think it would be awesome to have a function to generate the case with a calldata load as a condition, for simplicity.

marioevz (approved):
Looks excellent!

I had a couple of suggestions and opened a PR to your branch for you to take a look, but feel free to merge as is if the suggestions don't fit.

src/ethereum_test_tools/code/generators.py
danceratopz: :100: We have a `Switch`, not a `BytecodeSwitch`!

src/ethereum_test_tools/tests/test_code.py
danceratopz: Great idea. I refactored your `case_calldata` helper function to a to a `CalldataCase` for a more consistent interface: 
https://github.com/danceratopz/execution-spec-tests/pull/25/commits/1473c07bd52c6a828808b2826dd3d8fc1942bd13#diff-fd9295703f4f6a3a2e215448c539eb882ded71beb144c7e2b9afebd2b1f4cbfbR260-R286

I didn't previously know about `field(init=False)`, but I think this seems like a valid and good use case.

I'll merge this PR as is, but plz speak up if you have other ideas, happy to add/modfy later.

PR #297, opened 2023-09-13
tests/cancun/eip_4788: Update predeploy address(0xbEAC02..7d7a), add additional tests

spencer-tb:
Tagging updates to 4788 from the following PR: https://github.com/ethereum/EIPs/pull/7672

### PR Includes
- Update to the beacon root address & contract bytecode.
- **Additional tests cases:**
  - `test_invalid_beacon_root_calldata_value` - asserts that zero value calldata causes revert.
  - `test_beacon_root_selfdestruct` - checks that self destructing to the beacon root transfers a balance.


marioevz (approved):
LGTM :)

If we change the address again, I guess we can make another PR 🙃

PR #299, opened 2023-09-17
tests/cancun/eip_1153: Remove conftest '+1153' in network field.

spencer-tb:
Removes the "+1153" eip number from the fixture network field, as 1153 is now a part of Cancun.


danceratopz (approved):
Thanks! LGTM!

PR #300, opened 2023-09-18
improvement: Deprecate empty accounts within framework.

spencer-tb:
Following the future improvement to deprecate empty accounts -> [here](https://notes.ethereum.org/@RG8j1GXNQ-iPkkFqXEouVA/By-ep0ry6), an assertion check has been added to detect empty accounts within the pre and post allocation within tests.

Note that an empty account is one which has zero nonce, code, balance and an empty storage:
``` 
"0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
                "nonce": "0x00",
                "balance": "0x00",
                "code": "0x",
                "storage": {}
            }
 ```
 
The beacon root tests have been updated to account for the latter (due to the inclusion of the system address), alongside a few withdrawals tests.

### DIFF
--- a/tests/cancun/eip4788_beacon_root/conftest.py
+++ b/tests/cancun/eip4788_beacon_root/conftest.py
@@ -198,7 +198,7 @@ def pre(
     }
     if system_address_balance > 0:
         pre_alloc[to_address(SYSTEM_ADDRESS)] = Account(
-            nonce=1,
+            nonce=0,
             balance=system_address_balance,
         )
     return pre_alloc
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/cancun/eip4788_beacon_root/conftest.py [hunk 1]
```diff
@@ -189,17 +189,19 @@ def pre(
...
         ),
         caller_address: contract_call_account,
-        SYSTEM_ADDRESS: Account(
-            nonce=0,
-            balance=system_address_balance,
-        ),
     }
+    if system_address_balance > 0:
+        pre_alloc[to_address(SYSTEM_ADDRESS)] = Account(
+            nonce=1,
```
marioevz, line 201: This one is actually not quite true, since the receiving of the balance does not increase the nonce.
In reality, the system account will never have a nonce greater than zero, and only its balance might increase.
I think we can simply leave it at zero unless it's causing issues.
```suggestion
            nonce=0,
```

marioevz (changes requested):
Looks great!
Just a small change request :)

marioevz (approved):
LGTM!

PR #301, opened 2023-09-18
improvement: Fixture generation split based on hive specificity

spencer-tb:
This PR updates fixture generation such that it is split into 2 seperate formats. The format is determined by the `--enable-hive` flag:
1. The `--disable-hive` flag has been replaced by the `--enable-hive` flag. I decided to flip this due to increasing queries about the extra hive fields within the fixtures.
2. Without the `--enable-hive` flag, the default format is used. Client teams use this within there CI/CD.
3. With the `--enable-hive` an new fixture format is generated, similar to the current but only with the fields required by the hive pyspec simulator.
4. Its worth noting that we require https://github.com/ethereum/execution-spec-tests/pull/256 to be merged first, which adds the forkchoice version information for the hive pyspec simulator.

Note: after we merge this PR we will have to start including 2 tar files within every new release. One for the default fixture format `fixtures.tar.gz` and another for the hive fixture format `fixtures_hive.tar.gz`.


spencer-tb:
Tests for the hive fixture format have been added for reference, for any additional changes.

The hive fixture format is consumable in the pyspec simulator - tagging: https://github.com/ethereum/hive/pull/844

These have been updated for both solc version `0.8.20` &`0.8.21` to still account for padding differences.

marioevz (approved):
LGTM!

We can leave the fixture refactor for a follow up PR.

PR #302, opened 2023-09-18
feature: Run development fork tests on github actions

marioevz:
Adds a `evm-config.yaml` file which contains a list of possible build configurations for geth's evm binary.

- Main: Runs stable forks using ethereum/go-ethereum@master
- Develop: Runs development forks using marioevz/go-ethereum@cancun-t8n

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,9 @@ env_list =
     tests
     docs
 
+[main]
+development_fork = Cancun
+
 [testenv:framework]
 description = Run checks on helper libraries and test framework
 
@@ -28,9 +31,7 @@ extras = {[testenv:framework]extras}
 allowlist_externals = {[testenv:framework]allowlist_externals}
 commands = {[testenv:framework]commands}
 
-[testenv:tests]
-description = Run checks on the test cases in tests/
-
+[testenv:tests-base]
 extras =
     test
     lint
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -41,8 +42,26 @@ commands =
...
 
+[testenv:tests-develop]
+description = Run checks on the test cases in tests/, including tests for development forks
+
+extras =
+    {[testenv:tests-base]extras}
+
+commands =
+    pytest -n auto --until={[main]development_fork}
+
```
danceratopz, line 64: I wonder if we could optionally specify an evm binary to tox either by specifying it via:
1. A command-line flag.
2. Environment variable.
and then provide this value to `--evm-bin` in the pytest call?

That would make it even more useful.

We could do that for the non-development version, too.

danceratopz (commented):
This is amazing :100:

tox.ini
marioevz: It seems like it's possible by doing this: https://tox.wiki/en/3.3.0/example/general.html

Do you mean that we should build the two binaries from the start and then call tox twice only changing `--evm-bin` ?

danceratopz: > It seems like it's possible by doing this: https://tox.wiki/en/3.3.0/example/general.html

Nice, thanks for checking!

> Do you mean that we should build the two binaries from the start and then call tox twice only changing `--evm-bin` ?

This is one possibility, but as I mentioned in the follow-up comment, it's probably not worth adding this now. It'll add additional complexity that we won't need in the long run (as later I imagine we'll start specifying which evm-version to use locally in some test paths that target functionality for forks under development).

I really like the simplicity of this approach as it stands.

danceratopz (approved):
LGTM! Really like the modularity of the new Github actions.

PR #306, opened 2023-09-22
tests/cancun/eip_4844: Fix invalid blob tx pre cancun tests.

spencer-tb:
`engine_newPayloadV2` should return `INVALID` for the case where blob txs (type 3 txs) exist in the txs rlp.


marioevz (approved):
LGTM :)

Issue #43, opened 2023-01-24
Refactor Blockchain/StateTests (ethereum_test_tools/spec) for clearer Environment Abstration


spencer-tb:
From https://github.com/ethereum/execution-spec-tests/pull/37, via marioevz: 

> Looking at my code on make_block in BlockchainTests it seems like was really easy to make this mistake in the first place because of how the block and env are just mixed all over the place. I think it would be nice to refactor this a bit on a different PR to make a clear distinction that block structure is the source of the test information, but env is the structure that should contain all the correct information at some point.
> 
> I think this same mistake could be present in the StateTests, it's just that we don't have any withdrawals tests written in the StateTests format.

danceratopz:
See #307.

spencer-tb:
We are well overdue to closing this one :)

PR #307, opened 2023-09-22
improvement: Refactor of spec for blockchain & state test generation

spencer-tb:
## 🗒️ Description
The refactor proposal is to use `make_fixture()` and `make_hive_fixture()` within `fill.py`, to abstract all the filling logic to both `state_test.py` and `blockchain_test.py`. These additions should make it easier to add the correct StateTest fixture format in the future.

## 🔗 Related Issues
This is a follow up PR to address some comments on https://github.com/ethereum/execution-spec-tests/pull/301.

Resolves the following issue #43.

## ✅ Checklist
- [x] All: Added an entry to [CHANGELOG.md](../docs/CHANGELOG.md).


danceratopz:
Rebased onto main following the merge of #322 (#322 used a squash commit).

src/pytest_plugins/test_filler/test_filler.py
```diff
@@ -263,8 +263,6 @@ def get_module_dir(item) -> str:
         if not m:
             raise Exception("Could not parse test name: " + item.name)
         name = m.group(1)
-        if fixture.name:
-            name += "-" + fixture.name
         jsonFixture = fixture.to_json()
         self.all_fixtures[module_dir].append((name, jsonFixture))
```
danceratopz, line 267: I think removing L266-267 has removed the test function name from the filler name. But, we did briefly discuss in our weekly that we could change the `name` to be the test's pytest Item `nodeid`. 

This would mean deleting L265-L268 and changing (tested locally):

```suggestion
        self.all_fixtures[module_dir].append((item.nodeid, jsonFixture))
```

The motivation for this change is as following: The `nodeid` can be directly provided to fill to execute this test case in isoloation, e.g.

```
# get the nodeid's using --collect-only -q
✦1 ➜ fill --collect-only -q -k test_warm_coinbase_call_out_of_gas
tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALL-sufficient_gas]
tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALL-insufficient_gas]
tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALLCODE-sufficient_gas]
tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALLCODE-insufficient_gas]
tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-DELEGATECALL-sufficient_gas]
tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-DELEGATECALL-insufficient_gas]
tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-STATICCALL-sufficient_gas]
tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-STATICCALL-insufficient_gas]

1/228 tests collected (227 deselected) in 0.19s
```
and then execute a single test via its nodeid:
```
fill tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALL-sufficient_gas]
```

This will result in fixture names:
```
{
    "000-tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py::test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALL-sufficient_gas]": {
        "_info": {
            "filling-transition-tool": "evm version 1.13.0-unstable-fd389791-20230926",
            "reference-spec": "https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3651.md",
            "reference-spec-version": "d94c694c6f12291bb6626669c3e8587eef3adff1"
        },
        "network": "Shanghai",
        ....
```

This is strictly not a refactor, but perhaps we can sneak it in here? :slightly_smiling_face: cc @marioevz 

Do we need counter prefix here? `000-`, `001-`,... in the fixture name?

danceratopz (changes requested):
Nice, looks cleaner to me! 

Nice to see `engine` / "NoProof" removed from the pytest interface... that seemed unnecessary back when we did the pytest port.

src/pytest_plugins/test_filler/test_filler.py
spencer-tb: I like this a lot Dan 💯 

My vote is to remove the `0XX-` prefix :D

marioevz: Correct me if I'm wrong, but I think we could even leave it as "test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALL-sufficient_gas]" because the "tests/shanghai/eip3651_warm_coinbase/test_warm_coinbase.py" part is redundant since it can be deduced from the file location.

We could make a dry run and test that there are no files with two different tests using the same name just to be sure that this enough. What do you think?

Also if we do this, on pyspecs sim in hive, the name of the test must now include the file path to make sure two tests do not have the same name.

marioevz (commented):
I really like the changes, I'd like to diff the fixtures against the ones currently generated on main, just to double check we are not breaking anything :)

src/pytest_plugins/test_filler/test_filler.py
danceratopz: I understand the desire to keep this minimal, but the motivation to use nodeid is because, the following doesn't work:
```
fill test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALL-sufficient_gas]
```
-> 
```
ERROR: file or directory not found: test_warm_coinbase_call_out_of_gas[fork=Shanghai-CALL-sufficient_gas]
```
It only seems to work with the full nodeids. Note, this is not using `-k`.

danceratopz:
> I really like the changes, I'd like to diff the fixtures against the ones currently generated on main, just to double check we are not breaking anything :)

I diff'd 52f85140. 
* Without `--enable-hive`: The only differences I see are changes in fixture names (as mentioned above).
* With `--enable-hive`: Also differences in names. Only other difference is the presence of `engineFcuVersion` where it was previously missing (presumably a good thing? :))
```diff
diff '--color=auto' -r fixtures-geth-main-hive-ordered/shanghai/eip4895_withdrawals/withdrawals/use_value_in_tx.json fixtures-refactor-spec-hive-ordered/shanghai/eip4895_withdrawals/withdrawals/use_value_in_tx.json
7a8
>         "engineFcuVersion": "2",
187a189
>         "engineFcuVersion": "3",
```
image

danceratopz (approved):
LGTM!

For the resolution of using `nodeid` for fixture names in JSON, see: https://github.com/ethereum/execution-spec-tests/pull/307#discussion_r1354502252

PR #309, opened 2023-09-24
chore: add python 3.12 support

danceratopz:
## 🗒️ Description
- Adds Python 3.12 support.
- 3.12 tested (currently pre-release) in Github Actions.
- Upgrades flake8.

Tested with 3.12.0rc3 built via pyenv.

The same version of flake8 (6.0.0) produced more warnings on Python 3.12 vs 3.10/3.11 (so these are fixed). But, since our flake8 version requirement in setup.cfg didn't make any sense (`flake8>=5,<=6`, note the `<=6`), I bumped this version to ensure we use patches/improvements in v6. Practically as of now, it means we'll use 6.1.0 instead of 6.0.0.

## ✅ Checklist
- [x] All: Added an entry to [CHANGELOG.md](../docs/CHANGELOG.md).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,7 +14,7 @@ jobs:
             evm-type: 'main'
             tox-cmd: 'tox'
           - os: ubuntu-latest
-            python: '3.11'
+            python: '3.12'
             solc: '0.8.21'
             evm-type: 'main'
             tox-cmd: 'tox'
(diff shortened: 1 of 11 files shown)


spencer-tb (approved):
LGTM :) 

Tested on macOS using `pyenv install 3.12.0rc1`. 

I think we just need to rebase onto main, and re-update the changelog

PR #328, opened 2023-10-11
move evmone evaluate into t8n base class (stdin and filesystem versions of t8n interaction)

winsvega:
- Moves the `evaluate()` method from the evmone interface (which receives its input and outputs from the filesytem instead of stdin/stdout) to the base class so that it can be re-used by other transition tools.
- Adds a new attribute `t8n_use_stream` to the base class with a default value of `False`, so that other `t8n` tools still use stdin/stdout.

### DIFF
--- a/src/evm_transition_tool/evmone.py
+++ b/src/evm_transition_tool/evmone.py
@@ -1,27 +1,13 @@
 """
 Evmone Transition tool interface.
 """
-import json
-import os
-import shutil
-import subprocess
-import tempfile
-import textwrap
 from pathlib import Path
 from re import compile
 from typing import Any, Dict, List, Optional, Tuple
 
 from ethereum_test_forks import Fork
 
-from .transition_tool import TransitionTool, dump_files_to_directory
-
-
-def write_json_file(data: Dict[str, Any], file_path: str) -> None:
-    """
-    Write a JSON file to the given path.
-    """
-    with open(file_path, "w") as f:
-        json.dump(data, f, ensure_ascii=False, indent=4)
+from .transition_tool import TransitionTool
 
 
 class EvmOneTransitionTool(TransitionTool):
@@ -31,6 +17,7 @@ class EvmOneTransitionTool(TransitionTool):
 
     default_binary = Path("evmone-t8n")
     detect_binary_pattern = compile(r"^evmone-t8n\b")
+    t8n_use_stream = False
 
     binary: Path
     cached_version: Optional[str] = None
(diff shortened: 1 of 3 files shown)


danceratopz:
@winsvega could you fix up tox? All the tests are currently failing in Github Actions.

There's some help here, if you get stuck, I can help:
https://ethereum.github.io/execution-spec-tests/main/writing_tests/verifying_changes/

src/evm_transition_tool/geth.py
```diff
@@ -51,58 +48,3 @@ def is_fork_supported(self, fork: Fork) -> bool:
         If the fork is a transition fork, we want to check the fork it transitions to.
         """
         return fork.fork() in self.help_string
-
-    def verify_fixture(
```
winsvega, line 55: this got deleted with rebase. is it correct?

winsvega:
idk the checks still dont like smth.

danceratopz (commented):
I think it's a great idea to move the "file system" implementation of `evaluate()` from evmone into `TransitionTool` so that other tools can use it.

However, I would prefer to not have Geth as a default and instead insist that all `t8n` tools provide an interface and sub-class from `TransitionTool`. I don't think this is unreasonable, as the interface module is < 40 lines (see the evmone interface, in this PR, for example).

So while I'd like to keep `_evaluate_stream()` and `_evaluate_filesystem()`, I would prefer to add a separate interface for your custom `t8n` which must provide either a `-v` or `--version` flag which contains its name to make it identifiable. 

The reason being, is that I'd like to avoid ambiguity about which features a `t8n` tools supports and which not, like tracing, for example. I'd prefer to have this explicit in the implementation and not stumble over bugs/inconsistencies because the tools are not consistent.

Further reasoning, is that we still need to signal to `GethTransitionTool` which version of the evaluate function to use. Currently, in this implementation, it defaults to "filesystem" which changes the current behavior, which should stay as "stream", imo. To be entirely flexible with a "default" fallback tool, we'd have to add a command-line flag which defines whether to use "stream" or "filesystem", this clutters the interface; again, I'd prefer to keep it consistent.

How would you like to proceed. Shall I finish up this refactor and then we can add your interface here or in a separate PR? 

Also happy to discuss via chat!

winsvega:
default is stream. 
yes it is ok to have this classes and -v recognition. 
just in case the autodetection does not work, it defaults to geth filesystem interface. 

what we are talking about is basically same as I did with retesteth configs. defclaring for each client how does it work.
(which forks support, is it a socket listening, a t8n, an rpc, does it support txs.rlp or .json, does it calculate basefee or wait it to be provided and so on. and what exception strings it throws.)

danceratopz (approved):
Thanks! LGTM, I just updated the docstrings in 24a7566b to reflect the change.

PR #339, opened 2023-10-25
feat(pytest): add a consume command that acts as a test runner for client statetest/blocktest commands

danceratopz:
## 🗒️ Description

Adds a new command `consume` that takes a directory contain JSON fixture files and an evm tool as command-line arguments:
```
consume --fixture-directory=fixtures --evm-bin=/usr/bin/evm
```
This is another pytest-based command that dynamically generates a test that executes `evm blocktest` with every fixture file located in the fixture directory. For example:
image

Later when `evm blocktest` commands obtain an option analogous to pytest's `-k` command then it also generates a single test case for each fixture in every file, this can be simulated now via the `--multiple-tests-per-file` command-line option, but as it stands the test will simply be repeated and executed for all fixtures within the file, not actually per single fixture.

For example:
image

Further ideas:
- Check that the fixtures have the correct format.
- Check that the fixtures are not in hive format. 
- Write fixture format into the `_info` section and apply `evm blocktest` or `evm statetest` accordingly.
- Write metadata to fixtures' `_info` sections using `fill` that can be used when generating the tests in fill, for example, which eips are required to run the test; xfail if not supported. 
- In CI/CD this can be used with pytest's `--junit-xml=test_results.xml` option and the [publish-test-results](https://github.com/marketplace/actions/publish-test-results) Github Action, for example.

### Bonus

The fill, consume and tf entry points now point to a `click` cli that is combined within one module and allows code reuse AND easy custom tweaking of command-line options:
 - `fill --help` now only prints help for our custom options.
 - `fill --pytest-help` prints the default pytest help message.

TODO:
- [ ] Update doc. This will require improving our language as we'll have two commands that "test": `fill` and `consume`. Need to make it clear that `fill` generates test cases and `consume` executes them as a module test. 

## ✅ Checklist


setup.cfg
```diff
@@ -44,8 +45,10 @@ evm_transition_tool =
 
 [options.entry_points]
 console_scripts =
-    fill = entry_points.fill:main
-    tf = entry_points.tf:main
+    fill = entry_points.cli:fill
+    phil = entry_points.cli:fill
```
spencer-tb, line 49: <3

setup.cfg
```diff
@@ -44,8 +45,10 @@ evm_transition_tool =
...
+    tf = entry_points.cli:tf
```
spencer-tb, line 50: I'm up for deprecating `tf` now if you are

src/entry_points/cli.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    The `tf` command, deprecated as of 2023-06.
+    """
+    print(
+        "The `tf` command-line tool has been superseded by `fill`. Try:\n\n"
+        "fill --help\n\n"
+        "or see the online docs:\n"
+        "https://ethereum.github.io/execution-spec-tests/getting_started/executing_tests_command_line/"  # noqa: E501
+    )
+    sys.exit(1)
```
spencer-tb, line 52: Happy to remove this re. above :D

whitelist.txt
```diff
@@ -339,6 +340,7 @@ subclasses
 subcommand
 substring
 substrings
+tf
```
spencer-tb, line 343: Re. deprecation above

spencer-tb (commented):
Really appreciate this PR - its going to be very useful in the future once every client has a `blocktest`-like sub-command! Let's get this CI up and running :D

All further ideas are great as well! We should add the coverage to this weekly testing report for each client team

setup.cfg
danceratopz: Before the workshop I would have :100: agreed! But someone there literally mentioned "calling tf". So inclined to leave it for. Doesn't really hurt.

whitelist.txt
danceratopz: Good catch! (if we remove)

tests_consume/test_via_rlp.py
```diff
@@ -0,0 +1,244 @@
...
+            test_passed = False
+        test_result_details = "All good." if test_passed else "Oops, test failed."
+        test_result = HiveTestResult(test_pass=test_passed, details=test_result_details)
+    except Exception as e:
+        test_result = HiveTestResult(test_pass=False, details=str(e))
+    test.end(result=test_result)
+
+
+@pytest.fixture(scope="session")
+def network(test_suite):
```
marioevz, line 81: This might be unnecessary, hive automatically connects the clients to the bridge docker network, which is all we need to make requests from it.

tests_consume/test_via_rlp.py
```diff
@@ -0,0 +1,244 @@
...
+def hive_test(request, test_suite: HiveTestSuite):
+    test_parameter_string = request.node.nodeid.split("[")[-1].rstrip("]")
+    test: HiveTest = test_suite.start_test(
+        name=test_parameter_string, description="TODO: This should come from the '_info' field."
+    )
+    yield test
+    test_result: HiveTestResult
+    try:
+        if hasattr(request.node, "rep_call"):
+            test_passed = request.node.rep_call.passed
```
marioevz, line 70: It seems like `node` never contains `rep_call` for me.

tests_consume/test_via_rlp.py
```diff
@@ -0,0 +1,244 @@
...
+from hive.testing import HiveTest, HiveTestResult, HiveTestSuite
+from tenacity import retry, stop_after_attempt, wait_exponential
+
+from ethereum_test_tools.common.types import Fixture
+from pytest_plugins.consume_via_rlp.consume_via_rlp import TestCase
+from pytest_plugins.consume_via_rlp.network_ruleset_hive import ruleset
+
+
+@pytest.fixture(scope="function")
+def temp_dir() -> tempfile.TemporaryDirectory:
```
marioevz, line 21: I think we could skip the temp dir by putting the file's contents in a `bytes`, I think hive.py should support this.

src/pytest_plugins/consume_via_engine_api/consume_via_engine_api.py
```diff
@@ -0,0 +1,5 @@
+"""
```
marioevz, line 1: Just a thought that these tests are going to require the hive version of the fixtures.

marioevz (commented):
Few comments, but overall it's really impressive!

tests_consume/test_via_rlp.py
danceratopz: Improved in [30e2242](https://github.com/ethereum/execution-spec-tests/pull/339/commits/30e2242c903bdf6510272177be25d71865802816).

Probably needs further refinement for xfail/skip though.

src/pytest_plugins/pytest_hive/pytest_hive.py
```diff
@@ -0,0 +1,73 @@
...
+        test_suite_name = request.getfixturevalue("test_suite_name")
+        test_suite_description = request.getfixturevalue("test_suite_description")
+    except pytest.FixtureLookupError:
+        pytest.exit(
+            "Error: The 'test_suite_name' and 'test_suite_description' fixtures are not defined "
+            "by the hive simulator pytest plugin using this ('test_suite') fixture!"
+        )
+
+    suite = simulator.start_suite(name=test_suite_name, description=test_suite_description)
+    # TODO: Can we share this fixture across all nodes using xdist? Hive uses different suites.
```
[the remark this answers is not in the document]
danceratopz: Understand. We could create a test suites to group our test cases by:
1. EIP, or,
2. Fork introduced, or
3. Functionality (based on a pytest marker from fill that is saved to the `_info` section).
4. Test format: BlockchainTest/StateTest(AsBlockchainTest).

I would probably go for 1. EIP, but there will no doubt be lots of ethereum/tests that won't have an EIP. 

I would using matching suites across both the engine and the rlp simulators.

I would suggest we leave it out of those PR and see how The Port goes and what metadata we land up writing to `_info`.

src/pytest_plugins/consume_via_engine_api/consume_via_engine_api.py
danceratopz: We should rebrand this to "engine" version of the fixtures :wink: 

In due course, we should probably fix the naming of `fill`'s `--enable-hive` flag. Let's decide how we package our fixtures first though.

danceratopz: cc @spencer-tb

tests_consume/test_via_rlp.py
danceratopz: Yes, this does indeed seem unnecessary. 

If I remove the network setup and the explicit client connection/disconnection, then a few additional log entries are missing, but there is no change in the behavior of the test, except for a few log entries missing (which appears to be expected). To be precise, these lines are no longer present:
```
INFO[11-30|12:27:13] API: network created                     name=execution_client_network
...
INFO[11-30|12:27:14] API: container connected to network      network=execution_client_network container=47a7b34b
INFO[11-30|12:27:14] API: container disconnected              network=execution_client_network container=47a7b34b
INFO[11-30|12:27:14] removing docker network                  name=execution_client_network
```
As you indicate, I would assume that Hive is doing this for us already.

Removed in [d18b026](https://github.com/ethereum/execution-spec-tests/pull/339/commits/d18b0261c187b734ff54c0c4e3bf872f06bd4965).

src/ethereum_test_tools/common/types.py
```diff
@@ -45,6 +45,33 @@
 from .json import JSONEncoder, SupportsJSON, field, to_json
 
 
+def load_dataclass_from_json(dataclass_type, json_as_dict: dict):
```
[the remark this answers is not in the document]
danceratopz: :100: [9f04df9](https://github.com/ethereum/execution-spec-tests/pull/339/commits/9f04df9dc34aca37e0d2b3e47c454c1f010b4846).

src/pytest_plugins/consume_via_engine_api/consume_via_engine_api.py
spencer-tb: I still feel like we should deprecate `--enable-hive`, and remove the hive fixtures completely. As now we can extract and process all this information directly from the default fixtures within the `consume_via_engine_api.py` simulator. 

I'll make 2 seperate PRs and we can see what we prefer - more discussion on this would be a good idea I think!

spencer-tb: It makes me happy to remove the code from that PR! Will opt for 1 :D

tests_consume/test_via_rlp.py
```diff
@@ -0,0 +1,209 @@
...
+    env = {
+        "HIVE_FORK_DAO_VOTE": "1",
+        "HIVE_CHAIN_ID": "1",
+    }
+    assert test_case.fixture.fork in ruleset, "Oops, should never get here"
+    for k, v in ruleset[test_case.fixture.fork].items():
+        env[k] = f"{v:d}"
+    if test_case.fixture.seal_engine == "NoProof":
+        env["HIVE_SKIP_POW"] = "1"
+    return env
```
spencer-tb, line 136: I think we can remove the PoW check here, as I'm pretty sure we don't need to use `HIVE_SKIP_POW` within hive any more. I actually don't think we even need to add `sealEngine: "NoProof"` to our fixtures as well :D

Suggestion below. I'd still really like to change env prefix from `HIVE_FORK_DAO_VOTE` to `CLIENT_FORK_DAO_VOTE`, for client specific environment variables. Although I do admit having `HIVE_` makes the abstraction clearer within EEST 🤙

```suggestion
@pytest.fixture
def client_environment(test_case: TestCase) -> dict:
    """
    Defines the environment that hive will start the client with
    using the fork rules specific for the simulator.
    """
    client_env = {
        "HIVE_FORK_DAO_VOTE": "1",
        "HIVE_CHAIN_ID": "1",
        "HIVE_NODETYPE": "full",
        **{k: f"{v:d}" for k, v in forks.get(test_case.fixture.fork, {}).items()},
    }
    return client_env
```

spencer-tb: Note I had issues with clients in the engine (pyspec) sim in the past hence specifying that the client should be a full node.

I wondered if we would get a could get a speed up setting `"HIVE_NODETYPE": "snap",` but it had no noteable difference. 

With regards to speed ups: using `xdist` alongside copying the pre-loaded docker-ized client would be amazing. If we can find a way within `hive` or `hive.py` to first load the built client image to the point before a test is consumed, we can keeping copying this container over and over. If we have `>10000` test cases the speed up will be worth it.

danceratopz:
Force pushed to fix commit message.

marioevz (approved):
LGTM! I think we should be able to merge directly to main.

src/ethereum_test_tools/spec/consume/types.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+        return cls(root=test_cases)
+
+    @classmethod
+    def from_index_file(cls, index_file: Path) -> "TestCases":
+        """
+        Create a TestCases object from an index file.
+        """
+        with open(index_file, "r") as fd:
+            index = IndexFile(**json.load(fd))
```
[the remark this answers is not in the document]
marioevz: Ah, I fixed this already in a commit.

tests_consume/test_via_rlp.py
danceratopz: Thanks for these! :pray: 

Added your suggestions to the `environment` fixture in a3836e96. Nice dictionary comprehension!

Made a note on copying the pre-loaded dockerized client - I'd forgotten about that!

src/ethereum_test_tools/spec/consume/types.py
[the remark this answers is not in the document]
danceratopz: Thanks!

PR #341, opened 2023-11-03
move huge similar _evalue logic into separate file

winsvega:
## 🗒️ Description
The intent is to split huge files with a lot of logic
Devide the logic into separate files and refactor into a smaller functions
Making the overall code more readable

I do by small steps as I never used python, to get feedback from the team


src/evm_transition_tool/transition_tool_methods/_filesystem.py
```diff
@@ -0,0 +1,40 @@
+import os
```
danceratopz, line 1: I think it's a great idea to move these generic helper utils to a separate module, but I'd skip the sub-directory and remove the underscore `_`, and call it something like
`src/evm_transition_tool/utils.py` or `src/evm_transition_tool/file_utils.py`

Perhaps we can put it somewhere more neutral, but we can worry about that later.

src/evm_transition_tool/transition_tool_methods/_filesystem.py
```diff
@@ -0,0 +1,40 @@
...
+import stat
+import json
+from json import dump
+from typing import Any, Dict
+
+
+def _dump_files_to_directory(output_path: str, files: Dict[str, Any]) -> None:
```
danceratopz, line 8: This is only convention and is not enforced, but class methods that start with `_` are intended to be private to the class.

These functions aren't part of a class, so I'd remove the `_` prefix.

```suggestion
def dump_files_to_directory(output_path: str, files: Dict[str, Any]) -> None:
```

src/evm_transition_tool/transition_tool_methods/_filesystem.py
```diff
@@ -0,0 +1,40 @@
...
+            else:
+                dump(file_contents, f, ensure_ascii=True, indent=4)
+        if flags:
+            file_mode = os.stat(file_path).st_mode
+            if "x" in flags:
+                file_mode |= stat.S_IEXEC
+            os.chmod(file_path, file_mode)
+
+
+def _write_json_file(data: Dict[str, Any], file_path: str) -> None:
```
danceratopz, line 35:
```suggestion
def write_json_file(data: Dict[str, Any], file_path: str) -> None:
```

src/evm_transition_tool/transition_tool_methods/_evaluate.py
```diff
@@ -0,0 +1,262 @@
...
+import shutil
+import subprocess
+import tempfile
+import textwrap
+import textwrap
+from typing import Any, Dict, List, Optional, Tuple
+from ._filesystem import _dump_files_to_directory, _write_json_file
+
+def _evaluate_stream(
+        self,
```
danceratopz, line 15: This won't work as `self` is not declared in this scope. I think it'd be best to do the refactoring within `transition_tool.py`, keeping all class definitions in one file and then move the results out to a separate file(s) as appropriate afterwards.

danceratopz (changes requested):
Thanks, I appreciate the intention! The `TransitionTool` class and evaluate functions definitely need some clean-up!

Imo, there's not that much to be gained from moving them to their own file (but I think I see where you're going with this). It would def be great to refactor them into smaller functions, but also happy to keep these in `transition_tool.py` or move them to a `utils.py`, or something else, as appropriate.

You previously mentioned in a discussion about sub-classing in order to spread a class implementation over separate files. I don't think it's good form to subclass for the evaluate methods just to achieve this goal and this is not typically done in Python.

However, the `TransitionTool` class is in need of a larger refactor, as we're starting to add interfaces to other EVM tools (such as `blocktest`, see its `verify_fixtures()` method) which has nothing to do with `t8n` tools. So in order to achieve something similar to what I think you'd like to do, we could make an `EVMTool` base class and then make `TransitionTool` and `FixtureConsumer` subclasses (perhaps we can find a better name than `FixtureConsumer`, but it could potentially interface to both the `blocktest` and `statetest` commands). 

Few other comments below.

Happy to jump on a call and discuss how to proceed!

src/evm_transition_tool/transition_tool.py
```diff
@@ -293,32 +258,32 @@ def collect_traces(
                 traces.append(tx_traces)
         self.append_traces(traces)
 
+    @dataclass
+    class ToolData:
```
danceratopz, line 262: Looks good to me. I think I prefer the name `TransitionToolData` to make it clear it's only used in `t8n`. Also with a view to a future clean-up regarding the misplaced `verify_fixture()` function in a "`TransitionTool`" class.

src/evm_transition_tool/transition_tool.py
```diff
@@ -527,6 +435,61 @@ def _evaluate_stream(
 
         return output["alloc"], output["result"]
 
+    def construct_args_stream(self, t8n_data: ToolData, temp_dir):  # noqa: D102
```
danceratopz, line 438:
```suggestion
    def construct_args_stream(self, t8n_data: ToolData, temp_dir: tempfile.TemporaryDirectory):  # noqa: D102
```

src/evm_transition_tool/transition_tool.py
```diff
@@ -527,6 +435,61 @@ def _evaluate_stream(
...
+            f"--state.reward={t8n_data.reward}",
+        ]
+
+        if self.trace:
+            args.append("--trace")
+            args.append(f"--output.basedir={temp_dir.name}")
+        return args
+
+    def dump_debug_stream(
+        self, debug_output_path, temp_dir, stdin, args, result: subprocess.CompletedProcess
```
danceratopz, line 461: Can you add the missing typehints here?

danceratopz (changes requested):
LGTM. Could you fix up tox? 

And add the missing typehints? That can sometimes be tricky, feel free to reach out!

src/evm_transition_tool/transition_tool.py
winsvega: Here already too many args built up too

PR #342, opened 2023-11-03
feat(pytest): use node id as fixture name

### LINKED ISSUE #329: feat(fw): use pytest nodeids as test names in json fixtures
More info in this discussion: https://github.com/ethereum/execution-spec-tests/pull/307#discussion_r1345986884

This will also mean removing the `tag` keyword arguments from tests, which is currently appended to the test names. I think we should add tags though, but not as a name, rather as metadata to allow filtering of tests, by opcode for example, something like `tags=("create", "tstore", "tload", "reentrancy")`.

danceratopz:
## 🗒️ Description
This PR:
- Changes the test fixture names (the keys in the top-level dict used in JSON fixture files) to use the corresponding pytest node id.
- Replaces "=" with "_", although this slightly impacts readability, this will reduce friction along the tool chain. Rationale here: 
  - https://github.com/ethereum/execution-spec-tests/issues/329#issuecomment-1790826781 (dotnet), 
  - https://github.com/ethereum/execution-spec-tests/issues/329#issuecomment-1790861352 (pytest).
 
## 🔗 Related Issues
Fixes #329.

## ✅ Checklist


marioevz (approved):
LGTM! Thanks :)

docs/CHANGELOG.md
```diff
@@ -11,13 +11,28 @@ Test fixtures for use by clients are available for each release on the [Github r
...
 
 ### 🔧 EVM Tools
 
 ### 📋 Misc
 
 - 🔀 Docs: Update `t8n` tool branch to fill tests for development features in the [readme](https://github.com/ethereum/execution-spec-test) ([#338](https://github.com/ethereum/execution-spec-tests/pull/338)).
 
+## Breaking Changes
+
+1. In this release the pytest node ID is now used for fixture names (previously it only the test parameters were used), this should not be breaking. However, "=" in both node IDs, and therefore fixture names, has been replaced with "_", which may break tooling that depends on "=" character.
```
spencer-tb, line 25:
```suggestion
1. In this release the pytest node ID is now used for fixture names (previously only the test parameters were used), this should not be breaking. However, "=" in both node IDs and therefore fixture names, have been replaced with "_", which may break tooling that depends on the "=" character.
```

spencer-tb (approved):
LGTM - I like it a lot! 

Maybe we could rename `test_filler.py` to:
- `pytest_filler.py`
- `plugin_filler.py`

Then for `src/ethereum_test_tools/tests/test_filler.py`, we could rename this to `framework_filler.py`.

Then again `test_test_filler.py` makes me smile :)

docs/CHANGELOG.md
danceratopz: Thanks!

danceratopz:
> LGTM - I like it a lot!
> 
> Maybe we could rename `test_filler.py` to:
> 
>     * `pytest_filler.py`
> 
>     * `plugin_filler.py`
> 
> 
> Then for `src/ethereum_test_tools/tests/test_filler.py`, we could rename this to `framework_filler.py`.
> 
> Then again `test_test_filler.py` makes me smile :)

Hehe, as mentioned offline, one possibility is renaming `test_filler` to `fixture_filler` (which fits well with the `fixture_consumer` in #339). The `test_` prefix for the pytest plugin gets me every time :)

PR #356, opened 2023-12-07
all: Support `ethereum/tests` StateTest format

### LINKED ISSUE #343: Zero `max_fee_per_blob_gas` test is ineffective
Test [test_invalid_tx_max_fee_per_blob_gas](https://github.com/ethereum/execution-spec-tests/blob/14a629d085b97c0c2e87d5e6ef37916a1ec58930/tests/cancun/eip4844_blobs/test_blob_txs.py#L520C5-L520C41) is passing even when there is an issue in the verification logic as in this commit here: https://github.com/ethereum/go-ethereum/pull/28470/commits/5f39342a76ed32059bf340a8679d6f56fef9d4d9

The reason is that the state root included in the test is not the same as the one calculated by geth with the faulty verification logic, because the transaction is executed, and t8n, when there is an invalid transaction in the list of transactions, simply skips the execution and outputs an state root as if the transaction was never executed, but this will almost never be the case if there is a faulty verification, and therefore the block will be still rejected, and the test will pass and won't detect the issue.

There are multiple things that we can do to improve this:
- Better check the error that the client returns when the blockchain test is executed, and verify that the client detected the invalid transaction
- Include the test type `TransactionTest` in `execution-spec-tests` for this kind of tests, and also `StateTest`

marioevz:
## 🗒️ Description

- [x] Refactor fixture classes to use an abstract base class
- [x] Refactor the fixture collector to apply different logic depending on the fixture class
- [x] Implement StateTest format fixture classes
- [x] Refactor existing BlockchainTest formatted tests into StateTest format when possible
- [ ] Add documentation about the fixture formats, and per-fork format changes, for clients to implement test consumers which can be matched to an execution-spec-tests version (To be added on #375, and merged directly to main)
- [x] Fix github workflows to create a single file on releases

## 🔗 Related Issues
Fixes #343

## ✅ Checklist


marioevz:
Rebased to main, and fixed fork's `is_fork` usages.

marioevz:
Force-push: Rebased to main.

marioevz:
Rebased to main.

spencer-tb (approved):
LGTM! All passing on hive pyspec or as expected. 
I've not tried `consume direct`  using statetest but can fix any issues arising from that in a seperate PR :)

My pedantic comments as usual 😆

PR #366, opened 2023-12-20
forks: Add fork-specific, non-inherited class variables

marioevz:
## 🗒️ Description
Introduces fork-specific variables that are not inherited by subclass forks, containing the following variables:
- `transition_tool_name`: The name that the transition tool (`evm t8n`) expects for this fork.
- `blockchain_test_network_name`: The `network` value that must be included in the `BlockchainTest` fixture format for this fork.
- `solc_name`: The name that the solidity compiler (`solc`) expects for this fork.

This is very useful to name forks differently than the transition tool actually expects (e.g. when we rename `Merge` to `Paris`, the transition tool will still expect `Merge` as the fork to execute the tests).

This was achieved by using the `__init_subclass__` special method in the `BaseFork` class, with the class variables as parameters, and then, when defining the fork subclass, the parameters are passed when sub-classing the parent fork:

```python
class Merge(
    London,
    solc_name="Paris",
):
```
Can be converted to:
```python
class Paris(
    London,
    transition_tool_name="Merge",
    blockchain_test_network_name="Merge",
):
```
And will still behave exactly the same to the transition tool (and to the output of the `BlockchainTest` format).

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_forks/base_fork.py
+++ b/src/ethereum_test_forks/base_fork.py
@@ -2,7 +2,7 @@
 Abstract base class for Ethereum forks
 """
 from abc import ABC, ABCMeta, abstractmethod
-from typing import Any, List, Mapping, Optional, Protocol, Type
+from typing import Any, ClassVar, List, Mapping, Optional, Protocol, Type
 
 from .base_decorators import prefer_transition_to_method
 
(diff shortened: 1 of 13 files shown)


spencer-tb (approved):
LGTM :)

Very useful, I don't know why we didn't do this before.

I think in the future we can extend the naming to be specific to the t8n that is being used, i.e if Nethermind choose a different network name than geth. **But only if we need to cross that bridge.**

PR #368, opened 2023-12-22
feat/statetest: Implement StateTestOnly, Change Blob Tx 4844 Tests to StateTest, Use latest geth master

marioevz:
### Included changes
- Fixes all issues in 4844 tests.
- Partly converts tests to state tests when possible: the only tests converted in this PR are the ones that needed updates due to failures in the master geth branch, the rest of the tests (4844 and others) will be converted to StateTest in a separate PR
- Required geth changes are now in master ~~Requires this geth branch: https://github.com/marioevz/go-ethereum/tree/t8n-statetest-fixes~~

### Important considerations
One issue I have not been able to figure out is the use of parent blob gas used in BlockchainTest vs StateTest:

In the state test we can modify `currentExcessBlobGas` to state the blob gas in the current header of the environment where the transaction is supposed to be executed.

In the blockchain test, we cannot set this directly, so we generate a Block 1 to introduce blob transactions and bump the value of block 2 for the test to actually happen.

Right now, since the `env` for each test is different, I could not think of a better solution than to create two different test functions:
- One that generates the blockchain test
- A second one that uses an special class called `StateTestOnly` which generates only the state test, but not a blockchain test.

I'll keep thinking of a way to solve this.

### Merging
I'm planning to merge this using rebase, because the three commits included all do very different things, but they are all required to pass the CI tests (updating to master breaks some test fillers).

## ✅ Checklist


tests/cancun/eip4844_blobs/test_blob_txs.py
```diff
@@ -305,22 +306,50 @@ def pre(  # noqa: D103
     """
     return {
         TestAddress: Account(balance=total_account_minimum_balance + account_balance_modifier),
+        TestAddress2: Account(balance=10**9),
```
spencer-tb, line 309: This seems to only be used by the blockchain test version of `test_invalid_tx_max_fee_per_blob_gas`, leading me to take a look at the usage of `non_zero_blob_gas_used_genesis_block` from `conftest.py`. Which also is only used in that test within `blob_txs.py`.

I think this threw me off when reviewing! My preference would be to keep the test how it was before, purely for readability. It also has added benefit of keeping the fixtures slightly smaller

 More below on the latter :)

tests/cancun/eip4844_blobs/test_blob_txs.py
```diff
@@ -396,27 +426,86 @@ def expected_excess_blob_gas(
...
+        txs=txs,
+        exception=block_error,
+        engine_api_error_code=engine_api_error_code,
+        header_verify=header_verify,
+        rlp_modifier=rlp_modifier,
+    )
+
+
+@pytest.fixture
+def blocks(
```
spencer-tb, line 497: We could remove `blocks` here only retaining `test_block`, assuming the latter below

tests/cancun/eip4844_blobs/test_blob_txs.py
```diff
@@ -545,6 +627,49 @@ def test_invalid_tx_max_fee_per_blob_gas(
...
+)
+@pytest.mark.valid_from("Cancun")
+def test_invalid_tx_max_fee_per_blob_gas_state(
+    state_test_only: StateTestFiller,
+    state_env: Environment,
+    pre: Dict,
+    txs: List[Transaction],
+):
+    """
+    Reject blocks with invalid blob txs due to:
```
spencer-tb, line 659:
```suggestion
    Reject an invalid blob transaction due to:
```

tests/cancun/eip4844_blobs/test_blob_txs.py
```diff
@@ -500,43 +589,36 @@ def test_valid_blob_tx_combinations(
...
 ):
     """
     Reject blocks with invalid blob txs due to:
 
     - tx max_fee_per_blob_gas is barely not enough
     - tx max_fee_per_blob_gas is zero
     """
-    if parent_blobs:
-        pre[TestAddress2] = Account(balance=10**9)
-        blocks.insert(0, non_zero_blob_gas_used_genesis_block)
```
spencer-tb, line 536: My preference would be to keep this here as its specific to this test within the module

spencer-tb (commented):
Really nice! Only small suggestions - mostly based on personal preference :)

PR #371, opened 2024-01-05
feat(tests): callcode gas test when call stipend is applied to gas limit

acolytec3:
### Correctly evaluate CALLCODE gas usage when a call stipend is applied to gas limit
Adds a new test to cover a bug that was discovered in the EthereumJS EVM implementation of the `CALLCODE` opcode where we were incorrectly applying the call stipend to the gas limit applied to the new call frame before evaluating the gas consumed by the call frame so would incorrectly continue execution of the CALLCODE call frame instead of going OOG.

## 🔗 Related Issues
Covers the scenario described in [this issue](https://github.com/ethereumjs/ethereumjs-monorepo/issues/3194).

## ✅ Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -34,6 +34,7 @@ Test fixtures for use by clients are available for each release on the [Github r
 - 🐞 Fixed `tests/cancun/eip4844_blobs/test_blob_txs.py:test_insufficient_balance_blob_tx` to correctly calculate the minimum balance required for the accounts ([#379](https://github.com/ethereum/execution-spec-tests/pull/379)).
 - ✨ Add `tests/cancun/eip4844_blobs/test_blob_txs.py::test_sufficient_balance_blob_tx` and `tests/cancun/eip4844_blobs/test_blob_txs.py::test_sufficient_balance_blob_tx_pre_fund_tx` ([#379](https://github.com/ethereum/execution-spec-tests/pull/379)).
 - 🐞 Fix and enable `tests/cancun/eip4844_blobs/test_blob_txs.py::test_invalid_blob_tx_contract_creation` ([#379](https://github.com/ethereum/execution-spec-tests/pull/379)).
+- ✨ Adds a callcode gas test when call stipend is applied to gas limit. Covers a bug found in the EthereumJS EVM. ([#371](https://github.com/ethereum/execution-spec-tests/pull/371))
 
 ### 🛠️ Framework
 
--- a/tests/frontier/opcodes/test_call_and_callcode_gas_calculation.py
+++ b/tests/frontier/opcodes/test_call_and_callcode_gas_calculation.py
@@ -1,0 +1,160 @@
+"""
+abstract: Tests the nested CALL/CALLCODE opcode gas consumption with a positive value transfer.
+
+    This test is designed to investigate an issue identified in EthereumJS, as reported in:
+    https://github.com/ethereumjs/ethereumjs-monorepo/issues/3194.
+
+    The issue pertains to the incorrect gas calculation for CALL/CALLCODE operations with a
+    positive value transfer, due to the pre-addition of the gas stipend (2300) to the currently
+    available gas instead of adding it to the new call frame. This bug was specific to the case
+    where insufficient gas was provided for the CALL/CALLCODE operation. Due to the pre-addition
+    of the stipend to the currently available gas, the case for insufficient gas was not properly
+    failing with an out-of-gas error.
+
+    Test setup: Given two smart contract accounts, 0x0A (caller) and 0x0B (callee):
+    1) An arbitrary transaction calls into the contract 0x0A.
+    2) Contract 0x0A executes a CALL to contract 0x0B with a specific gas limit (X).
+    3) Contract 0x0B then attempts a CALL/CALLCODE to a non-existent contract 0x0C,
+       with a positive value transfer (activating the gas stipend).
+    4) If the gas X provided by contract 0x0A to 0x0B is sufficient, contract 0x0B
+       will push 0x01 onto the stack after returning to the call frame in 0x0A. Otherwise, it
+       should push 0x00, indicating the insufficiency of gas X (for the bug in EthereumJS, the
+       CALL/CALLCODE operation would return 0x01 due to the pre-addition of the gas stipend).
+    5) The resulting stack value is saved into contract 0x0A's storage, allowing us to
+       verify whether the provided gas was sufficient or insufficient.
+"""
+
+from dataclasses import dataclass
+from typing import Dict
+
+import pytest
+
+from ethereum_test_tools import (
+    Account,
+    Environment,
+    StateTestFiller,
+    TestAddress,
+    Transaction,
+    to_address,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+"""
+PUSH opcode cost is 3, GAS opcode cost is 2.
+We need 6 PUSH's and one GAS to fill the stack for both CALL & CALLCODE, in the callee contract.
+"""
+CALLEE_INIT_STACK_GAS = 6 * 3 + 2
+
+"""
+CALL gas breakdowns: (https://www.evm.codes/#f1)
+memory_exp_cost + code_exec_cost + address_access_cost + positive_value_cost + empty_account_cost
+= 0 + 0 + 2600 + 9000 + 25000 = 36600
+"""
+CALL_GAS = 36600
+CALL_SUFFICIENT_GAS = CALL_GAS + CALLEE_INIT_STACK_GAS
+
+"""
+CALLCODE gas breakdowns: (https://www.evm.codes/#f2)
+memory_exp_cost + code_exec_cost + address_access_cost + positive_value_cost
+= 0 + 0 + 2600 + 9000 = 11600
+"""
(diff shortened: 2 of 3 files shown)


acolytec3:
This new test covers a bug that was identified in the EthereumJS EVM implementation that was recently fixed by @jochem-brouwer in our code.  I'm not 100% sure where to place this test in the broader test suite so have opened this PR to get conversation started and will work through the checklist over the next day or two.

spencer-tb:
Thanks for the contribution :)

I've added a refactor in a PR to your branch here: https://github.com/acolytec3/execution-spec-tests/pull/1
This uses some pytest parameterization, alongside the `Opcodes` class to replace the bytecode with.

Currently test placement is allocated based on when the feature/EIP was introduced.
As `CALLCODE` was introduced in frontier, I moved it there!

acolytec3:
> Thanks for the contribution :)
> 
> I've added a refactor in a PR to your branch here: [acolytec3#1](https://github.com/acolytec3/execution-spec-tests/pull/1) This uses some pytest parameterization, alongside the `Opcodes` class to replace the bytecode with.
> 
> Currently test placement is allocated based on when the feature/EIP was introduced. As `CALLCODE` was introduced in frontier, I moved it there!

Thanks!  I wasn't sure where to put it but have merged your PR.  Anything else you need from me on it?

acolytec3:
Done.  Looks like I mucked up something in the changelog when I squashed things :thinking: Any ideas here?

spencer-tb:
All good! 💯

tests/frontier/opcodes/test_callcode.py
```diff
@@ -0,0 +1,117 @@
...
+@pytest.mark.parametrize(
+    "caller_gas",
+    [0x2D5F, 0x3D50],
+    ids=["insufficient", "sufficient"],
+)
+@pytest.mark.parametrize(
+    "caller_type",
+    [Op.CALL, Op.CALLCODE],
+)
+@pytest.mark.valid_from("London")
```
jochem-brouwer, line 106: These tests are now inside the Frontier folder but they are only valid from London... is this allowed?

spencer-tb: Yeah, we've been discussing test placement this week! I do think what we have can be a little confusing, and very open to changing it.

The general rule for now: tests should be added to the fork folder where the feature being tested was added. So in this case, as the callcode opcode was added in Frontier, we add the test here, even though the test is only valid from London.
```
mark_valid_from_fork >= feature_introduced_fork
```
This is to sync more with the structure of eip tests, i.e the 4844 eip tests are in the Cancun folder, as 4844 is introduced in Cancun.

We will add formal rules to the docs and checks to the framework for this (soon-ish).

spencer-tb: Ideally in the future we will have a lot more callcode tests. So the primary usecase for this structure is to keep all the callcode tests in one place, as opposed to multiple different folders.

Will think on this some more!

jochem-brouwer: Gave this some more thought but this actually makes a lot of sense. For instance, if we would now move these tests into London, then later on if we want to add a specific Chainstart/Frontier test then we would have to `git mv` it into that hardfork folder. So I agree, put the tests in the hardfork folder where the opcode has been introduced. And of course, changes later (for instance gas cost of callcode) could be marked specific in a certain hardfork range if those are added in the same file.

spencer-tb: I'm glad you agree :) and in that case we shall stick to this! I'm hoping to add this information to future docs very soon, alongside checks within the framework for the: `mark_valid_from_fork` pytest directive.

tests/frontier/opcodes/test_callcode.py
```diff
@@ -0,0 +1,117 @@
...
+
+@pytest.fixture
+def callee_code():
+    """
+    Code called by the caller contract:
+        PUSH1 0x00 * 4
+        PUSH1 0x00
+        PUSH1 0x00
+        PUSH2 0xAACC
+        PUSH2 0x1a90
```
danceratopz, line 50: Could you add a comment on the value chosen here for gas? Sorry, if I've missed the obvious here.

tests/frontier/opcodes/test_callcode.py
```diff
@@ -0,0 +1,117 @@
...
+@pytest.fixture
+def post(caller_gas) -> Dict[str, Account]:  # noqa: D103
+    return {
+        to_address(0xAAAB): Account(storage={0x00: (0x01 if caller_gas >= 0x3D50 else 0x00)}),
+    }
+
+
+@pytest.mark.parametrize(
+    "caller_gas",
+    [0x2D5F, 0x3D50],
```
danceratopz, line 99: I'm just wondering why `0x2D5F` is insufficient and `0x3D50` is sufficient. Can this be made tighter? Or why is there such a jump between the two values. Increasing `0x2D5F` doesn't cause the test to fail; it would be nice if this was a hard limit. Thanks for explaining, perhaps I've missed something here!

danceratopz (commented):
This is looking very good! I just proposed a small change in https://github.com/acolytec3/execution-spec-tests/pull/2 for readability, but it's not essential.

One thing that would be helpful would be a comment on the gas values used, see comment below.

tests/frontier/opcodes/test_callcode.py
spencer-tb: I don't think the value matters and is just arbitrary.

I believe we can copy over this: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3214/files#diff-ca2747b296607c2b77ee28d7067ae931408fd2e023d421b588bed952d9cb035fR747

tests/frontier/opcodes/test_callcode.py
spencer-tb: Glad you flagged this up - I assumed both CALL/CALLCODE would have the same insufficient gas values when refactoring. I checked in hive, and its not working properly. Apologies

From: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3214/files#diff-ca2747b296607c2b77ee28d7067ae931408fd2e023d421b588bed952d9cb035fR729-R732
```
    ['f1', 36600, '0x'], // 36600 is CALL fee
    ['f2', 11600, '0x'], // 11600 is CALLCODE fee
    ['f1', 36600 + 7 * 3, '0x01'], // 36600 is CALL fee + 7 * 3 gas for 7 PUSH opcodes
    ['f2', 11600 + 7 * 3, '0x01'], // 11600 is CALLCODE fee + 7 * 3 gas for 7 PUSH opcodes
```
We should be using these values! I'll update this just now.

jochem-brouwer: Ah right, great find - yes especially the last two gas constants ensure that exactly enough gas is provided for CALL(CODE), such that that call frame does not run OOG

spencer-tb:
I added a PR re. the recent comments: https://github.com/acolytec3/execution-spec-tests/pull/3, but after looking over this again, I get a mismatch for the CALL opcode when filling with geth. That is when using the values specified in:
```
['f1', 36600, '0x'], // 36600 is CALL fee
['f2', 11600, '0x'], // 11600 is CALLCODE fee
['f1', 36600 + 7 * 3, '0x01'], // 36600 is CALL fee + 7 * 3 gas for 7 PUSH opcodes
['f2', 11600 + 7 * 3, '0x01'], // 11600 is CALLCODE fee + 7 * 3 gas for 7 PUSH opcodes
```

What values do you guys get in the dynamic gas formulas?
```
CALL gas breakdowns: (https://www.evm.codes/#f1)
memory_exp_cost + code_exec_cost + address_access_cost + positive_value_cost + empty_account_cost

CALLCODE gas breakdowns: (https://www.evm.codes/#f2)
memory_exp_cost + code_exec_cost + address_access_cost + positive_value_cost
```

tests/frontier/opcodes/test_callcode.py
danceratopz: If the value is arbitrary then I think using `GAS` is a better choice here.

danceratopz (commented):
The renaming made in https://github.com/acolytec3/execution-spec-tests/pull/3 got me thinking (as gets trickier to give a good name to tests targeting multiple opcodes!).

Does it make sense to expand this test to all execution opcodes, i.e., `DELEGATECALL` and `STATICCALL` while we're at it?

jochem-brouwer:
This test is specific for the call stipend which we erroneously added to the current gas limit. This only applies to CALL opcodes which allow endowment (sending value) so the other candidate besides CALLCODE is CALL. 😊

tests/frontier/opcodes/test_callcode.py
jochem-brouwer: Agreed, to avoid confusion :smile:

jochem-brouwer:
I wanted to resolve the merge conflict but for some reason the PR does not show up right. Could you cherry-pick https://github.com/acolytec3/execution-spec-tests/pull/4/commits/76ed6432434e64281baa9e085fa26af886c466dd in here?

spencer-tb (approved):

danceratopz:
Looks great! Thanks a lot for the gas calculation explanations / sufficient/insufficient gas improvements!

danceratopz (approved):
LGTM!

PR #378, opened 2024-01-10
fix(tools): Fix state_tests filled json generation

winsvega:


src/ethereum_test_tools/common/types.py
```diff
@@ -1688,7 +1695,7 @@ def with_signature_and_sender(self) -> "Transaction":
                 tx.v += 27
 
         # Remove the secret key because otherwise we might attempt to sign again (?)
-        tx.secret_key = None
+        # tx.secret_key = None
```
marioevz, line 1698: We should simply remove these two lines instead of commenting them.

marioevz: This also makes a lot of tests fail (see `tox -e framework`), so we might need to remove this check: https://github.com/ethereum/execution-spec-tests/blob/641470d21e94adf2fe6656800b52d8dc194c4a06/src/ethereum_test_tools/common/types.py#L1334

marioevz (commented):
Some comments. Thanks!

src/ethereum_test_tools/spec/state/types.py
```diff
@@ -44,49 +43,56 @@ class FixtureEnvironment:
...
     excess_blob_gas: Optional[NumberConvertible] = field(
         default=None,
         json_encoder=JSONEncoder.Field(
-            name="currentExcessBlobGas",
-            cast_type=Number,
+            name="parentExcessBlobGas",
+            cast_type=ZeroPaddedHexNumber,
+        ),
+    )
+    previous_hash: Hash = field(
```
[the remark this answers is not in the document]
winsvega: Yes, the field is an artifact. But then we should remove it from all the tests as an upgrade

src/ethereum_test_tools/common/types.py
winsvega: If I copy it to another field then is no problem

winsvega: changed to use a separate field so not to mess any existing logic (as it fails if I comment this line)

src/ethereum_test_tools/common/types.py
```diff
@@ -995,6 +995,13 @@ class Environment:
             cast_type=Hash,
         ),
     )
+    previous_hash: FixedSizeBytesConvertible = field(
```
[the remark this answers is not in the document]
winsvega: this is so not to change existing test format.

winsvega:
added this `Transaction().with_signature_and_sender_and_private_key()`  
which is the same but adds a private_key to the structure 

```
        secret_key = self.secret_key
        tx = self.with_signature_and_sender()
        tx.secret_key_test = secret_key
```

this one is called to construct state test transactions

src/ethereum_test_tools/spec/state/types.py
```diff
@@ -44,49 +43,63 @@ class FixtureEnvironment:
...
+            cast_type=ZeroPaddedHexNumber,
+        ),
+    )
+    beacon_root: Optional[FixedSizeBytesConvertible] = field(
+        default=None,
+        json_encoder=JSONEncoder.Field(
+            name="currentBeaconRoot",
+            cast_type=Hash,
+        ),
+    )
```
marioevz, line 97: I think we need to discuss this a bit further because this is a block property, so it requires the state test consumers to perform a block action before executing the transactions, while we could simply put this result in the state (just like withdrawals for example).

src/ethereum_test_tools/spec/state/types.py
```diff
@@ -44,49 +43,63 @@ class FixtureEnvironment:
...
-            cast_type=Number,
+            cast_type=ZeroPaddedHexNumber,
         ),
     )
     excess_blob_gas: Optional[NumberConvertible] = field(
         default=None,
         json_encoder=JSONEncoder.Field(
-            name="currentExcessBlobGas",
-            cast_type=Number,
+            name="parentExcessBlobGas",
```
marioevz, line 87: I think this should be `currentExcessBlobGas`, because adding the `parentExcessBlobGas`, apart from it requiring a `parentBlobGasUsed`, requires a block header computation, which I think is outside of the scope of a state test.

By using `currentExcessBlobGas`, the state test consumer can directly know how much the executing transaction must be charged with per blob.

marioevz (approved):

danceratopz:
No changelog/summary?

marioevz:
> No changelog/summary?

You're right, sorry about this, I'll be more careful when merging in the future :raised_hands:

Issue #376, opened 2024-01-09
Fix `test_invalid_blob_tx_contract_creation` by including the test directly as RLP with empty `to`


marioevz:
https://github.com/ethereum/execution-spec-tests/blob/51d30bdd4e77199e7bddb70857fe8f5cc5475d63/tests/cancun/eip4844_blobs/test_blob_txs.py#L767

This is a less-than-ideal workaround to verify that a block is rejected when a transaction type-3 with an empty `to` field.

Basically we trick t8n into executing the transaction with a populated `to` field and then replace the RLP on the block in the test.

Problem is that `stateRoot` `gasUsed` block header values will make no sense, so an invalid block test, we need to check the error to make sure the rejection reason is correct.

Is there any other way we can achieve this @MariusVanDerWijden  ?

PR #379, opened 2024-01-10
fix(tests): Fix minimum balance required on 4844 blob tx gas tests

marioevz:
## 🗒️ Description
Fixes:
- Gas limit on blob transactions, where the value was hard-coded to 21,000 without taking `tx.data` into account, or access lists
- Minimum balance calculation that the account requires for the transaction to be valid

Adds tests:
- Prefund the blob tx sender account within the same block using a prepended transaction (block should be valid)
- Test more combinations of `max_fee_per_gas` (including value higher than current block base fee)
- Test blob transactions with access lists

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/tests/cancun/eip4844_blobs/test_blob_txs.py
+++ b/tests/cancun/eip4844_blobs/test_blob_txs.py
@@ -23,6 +23,7 @@ import pytest
 
 from ethereum_test_forks import Fork
 from ethereum_test_tools import (
+    AccessList,
     Account,
     Block,
     BlockchainTestFiller,
@@ -49,6 +50,9 @@ from .spec import Spec, SpecHelpers, ref_spec_4844
 REFERENCE_SPEC_GIT_PATH = ref_spec_4844.git_path
 REFERENCE_SPEC_VERSION = ref_spec_4844.version
 
+TestPrefundingKey = "0x0b2986cc45bd8a8d028c3fcf6f7a11a52f1df61f3ea5d63f05ca109dd73a3fa0"
+TestPrefundingAddress = "0x97a7cb1de3cc7d556d0aa32433b035067709e1fc"
+
 
 @pytest.fixture
 def destination_account() -> str:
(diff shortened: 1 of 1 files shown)


marioevz:
Fixes #376

spencer-tb (approved):
LGTM! :)

tests/cancun/eip4844_blobs/test_blob_txs.py
```diff
@@ -925,23 +1056,34 @@ def test_invalid_blob_hash_versioning_multiple_txs(
 
 
 @pytest.mark.parametrize(
-    "destination_account,tx_error", [(None, "no_contract_creating_blob_txs")], ids=[""]
-)
-@pytest.mark.skip(reason="Unable to fill due to invalid field in transaction")
+    "tx_gas", [500_000], ids=[""]
+)  # Increase gas to account for contract creation
 @pytest.mark.valid_from("Cancun")
 def test_invalid_blob_tx_contract_creation(
```
spencer-tb, line 1062: Re. chat with geth could we maybe call this `test_decode_blob_tx_contract_creation`?

With a future plan where we highlight the test type in the name:
1) `test_invalid...`, tests for expected invalidity in a block. `evm blocktest` response must be an error and match that of hive.
2)` test_decode...`, tests for the rejection of an invalid block due to a decoding error. These must pass `evm blocktest` as the block is not decoded successfully, although the logs checking the latter must match an expected error in hive.

marioevz: I'm not sure if for this specific case we could make a `decode` test, because other clients might choose to not modify their decoders for type-3 transactions and simply add a check that the value is not null after decoding.

I think both ways of implementing this check are valid.

If we create a `test_decode` fixture format, it could be a subset of invalid block tests, and we should be able to produce a blockchain test with such tests, but in this very specific case, I think should still be a blockchain test only for the reason I mentioned above.

spencer-tb: > I'm not sure if for this specific case we could make a decode test, because other clients might choose to not modify their decoders for type-3 transactions and simply add a check that the value is not null after decoding.

Totally agree with this, both could be considered valid

PR #380, opened 2024-01-11
Unit test/small refactor

spencer-tb:
Refactor of the tests:
- Added a folder for tests specific to the filling of the framework.
- Merge both storage and accounts tests, as to me they are all related.
- Used pytest parameterization for tests to align better with the other framework tests.


danceratopz:
I think we should improve the underlying exception classes in order to make them easier to test. It took me a bit longer than expected to understand what is being tested :laughing: 

I always manage to find more work :sweat_smile: 

Will get back to this with an idea.

danceratopz:
> Oh sounds interesting! More exception classes sound ideal to be honest

No new classes needed upfront. But if you make the Exception class a dataclass, you get an `__eq__()` method for free, which you can then directly use to check for equality. Currently, we can't do this. This is a very small change, just verifying it doesn't break anything.

Issue #381, opened 2024-01-12
test: Add more tests for `CREATE2` <>`SELFDESTRUCT` interactions


marioevz:
We currently have `CREATE2` -> `SELFDESTRUCT` in `tests/cancun/eip6780_selfdestruct/test_selfdestruct.py::test_create_selfdestruct_same_tx`, and the self-destructing contract is called several times to do a self-destruct, but we don't have a test that attempts to do `CREATE2` -> `SELFDESTRUCT` -> `CREATE2`.

We can improve the test coverage with the following tests:

- [ ] Perform a `CREATE2`, make sure that the initcode sets at least a couple of storage keys, then on a different call, in the same tx, perform a self-destruct. Then:
  - [ ] a) on the same tx, attempt to recreate the contract
  - [ ] b) on a different tx, attempt to recreate the contract
- [ ] Perform a `CREATE2`, make sure that the initcode sets at least a couple of storage keys, then in a different tx, perform a self-destruct. Then:
  - [ ] a) on the same tx, attempt to recreate the contract
  - [ ] b) on a different tx, attempt to recreate the contract

PR #382, opened 2024-01-15
feat(tests): dynamic create2 selfdestruct collision 

winsvega:
first case from
https://github.com/ethereum/execution-spec-tests/issues/381

### DIFF
--- a/tests/cancun/eip6780_selfdestruct/test_dynamic_create2_selfdestruct_collision.py
+++ b/tests/cancun/eip6780_selfdestruct/test_dynamic_create2_selfdestruct_collision.py
@@ -1,0 +1,158 @@
+"""
+Suicide scenario requested test
+https://github.com/ethereum/execution-spec-tests/issues/381
+"""
+
+from typing import Dict, Union
+
+import pytest
+
+from ethereum_test_forks import Fork
+from ethereum_test_tools import (
+    Account,
+    Environment,
+    Initcode,
+    StateTestFiller,
+    TestAddress,
+    Transaction,
+    compute_create2_address,
+    to_address,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "EIPS/eip-6780.md"
+REFERENCE_SPEC_VERSION = "2f8299df31bb8173618901a03a8366a3183479b0"
+
+
+@pytest.fixture
+def env():  # noqa: D103
+    return Environment(
+        coinbase="0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
+        difficulty=0x020000,
+        gas_limit=71794957647893862,
+        number=1,
+        timestamp=1000,
+    )
+
+
+@pytest.mark.valid_from("Merge")
+@pytest.mark.parametrize(
+    "create2_dest_already_in_state",
+    (False, True),
+)
+def test_dynamic_create2_selfdestruct_collision(
+    env: Environment,
+    fork: Fork,
+    create2_dest_already_in_state: bool,
+    state_test: StateTestFiller,
+):
+    """Dynamic Create2->Suicide->Create2 collision scenario:"""
+    """
+    Perform a CREATE2, make sure that the initcode sets at least a couple of storage keys,
+    then on a different call, in the same tx, perform a self-destruct.
+    Then:
+        a) on the same tx, attempt to recreate the contract   <=== Covered int this test
+            1) and create2 contract already in the state
+            2) and create2 contract is not in the state
+        b) on a different tx, attempt to recreate the contract
+    Perform a CREATE2, make sure that the initcode sets at least a couple of storage keys,
+    then in a different tx, perform a self-destruct.
+    Then:
+        a) on the same tx, attempt to recreate the contract
+        b) on a different tx, attempt to recreate the contract
+    Verify that the test case described
+    in https://wiki.hyperledger.org/pages/viewpage.action?pageId=117440824 is covered
+    """
+
+    first_create2_result = 1
+    second_create2_result = 2
+
+    address_to = to_address(0x0600)
+    address_code = to_address(0x0601)
+    address_create2_storage = to_address(0x0512)
+    suicide_destination = to_address(0x03E8)
+
+    create2_salt = 1
+    initcode = Initcode(
+        deploy_code=Op.SELFDESTRUCT(Op.PUSH20(suicide_destination)),
+        initcode_prefix=Op.SSTORE(1, 1)
+        + Op.CALL(Op.GAS(), Op.PUSH20(address_create2_storage), 0, 0, 0, 0, 0),
+    )
+    create2_address = compute_create2_address(address_code, create2_salt, initcode)
+
+    pre = {
+        address_to: Account(
+            balance=100000000,
+            nonce=0,
+            code=Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+            + Op.SSTORE(
+                1, Op.CALL(Op.GAS(), Op.PUSH20(address_code), 0, 0, Op.CALLDATASIZE(), 0, 0)
+            )
+            + Op.SSTORE(2, 1),
+            storage={0x01: 0xFF},
+        ),
+        address_code: Account(
+            balance=1000000000000000000,
+            nonce=0,
+            code=Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+            + Op.SSTORE(first_create2_result, Op.CREATE2(7, 0, Op.CALLDATASIZE(), create2_salt))
+            + Op.CALL(Op.GAS(), Op.SLOAD(1), 3, 0, 0, 0, 0)
+            + Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+            + Op.SSTORE(second_create2_result, Op.CREATE2(5, 0, Op.CALLDATASIZE(), create2_salt)),
+            storage={0x02: 0xFF},
+        ),
+        address_create2_storage: Account(
+            balance=7000000000000000000,
+            nonce=0,
+            code=Op.SSTORE(1, 1),
+            storage={},
+        ),
+        TestAddress: Account(
+            balance=7000000000000000000,
+            nonce=0,
+            code="0x",
+            storage={},
+        ),
+    }
+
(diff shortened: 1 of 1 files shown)


tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py
```diff
@@ -0,0 +1,112 @@
...
+        b) on a different tx, attempt to recreate the contract
+    Perform a CREATE2, make sure that the initcode sets at least a couple of storage keys,
+    then in a different tx, perform a self-destruct.
+    Then:
+        a) on the same tx, attempt to recreate the contract
+        b) on a different tx, attempt to recreate the contract
+    Verify that the test case described in https://wiki.hyperledger.org/pages/viewpage.action?pageId=117440824 is covered
+    """
+
+    address_to = TestAddress2
```
marioevz, line 57: There is nothing wrong with this at the moment, but this is an account that tests normally send transactions from.

In the case that in the future we'd like to combine tests somehow, this could cause minor issues, and it's really simple to just use any other address right now.

tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py
```diff
@@ -0,0 +1,112 @@
...
+    suicide_d = to_address(0x03E8)
+    create2_address = to_address(0x15A1593FA2C3A4FAA0DD9109C64860E906ECD6)
+
+    pre = {
+        address_to: Account(
+            balance=1000000000000000000,
+            nonce=0,
+            code=Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+            + Op.MSTORE(0, Op.CREATE2(0, 0, Op.CALLDATASIZE(), 1))
+            + Op.CALL(Op.GAS(), Op.MLOAD(0), 0, 0, 0, 0, 0)
```
marioevz, line 68: We could send some value in this call and then verify that `sucide_d` has it in the post state.

tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py
```diff
@@ -0,0 +1,112 @@
...
+        ),
+    }
+
+    post = {
+        # Make sure the CREATE2 address performed storage operations
+        address_s: Account(storage={0x01: 0x01}),
+        # Make sure the second CREATE2 attempt is failing
+        address_to: Account(storage={0x01: 0x00}),
+        # Make sure selfdestruct cleans the account
+        create2_address: Account.NONEXISTENT,
```
marioevz, line 93: We could parametrize another flavor of this test where the `create2_address` is already created before this block, and we test that even though create2 was called in this same tx, the self-destruct later is still inhibited.

As per @MariusVanDerWijden, there could be a bug in a client where they set a `create2` flag somewhere when the opcode is called, and then they could mistakenly try to delete the account even though it was created before the tx.

marioevz (commented):
Some comments! Thanks!

tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py
```diff
@@ -0,0 +1,129 @@
...
+
+    address_to = to_address(0x0600)
+    address_s = to_address(0x0512)
+    suicide_d = to_address(0x03E8)
+
+    initcode = Initcode(
+        deploy_code=Op.SELFDESTRUCT(Op.PUSH20(suicide_d)),
+        initcode_prefix=Op.SSTORE(1, 1) + Op.CALL(Op.GAS(), Op.PUSH20(address_s), 0, 0, 0, 0, 0),
+    )
+    create2_address = compute_create2_address(address_to, 0, initcode)
```
marioevz, line 71:
```suggestion
    create2_salt = 1
    create2_address = compute_create2_address(address_to, create2_salt, initcode)
```

Added a variable here because it seems that we were using different salts on the address calculation and on the actual creation, so a variable here is just a nice way to make sure this is not an issue.

tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py
```diff
@@ -0,0 +1,129 @@
...
+
+    pre = {
+        address_to: Account(
+            balance=1000000000000000000,
+            nonce=0,
+            code=Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+            + Op.MSTORE(0, Op.CREATE2(7, 0, Op.CALLDATASIZE(), 1))
```
marioevz, line 78:
```suggestion
            + Op.MSTORE(0, Op.CREATE2(7, 0, Op.CALLDATASIZE(), create2_salt))
```

tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py
```diff
@@ -0,0 +1,129 @@
...
+            + Op.CALL(Op.GAS(), Op.MLOAD(0), 3, 0, 0, 0, 0)
+            + Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+            + Op.SSTORE(1, Op.CREATE2(5, 0, Op.CALLDATASIZE(), 1)),
```
marioevz, line 81:
```suggestion
            + Op.SSTORE(1, Op.CREATE2(5, 0, Op.CALLDATASIZE(), create2_salt)),
```

marioevz (commented):
I ran the tests with the salt fix and it seems like we are running out of gas on the tests where the create2 contract is already in the pre-state for some reason.

tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py
```diff
@@ -0,0 +1,112 @@
...
+    Then:
+        a) on the same tx, attempt to recreate the contract
+        b) on a different tx, attempt to recreate the contract
+    Verify that the test case described in https://wiki.hyperledger.org/pages/viewpage.action?pageId=117440824 is covered
+    """
+
+    address_to = TestAddress2
+    address_s = to_address(0x0512)
+    suicide_d = to_address(0x03E8)
+    create2_address = to_address(0x15A1593FA2C3A4FAA0DD9109C64860E906ECD6)
```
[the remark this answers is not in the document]
winsvega: the question is what if I have multiple creates, how do I manage initcode with Op. ?

tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py
```diff
@@ -0,0 +1,129 @@
...
+        post[create2_address] = Account.NONEXISTENT
+
+    tx = Transaction(
+        ty=0x0,
+        chain_id=0x0,
+        nonce=0,
+        to=address_to,
+        gas_price=10,
+        protected=False,
+        data=initcode.bytecode,
```
danceratopz, line 124: #### Error 
```
tests/cancun/eip6780_selfdestruct/test_create2_collision_selfdestruct.py:126: error: Argument "data" to "Transaction" has incompatible type "Optional[bytes]"; expected "Union[str, bytes, SupportsBytes, List[int]]"
```

#### Explanation
`Transaction.data` should be one of `str, bytes, SupportsBytes, List[int]` but `Initcode` inherits from `Code` which (unfortunately) allows `bytecode` to be of type `None`, which is not expected. 

#### Solution
This resolves the mypy issue, but is also burdensome and very ugly, as it implies in this test code that the bytecode could be `None`, which it never is.
```suggestion
        data=initcode.bytecode if initcode.bytecode is not None else bytes(),
```
 

#### Possible Improvement
I think we should make the default value for the optional `Transaction.data` compatible with the default value for the optional value for `Code.bytecode`.

danceratopz (commented):
Hey @winsvega, I looked at the mypy/typehint problems in Github Actions.

Both of these problems require an unnecessary amount of familiarity with typehints and our framework code in order to avoid:
- The `post` error fix is just cumbersome and clunky (if you search through our code, many test implementers have simply used `# type: ignore` to resolve this, as it's not obvious as all!
- The `Transaction.data` fix is even worse. Both clunky and misleading.

We should improve the framework so that test implementers don't trip other these!

See suggestions below that resolve the error messages! As the framework stands now.

tests/cancun/eip6780_selfdestruct/test_dynamic_create2_selfdestruct_collision.py
```diff
@@ -0,0 +1,158 @@
...
+        ),
+        address_code: Account(
+            balance=1000000000000000000,
+            nonce=0,
+            code=Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+            + Op.SSTORE(first_create2_result, Op.CREATE2(7, 0, Op.CALLDATASIZE(), create2_salt))
+            + Op.CALL(Op.GAS(), Op.SLOAD(1), 3, 0, 0, 0, 0)
+            + Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+            + Op.SSTORE(second_create2_result, Op.CREATE2(5, 0, Op.CALLDATASIZE(), create2_salt)),
+            storage={0x02: 0xFF},
```
marioevz, line 102:
```suggestion
            storage={
                first_create2_result: 0xFF,
                second_create2_result: 0xFF,
            }
```
We need to also preset `0x01` to something non-zero, to make sure that the transaction did not OOG'd.

I tried this locally and at the moment it is still running out of gas if the contract was pre-existing, because the `CREATE2` fails and then consumes 63/64 of the remaining tx's gas, and it runs out of gas by the `Op.CALL(...`.

I still think we need to make a subcall to a contract that only does:
```python
Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+ Op.MSTORE(0, Op.CREATE2(7, 0, Op.CALLDATASIZE(), create2_salt))
+ Op.RETURN(0, 32)
```
And then call that contract with a fraction of the tx's total gas, so it only consumes that fraction with the failing `CREATE2` and it let's the top level contract finish.

winsvega:
I get confused when you rewrite my code.
its ok now, but in the future please leave the comments.

marioevz:
> I get confused when you rewrite my code. its ok now, but in the future please leave the comments.

Sorry, will do!

marioevz (approved):

Issue #385, opened 2024-01-16
test: Check balance of an account sending a type-3 transaction mid-execution


marioevz:
Add a test where we store the balance of the sender mid-execution when the transaction is a type-3 transaction:

```python
contract_code = Op.SSTORE(0, Op.BALANCE(Op.ORIGIN))
...
tx = Transaction(
            ty=Spec.BLOB_TX_TYPE,
            nonce=tx_i,
            to=contract_address,
            value=tx_value,
            gas_limit=tx_gas,
            data=tx_calldata,
            max_fee_per_gas=tx_max_fee_per_gas,
            max_priority_fee_per_gas=tx_max_priority_fee_per_gas,
            max_fee_per_blob_gas=tx_max_fee_per_blob_gas,
            access_list=tx_access_list,
            blob_versioned_hashes=blob_hashes,
            error=tx_error if tx_i == (len(blob_hashes_per_tx) - 1) else None,
        )
```

Test using a similar parametrization to `tests/cancun/eip4844_blobs/test_blob_txs.py::test_insufficient_balance_blob_tx`.

This is to verify that all clients are subtracting the blob gas from the account before executing the transaction.

marioevz:
Also add a case where the type-3 sender receives funds mid-execution.

PR #407, opened 2024-01-28
feat(tests): Add blob gas subtraction order tests

spencer-tb:
## 🗒️ Description
Adds parameterized tests to ensure that the blob gas fee is subtracted from the sender's balance before a blob transaction is executed. Includes the case where the sender recieves funds before the blob tx is executed.

## 🔗 Related Issues
As requested by @marioevz: https://github.com/ethereum/execution-spec-tests/issues/385

## ✅ Checklist

### DIFF
--- a/tests/cancun/eip4844_blobs/test_blob_txs.py
+++ b/tests/cancun/eip4844_blobs/test_blob_txs.py
@@ -482,33 +482,16 @@ def header_verify(
 
 
 @pytest.fixture
-def all_blob_gas_used(
-    fork: Fork,
-    txs: List[Transaction],
-    block_number: int,
-    block_timestamp: int,
-) -> Optional[int | Removable]:
-    """
-    Calculates the blob gas used by the test block taking into account failed transactions.
-    """
-    if not fork.header_blob_gas_used_required(
-        block_number=block_number, timestamp=block_timestamp
-    ):
-        return Header.EMPTY_FIELD
-    return sum([Spec.get_total_blob_gas(tx) for tx in txs])
-
-
-@pytest.fixture
 def rlp_modifier(
-    all_blob_gas_used: Optional[int | Removable],
+    expected_blob_gas_used: Optional[int | Removable],
 ) -> Optional[Header]:
     """
     Header fields to modify on the output block in the BlockchainTest.
     """
-    if all_blob_gas_used == Header.EMPTY_FIELD:
+    if expected_blob_gas_used == Header.EMPTY_FIELD:
         return None
     return Header(
-        blob_gas_used=all_blob_gas_used,
+        blob_gas_used=expected_blob_gas_used,
     )
 
 
--- a/tests/cancun/eip4844_blobs/test_blob_txs.py
+++ b/tests/cancun/eip4844_blobs/test_blob_txs.py
@@ -218,8 +218,10 @@ def total_account_minimum_balance(  # noqa: D103
 def total_account_transactions_fee(  # noqa: D103
     tx_gas: int,
     tx_value: int,
-    block_fee_per_gas: int,
     blob_gasprice: int,
+    block_fee_per_gas: int,
+    tx_max_fee_per_gas: int,
+    tx_max_priority_fee_per_gas: int,
     blob_hashes_per_tx: List[List[bytes]],
 ) -> int:
     """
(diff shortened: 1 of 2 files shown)


spencer-tb:
I initially tried to verify the balance values stored within the post state of the `destination_account`.

This was easy without the additional parameterization of the `max_fee_per...` fields. 
You can find the following revision here: https://github.com/spencer-tb/execution-spec-tests/commit/4be8852d8c2084a0282d73baf7e21ef69f6184e3
Within this the entire total account minimum fee, required to execute the tx was removed (incl. the blob gas fee) before executing the tx and contract code.

However when it came to using the `max_fee_per...` fields, I maybe spend a day trying to figure out the correct formula for the balance subtraction, as the amount subtracted is less than the value for the total account minimum fee.

tests/cancun/eip4844_blobs/test_blob_txs.py
```diff
@@ -892,6 +892,59 @@ def test_sufficient_balance_blob_tx_pre_fund_tx(
...
+@pytest.mark.parametrize(
+    "tx_calldata",
+    [b"", b"\x01"],
+    ids=["no_calldata", "single_zero_calldata"],
+)
+@pytest.mark.parametrize("tx_max_fee_per_blob_gas", [1, 100])
+@pytest.mark.parametrize(
+    "tx_gas", [500_000], ids=[""]
+)  # Increase gas to account for contract code
+@pytest.mark.parametrize("mid_tx_send_amount", [0, 100])
```
marioevz, line 912: I think we should remove this parameterization in favor of always sending 100 wei, just to reduce the number of tests this produces.

marioevz:
> I initially tried to verify the balance values stored within the post state of the `destination_account`.
> 
> This was easy without the additional parameterization of the `max_fee_per...` fields. You can find the following revision here: [spencer-tb@4be8852](https://github.com/spencer-tb/execution-spec-tests/commit/4be8852d8c2084a0282d73baf7e21ef69f6184e3) Within this the entire total account minimum fee, required to execute the tx was removed (incl. the blob gas fee) before executing the tx and contract code.
> 
> However when it came to using the `max_fee_per...` fields, I maybe spent a day trying to figure out the correct formula for the balance subtraction, as the amount subtracted is less than the value for the total account minimum fee.
> 
> Would you like me continue and add a check for the balance? Similarly sstore the return value of the call to send funds?

So if we try to check the stored balance, the balance should be zero only for cases where
a) `tx.max_fee_per_gas == block.base_fee_per_gas`
b) `tx.max_fee_per_blob_gas == block.blob_base_fee_per_gas`

both (a) and (b) are true, because if you check the validation steps in 1559 (which are very similar to 4844) https://eips.ethereum.org/EIPS/eip-1559, the formula for the minimum balance for the transaction to be intrinsically valid takes `tx.max_fee_per_gas` and `tx.max_fee_per_blob_gas` into account, whereas the actual balance subtraction, which happens right before the transaction execution, uses `block.base_fee_per_gas` and `block.blob_base_fee_per_gas` respectively.

So this is why we fund the account with the maximum amount it would have to pay in any block, because otherwise the transaction is invalid.

tests/cancun/eip4844_blobs/test_blob_txs.py
```diff
@@ -892,6 +892,59 @@ def test_sufficient_balance_blob_tx_pre_fund_tx(
...
+    assert len(txs) == 1
+    pre[destination_account] = Account(
+        balance=mid_tx_send_amount,
+        code=Op.SSTORE(0, Op.BALANCE(Op.ORIGIN))
+        + Op.CALL(Op.GAS, Op.ORIGIN, mid_tx_send_amount, 0, 0, 0, 0)
+        + Op.SSTORE(1, Op.BALANCE(Op.ORIGIN)),
+    )
+    state_test(
+        pre=pre,
+        post={},
```
marioevz, line 942: My only worry about not checking anything in the post is that the tx reverts for any reason, then we didn't actually verify anything.

marioevz (commented):
Looks really good, just a few comments.

marioevz:
To calculate the actual balance we can use a similar approach to function `total_account_minimum_balance`, but instead of calculating using `tx_max_fee_per_blob_gas` and `tx_max_fee_per_gas`, we use the actual blob base fee and normal base fee of the block (functions `blob_gasprice` and `block_fee_per_gas` pytest fixtures), and we subtract that from the minimum account balance.

spencer-tb:
We now have the `total_account_transactions_fee` fixture within: https://github.com/ethereum/execution-spec-tests/pull/407/commits/b49c5b043cbb5d1b7ca7ec9a83ba503feba11782 alongside the updates to the tests for the post state balance checks. Some small naming tweaks were made to the `total_account_minimum_balance` fixture.

Within: https://github.com/ethereum/execution-spec-tests/pull/407/commits/f226e5473845e42a6926947eb180772a86b2663f I removed a duplicate fixture.

I am just working on adding the `max_priority_fee_per_gas` parameterization to the calculation.

spencer-tb:
We should be a go now!

marioevz (approved):
Awesome! Thanks!

PR #418, opened 2024-01-30
bug(pytest, forks): fix `yul_test` marker and improve `solc --evm-version` param handling

danceratopz:
## 🗒️ Description

1. Fixes `fill -m yul_test` which failed to filter tests that were (dynamically) marked as a yul test.
2. Defines which forks our supported solc versions can use as an argument to `--evm-version`.
3. Uses the definition from 2. to better pick an argument to `--evm-version` if the current fork is not supported.
4. Exit `fill` early in case of raising an incompatible solc. Previously this would return an error for every collected test:
    ```
    ✦ ➜ fill tests --solc-bin ~/bin/solc-0.8.19
    Exit: Unsupported solc version: 0.8.19+commit.7dd6d404.Linux.gpp. Minimum required version is 0.8.20
    ```

Opened as draft, remaining to do:
- [x] Add tests for new `ethereum_test_forks.helpers`.

Won't add pytest framework tests for `pytest_plugins.test_filler`. Would be nice, but we currently don't currently manage available solc versions, so results are undefined.

## 🔗 Related Issues
None.

## ✅ Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,6 +4,19 @@ Test fixtures for use by clients are available for each release on the [Github r
 
 **Key:** ✨ = New, 🐞 = Fixed, 🔀 = Changed, 💥 = Breaking change.
 
+## 🔜 [Unreleased] https://github.com/ethereum/execution-spec-tests/releases/tag/v-Unreleased) - 2024-xx-xx
+
+### 🧪 Test Cases
+
+### 🛠️ Framework
+
+- ✨ Improve handling of the argument passed to `solc --evm-version` when compiling Yul code ([#418](https://github.com/ethereum/execution-spec-tests/pull/418)).
+- 🐞 Fix `fill -m yul_test` which failed to filter tests that are (dynamically) marked as a yul test ([#418](https://github.com/ethereum/execution-spec-tests/pull/418)).
+
+### 🔧 EVM Tools
+
+### 📋 Misc
+
 ## [v2.1.0](https://github.com/ethereum/execution-spec-tests/releases/tag/v2.1.0) - 2024-01-29: 🐍🏖️ Cancun
 
 Release [v2.1.0](https://github.com/ethereum/execution-spec-tests/releases/tag/v2.1.0) primarily fixes a small bug introduced within the previous release where transition forks are used within the new `StateTest` format. This was highlighted by @chfast within #405 (https://github.com/ethereum/execution-spec-tests/issues/405), where the fork name `ShanghaiToCancunAtTime15k` was found within state tests.
--- a/src/ethereum_test_forks/base_fork.py
+++ b/src/ethereum_test_forks/base_fork.py
@@ -8,28 +8,6 @@ from semver import Version
 
 from .base_decorators import prefer_transition_to_method
 
-SOLC_FORKS_0_8_20 = [
-    "homestead",
-    "tangerineWhistle",
-    "spuriousDragon",
-    "byzantium",
-    "constantinople",
-    "petersburg",
-    "istanbul",
-    "berlin",
-    "london",
-    "paris",
-    "shanghai",
-]
-SOLC_VERSION_TO_SUPPORTED_FORKS = {
-    Version.parse("0.8.20"): SOLC_FORKS_0_8_20,
-    Version.parse("0.8.21"): SOLC_FORKS_0_8_20,
-    Version.parse("0.8.22"): SOLC_FORKS_0_8_20,
-    Version.parse("0.8.23"): SOLC_FORKS_0_8_20,
-    # TBA: 0.8.24 is the current dev version.
-    Version.parse("0.8.24"): SOLC_FORKS_0_8_20,
-}
-
 
 class ForkAttribute(Protocol):
     """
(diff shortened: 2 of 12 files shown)


src/ethereum_test_forks/base_fork.py
```diff
@@ -4,8 +4,32 @@
...
+    "shanghai",
+]
+SOLC_VERSION_TO_SUPPORTED_FORKS = {
+    Version.parse("0.8.20"): SOLC_FORKS_0_8_20,
+    Version.parse("0.8.21"): SOLC_FORKS_0_8_20,
+    Version.parse("0.8.22"): SOLC_FORKS_0_8_20,
+    Version.parse("0.8.23"): SOLC_FORKS_0_8_20,
+    # TBA: 0.8.24 is the current dev version.
+    Version.parse("0.8.24"): SOLC_FORKS_0_8_20,
+}
```
danceratopz, line 31: Would have preferred to have this in `ethereum_test_tools.code.yul`. I tried and had an issue with circular imports, wasn't sure how best to resolve it.

src/ethereum_test_forks/forks/forks.py
```diff
@@ -150,7 +150,7 @@ def pre_allocation(cls, block_number: int = 0, timestamp: int = 0) -> Mapping:
         return {}
 
 
-class Homestead(Frontier):
+class Homestead(Frontier, solc_name="homestead"):
```
danceratopz, line 153: @marioevz I decided to explicitly specify the `solc_name` as the `solc_name()` method didn't actually return the correct name (even though it was correctly handled in `ethereum_test_tools.code.yul` with `lower()`).

src/ethereum_test_forks/base_fork.py
```diff
@@ -258,6 +282,16 @@ def solc_name(cls, block_number: int = 0, timestamp: int = 0) -> str:
         """
         pass
 
+    @classmethod
+    def is_supported_by_solc(cls, version: Version) -> bool:
+        """
+        Returns True if the fork is supported by solc (--evm-version).
+        """
+        assert (
+            version in SOLC_VERSION_TO_SUPPORTED_FORKS.keys()
```
danceratopz, line 291: It's really interesting that this works even if you parse a very specific build, e.g., `0.8.20+commit.a1b79de6.Linux.g++`, it will match 0.8.20. `semver` librrary magic.

src/pytest_plugins/test_filler/test_filler.py
```diff
@@ -487,19 +497,8 @@ def pytest_collection_modifyitems(config, items):
...
-    """
-    if "yul" in item.fixturenames:
-        marker = pytest.mark.yul_test()
-        item.add_marker(marker)
-        if any([fork.name() in item.name for fork in get_development_forks()]):
-            if item.config.getoption("verbose") >= 1:
-                warnings.warn("Compiling Yul for with Shanghai, not the target fork.")
-            item.add_marker(pytest.mark.compile_yul_with("Shanghai"))
+        if "yul" in item.fixturenames:
+            item.add_marker(pytest.mark.yul_test)
```
danceratopz, line 502: The `compile_yul_with()` functionality has been directly added to the `yul` fixture. It's more logical to have it there.

src/pytest_plugins/test_filler/test_filler.py
```diff
@@ -462,7 +473,6 @@ def pytest_generate_tests(metafunc):
             )
 
 
-@pytest.hookimpl(trylast=True)
```
danceratopz, line 465: This was one part necessary to fix `fill -m yul_test`.

src/pytest_plugins/test_filler/test_filler.py
```diff
@@ -487,19 +497,8 @@ def pytest_collection_modifyitems(config, items):
...
-    """
-    if "yul" in item.fixturenames:
-        marker = pytest.mark.yul_test()
-        item.add_marker(marker)
-        if any([fork.name() in item.name for fork in get_development_forks()]):
-            if item.config.getoption("verbose") >= 1:
-                warnings.warn("Compiling Yul for with Shanghai, not the target fork.")
-            item.add_marker(pytest.mark.compile_yul_with("Shanghai"))
+        if "yul" in item.fixturenames:
+            item.add_marker(pytest.mark.yul_test)
```
danceratopz, line 501: And this adds the `yul_test` marker, which was previously incorrectly added in the `pytest_runtest_setup` hook; in `pytest_runtest_setup` it was added too late (after collection had finished) and therefore it was no longer possible to filter by the marker.

danceratopz (commented):
Just a few comments.

src/ethereum_test_forks/base_fork.py
```diff
@@ -258,6 +260,14 @@ def solc_name(cls, block_number: int = 0, timestamp: int = 0) -> str:
         """
         pass
 
+    @classmethod
+    @abstractmethod
+    def solc_min_version(cls, block_number: int = 0, timestamp: int = 0) -> Version:
```
danceratopz, line 265: Couldn't we omit the `block_number` and `timestamp` args here?

spencer-tb (approved):
LGTM! I really like the approach with `solc_min_version()`

docs/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ Test fixtures for use by clients are available for each release on the [Github r
 
 **Key:** ✨ = New, 🐞 = Fixed, 🔀 = Changed, 💥 = Breaking change.
 
+## 🔜 [Unreleased] https://github.com/ethereum/execution-spec-tests/releases/tag/v-Unreleased) - 2024-xx-xx
```
spencer-tb, line 7:
```suggestion
## 🔜 [Unreleased] (https://github.com/ethereum/execution-spec-tests/releases/tag/v-Unreleased) - 2024-xx-xx
```

src/ethereum_test_forks/base_fork.py
marioevz: The problem is on the transition forks generator, if it's an abstract method, then it will try to perform some actions on it, and it needs the parameters to be `block_number` and `timestamp` precisely.

See here: https://github.com/ethereum/execution-spec-tests/blob/77f887a7d3d1ea446d282c570fb394e276e8d5f3/src/ethereum_test_forks/transition_base_fork.py#L71-L77

I think we could get away with not marking this method as abstract maybe?

danceratopz: Nice!

PR #422, opened 2024-01-31
refactor(fw,tests): fix address and hash types

### LINKED ISSUE #387: chore: Automatically handle Address types on Opcode calls (`PUSH20`)
We currently have to use `PUSH20` in order to properly push an address into the stack when using an address into an opcode.

E.g.
```python
some_address = to_address(0x600)
...
code = Op.CALL(Op.GAS(), Op.PUSH20(some_address), 0, 0, 0, 0, 0)
```

The `Op.PUSH20` seems confusing, because if we do:

```python
some_address = 0x600
...
code = Op.CALL(Op.GAS(), some_address, 0, 0, 0, 0, 0)
```

then it's unnecessary, because integers are treated differently when put into the stack, but `to_address` returns a string.

So we either:
- Change `to_address` to return an `Address` type, and then handle `Address` types as another special case when it's received as a parameter of an `Op` call
- Don't ever use `to_address` or `Address` in tests and always treat them as integers

First approach seems better for me, because it's nice to have the a clearly defined type for something as special as addresses.

marioevz:
## 🗒️ Description
- Creates the `common/base_types.py` file, which includes basic types other complex types use, such as `Address`, `Hash`, and other types that do not have any other dependency.
- Completely removes `to_address`, `to_hash` and `to_hash_bytes` in favor of using the `Address` and `Hash` classes directly.
- Adds `Address` to the list of types an `Op` callable can receive, and, instead of treating the `Address` as bytecode, it automatically detects the type and adds the appropriate `Op.PUSH20`.
- `FixedSizeBytes` types can now be compared to strings and integers
- Removes `Op.PUSH20` across all tests when an `Address` type is used as a parameter of an `Op`.

Nice thing about this change is that it produces zero fixture changes, and slightly reduces lines of code.

TODO: Update docs.

## 🔗 Related Issues
Fixes #387

## ✅ Checklist


spencer-tb (approved):
LGTM! Had another look locally to see if I could find anything else, got matching fixtures too :D

spencer-tb:
Not sure what docs should be added, the only thing I can think of adding the usage of the `Address` type when initializing accounts, within pre & post states etc here:
https://github.com/ethereum/execution-spec-tests/blob/main/docs/writing_tests/writing_a_new_test.md?plain=1#L103-L206

This PR got me thinking: I wonder if we should force all addresses to be of the `Address` type, pre and post to be `Alloc` types, and storage to be `Storage`. If we check for this when filling, it **_might_** catch some an unintentional additions.

Issue #417, opened 2024-01-30
invalid block json description


winsvega:
`(cancun/eip4844_blobs/tests/tmp/tmptest_e63dbf6c_a8b3_42b0_92dd_7f524cd5686f/test_excess_blob_gas_fork_transition.py::test_invalid_post_fork_block_without_blob_fields[fork_ShanghaiToCancunAtTime15k-blockchain_test-excess_blob_gas_missing_False-blob_gas_used_missing_True], step: BlockchainTest)`

```
Error: BlockchainTestInFilled convertion error: BlockchainTestBlock convertion error: BlockHeader4844 parse error: Expected field 'blobGasUsed' not found in config: BlockHeader4844 blockHeader
"blockHeader" : {
    "baseFeePerGas" : "0x07",
    "bloom" : "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "coinbase" : "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
    "difficulty" : "0x00",
    "excessBlobGas" : "0x00",
    "extraData" : "0x",
    "gasLimit" : "0x016345785d8a0000",
    "gasUsed" : "0x00",
    "hash" : "0x5155c84b69a8af6ee6f06d7414f1e434d01dbd8a5eab1ac429b246525e319e16",
    "mixHash" : "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce" : "0x0000000000000000",
    "number" : "0x10",
    "parentBeaconBlockRoot" : "0x0000000000000000000000000000000000000000000000000000000000000000",
    "parentHash" : "0x64177ba1942c915cc9c4d95a69b27befb29a3e0c2a56882fba4fb8c5bf367fea",
    "receiptTrie" : "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "stateRoot" : "0x28415a6a7e08c2fb01beaa4d4e8397ed3c51f8c7400d3619803652afedf658c1",
    "timestamp" : "0x3a98",
[...]

Issue #415, opened 2024-01-30
rlp_decoded vs provided rlp inconsistency


winsvega:
have this issue:
```
Error: Error checking invalid block rlp.tx vs rlp_decoded tr `{
    "data" : "0x",
    "gasLimit" : "0x07a120",
    "nonce" : "0x00",
    "to" : "",
    "value" : "0x01",
    "v" : "0x01",
    "r" : "0x990fd9e8068282f2f657fda1fc17ba168b6ccb064a796796a04647268efdc0ba",
    "s" : "0x5497398a0d1bfe95930e450561b2a88ece16b0212a4784e92b0a7cd3921852c4",
    "chainId" : "0x01",
    "type" : "0x03",
    "maxFeePerGas" : "0x07",
    "maxPriorityFeePerGas" : "0x00",
    "accessList" : [
    ],
    "maxFeePerBlobGas" : "0x01",
    "blobVersionedHashes" : [
        "0x0100000000000000000000000000000000000000000000000000000000000000"
    ]
}` 
is different to rlp_decoded tr 
`{
    "data" : "0x",
    "gasLimit" : "0x07a120",
    "nonce" : "0x00",
    "to" : "0x0000000000000000000000000000000000000100",
    "value" : "0x01",
    "v" : "0x01",
    "r" : "0x8633745d9ccae895b1cd9b35473bf2c2f19230b32ebecf93fcf6e98e08868a7f",
    "s" : "0x2df5e81e127ad763e99e70e134ea214d74d47e80787950b1959551c7b61f68ea",
    "sender" : "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b",
    "chainId" : "0x01",
    "type" : "0x03",
    "maxFeePerGas" : "0x07",
    "maxPriorityFeePerGas" : "0x00",
    "accessList" : [
    ],
    "maxFeePerBlobGas" : "0x01",
    "blobVersionedHashes" : [
        "0x0100000000000000000000000000000000000000000000000000000000000000"
    ]
}`) (cancun/eip4844_blobs/tests/tmp/tmptest_f077683c_5413_4e48_a90e_1c767f88f5d6/test_blob_txs.py::test_invalid_blob_tx_contract_creation[fork_Cancun-blockchain_test-], fork: Cancun, block: 1)
```

the transaction that appars in rlp_decoded is different to the one that is actually passed in rlp field.

winsvega:
actually I thought about rlp_decoded. 
[...]

PR #427, opened 2024-02-07
feat(fixtures): do not export rlp_decoded if there is BlockException.RLP_STRUCTURES_ENCODING

winsvega:
## 🗒️ Description
Do not export rlp_decoded filled blockchain field if there is BlockException.RLP_STRUCTURES_ENCODING exception

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/415
https://github.com/ethereum/execution-spec-tests/issues/417

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/exceptions/exceptions.py
+++ b/src/ethereum_test_tools/exceptions/exceptions.py
@@ -43,6 +43,12 @@ class ExceptionBase(Enum):
     Base class for exceptions.
     """
 
+    def __contains__(self, exception) -> bool:
+        """
+        Checks if provided exception is equal to this
+        """
+        return self == exception
+
     def __or__(
         self,
         other: Union["TransactionException", "BlockException", ExceptionList],
@@ -146,6 +152,10 @@ class BlockException(ExceptionBase):
     INCORRECT_EXCESS_BLOB_GAS = auto()
     """
     Block's excess blob gas in header is incorrect.
+    """
+    RLP_STRUCTURES_ENCODING = auto()
+    """
+    Block's rlp encoding is valid but ethereum structures in it are invalid
     """
 
 
--- a/src/ethereum_test_tools/spec/blockchain/blockchain_test.py
+++ b/src/ethereum_test_tools/spec/blockchain/blockchain_test.py
@@ -38,6 +38,7 @@ from ..base.base_test import (
 from ..debugging import print_traces
 from .types import (
     Block,
+    BlockException,
     Fixture,
     FixtureBlock,
     FixtureEngineNewPayload,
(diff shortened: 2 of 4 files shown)


src/ethereum_test_tools/spec/blockchain/types.py
```diff
@@ -1171,3 +1171,17 @@ def format(cls) -> FixtureFormats:
         fixture.
         """
         return FixtureFormats.BLOCKCHAIN_TEST_HIVE
+
+
+def contains_exception(
```
marioevz, line 1176: I think we could instead implement `def __contains__(self, item) -> bool` in `class ExceptionBase` (see [docs.python.org](https://docs.python.org/2/reference/datamodel.html#object.__contains__)), which could simply compare for equality, and we should be able to just check by doing `BlockException.RLP_ENCODING in block.exception`.

src/ethereum_test_tools/exceptions/exceptions.py
```diff
@@ -147,6 +147,10 @@ class BlockException(ExceptionBase):
     """
     Block's excess blob gas in header is incorrect.
     """
+    RLP_ENCODING = auto()
```
marioevz, line 150: I think `RLP_ENCODING` should replace the existing `INCORRECT_BLOCK_FORMAT`, unless there's a block formatting error that does not result from an RLP encoding issue?

src/ethereum_test_tools/exceptions/exceptions.py
```diff
@@ -147,6 +147,10 @@ class BlockException(ExceptionBase):
...
+    """
```
marioevz, line 151: We should also add `RLP_ENCODING` to `class TransactionException` and use that for tests that define the exception at the `Transaction` object.

We would then also the need to check like:
```python
if TransactionException.RLP_ENCODING in block.exception or BlockException.RLP_ENCODING in block.exception:
   ...
```

marioevz (commented):
Thanks! I added some comments for a suggested refactor :+1:

src/ethereum_test_tools/exceptions/exceptions.py
winsvega: You right this sound like an exception related to rlp itself.

1) rlp invalid - RLP_ENCODING  (currently there are no such tests)
2) eth structure invalid (failure to read rlp into eth structures)

src/ethereum_test_tools/spec/blockchain/types.py
marioevz: I referenced a super old python version in here :see_no_evil: 
Correct link: https://docs.python.org/3.10/reference/datamodel.html#object.__contains__

src/ethereum_test_tools/exceptions/exceptions.py
winsvega: I emphasized on that this exception is about ethereum structures not RLP being invalid

PR #431, opened 2024-02-08
feat(fw): add optional `verify_sync` flag to hive blockchain tests

marioevz:
## 🗒️ Description
Adds an optional `verify_sync` flag for hive (or blockchain Engine API tests) to pick up and test a second client syncing to the first client.

Syncing is tricky to some clients and we actually need to send a full header via `engine_newPayloadVN` for some of them to start the sync process.

So this change also appends an empty block on top of the latest head of the test to use as syncing head.

## 🔗 Related Issues
None.

## ✅ Checklist


spencer-tb:
I like this a lot! 💯 

Are you thinking that syncing should be specific to certain tests, after processing and setting the head post complex scenario, or applied to the end of all tests? Or even simple specific sync tests for each fork?

I tried running this modification for some the `blob_txs` tests, using your hive branch and it worked well! Personally it seems reasonable to do this for all tests.

One question arises for me: instead of creating a sync payload could we not use the last engine new payload in the test?

marioevz:
> One question arises for me: instead of creating a sync payload could we not use the last engine new payload in the test?

The problem is that if the last engine payload is the one that triggers some special scenario you'd like to test under sync, and you provide this payload via the engine API, then we are not really testing syncing because the payload arrives in full via the engine api. Even in cases where we have more than one payload, we won't be testing syncing for the last one.

spencer-tb (approved):
LGTM - added the changelog.

PR #434, opened 2024-02-09
chore(fw,tests): Transaction nonce as `Iterator[int]`

marioevz:
## 🗒️ Description
Allows more broad usage of iterators in some fields such as:
- BlockchainTest: blocks -> Iterable
- Block: txs -> Iterable
- Block: withdrawals -> Iterable
- Transaction: nonce -> Iterator

This is now possible:
```python
nonce = count()
...
tx1 = Transaction(nonce=nonce)
tx2 = Transaction(nonce=nonce)
```

Although, when refactoring some of the tests I realized most usages of this pattern didn't really need it :upside_down_face: 

Anyway, it's nice to allow it. Wdyt? @spencer-tb @danceratopz 

## 🔗 Related Issues
None

## ✅ Checklist


marioevz:
Another pattern I saw is something like this:
```python
txs= [
Transaction(
        ty=0x0,
        nonce=0,
        to=account1,
        data=data1,
        gas_limit=500000,
        gas_price=10,
        protected=True,
    ),
Transaction(
        ty=0x0,
        nonce=0,
        to=account2,
        data=data2,
        gas_limit=500000,
        gas_price=10,
        protected=True,
    ),
```
I think we can replace it with something like
```python
Transactions(
nonce=count(),
to=[account1, account2],
data=[data1,data2]
gas_limit=500_000,
gas_price=10,
)
```
Would this be more useful? Basically just a generator that checks all kwargs and if the value is:
- `isinstance(Iterator, arg)`, nothing to do (use `next(arg)` on each iteration)
- `isinstance(Iterable, arg)` (but not string, nor bytes), do `iter(arg)`
- Neither, use `itertools.cycle`

Until any of the iterators is exhausted. `nonce` can be optional and just defaults to a new `itertools.count` if `None`.

marioevz:
Implemented previous comment in #435, and it's much more useful, and supersedes all changes in this PR.

Issue #442, opened 2024-02-16
feat(fw): add versioning to the `_info` section of fixtures


spencer-tb:
### Description

1) Create a `fill` version flag `--version` that displays the existing framework version:
    - Ideally this will adhere to the latest release of the repo plus the latest commit
    - Something like: `v0.1.0-06bd62b`
 2) Add this version to the  `_info` section of fixtures, within `BaseTest` during the filling procedure.

PR #436, opened 2024-02-13
feat(fw|ci): add fixture diff utility within CI and local CLI

spencer-tb:
## 🗒️ Description

I thought I would give this a try. So far I've added an entry point `diff_fixtures.py` (like @danceratopz) suggested. This file will be used for both CI workflows and additionally for the CLI interface for local fixture diff'ing.

### CI Workflows

`diff_fixtures.py` will have functions that the CI/CD workflows can use. My understanding is that we only need to be able to tell what files have changed. So far I am planning on adding additional json files that are created using these functions within the workflows, containing:
1) An **ordered** map of the fixture file name, to its content hashed (minus the `_info` section). Allowing us to check what files have changed specifically.
2) A cumulative hash of each fixture hash. Allowing us to simply check anything fixtures have changed.

Here is an example structure:
```json
{
    "cumulative_hash": "ae79b35b75b95cea2fca707cb0750d08a65950160a63b50e5d9d4bba360e3a7e",
    "hash_map": {
        "cancun/eip4844_blobs/point_evaluation_precompile/point_evaluation_precompile_before_fork.json": [
            "6cb959f9baa5c76d60c5042c2434981bbc5994d065f2122b30f4998c9e075fa6"
        ],
        "cancun/eip4844_blobs/point_evaluation_precompile/point_evaluation_precompile_calls.json": [
            "e4e1affc2687a817d5cc971b3a6a73583b001c16091fefef0500d4f67514dbb5"
        ],
        ...
```
An additional step will be added to the build and release workflow, so that every time a PR is merged into main a new json file is created with the hashes. Similarly, I will add an equivalent file that is generated for each commit within a PR. This will be a separate workflow that will be added as an extenstion to the existing tox verifacations.

At this point we have 2 files that a comparison can be drawn from, using the cumulative hash first, and then the file by file hash afterwards. If we want to be more granular with diffing we can in the future. We can also only diff the files that we expect to change, to speed up the workflow.

Maybe a trigger within the PR will be nice to determine when the fixture diff should occur.

### CLI Interface

The CLI command, I'm calling it `dfx` for now will be able to generate the same file locally and pull in the the json file generated from main, such that a diff can be created.

This will be more granular with more information, where as the workflows will ideally signal if a diff exists.

If we want we can add this to tox, but my gut feeling is that it should be a seperate utility.

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/388

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
[...]

PR #443, opened 2024-02-16
feat(fw|ci): add EEST versioning to framework

spencer-tb:
## 🗒️ Description
Adds versioning to the framework. Uses the assumption that the current version of the framework matches that of the latest release. The full version is denoted as: `execution-spec-tests v2.1.0-1e4ef51`.

- A new field is added within the fixtures `_info` section during filling:
```json   
"_info": {
     "framework-version": "execution-spec-tests v2.1.0-1e4ef51"
 }
 ```
 
- A CLI flag is added to fill, `fill --version-info` which outputs the following:
 ```
 Exit: 'Displaying framework version information'
> execution-spec-tests v2.1.0-1e4ef51c
```

- A new workflow `release.yaml` is added with jobs to perform when a release is published. 
  - The versioning is all achieved locally by updating the `version` field within `setup.cfg`. 
  - To avoid us having to update this manually, the workflow currently bumps the version within `setup.cfg`. 
  - It additionally updates the changelog, completing jobs we must perform manully after publishing a release. 
  - The workflow is finalized by creating a PR with the relavant changes to `main`.

## 🔗 Related Issues
Resolves: https://github.com/ethereum/execution-spec-tests/issues/442
Required for: https://github.com/ethereum/execution-spec-tests/pull/436

#### Pre-Review
- [x] Consider moving `get_framework_version()` elsewhere and not within base spec. (Created `ethereum_test_tools/utilities`)
- [x] Consider adding additional jobs to perform within the release workflow, including automatic changelog updates. (Added changelog updates to workflow)
#### Post-Review
- [ ] Discuss whether we want seperate versioning for fixtures and test/framework code.
- [ ] Replace custom versioning solution with existing solution (versioneer).
- [ ] Discuss optimal method or flow for creating releases.
- [ ] ...

## ✅ Checklist
- [] All: Considered squashing commits to improve commit history.


spencer-tb:
Requires that settings within https://github.com/ethereum/execution-spec-tests/settings/actions, are set to the following:
Screenshot 2024-02-17 at 11 43 32

spencer-tb:
The workflow should leave us with a PR to main with something like the following: https://github.com/spencer-tb/execution-spec-tests/pull/29

src/ethereum_test_tools/spec/base/base_test.py
```diff
@@ -100,6 +102,7 @@ def fill_info(
         self.info["filling-transition-tool"] = t8n.version()
         if ref_spec is not None:
             ref_spec.write_info(self.info)
+        self.info["framework-version"] = get_framework_version()
```
marioevz, line 105: I think we should pass this to the filler as a `str | None`, and ignore if `None`, but the main advantage would be avoid calling `get_framework_version` every fixture we need to fill info for, and reusing the data for all tests.

setup.cfg
```diff
@@ -1,9 +1,9 @@
 [metadata]
-name = test-filler
+name = execution-spec-tests
 description = Ethereum execution client test authoring framework
 long_description = file: README.md
 long_description_content_type = text/markdown
-version = 1.0.0
+version = v2.1.0
```
marioevz, line 6: Oh man, we had not updated this in a long time.

.github/workflows/post_release.yaml
```diff
@@ -0,0 +1,65 @@
...
+          fetch-depth: 0
+
+      - name: Configure git user for actions
+        run: |
+          git config --global user.email "action@github.com"
+          git config --global user.name "GitHub Action"
+
+      - name: Set environment variables
+        run: |
+          echo "NEW_VERSION=${{ github.event.release.tag_name }}" >> $GITHUB_ENV
```
marioevz, line 26: I think the problem is probably going to be that the tag points to a commit that is parent to the commit that the created PR is trying to merge, or I might be misunderstanding something.

Maybe if we do instead a manually triggered workflow that takes the new version number as parameter and does everything, even creating the tag, automatically?

marioevz (changes requested):
Some comments but looks great overall!

src/ethereum_test_tools/spec/base/base_test.py
spencer-tb: Nice, this could definitely be called only once

.github/workflows/post_release.yaml
spencer-tb: I thought on this a bit. This is true, these changes would create a PR to main after the release. So they would not be included in the release - I am not sure what was going through my mind here. Ideally we do this before the release, so everything is included before we publish.

Following your comment:
> Maybe if we do instead a manually triggered workflow that takes the new version number as parameter and does everything, even creating the tag, automatically?

I am thinking we first rename/move `CHANGELOG.md` to `RELEASE_NOTES.md`, this would now hold all the release notes and changes for every release so far. We then use`CHANGELOG.md` as the current log and release notes for what is being worked on. So in a workflow we can simply copy the contents of the changelog to the full set of release notes.

The release process could then be:

1) We update the new changelog file with everything we want to include in the new release notes, including the description. This would be the final manual PR before drafting a new release, where it allows us all a way to review what the release will look like. This PR would be labelled with something like `release`, such that it can be used to trigger a workflow when merged.

2) When this PR is merged it would start the `pre_release.yaml` workflow.  This would essentially automate all the relavant changes that have to occur before a release. First the changelog would be copied into the cumulative release notes file, and the changelog template copied into the changelog file. Secondly the version would be bumped within `setup.cfg`. This would then create a PR to main (as we are already doing here) allowing us to review these changes quickly. 

3) When the automated`pre_release.yaml` PR is merged to main, this would then trigger the `draft_release.yaml` workflow. Essentially doing what you suggested. Extract the new version, create/push the new tag with the version, and draft the new release copying the contents within the release notes. We can then review/edit the draft and publish.

What do you think? It should be pretty easy for me to add now :)

marioevz: I don't know how easy it would be to detect the PR name, with manually trigger I meant to go to https://github.com/ethereum/execution-spec-tests/actions, and under "Workflows", like in this image:
image

We can create a new workflow that is called "Prepare Release and Tag", which takes the version number here:

image

Does the automatic commit with all the appropriate changes, and pushes to main (which could be dangerous but we can bail on any minimal error) and then creates the tag that points to the commit that was just pushed.

danceratopz (commented):
Really like this! There's two things I'd like to discuss before going into details!

1. This suggestion tightly couples the execution-spec-tests package version with the fixture release version (i.e. they have the same version :slightly_smiling_face: ). This is probably ok, but it's worth briefly discussing whether they should be distinct. This is only relevant if developers start using execution-spec-tests in their packages. Then it could be worth using two semvers: one for fixtures and one the code/api. For fixtures, we should aim (imo) to reserve a major version bump for changes that require client teams to re-work their CI, a minor bump for new tests and a patch for bug-fixes to a test/fixture release. There would often be an overlap, but often a fixture release wouldn't involve a package release. If we couple these, then downstream users of an execution-spec-tests package will have a hard time making sense of versioning.

2. Should EEST use a (popular) Python package to help versioning instead of implementing a custom solution? Couple of examples below.
    There's a few reasons I'd suggest to go this route. 
    i. Keep things "more" standard, instead of a completely custom solution, we would follow the best practices used by a common package. There is no single best practice/package in Python as far as I can tell. But typically, for example, a package would get a `__version__` property that can be used within the source code. 
    ii. I think the version string returned below isn't quite sufficient (see comment below). This could be improved without an external package of course. How about something like the following:
        - `v2.1.1-g1076c97` - if the current head is a tag (in this case `v0.1.1`)
        - `v2.1.1+2.g1076c97-dev` - if the current head is 2 commits past the last tag (release)   
        - `v2.1.1+2.g1076c97.dirty` - if the current repo status is 2 commits past the last tag and dirty (local edits).
    Everything before the `+` is a semver; everything after is to be interpreted as "build metadata".
    ```
    iii. Remove custom github actions code/dependency on Github Actions. Would enable creating a release locally on the command-line (or doing a dry run of creating a release for verification purposes).

Examples of versioning packages:
- https://pypi.org/project/versioneer/, 
- https://docs.openstack.org/pbr

spencer-tb:
Requires: https://github.com/ethereum/execution-spec-tests/pull/454

setup.cfg
```diff
@@ -67,6 +67,7 @@ lint =
     flake8>=6.1.0,<7
     pep8-naming==0.13.3
     fname8>=0.0.3
+    GitPython>=3.1
```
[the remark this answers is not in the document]
spencer-tb: Whoops! Nice catch

src/pytest_plugins/test_filler/test_filler.py
```diff
@@ -126,6 +133,14 @@ def pytest_addoption(parser):
         help="Path to dump the transition tool debug output.",
     )
 
+    version_group = parser.getgroup("version_info", "Versioning of the EEST framework")
+    version_group.addoption(
+        "--version-info",
```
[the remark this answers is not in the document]
spencer-tb: If we use `--version` it conflicts with the existing flag from pytest

I assumed thats why we used `fill --test-help` instead of `fill --help`

I can look into a workaround

src/pytest_plugins/test_filler/test_filler.py
```diff
@@ -175,6 +190,10 @@ def pytest_configure(config):
             f"{Frontier.solc_min_version()}",
             returncode=pytest.ExitCode.USAGE_ERROR,
         )
+    if config.getoption("version_info"):
+        pytest.exit(
+            f"'Displaying framework version information'\n" f"> {get_framework_version()}\n"
+        )
```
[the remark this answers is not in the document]
spencer-tb: If we use: `pytest.exit(get_framework_version())`

We get:
```
➜ fill --version-info
Exit:
execution-spec-tests v2.1.0-c5a62ea
```

I figured it would be nice to have a reason for the exit as there is no way of silencing the `"Exit:\n"`

spencer-tb:
>     2. Should EEST use a (popular) Python package to help versioning instead of implementing a custom solution? Couple of examples below.
>        There's a few reasons I'd suggest to go this route.
>        i. Keep things "more" standard, instead of a completely custom solution, we would follow the best practices used by a common package. There is no single best practice/package in Python as far as I can tell. But typically, for example, a package would get a `__version__` property that can be used within the source code.
>        ii. I think the version string returned below isn't quite sufficient (see comment below). This could be improved without an external package of course. How about something like the following:
>        - `v2.1.1-g1076c97` - if the current head is a tag (in this case `v0.1.1`)
>        - `v2.1.1+2.g1076c97-dev` - if the current head is 2 commits past the last tag (release)
>        - `v2.1.1+2.g1076c97.dirty` - if the current repo status is 2 commits past the last tag and dirty (local edits).
>        Everything before the `+` is a semver; everything after is to be interpreted as "build metadata".

Big fan of this! 

Versioneer looks ideal for a versioning package :) 

I would love to discuss 1) & 3) on Wednesday's call, to determine what the best options are together.

winsvega:
UP @spencer-tb

spencer-tb:
Closing. Requirements will change in an EELS/EEST world.

Issue #447, opened 2024-02-21
pyspec makes tests with pre state accounts having zero value storage keys


winsvega:
At least Pyspecs/cancun/eip1153_tstore.tload_after_tstore still contains zero storage entries

this better be removed to better parsing in alloc.json by t8n

marioevz:
Closing as there is a new issue that should track this problem (#475).

PR #448, opened 2024-02-21
fix(tests): Remove zero pre-alloc from tests

marioevz:
## 🗒️ Description
Remove zeros from pre-alloc of some of the tests.

Assigning a zero in the `pre` of a test is unnecessary, and can produce a different behavior in some clients (should actually be defined and fixed).

Assigning a zero in `post` is necessary to verify empty keys in storage in some of the tests.

## 🔗 Related Issues
#447

## ✅ Checklist

### DIFF
--- a/tests/cancun/eip1153_tstore/test_tstorage.py
+++ b/tests/cancun/eip1153_tstore/test_tstorage.py
@@ -73,7 +73,7 @@ def test_tload_after_tstore(state_test: StateTestFiller):
 
     pre = {
         TestAddress: Account(balance=10_000_000),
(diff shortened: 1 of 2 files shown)


marioevz:
I have a test code that automatically checks for this to not happen in the future, but I'm reluctant to add it because it's just extra processing at the start of every test, I'm open to discuss it though.

spencer-tb (approved):
LGTM - personally I feel like the check is worth adding!

Although this got me thinking about the future - I wonder if we could add custom mypy plugins specific to EEST types. For this case I think we would have to define the `Pre` type. We could probably do this for other checks too.

Issue #455, opened 2024-02-25
feauture request: make Op.OOG pseudo Opcode


winsvega:
in the test we can use this parametrization:

```
@pytest.mark.parametrize("call_return", [Op.REVERT, Op.RETURN])
```

which easilty types with 

```
action=Op.MSTORE(0, Op.TLOAD(0)) + call_return(0, 32),
```

but then I need to check one more revert scenario is when the call goes OOG. 
would be nice to have just one more `Op.OOG`  that would  just put this bytes into code `Op.SHA3(0, Op.SUB(0, 1)) ` and ignore arguments

marioevz:
Closing as PR was merged.

PR #440, opened 2024-02-15
feat(tests): EIP-1153: Convert tests from `ethereum/tests`

winsvega:
## 🗒️ Description
convert vectors 
1,2,3,4,5,6,7,8,9      11,12,     16,18,    ,20
add a little parametrisation around calls
need review

EVMONE calculated coverage diff. on all mentioned converted files vs pyspec generated files
http://retesteth.ethdevops.io/temp/evmone/DIFF/
- CBC - Covered Baseline Code.
- G/L/UBC - Gained/Lost/Uncovered Baseline Coverage.
- GBC means new lines and functions covered.
- Absense of LBC red color means no coverage lost (with test_yul_coverage.json that executes a few evm instructions not invoked by Op.

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/pull/437
Tests remove PR: https://github.com/ethereum/tests/pull/1383

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](../docs/) directory.
- [x] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [x] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).

PR #457, opened 2024-02-26
feat(tests): extend Op with Op.OOG macro opcode

winsvega:
## 🗒️ Description
I think it would be handy to have test purpose Op that will trigger OOG if called
This will simplify parametrisations and code in tests when paramtrising over subcall return types
Such as RETURN, REVERT

Op.OOG(x, y)  is replaced with Op..SHA3(0, 100000000000)
alowing to use
```
@pytest.mark.parametrize("call_return", [Op.RETURN, Op.REVERT, Op.OOG])
...
contract_code = Op.MSTORE(0, Op.TLOAD(0)) + call_return(0, 32)
```

Example: 
https://github.com/ethereum/execution-spec-tests/pull/440/

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/455

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/vm/opcode.py
+++ b/src/ethereum_test_tools/vm/opcode.py
@@ -4932,7 +4932,7 @@ class Macros(Macro, Enum):
     OOG(args)
     ----
 
-    Halt execution by consuming all available gas
+    Halt execution by consuming all available gas.
 
     Inputs
     ----
@@ -4946,9 +4946,9 @@ class Macros(Macro, Enum):
     ----
     This operation will result in gasprice = 19073514453125027
     Add a little more and geth report gasprice = 30 with oog exception
-    Make it 0 - 1 and geth report gasprice > u64 error
+    Make it 0 - 1 and geth report gasprice > u64 error.
 
-    Source
+    Bytecode
     ----
     SHA3(0, 100000000000)
     """
--- a/src/ethereum_test_tools/vm/opcode.py
+++ b/src/ethereum_test_tools/vm/opcode.py
@@ -48,14 +48,13 @@ class OpcodeMacroBase(bytes):
         """
         return super().__new__(cls, *args)
 
-    def __call__(self, *args_t: Union[int, bytes, str, "Opcode", FixedSizeBytes]) -> bytes:
+    def __call__(self, *_: Union[int, bytes, str, "Opcode", FixedSizeBytes]) -> bytes:
         """
         Make OpcodeMacroBase callable, so that arguments can directly be
         provided to an Opcode in order to more conveniently generate
         bytecode (implemented in the subclass).
         """
         # ignore opcode arguments
-        args_t.count
         return bytes(self)
 
     def __str__(self) -> str:
--- a/src/ethereum_test_tools/vm/opcode.py
+++ b/src/ethereum_test_tools/vm/opcode.py
@@ -62,7 +62,7 @@ class Opcode(bytes):
         """
         Creates a new opcode instance.
         """
-        if type(opcode_or_byte) is Opcode:
+        if type(opcode_or_byte) is Opcode or type(opcode_or_byte) is Macro:
             # Required because Enum class calls the base class with the instantiated object as
             # parameter.
             return opcode_or_byte
@@ -73,7 +73,7 @@ class Opcode(bytes):
             obj.min_stack_height = min_stack_height
             obj.data_portion_length = data_portion_length
             return obj
-        assert False
+        raise TypeError("Opcode constructor '__new__' didn't return an instance!")
 
     def __call__(self, *args_t: Union[int, bytes, str, "Opcode", FixedSizeBytes]) -> bytes:
         """
@@ -4582,8 +4582,8 @@ class Opcodes(Opcode, Enum):
     Inputs
     ----
     - value: value in wei to send to the new account
-    - offset: byte offset in the memory in bytes, the initialisation code for the new account
-    - size: byte size to copy (size of the initialisation code)
+    - offset: byte offset in the memory in bytes, the initialization code for the new account
+    - size: byte size to copy (size of the initialization code)
 
     Outputs
     ----
@@ -4769,8 +4769,8 @@ class Opcodes(Opcode, Enum):
     Inputs
     ----
     - value: value in wei to send to the new account
-    - offset: byte offset in the memory in bytes, the initialisation code of the new account
-    - size: byte size to copy (size of the initialisation code)
+    - offset: byte offset in the memory in bytes, the initialization code of the new account
+    - size: byte size to copy (size of the initialization code)
     - salt: 32-byte value used to create the new account at a deterministic address
 
     Outputs
@@ -4913,7 +4913,7 @@ class Opcodes(Opcode, Enum):
 
     OOG = Macro(SHA3(0, 100000000000))
     """
-    OOG(args)
+    OOG(args) [Macro]
     ----
 
     Halt execution by consuming all available gas
@@ -4928,6 +4928,11 @@ class Opcodes(Opcode, Enum):
 
     Gas
     ----
-    
-    Source: SHA3(0, 100000000000)
+    This operation will result in gasprice = 19073514453125027
+    Add a little more and geth report gasprice = 30 with oog exception
+    Make it 0 - 1 and geth report gasprice > u64 error
+
+    Bytecode
+    ----
+    SHA3(0, 100000000000)
     """
--- a/src/ethereum_test_tools/vm/opcode.py
+++ b/src/ethereum_test_tools/vm/opcode.py
@@ -70,7 +70,7 @@ class OpcodeMacroBase(bytes):
         """
         if isinstance(other, OpcodeMacroBase):
             return self._name_ == other._name_
-        return NotImplemented
+        return False
 
 
 class Opcode(OpcodeMacroBase):
--- a/src/ethereum_test_tools/vm/opcode.py
+++ b/src/ethereum_test_tools/vm/opcode.py
@@ -30,7 +30,51 @@ def _get_int_size(n: int) -> int:
 _push_opcodes_byte_list = [bytes([0x5F + x]) for x in range(33)]
 
 
-class Opcode(bytes):
+class OpcodeMacroBase(bytes):
+    """
+    Base class for Macro and Opcode, inherits from bytes.
+
+    This class is designed to represent a base structure for individual evm opcodes
+    and opcode macros.
+    """
+
+    _name_: str
+
+    def __new__(cls, *args):
+        """
+        Since OpcodeMacroBase is never instantiated directly but through
+        subclassing, this method simply forwards the arguments to the
+        bytes constructor.
+        """
+        return super().__new__(cls, *args)
+
+    def __call__(self, *args_t: Union[int, bytes, str, "Opcode", FixedSizeBytes]) -> bytes:
+        """
+        Make OpcodeMacroBase callable, so that arguments can directly be
+        provided to an Opcode in order to more conveniently generate
+        bytecode (implemented in the subclass).
+        """
+        # ignore opcode arguments
+        args_t.count
+        return bytes(self)
+
+    def __str__(self) -> str:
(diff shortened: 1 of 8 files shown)


src/ethereum_test_tools/vm/opcode.py
```diff
@@ -25,6 +26,15 @@ def _get_int_size(n: int) -> int:
 _push_opcodes_byte_list = [bytes([0x5F + x]) for x in range(33)]
 
 
+class SpecialCodes(Enum):
```
marioevz, line 29: I think that what we're doing here is more akin to a macro maybe?
```suggestion
class Macro(Enum):
```

marioevz: Another thing is that we could leave the `class Opcodes` without any changes and define this further below as it's own separate class:

```python
class Macro(bytes):
    """
    Represents... etc..
    """
    def __call__(self, *) -> bytes:
        return self
```

```python
class Macros(Macro, Enum):
    """
    Enum containing all macros
    """
    OOG = Opcodes.SHA3(0, 100000000000)
```

So I think we should not need the `special_code` parameter in the `Opcode` class.

The only reason why we would need the `class Macro` is to make it compatible with `Opcodes` by making it callable.

The downside of this is typing: If any test explicitly wants to create, e.g. a list of opcodes that includes a macro, they need to define it as `List[Opcodes | Macros]`, but it might not be a big deal.

marioevz (commented):
I really like the idea! Just a couple of suggestions so let me know if it sounds ok to you and if it still achieves the original purpose you were thinking of.

src/ethereum_test_tools/vm/opcode.py
winsvega: Yes, I also was thinking to nest the class

winsvega:
what about this part? 

```
+    def __eq__(self, other):
+        """
+        Required to differentiate between SELFDESTRUCT and SENDALL type of cases
+        """
+        if not isinstance(other, Opcode):
+            return NotImplemented
+        return self._name_ == other._name_
```

I see that if you use SENDALL it will still be trated as SELFDESTRUCT 
not an issue, but if you assign something in the constructor of SENDALL opcode in enum list
it won't have an effect, because it falls back to SELFDESTRUCT constructor

src/ethereum_test_tools/vm/opcode.py
winsvega: I managed to encode it in the same Op enum

src/ethereum_test_tools/vm/opcode.py
```diff
@@ -188,6 +190,54 @@ def __str__(self) -> str:
...
+        """
+        Required to differentiate between SELFDESTRUCT and SENDALL type of cases
+        And to register the Macro opcodes which are defined from same bytes
+        """
+        if isinstance(other, Opcode):
+            return self._name_ == other._name_
+        return NotImplemented
+
+
+class Macro(Opcode):
```
marioevz, line 203: I still don't quite get the point of `Macro` inheriting `Opcode`.

In my opinion, I feel like it should be it's own thing separate from `Opcode`:
```python
class Macro(bytes):
    ...

class Macros(Macro, Enum):
   ...
```

I think the only advantage of having macros in the same `Enum` is that you don't have to import two different things when writing the test, but is that enough reason to have them joined together?

I'm open to discussion, is just that I wanted to see the arguments.

marioevz (commented):
Still not completely on-board with the idea that the opcodes and the macros should live in the same `Enum`, but definitely open to discuss it!

What do you guys think? @spencer-tb @danceratopz

src/ethereum_test_tools/vm/opcode.py
winsvega: Yes. I don't want to write unions everywhere.

It is easy to import, define types, and write compare statements if it all lives just in one enum.

marioevz (commented):
Some small comments.

src/ethereum_test_tools/vm/opcode.py
danceratopz: I like this addition! 

But I also don't think `OOG` should be in `Opcode`, nor should `Macro` inherit from `Opcode`. Rather `Macros` should be a separate enum class. 

I don't like how this reads as though `OOG` is actually an opcode; I'd rather be explicit:
https://github.com/ethereum/execution-spec-tests/blob/d27fe5065db560648ae7bf8edad2d182b03a8935/tests/cancun/eip1153_tstore/test_tload_reentrancy.py#L30-L31

I don't think it's worth sacrificing being explicit for the sake of saving an import. With regards to typehints, we can define a new type (name suggestions welcome! :) )
```python
OpcodeOrMarco = Opcode | Macro
```

danceratopz (changes requested):
I really like this, but I'd rather see `Macro` defined as its own, independent type.

winsvega:
Why Macro can't be inherited from Opcode?

src/ethereum_test_tools/vm/opcode.py
```diff
@@ -4861,3 +4910,24 @@ class Opcodes(Opcode, Enum):
 
     Source: [evm.codes/#FF](https://www.evm.codes/#FF)
     """
+
+    OOG = Macro(SHA3(0, 100000000000))
+    """
+    OOG(args)
```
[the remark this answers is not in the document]
danceratopz: Cleaned up in a517bda2.

src/ethereum_test_tools/vm/opcode.py
```diff
@@ -30,7 +30,51 @@ def _get_int_size(n: int) -> int:
...
+        return self._name_
+
+    def __eq__(self, other):
+        """
+        Required to differentiate between SELFDESTRUCT and SENDALL type of cases
+        And to register the Macro opcodes which are defined from same bytes
+        """
+        if isinstance(other, OpcodeMacroBase):
+            return self._name_ == other._name_
+        return NotImplemented
```
winsvega, line 74: @marioevz  if I raise NotImplementedError here it breaks many places in unit tests.

src/ethereum_test_tools/vm/opcode.py
```diff
@@ -4861,3 +4920,35 @@ class Opcodes(Opcode, Enum):
...
+
+    Fork
+    ----
+    Frontier
+
+    Gas
+    ----
+    This operation will result in gasprice = 19073514453125027
+    Add a little more and geth report gasprice = 30 with oog exception
+    Make it 0 - 1 and geth report gasprice > u64 error.
```
danceratopz, line 4949:
```suggestion
    `SHA3(0, 100000000000)` results in 19073514453125027 gas used and an OOG
    exception.
    
    Note: 
    - If a value > `100000000000` is used as second argument, the resulting geth
     trace reports gas `30` and an OOG exception.
    - `SHA3(0, SUB(0, 1))` causes a gas > u64 exception.
```

danceratopz (approved):
Looks good to me! I like the new separation between Macro and Opcode.

@winsvega Could you just check the suggested docstring change before we merge?

src/ethereum_test_tools/vm/opcode.py
```diff
@@ -30,7 +30,51 @@ def _get_int_size(n: int) -> int:
...
+
+    def __call__(self, *args_t: Union[int, bytes, str, "Opcode", FixedSizeBytes]) -> bytes:
+        """
+        Make OpcodeMacroBase callable, so that arguments can directly be
+        provided to an Opcode in order to more conveniently generate
+        bytecode (implemented in the subclass).
+        """
+        # ignore opcode arguments
+        args_t.count
+        return bytes(self)
```
marioevz, line 59:
```suggestion
    def __call__(self, *_: Union[int, bytes, str, "Opcode", FixedSizeBytes]) -> bytes:
        """
        Make OpcodeMacroBase callable, so that arguments can directly be
        provided to an Opcode in order to more conveniently generate
        bytecode (implemented in the subclass).
        """
        # ignore opcode arguments
        return bytes(self)
```
This seems to work to gracefully ignore the parameters (using a simple underscore as the name of the parameter).

src/ethereum_test_tools/vm/opcode.py
marioevz, line 74:
```suggestion
        return False
```
Solution seems to be to simply return `False` in this case. It seems like python internally is comparing our type to some other enum item, which I didn't thought when I suggested to raise here.

winsvega: this change breaks the tests

winsvega: how about we keep it 
`return bytes(self) == bytes(other)`

danceratopz: Makes sense to me!

src/ethereum_test_tools/vm/opcode.py
```diff
@@ -70,7 +70,7 @@ def __eq__(self, other):
         """
         if isinstance(other, OpcodeMacroBase):
             return self._name_ == other._name_
-        return NotImplemented
+        return False
```
winsvega, line 73: this breaks the tests

src/ethereum_test_tools/vm/opcode.py
danceratopz: 84e3072e693516f34e5021c82906a0bcb1ff46ba returns False instead of raising or returning `NotImplementedError`

danceratopz: So the fail wasn't in existing tests (tox is green), it's in the new tests in @winsvega's  `eip1153` branch. It's due to a subtle (previously unexpected?) use of `Opcode` syntax: 

https://github.com/ethereum/execution-spec-tests/blob/4e41d537664d085e4432e246e8fc7bdecaeba0f9/tests/cancun/eip1153_tstore/test_tload_reentrancy.py#L32

Here, **calling** the `Op.ADDRESS` enum (note the brackets) causes it to return the bytecode (a `bytes` object), instead of just creating an instance of `Op.ADDRESS` which is of type `Opcodes`. 

This is an easy trap to fall into, so I think this could be a good solution:
```python
    def __eq__(self, other):
        """
        Required to differentiate between SELFDESTRUCT and SENDALL type of cases
        And to register the Macro opcodes which are defined from same bytes
        """
        if isinstance(other, OpcodeMacroBase):
            return self._name_ == other._name_
        if isinstance(other, bytes):
            bytes(self) == other
        return NotImplementedError(f"Unsupported type for comparison f{type(other)}")
```

src/ethereum_test_tools/vm/opcode.py
```diff
@@ -30,7 +30,50 @@ def _get_int_size(n: int) -> int:
...
+        return self._name_
+
+    def __eq__(self, other):
+        """
+        Required to differentiate between SELFDESTRUCT and SENDALL type of cases
+        And to register the Macro opcodes which are defined from same bytes
+        """
+        if isinstance(other, OpcodeMacroBase):
+            return self._name_ == other._name_
+        return False
```
danceratopz, line 73:
```suggestion
        if isinstance(other, OpcodeMacroBase):
            return self._name_ == other._name_
        if isinstance(other, bytes):
            bytes(self) == other
        return NotImplementedError(f"Unsupported type for comparison f{type(other)}")
```

src/ethereum_test_tools/vm/opcode.py
danceratopz: See https://github.com/ethereum/execution-spec-tests/pull/457#discussion_r1538971353

danceratopz: See https://github.com/ethereum/execution-spec-tests/pull/457#discussion_r1538971353

src/ethereum_test_tools/vm/opcode.py
winsvega: I explicitly stated that call_dest is bytes variable. why did it compare still?

danceratopz (changes requested):
The suggested change has nothing to do with the addition of the `Macro` class, but came up whilst reviewing #440,

Issue #395, opened 2024-01-23
Make use of features from solc 0.8.23 release


richardgreg:
Most specifically the following feature
> Commandline Interface: An empty --yul-optimizations sequence can now be always provided.

Ref: #345

richardgreg:
Hi @danceratopz, I've gone through the docs and have a fair understanding of the program, though my lack of blockchain background holds me back a bit. 

If I'm not wrong, the files in `ethereum_test_tools` help with test execution by customising pytest, right? Could you give me an example of such a command, using the `solc` command line?

danceratopz:
Hey @richardgreg :+1: 

The pytest-specific config/customization is in `src/pytest_plugins/`. The libraries in `src/ethereum_test_tools/` help generate test vectors/fixtures (that are subsequently used by ethereum execution clients) but try to be independent of pytest (i.e., you could use a different test framework/runner but still use these libraries to generate test fixtures). 

`ethereum_test_tools` does many things such as defining all the different types required to define test fixtures (accounts, blocks, transactions, etc), define the "test fixture" specifications and help write these resulting fixtures to JSON files. 

As you might have seen, the `Yul` class from `ethereum_test_tools` helps compile any Yul source constained in tests, here are the relevant lines.
https://github.com/ethereum/execution-spec-tests/blob/ccf25a322bf37a1c0fd9d6498c5e9edfaeb0d098/src/ethereum_test_tools/code/yul.py#L83-L97

If you'd like to see how the framework is calling `solc`, you can set some breakpoints or add some print statements here, for exampe:
```
print('\necho "' + self.source + '" | ' + " ".join(str(arg) for arg in solc_args))
```
which results in:
```
echo "
              {
                  function f(a, b) -> c {
                      c := add(a, b)
                  }

                  sstore(0, f(1, 2))
                  return(0, 32)
              }
              " | solc --evm-version shanghai --assemble -
```

(running `fill tests/homestead/yul/test_yul_example.py --fork=Shanghai -s -v`).

danceratopz:
See https://github.com/ethereum/execution-spec-tests/pull/467#issuecomment-1999531425.

PR #467, opened 2024-03-04
Make use of features from solc 0.8.23 release

richardgreg:
## 🗒️ Description

Most specifically the following feature:
Commandline Interface: An empty --yul-optimizations sequence can now be always provided.

## 🔗 Related Issues

Fixes: #395

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/code/yul.py
+++ b/src/ethereum_test_tools/code/yul.py
@@ -80,6 +80,14 @@ class Yul(SupportsBytes, Sized):
                 )
             else:
                 solc_args = (self.binary, *DEFAULT_SOLC_ARGS)
+
+            solc_version = self.version()
+            if solc_version >= Version("0.8.23"):
+                solc_args += (
+                    "--yul-optimizations",
+                    "",
+                )
+
             result = run(
                 solc_args,
                 input=str.encode(self.source),


src/ethereum_test_tools/code/yul.py
```diff
@@ -80,6 +80,14 @@ def __bytes__(self) -> bytes:
                 )
             else:
                 solc_args = (self.binary, *DEFAULT_SOLC_ARGS)
+
+            solc_version = self.version()
+            if solc_version >= Version("0.8.23"):
```
marioevz, line 85:
```suggestion
            if solc_version >= Version.parse("0.8.23"):
```

marioevz (commented):
Thanks for this, but I could not make it work, even after applying the suggested change, solc is constantly hitting an assertion, and simply refuses to compile.

Are you getting this same exception with 0.8.23+ ?

danceratopz:
> > Thanks for this, but I could not make it work, even after applying the suggested change, solc is constantly hitting an assertion, and simply refuses to compile.
> > Are you getting this same exception with 0.8.23+ ?
> 
> Same 😅
> 
> @danceratopz and I discussed the issue earlier and hoped he'd offer a perspective that led to a solution. If an empty `--yul-optimization` doesn't work then feel free to close the PR 🙂

Looking at the help output (below), it looks like the `--yul-optimizations` option is only relevant (resp. applicable) if `--optimize` is enabled, which it isn't in our case. So it doesn't look like this option helps us in any way. 

Nonetheless, thanks for taking the time to try it out @richardgreg!

The unhandled exception looks to be a minor bug in solc's command-line argument parsing: https://github.com/ethereum/solidity/issues/14946

```
Optimizer Options:
  --optimize           Enable optimizer.
  --optimize-runs n (=200)
                       The number of runs specifies roughly how often each 
                       opcode of the deployed code will be executed across the 
                       lifetime of the contract. Lower values will optimize 
                       more for initial deployment cost, higher values will 
                       optimize more for high-frequency usage.
  --optimize-yul       Enable Yul optimizer (independently of the EVM assembly 
                       optimizer). The general --optimize option automatically 
                       enables this unless --no-optimize-yul is specified.
  --no-optimize-yul    Disable Yul optimizer (independently of the EVM assembly
                       optimizer).
  --yul-optimizations steps
                       Forces Yul optimizer to use the specified sequence of 
                       optimization steps instead of the built-in one.
```

PR #485, opened 2024-03-25
[WIP] feat(fw): verkle t8n post state verification

spencer-tb:
## 🗒️ Description
Creating a draft PR here for visibility and discussion.

This implements a method for checking the post state of the verkle tree after t8n execution.

Currently it calls an additonal subcommand: `evm t8n verkle`, responsible for generating the verkle tree keys for account addresses and storage slots. These can be used for each account to generate an expected verkle post state using our existing post alloc, and compared against the actual vkt from t8n execution.

---
For example, once we have the account address key:
```
evm t8n verkle 0x0000000000000000000000000000000000000100
> 0x31b6...6300   # address
```
We can create the account balance, nonce and code hash keys, by adding 0x01, 0x02 and 0x03 respectively to get their keys for the vkt:
```
> 0x31b6...6301  # balance
> 0x31b6...6302  # nonce
> 0x31b6...6303  # code hash
```
For account storage keys we simply add the storage slot to the sub-command:
```
evm t8n verkle 0x0000000000000000000000000000000000000100 0x01
> 0x31b6...6341   # storage slot 1
```
Using these keys we can create our own vkt representation, adding there appropriate values from the post alloc. Note the address value is `Hash32(0)`. Numerical values are in little endian (balance/nonce). The code hash value is the keccak256 of the account code bytes.

---

Currently this method is not ideal - but it does work. As we call the subcommand around 260 times it takes around 120s to fill a single test. One proposed solution is update the subcommand, passing the entire post alloc to the subcommand - expecting an entire vkt conversion back. More comments are below.

## Usage

Use the following geth branch: gballet/t8n-merge-from-shanghai-to-prague

Run with:
```
fill --from Prague --until Prague --t8n-dump-dir=./tmp -k test_verkle_from_mpt_conversion[fork_ShanghaiToPragueVerkleTransition-blockchain_test]
```
Check the output folders with `vkt.json` for any debugging.

**Note it takes around 2 mins to fill the test.** 

## 🔗 Related Issues
N/A

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/spec/base/base_test.py
+++ b/src/ethereum_test_tools/spec/base/base_test.py
@@ -65,12 +65,37 @@ def verify_post_alloc(*, expected_post: Mapping, got_alloc: Mapping):
                     raise Exception(f"expected account not found: {address}")
 
 
-def verify_post_vkt(*, expected_post: Mapping, got_vkt: Mapping):
+def verify_post_vkt(transition_tool: TransitionTool, expected_post: Mapping, got_vkt: Mapping):
     """
     Verify that the final verkle tree mapping matches the expected post in the test.
     Raises exception on unexpected values.
     """
-    # TODO: Add the use the verkle subcommand to get the keys for each value in the tree.
+    # TODO: Add this check before filling within the filler cli plugin.
+    if not transition_tool.verkle_subcommand:
+        raise Exception("Only geth's evm tool is supported to verify verkle trees.")
+
+    # Convert the expected post alloc to a verkle tree for comparison.
+    # TODO: Maintain an intermediate type VerkleKeyMap: that maps verkle keys to alloc keys
+    expected_vkt = transition_tool.post_alloc_to_vkt(post_alloc=expected_post)
+
+    # Check for keys that are missing or unexpected in the actual VKT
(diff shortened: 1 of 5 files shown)


spencer-tb:
The question now that pops up - how do we best optimize this?

1) We update the `evm t8n verkle` subcommand to take in our post allocation as an input, where geth creates and returns our expected verkle tree instead of us creating it manually. This would solve the runtime issue. We'd only call the subcommand once. 

2) We do 1), but entirely within t8n and not an additional subcommand. Essentially we'd provide additional optional flags `--input.postAlloc` and `--output.postVkt`. We'd provide the input post alloc on our last t8n call. T8n would do the same conversion above and provide it as an output.

3) Can t8n provide alloc alongside the vkt during each transition? If so, why don't we use the alloc here. Maybe it could be prone to errors?

If the long term plan is to use the pedersen hash in python within EEST, then the current implementation style would be the way forward. Although, 1) or 2) using the latter in EELS seems more optimal.

spencer-tb:
CIFO: #507

All relavent changes are included there. Most code is however removed, as we are going to verify the post state vkt slightly differently.

PR #496, opened 2024-04-08
eof classes idea

winsvega:
Eof classes idea for Mario

also to be fliexible we should allow illegal values of those fields to be able to generate all kinds of malicious code.

### DIFF
--- a/src/ethereum_test_tools/eof/v1/eof_classes.py
+++ b/src/ethereum_test_tools/eof/v1/eof_classes.py
@@ -1,0 +1,114 @@
+"""
+EOF v1 specs
+https://github.com/ipsilon/eof/blob/main/spec/eof.md
+"""
+
+from typing import Generic, TypeVar, List
+
+class Byte:
+    def __init__(self, value=0):
+        self._value = None  # Initialize _value; will be set properly via the property setter
+        self.value = value  # Use the property setter to initialize value
+
+    @property
+    def value(self):
+        """Property getter for the byte value."""
+        return self._value
+
+    @value.setter
+    def value(self, new_value):
+        """Property setter for the byte value, with validation to ensure it fits in a byte."""
+        if not isinstance(new_value, int):
+            raise TypeError("Value must be an integer")
+        if not 0 <= new_value <= 255:
+            raise ValueError("Value must be between 0 and 255 inclusive")
+        self._value = new_value
+
+    def __repr__(self):
+        """String representation for debugging."""
+        return f"Byte({self.value})"
+
+class Bytes2:
+    def __init__(self, value=0):
+        self._value = None  # Initialize _value; will be set properly via the property setter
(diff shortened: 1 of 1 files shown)


src/ethereum_test_tools/eof/v1/eof_classes.py
```diff
@@ -0,0 +1,114 @@
...
+    code_size: NonEmptyList[Bytes2]
+    # [
+    kind_container: Byte
+    num_container_sections: Bytes2
+    container_size: NonEmptyList[Bytes2]
+    # ]
+
+    kind_data: Byte
+    data_size: Bytes2
+    terminator: Byte
```
marioevz, line 104: If we use pydantic we can skip having to create `NonEmptyList`, e.g.:

```python
from pydantic import BaseModel

class Header(BaseModel):
    magic: Bytes2
    ...
    code_size: List[Bytes2] = Field(..., min_length=1)
```
We still need to (potentially) declare a `__bytes__` function to transform all of the fields into binary form.

src/ethereum_test_tools/eof/v1/eof_classes.py
```diff
@@ -0,0 +1,114 @@
+"""
+EOF v1 specs
+https://github.com/ipsilon/eof/blob/main/spec/eof.md
+"""
+
+from typing import Generic, TypeVar, List
+
+class Byte:
```
marioevz, line 8: If you really need these two classes, I think an off-the-shelf solution could be to use the `base_types.FixedSizeBytes` class, you can declare something like:
```python
from ..common.base_types import FixedSizeBytes

class Byte(FixedSizeBytes[1]):  # type: ignore
    """
    Class that helps represent a single byte.
    """

    pass


class Bytes2(FixedSizeBytes[2]):  # type: ignore
    """
    Class that helps represent two bytes.
    """

    pass
```

An advantage of this is that these play nicely with pydantic.

src/ethereum_test_tools/eof/v1/eof_classes.py
```diff
@@ -0,0 +1,114 @@
...
+
+    kind_data: Byte
+    data_size: Bytes2
+    terminator: Byte
+
+class Body:
+    types_section: NonEmptyList[TypeSection]
+    code_section: NonEmptyList[bytes]
+    container_section: List[bytes]
+    data_section: bytes
```
marioevz, line 110:
```suggestion
from ..common.base_types import Bytes
class Body(BaseModel):
    types_section: List[TypeSection] = Field(..., min_length=1)
    code_section: List[Bytes] = Field(..., min_length=1)
    container_section: List[Bytes]
    data_section: Bytes
```

marioevz (commented):
Some comments. It would seem like all these classes would benefit a lot from pydantic.

I would also recommend adding a test file to test these classes when converted to `bytes`.

PR #502, opened 2024-04-12
new(test): add tests for EOF/EIP-663 DUPN SWAPN

chfast:
## 🗒️ Description

Adds new tests for EIP-663.

## ✅ Checklist

### DIFF
--- a/.github/workflows/fixtures.yaml
+++ b/.github/workflows/fixtures.yaml
@@ -19,11 +19,11 @@ jobs:
             fill-params: ''
             solc: '0.8.21'
             python: '3.11'
-          # - name: 'fixtures_develop'
-          #   evm-type: 'develop'
-          #   fill-params: '--until=Prague'
-          #   solc: '0.8.21'
-          #   python: '3.11'
+          - name: 'fixtures_develop'
+            evm-type: 'develop'
+            fill-params: '--until=Prague --ignore=./tests/prague/eip7692_eof_v1'
+            solc: '0.8.21'
+            python: '3.11'
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,8 +21,8 @@ jobs:
           - os: ubuntu-latest
             python: '3.11'
             solc: '0.8.21'
-            evm-type: 'main'  # 'develop'
-            tox-cmd: 'tox run-parallel --parallel-no-spinner'  # 'tox -e tests-develop'
+            evm-type: 'develop'
+            tox-cmd: 'tox -e tests-develop'
           - os: macos-latest
             python: '3.11'
             solc: '0.8.22'
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -24,6 +24,8 @@ Test fixtures for use by clients are available for each release on the [Github r
 - ✨ Added `--traces` support when running with Hyperledger Besu ([#511](https://github.com/ethereum/execution-spec-tests/pull/511)).
 - ✨ Use pytest's "short" traceback style (`--tb=short`) for failure summaries in the test report for more compact terminal output ([#542](https://github.com/ethereum/execution-spec-tests/pull/542)).
 - ✨ The `fill` command now generates HTML test reports with links to the JSON fixtures and debug information ([#537](https://github.com/ethereum/execution-spec-tests/pull/537)).
+- ✨ Add an Ethereum RPC client class for use with consume commands ([#556](https://github.com/ethereum/execution-spec-tests/pull/556)).
+- ✨ Add a "slow" pytest marker, in order to be able to limit the filled tests until release ([#562](https://github.com/ethereum/execution-spec-tests/pull/562)).
 
 ### 🔧 EVM Tools
 
--- a/evm-config.yaml
+++ b/evm-config.yaml
@@ -4,5 +4,5 @@ main:
   ref: master
 develop:
   impl: geth
-  repo: ethereum/go-ethereum
-  ref: master
+  repo: lightclient/go-ethereum
+  ref: prague-devnet-0
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,6 +3,8 @@ console_output_style = count
 minversion = 7.0
 python_files = *.py
 testpaths = tests/
+markers =
+    slow
 addopts = 
     -p pytest_plugins.test_filler.test_filler
     -p pytest_plugins.forks.forks
--- a/src/cli/pytest_commands.py
+++ b/src/cli/pytest_commands.py
@@ -150,19 +150,19 @@ def get_hive_flags_from_env():
     pytest_args = []
     xdist_workers = os.getenv("HIVE_PARALLELISM")
     if xdist_workers is not None:
-        pytest_args.extend("-n", xdist_workers)
+        pytest_args.extend(["-n", xdist_workers])
     test_pattern = os.getenv("HIVE_TEST_PATTERN")
     if test_pattern is not None:
         # TODO: Check that the regex is a valid pytest -k "test expression"
-        pytest_args.extend("-k", test_pattern)
+        pytest_args.extend(["-k", test_pattern])
     random_seed = os.getenv("HIVE_RANDOM_SEED")
     if random_seed is not None:
         # TODO: implement random seed
-        warnings.warning("HIVE_RANDOM_SEED is not yet supported.")
+        warnings.warn("HIVE_RANDOM_SEED is not yet supported.")
     log_level = os.getenv("HIVE_LOGLEVEL")
     if log_level is not None:
         # TODO add logging within simulators and implement log level via cli
-        warnings.warning("HIVE_LOG_LEVEL is not yet supported.")
+        warnings.warn("HIVE_LOG_LEVEL is not yet supported.")
     return pytest_args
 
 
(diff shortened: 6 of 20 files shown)


tests/prague/eip663_dupn_swapn_exchange/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+import pytest
+
+from ethereum_test_tools import Transaction
+
+REFERENCE_SPEC_GIT_PATH = "EIPS/eip-663.md"
+REFERENCE_SPEC_VERSION = "6aed382f86258f33603f5dc275956f739aaae096"
+
+
+@pytest.fixture
```
marioevz, line 14: For fixtures that are used in all tests, these can be added to a file called `conftest.py`.

Example: https://github.com/ethereum/execution-spec-tests/blob/main/tests/cancun/eip4844_blobs/conftest.py

Then there is no need to manually importing `tx` in each test file.

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
```diff
@@ -0,0 +1,57 @@
...
+
+from ..eip3540_eof_v1.spec import EOF_FORK_NAME
+from . import REFERENCE_SPEC_GIT_PATH, REFERENCE_SPEC_VERSION, tx
+
+REFERENCE_SPEC_GIT_PATH = REFERENCE_SPEC_GIT_PATH
+REFERENCE_SPEC_VERSION = REFERENCE_SPEC_VERSION
+
+
+@pytest.mark.valid_from(EOF_FORK_NAME)
+def test_dupn_all_valid_immediates(
```
marioevz, line 21: Linter is complaining about word `immediates`, we can simply add it to `whitelist.txt` in the root directory.

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
```diff
@@ -0,0 +1,57 @@
...
+    tx: Transaction,
+    state_test: StateTestFiller,
+):
+    """
+    Test case for all valid DUPN immediates.
+    """
+
```
marioevz, line 28:
```suggestion
```
Linter is complaining about this blank line. If you are currently using vscode, you can install `Flake8` extension to see these in real-time.
Otherwise you can run `tox -e tests` and see all the issues that the linter will complain about.

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
```diff
@@ -0,0 +1,57 @@
...
+"""  # noqa: E501
+
+import pytest
+
+from ethereum_test_tools import Account, Environment, StateTestFiller, TestAddress, Transaction
+from ethereum_test_tools.eof.v1 import Container, Section
+from ethereum_test_tools.eof.v1.constants import NON_RETURNING_SECTION
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+from ..eip3540_eof_v1.spec import EOF_FORK_NAME
```
marioevz, line 13: Linter is complaining about word `eip3540`, you can add this too to the `whitelist.txt`.

tests/prague/eip663_dupn_swapn_exchange/test_swapn.py
```diff
@@ -0,0 +1,58 @@
...
+
+@pytest.mark.valid_from(EOF_FORK_NAME)
+def test_swapn_all_valid_immediates(
+    tx: Transaction,
+    state_test: StateTestFiller,
+):
+    """
+    Test case for all valid SWAPN immediates.
+    """
+
```
marioevz, line 28:
```suggestion
```
Same issue with linter.

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    n = 256
+    values = range(0xD00, 0xD00 + n)
+
+    eof_code = Container(
+        sections=[
+            Section.Code(
+                code=b"".join(Op.PUSH2(v) for v in values)
+                + b"".join(Op.DUPN(x) + Op.PUSH2(x) + Op.SSTORE for x in range(0, n))
```
marioevz, line 36:
```suggestion
                + b"".join(Op.SSTORE(x, Op.DUPN[x]) for x in range(0, n))
```
There was a slight semantics change recently where we added square-brackets to the opcodes that have a data portion, so this should work as well.
Right now both `Op.DUPN(x)` and `Op.DUPN[x]` result in the exact same bytecode, but for others like `RJUMPV` it makes it much more readable, so we are trying to push for the new semantics.

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
```diff
@@ -0,0 +1,57 @@
...
+                + Op.STOP,
+                code_inputs=0,
+                code_outputs=NON_RETURNING_SECTION,
+                max_stack_height=258,
```
marioevz, line 40:
```suggestion
                max_stack_height=n+2,
```

tests/prague/eip663_dupn_swapn_exchange/test_swapn.py
```diff
@@ -0,0 +1,58 @@
...
+    n = 100
+    values = range(0x500, 0x500 + 257)
+
+    eof_code = Container(
+        sections=[
+            Section.Code(
+                code=b"".join(Op.PUSH2(v) for v in values)
+                + b"".join(Op.SWAPN(0xFF - x) + Op.PUSH2(x) + Op.SSTORE for x in range(0, n))
```
marioevz, line 36:
```suggestion
                + b"".join(Op.SSTORE(x, Op.SWAPN[0xFF - x]) for x in range(0, n))
```

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
```diff
@@ -0,0 +1,57 @@
...
+
+    n = 256
+    values = range(0xD00, 0xD00 + n)
+
+    eof_code = Container(
+        sections=[
+            Section.Code(
+                code=b"".join(Op.PUSH2(v) for v in values)
+                + b"".join(Op.DUPN(x) + Op.PUSH2(x) + Op.SSTORE for x in range(0, n))
+                + Op.STOP,
```
marioevz, line 37:
```suggestion
                + Op.POP * n
                + Op.STOP,
```
Not really sure about this one, but do we still need to clear the stack before stopping in EOF?

tests/prague/eip663_dupn_swapn_exchange/test_swapn.py
```diff
@@ -0,0 +1,58 @@
...
+                + Op.STOP,
```
marioevz, line 37: Same comment as in `test_dupn.py`, we might need to pop stack values, otherwise just leave as is.

marioevz (commented):
Looks good overall! A few comments.

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
chfast: Somehow, this expression doesn't work if changed `(v)` to `[v]`: `b"".join(Op.PUSH2(v) for v in values)`.

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
chfast: Not any more in the case where we end execution (e.g. `STOP` is used).

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
chfast: Somehow, this expression doesn't work if changed `(v)` to `[v]`: `b"".join(Op.PUSH2(v) for v in values)`.

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
```diff
@@ -0,0 +1,56 @@
...
+REFERENCE_SPEC_VERSION = REFERENCE_SPEC_VERSION
+
+
+@pytest.mark.valid_from(EOF_FORK_NAME)
+def test_dupn_all_valid_immediates(
+    tx: Transaction,
+    state_test: StateTestFiller,
+):
+    """
+    Test case for all valid DUPN immediates.
```
chfast, line 26: This test executes all DUPN instructions in single execution and stores the results in 256 storage slots.

Should I use "parametrize" to create 256 test cases instead?

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
marioevz: Yes, I tried to find a workaround but sadly there does not seem to be any straight-forward one because the resulting `Opcode` type from the subscript operation is not detected as something that can be bytes-concatenated.

For the time being just leave it as is:
```python
b"".join(Op.PUSH2(v) for v in values)
```

tests/prague/eip663_dupn_swapn_exchange/test_dupn.py
marioevz: I think this should be fine as is. When we parametrize we try to use edge values, e.g. 0, 256-1, 256, 256+1, but not many in-between.

chfast: ok then. Anything else to do here?

chfast:
Ping.

PR #503, opened 2024-04-12
git workflow evmone coverage script

winsvega:
## 🗒️ Description
git job to measure coverage of ported tests
if a new line introduced to converted_ethereum_tests, the script will run coverage and make an artifact

## 🔗 Related Issues
Feauture

## ✅ Checklist


.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,160 @@
...
+          path: ${{ github.workspace }}/evmoneimage
+          key: ${{ runner.os }}-evmone-${{ hashFiles('evmone-hash.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-evmone-
+
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }} # Checks out the PR branch
+          fetch-depth: 0 # Necessary to fetch all history for diff
```
marioevz, line 25:
```suggestion
      - name: Checkout code
        uses: actions/checkout@v3
```
I think this action by default checks out the current PR's branch, so the ref is unnecessary.
Also, we seem to be doing this twice at the start of the steps? (Line 12).

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,160 @@
...
+          #install tests
+          cd ${{ github.workspace }}
+          mkdir testpath
+          cd testpath
+          git init
+          git remote add origin https://github.com/ethereum/tests.git
+          git config core.sparseCheckout true
+          git sparse-checkout set GeneralStateTests
+          git pull origin develop
+          cd ..
```
marioevz, line 70: This could be converted into its own step with `actions/checkout`:
```yaml
- name: Checkout ethereum/tests
        uses: actions/checkout@v4
        with:
          repository: ethereum/tests
          path: testpath
          sparse-checkout: |
            GeneralStateTests
```

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,160 @@
...
+          #install go
+          curl -L --output go1.21.8.linux-amd64.tar.gz "https://go.dev/dl/go1.21.8.linux-amd64.tar.gz"
+          tar -xf go1.21.8.linux-amd64.tar.gz
+          sudo mv go /usr/local
+          sudo ln -s /usr/local/go/bin/go /usr/local/bin/go
+          git clone --depth 1 https://github.com/ethereum/go-ethereum.git
+          cd go-ethereum
+          go build ./cmd/evm
+          cp evm /usr/local/bin
+          cd ..
```
marioevz, line 53: We have an action to do this more easily:
```yaml
- name: Build EVM
        uses: ./.github/actions/build-evm
        id: evm-builder
        with:
          type: 'main'
```

marioevz (commented):
Some comments.

.github/workflows/coverage.yaml
winsvega: does it also install solc?

.github/workflows/coverage.yaml
winsvega: there must be 2 separate steps. one is to fetch branches, and enother one to restore cache. 
never used the git cache in this context, but it supposed to keep the docker file and only update it when the hash file is changed.

marioevz (approved):
LGTM, thanks for the simplification effort, it's much cleaner now!

PR #513, opened 2024-04-17
fix(fw,tests): Fixes to EOF

marioevz:
### EOF classes use pydantic
EOF classes `Section` and `Container` now use pydantic `BaseModel`.

### Opcodes with data portion are now subscriptable
Opcodes that have a data portion are now subscriptable:
- `Op.PUSH1[1]`
- `Op.RJUMP[1]`
- `Op.RJUMPI[1](Op.ORIGIN)`

Where the subscript is used to define the data portion of the opcode, and the call arguments can now be used exclusively for the stack.

Data portion can still be passed as the first argument to the call, so `Op.PUSH1(1) == Op.PUSH1[1]  == b"\x60\x01"` but `Op.PUSH1[1](1)` will fail (we can begin to look forward to deprecate setting the data portion from the call arguments).

This is specially useful for `Op.RJUMPV`, where we can define each jump destination as a parameter of the subscript:
- `Op.RJUMPV[5, 6, 7](Op.ORIGIN)`

In this example, the RJUMPV is converted into bytecode with three possible jump destinations, `+5`, `+6`, or `+7`, and the `max_index` operand is automatically set to `uint8(3)`, and `Op.ORIGIN` is executed to set the only required stack element.

### Fixes to EOF tests

All tests usages of `RJUMP`, `RJUMPI`, `RJUMPV` and `CALLF` have been fixed to properly use subscripts.

### `Section.Code`, `Section.Data`, `Section.Container` helpers

Three helpers have been added as class methods to the `Section` class, and can be used as helpers to create code, data or container sections respectively.

```python
Section(kind=SectionKind.CODE, data=Op.STOP)
```
can now simply be:
```python
Section.Code(code=Op.STOP)
```

Rest of the section arguments are passed along to the class creator.

### Notes

Fixed ~Tox is still failing due to file `tests/prague/eip3540_eof_v1/test_execution_function.py`~

## ✅ Checklist


danceratopz:
`tox -e tests-develop` is failing as initial Prague support was added to `t8n` in
https://github.com/ethereum/go-ethereum/commit/823719b9e1b72174cd8245ae9e6f6f7d7072a8d6, full support should be enabled with https://github.com/ethereum/go-ethereum/pull/29557

winsvega:
`validity_error=EOFException.DEFAULT_EXCEPTION,`

logically this field does not belong to container. the container itself should be agnostic about its validity. 
the validity lives on higher level

danceratopz (approved):
Pydantic and Opcode changes LGTM! Pretty clever to hijack `__getitem__` to create the RJUMPV look-up table.

winsvega:
> > because I want to write custom_terminator=0x01 or 0x00
> > but that does not work as it must be bytes.
> 
> The reason we use bytes here is because you might want to add many zeros at the end for testing purposes and if you make this an integer you cannot specify more than one byte (0x0000 just resolves to 0 in python and the leading zeros are discarded). Maybe the correct name for this field should then be `custom_terminator_bytes`.

then all the reset must be bytes too for consistancy. 
its a pity 0x0001  is treated as integer in python

PR #519, opened 2024-04-22
feat(fw): Call `evmone-eofparse` on EOF tests filling

marioevz:
## 🗒️ Description
Adds EOF exception checking using evmone-eofparse tool. Prints a nice message now if expected exception does not match returned by evmone or if valid eof test fails on evmone and so on.

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/spec/eof/eof_test.py
+++ b/src/ethereum_test_tools/spec/eof/eof_test.py
@@ -2,6 +2,9 @@
 Ethereum EOF test spec definition and filler.
 """
 
+from pathlib import Path
+from shutil import which
+from subprocess import CompletedProcess, run
 from typing import Callable, ClassVar, Generator, List, Optional, Type
 
 from ethereum_test_forks import Fork
@@ -11,6 +14,39 @@ from ...common.base_types import Bytes
 from ...exceptions import EOFException
 from ..base.base_test import BaseFixture, BaseTest
 from .types import Fixture
+
+
+class EOFParse:
+    """evmone-eofparse binary."""
+
+    binary: Path
+
+    def __new__(cls):
+        """Make EOF binary a singleton."""
+        if not hasattr(cls, "instance"):
+            cls.instance = super(EOFParse, cls).__new__(cls)
+        return cls.instance
+
+    def __init__(
+        self,
+        binary: Optional[Path | str] = None,
+    ):
+        if binary is None:
+            which_path = which("evmone-eofparse")
+            if which_path is not None:
+                binary = Path(which_path)
+        if binary is None or not Path(binary).exists():
+            raise Exception("""`evmone-eofparse` binary executable not found""")
+        self.binary = Path(binary)
+
+    def run(self, *args: str, input: str | None = None) -> CompletedProcess:
+        """Run evmone with the given arguments"""
+        return run(
+            [self.binary, *args],
+            capture_output=True,
+            text=True,
+            input=input,
+        )
 
 
 class EOFTest(BaseTest):
@@ -35,7 +71,7 @@ class EOFTest(BaseTest):
         """
         Generate the EOF test fixture.
         """
-        return Fixture(
+        fixture = Fixture(
             vectors={
                 "0": {
                     "code": self.data,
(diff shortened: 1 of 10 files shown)


winsvega:
this PR consumes https://github.com/ethereum/execution-spec-tests/pull/516  that introduced  this feature example use

src/ethereum_test_tools/spec/eof/eof_test.py
```diff
@@ -48,6 +84,51 @@ def make_eof_test_fixture(
...
+        for _, vector in fixture.vectors.items():
+            expectedResult = vector.results.get(str(fork))
+            result = eof_parse.run(input=str(vector.code))
+            self.verify_result(result, expectedResult, vector.code)
+
+        return fixture
+
+    def verify_result(self, result: CompletedProcess, expectedResult: Result, code: Bytes):
+        parser = EvmoneExceptionParser()
+        res_error = result.stdout.replace("\n", "")
```
winsvega, line 97: currently eof tool does not return an error code, so we analyze the stdout

src/ethereum_test_tools/spec/eof/eof_test.py
```diff
@@ -48,6 +84,51 @@ def make_eof_test_fixture(
...
+                    return
+
+                msg = "EOF code expected to fail with a different exception, than reported:"
+                formatted_message = (
+                    f"{msg} \n"
+                    f"{code} \n"
+                    f"Expected: {expRes} ({parser.parse_exception(expRes)}) \n"
+                    f"     Got: {parser.rev_parse_exception(res_error)} ({res_error})"
+                )
+                raise Exception(formatted_message)
```
winsvega, line 131: this exception verification feature can be done in the same manner for state tests/bc tests with different tools. 
would be nice to see it in consume.  and also when generating the tests via t8n, the exception string from t8n can be verified against our exception maps

shemnon:
It would be nice if this could be made generic in a manner like TransitionTool is made generic, I expect each client to expose their own tool.

Also, I would like a format that accepts required arguments.  Besu's equivelant is `evmtool code-validation` and that won't pass the binary string matching task ask it is a binary plus an argument.

Combine this with the fork being validated possibly being another argument.

marioevz:
> It would be nice if this could be made generic in a manner like TransitionTool is made generic, I expect each client to expose their own tool.
> 
> Also, I would like a format that accepts required arguments. Besu's equivelant is `evmtool code-validation` and that won't pass the binary string matching task ask it is a binary plus an argument.
> 
> Combine this with the fork being validated possibly being another argument.

I think @danceratopz might know a bit better than I do how to implement something along these lines.

We might need to create another package like we currently have `evm_transition_tool`, maybe a `eof_tool` with all different flavors of EOF tools implemented.

We might also need to implement new parameters to configure it in the `pytest_plugins/test_filler`.

My main ask when we implement this is that the normal test filling process does not fail if there no eof tool installed, only fails if we are trying to fill an EOF test.

src/ethereum_test_tools/spec/eof/eof_test.py
danceratopz: would be nice to have error codes!

winsvega (approved):

winsvega:
we stick to evmone-eofparse for verification for now till later when we refactor tooling capture

danceratopz (commented):
Very nice.

Really like how/where `eofparse` is called, I had a mental block on this one and wanted to add it to the `FixtureCollector` - it's much more logical in the fixture generator for this use case!

Love the exceptions raised by `verify_fixtures()` - really clear.

After aligning with @winsvega, I've refactored the code a bit in #538, which addresses the small concerns below. Please let me know what you think.

From the above comment from @shemnon:
> It would be nice if this could be made generic in a manner like TransitionTool is made generic, I expect each client to expose their own tool.

We will definitely do this! But will add this work in a separate PR as it requires quite a bit of code organization. Don't want to block this helpful addition though, so we should get this merged first.

danceratopz:
> My main ask when we implement this is that the normal test filling process does not fail if there no eof tool installed, only fails if we are trying to fill an EOF test.

@marioevz, Regarding this, we warn at the end of the test session if `evmone-eofparse` was not found:
image

src/ethereum_test_tools/spec/eof/eof_test.py
danceratopz: Yup, definitely!

PR #522, opened 2024-04-23
Test EIP-7069 Address Space Extension implications

shemnon:
## 🗒️ Description
Validate ASE behaviors on legacy and EOF opcodes for ASE, when targeting
EOAs, Contracts, and empty accounts.  Opcodes are EXTCALL/CALL,
EXTDELEGATECALL/DELEGATECALL,  EXTSTATICCALL/STATICCALL, and CALLCODE.

## ✅ Checklist


tests/prague/eip7676_address_space_extension/test_ase_opcodes.py
```diff
@@ -0,0 +1,219 @@
...
+
+    pre = {
+        TestAddress: Account(
+            balance=1000000000000000000000,
+            nonce=1,
+        ),
+        Address(0x100): Account(
+            code=(
+                Op.MSTORE(0, Op.PUSH32(address_bytes))
+                + Op.SSTORE(0, Op.CALL(50000, 0x200, 0, 0, 32, 0, 0))
```
winsvega, line 68: 0x200 account addresses can replace with meaningful variable names
like giving a hint what 0x200 address code is doing

tests/prague/eip7676_address_space_extension/test_ase_opcodes.py
```diff
@@ -0,0 +1,219 @@
...
+                + Op.SSTORE(1, Op.CALL(50000, 0x300, 0, 0, 32, 0, 0))
+                + Op.SSTORE(2, Op.CALL(50000, 0x400, 0, 0, 32, 0, 0))
+                + Op.SSTORE(3, Op.CALL(50000, 0x500, 0, 0, 32, 0, 0))
+                + Op.SSTORE(4, Op.CALL(50000, 0x201, 0, 0, 32, 0, 0))
+                + Op.SSTORE(5, Op.CALL(50000, 0x301, 0, 0, 32, 0, 0))
+                + Op.SSTORE(6, Op.CALL(50000, 0x401, 0, 0, 32, 0, 0))
+                + Op.SSTORE(7, Op.CALL(50000, 0x501, 0, 0, 32, 0, 0))
+                + Op.SSTORE(8, Op.CALL(50000, 0x601, 0, 0, 32, 0, 0))
+                + Op.STOP()
```
winsvega, line 77: storage slots 1...8 can also be variables with the names describing the meaning of what is stored in it
like this
```
code_worked = 9
Op.SSTORE(code_worked, 1)
...
```

winsvega (dismissed):

tests/prague/eip7676_address_space_extension/test_ase_opcodes.py
shemnon: Done.  `ADDRESS_<calltype>`

tests/prague/eip7676_address_space_extension/test_ase_opcodes.py
shemnon: Done.  `ID_<calltype>` identifies what call type the result is tracking.

tests/prague/eip7692_eof_v1/eip7676_address_space_extension/test_ase_opcodes.py
```diff
@@ -0,0 +1,262 @@
...
+BASE_ADDRESS = 100000000
+# ADDRESS_EOF_BALANCE = Address(BASE_ADDRESS + ID_EOF_BALANCE)
+# ADDRESS_BALANCE = Address(BASE_ADDRESS + ID_BALANCE)
+ADDRESS_EXTCALL = Address(BASE_ADDRESS + ID_EXTCALL)
+ADDRESS_CALL = Address(BASE_ADDRESS + ID_CALL)
+ADDRESS_EXTDELEGATECALL = Address(BASE_ADDRESS + ID_EXTDELEGATECALL)
+ADDRESS_DELEGATECALL = Address(BASE_ADDRESS + ID_DELEGATECALL)
+ADDRESS_EXTSTATICCALL = Address(BASE_ADDRESS + ID_EXTSTATICCALL)
+ADDRESS_STATICCALL = Address(BASE_ADDRESS + ID_STATICCALL)
+ADDRESS_CALLCODE = Address(BASE_ADDRESS + ID_CALLCODE)
```
marioevz, line 48: I'm not super excited about this pattern: my concern is regarding a potential change that we will introduce in the close future where the test function "requests" the pre-alloc to place a smart contract "somewhere" and return the address, like so:
```python
def test_function(state_test: StateTestFiller, pre: Alloc, ...):
    my_contract_address = pre.deploy_contract(Op.SSTORE(0, 1))
```
So in the future you cannot request a specific address as easily as now, so I think we should prepare for that by making changes to patterns like this.

marioevz (commented):
Just a comment about the pattern used to write these tests.

tests/prague/eip7692_eof_v1/eip7676_address_space_extension/test_ase_opcodes.py
shemnon: I'm not exactly sure what is being asked for as `Alloc` does not have the `deploy_contract` method right now,
but I did migrate the allocations to use `itertools.Count` to communicate it's looking for "whatever address is handed out".  This breaks ordering assumptions implicit in the BASE+ID code that existed before.

Do I need to move this into the test method instead of at top level?

shemnon:
Since 7979 is out of scope for EOFv1 I migrated this into a 7069 test that only tests the ASE aspects.

marioevz (commented):
It looks really good overall, but one change I would probably do is:

Instead of doing all types of calls in a single shot, maybe we could simply parameterize for each call opcode.

Reason being that, running the tests like this makes it difficult to single out the failing ones to a single opcode with an incorrect behavior, but if it was parameterized, it would be very easy to do this, even if it comes at the cost of having to run more tests.

shemnon:
Refactored so it's parameterized on opcode.  However the logic to determine the "correct" output is still in place.  The other alternative is 7 slightly different versions of the test and parameters, dropping most if the conditional building.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,226 @@
...
+                        max_stack_height=1 + len(call_suffix),
+                    )
+                ],
+            )
+            if ase_ready_opcode
+            else Op.SSTORE(
+                slot_target_call_status,
+                target_opcode(Op.GAS, Op.CALLDATALOAD(0), *call_suffix),
+            )
+            + Op.SSTORE(slot_target_returndata, Op.RETURNDATALOAD(0))
```
marioevz, line 139:
```suggestion
            + Op.RETURNDATACOPY(0, 0, Op.RETURNDATASIZE)
            + Op.SSTORE(slot_target_returndata, Op.MLOAD(0))
```
Had to make this change locally to make some of the tests work, I'm not sure if `RETURNDATALOAD` is available for legacy contracts, and also want to make sure this is not the purpose of the test.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,226 @@
...
+        ),
+        address_target: Account(
+            code=Container(
+                sections=[
+                    Section.Code(
+                        code=Op.SSTORE(
+                            slot_target_call_status,
+                            target_opcode(Op.CALLDATALOAD(0), *call_suffix),
+                        )
+                        + Op.SSTORE(slot_target_returndata, Op.RETURNDATALOAD(0))
```
marioevz, line 126: Same thing here, emvone is exceptionally aborting execution when reaching `RETURNDATALOAD`.
Changing to RETURNDATACOPY + RETURNDATASIZE + MLOAD solves the issue here too.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
shemnon: RETURNDATALOAD is part of EOF, this particular test shouldn't be run in legacy.  evmone is out of spec for this test. It should work per spec - https://github.com/ethereum/EIPs/pull/8617 just got merged into the EIP.  @chfast @pdobacz

This is why I don't think we should add blocking github actions on tests targeting forks that haven't flipped any of the long-term test nets.  The spec could change and implementations could be implementing different versions of the EIP.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
shemnon: See prior comment.  evmone is out of spect at the moment.

shemnon: This is also an emergent behavior of EXTDELEGATECALL failing to call a legacy contract, and thus zero returndata is in the return data buffer, hence all calls to RETURNDATALOAD should return zeros.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,226 @@
...
+                nonce=0,
+            )
+
+    target_storage: dict[int, int | bytes | Address] = {}
+    match target_account_type:
+        case "empty":
+            target_storage[slot_target_call_status] = 0 if ase_ready_opcode else 1
+        case "EOA":
+            pre[Address(stripped_address)] = Account(code="", balance=10**18, nonce=9)
+            target_storage[slot_target_call_status] = 0 if ase_ready_opcode else 1
```
marioevz, line 177: Also use `LEGACY_SUCCESS` and `EXTCALL_SUCCESS` constants here (if decided to be named them like that).

marioevz (commented):
Some comments, thanks!

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
pdobacz: I'm not 100% sure I'm following what's happening here but:
- evmone has RETURNDATALOAD implemented for EOF code only, like it should and like it is used in this test
- evmone doesn't just yet have the lax OoB for RETURNDATALOAD/COPY - [will be merged soon](https://github.com/ethereum/evmone/pull/909), causing the RETURNDATALOAD instruction to fail, however I think this in particular is slightly misleading, because:
- the change from LOAD to COPY seems to be covering up the real problem. How? I am guessing that the underlying `target_opcode` fails (see below why), and RETURNDATASIZE is 0, hence the change @marioevz made here changed semantics
- I am also guessing the underlying call fails because evmone doesn't yet implement ASE-ready EXT*CALL instructions - this is WIP right now

If I'm correct above, then what I think should happen is we should keep using RETURNDATALOAD and defer `ase_ready_opcode` testing until it is implemented in evmone. Additionally, since https://github.com/ethereum/EIPs/pull/8617 will soon become reality, we should make sure that the expected return data is not zero bytes, because this is what the RETURNDATA* instructions will give you even on accidental out-of-bounds read.

shemnon: I've removed the use of RETURNDATALOAD from these tests.  However they will return to a new test given the recent change to soft-zeroing in EOF, those need some explicit tests to validate correctness.

What remains for evmone are two checks:
* Revert on ASE addresses
* Revert on EXTDELEGATECALL 

While the delegate check may warrant an independent test, removing it from the matrix would leave a hole that is essentail to what is tested, whereas RETURNDATALOAD is arguably orthagonal.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,226 @@
...
+address_target = Address(next(_address_allocation))
+
+_slot = itertools.count(1)
+slot_top_level_call_status = next(_slot)
+slot_target_call_status = next(_slot)
+slot_target_returndata = next(_slot)
+
+value_legacy_success = 1
+value_eof_success = 0
+value_eof_fail = 1
```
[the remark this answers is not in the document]
shemnon: Done.  Went with CALL_* and EXTCALL_*

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
shemnon: done

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
pdobacz: > defer ase_ready_opcode testing until it is implemented in evmone

I phrased this badly. I should have written: "defer testing ase_ready_opcode in evmone", not that tests wait on evmone, sorry about this. BTW This has just [been merged](https://github.com/ethereum/evmone/pull/915).

Re
>What remains for evmone are two checks:
As of today's master, both are implemented. EXTDELEGATECALL check has been for a while, was it failing to fail on such calls?

---

I've read the test again carefully and see the point about RETURNDATALOAD - you don't care if the zeros returned are due to noop RETURNDATACOPY (RETURNDATASIZE==0) or from zero-padding on RETURNDATALOAD. because on success we expect the caller address to be there - all clear now.

chfast (changes requested):

tests/prague/eip7692_eof_v1/eip7069_extcall/spec.py
```diff
@@ -0,0 +1,11 @@
...
+EOF V1 Constants used throughout all tests
+"""
+
+EOF_FORK_NAME = "CancunEIP7692"
+
+CALL_FAILURE = 0
+CALL_SUCCESS = 1
+EXTCALL_SUCCESS = 0
+EXTCALL_FAILURE = 1
+EXTCALL_REVERT = 2
```
pdobacz, line 11: these should be swapped, revert is 1

> Push a status code on the stack:
> 
>     0 if the call was successful.
>     1 if the call has reverted (also can be pushed earlier in a light failure scenario).
>     2 if the call has failed.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,227 @@
...
+                    target_storage[slot_target_returndata] = stripped_address
+                case Op.CALLCODE | Op.DELEGATECALL:
+                    target_storage[slot_target_call_status] = CALL_SUCCESS
+                    # CALLCODE and DELEGATECALL call will call the stripped address
+                    # but will change the sender to self
+                    target_storage[slot_target_returndata] = address_caller
+                case Op.EXTCALL | Op.EXTSTATICCALL:
+                    target_storage[slot_target_call_status] = EXTCALL_SUCCESS
+                    # EXTCALL and EXTSTATICCALL will fault if calling an ASE address
+                    target_storage[slot_target_returndata] = 0 if ase_address else stripped_address
```
marioevz, line 189: Same as previous comment:
```suggestion
                    target_storage[slot_target_returndata] = exceptional_abort_canary if ase_address else stripped_address
```

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,227 @@
...
+        address_entry_point: Account(
+            code=(
+                Op.MSTORE(0, Op.PUSH32(target_address))
+                + Op.SSTORE(
+                    slot_top_level_call_status,
+                    Op.CALL(50000, address_caller, 0, 0, 32, 0, 0),
+                )
+                + Op.STOP()
+            ),
+            nonce=1,
```
marioevz, line 112: Add the canary here:
```suggestion
            nonce=1,
            storage={
                slot_top_level_call_status: exceptional_abort_canary,
            },
```

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,227 @@
...
+            )
+            if ase_ready_opcode
+            else Op.SSTORE(
+                slot_target_call_status,
+                target_opcode(Op.GAS, Op.CALLDATALOAD(0), *call_suffix),
+            )
+            + Op.RETURNDATACOPY(0, 0, Op.RETURNDATASIZE)
+            + Op.SSTORE(slot_target_returndata, Op.MLOAD(0))
+            + Op.STOP,
+            nonce=1,
```
marioevz, line 139: And here:
```suggestion
            nonce=1,
            storage={
                slot_target_call_status: exceptional_abort_canary,
                slot_target_returndata: exceptional_abort_canary,
            },
```

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,227 @@
...
+                            code_outputs=NON_RETURNING_SECTION,
+                            max_stack_height=2,
+                        )
+                    ],
+                ),
+                balance=0,
+                nonce=0,
+            )
+
+    target_storage: dict[int, int | bytes | Address] = {}
```
marioevz, line 169: I think this variable should be called `caller_storage` because the actual target's storage is not being modified.

marioevz (commented):
Some more comments after another review pass.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,227 @@
...
+    match target_account_type:
+        case "empty" | "EOA":
+            target_storage[slot_target_call_status] = (
+                EXTCALL_SUCCESS if ase_ready_opcode else CALL_SUCCESS
```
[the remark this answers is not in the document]
shemnon: In the case of EXTDELEGATECALL...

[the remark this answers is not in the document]
shemnon: updated to match the current spec.

tests/prague/eip7692_eof_v1/eip7069_extcall/spec.py
shemnon: Updated

tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
```diff
@@ -0,0 +1,227 @@
...
+                    target_storage[slot_target_call_status] = CALL_SUCCESS
+                    # CALL and STATICCALL call will call the stripped address
+                    target_storage[slot_target_returndata] = stripped_address
+                case Op.CALLCODE | Op.DELEGATECALL:
+                    target_storage[slot_target_call_status] = CALL_SUCCESS
+                    # CALLCODE and DELEGATECALL call will call the stripped address
+                    # but will change the sender to self
+                    target_storage[slot_target_returndata] = address_caller
+                case Op.EXTCALL | Op.EXTSTATICCALL:
+                    target_storage[slot_target_call_status] = EXTCALL_SUCCESS
```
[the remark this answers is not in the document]
shemnon: Added canary values and assert on canaries when ASE failures happen.

PR #534, opened 2024-04-30
eof tool class

winsvega:
WIP on eof tool class. 
on HOLD because we want to refactor tooling in general or perhaps design a Config class that will define all the binaries for a given client. and any client specific info in it.


winsvega:
I guess this is far behind the current state of the code

PR #537, opened 2024-05-01
feat(pytest): add pytest-html plugin for html reports

danceratopz:
## 🗒️ Description

- Adds the [pytest-html](https://pytest-html.readthedocs.io/en/latest/) plugin to EEST's dependencies so that html test reports can be generated for `fill` executions. 
- Generates html report by default. Default location is the root of `fill`'s output directory (`--output`, default is `./fixtures`).
- Adds `fill` command-line options and `t8n`/`solc` versions to the report Environment section.
- Adds links to the JSON fixture and `--evm-dump-dir` for each test (see screengrab).

Demo screengrab of report below.

### TODO

- [x] 1. Add relevant config/metadata (solc version, t8n version) to "Environment" Table (see screenshot).
- [x] 2. Enable by default and write to fixture output directory (we could use click to specify output directory, but perhaps the plugin has an option to help manage this).
- [x] 3.a. Add new "Test fixture" column to the "Summary" table (or in "Links" column?). Add link to the locally generated json fixture, if test is pass.
- [x] 3.b. Add new "Debug output dir" column to the "Summary" table with link (or in "Links" column?), if option is enabled.

### Other tasks beyond the scope of this PR:
- [ ] 4. Enable for framework tests?
- [ ] 5. Use in CI/CD?

*Considerations for use in CI/CD:*
- Filling with Cancun generates an html report of 5.9 M, even if all tests pass.
- There seems to be negligible impact on runtimes with reporting enabled.

### HTML Report Features Demo

fill_html_reports

## 🔗 Related Issues
None

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [ ] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
[...]

Issue #77, opened 2023-03-09
Error reporting is not informative 


winsvega:
I deliberatly messed up the expect section in py filler and got this

```
Traceback (most recent call last):
  File "/home/wins/Ethereum/execution-spec-tests/venv/bin/tf", line 8, in <module>
    sys.exit(main())
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_filling_tool/main.py", line 104, in main
    filler.fill()
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_filling_tool/filler.py", line 61, in fill
    future.result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_filling_tool/filler.py", line 128, in fill_fixture
    fixture = filler(t8n, b11r, "NoProof")
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_tools/filling/decorators.py", line 67, in inner
    return fill_test(
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_tools/filling/fill.py", line 53, in fill_test
    raise e
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_tools/filling/fill.py", line 43, in fill_test
    (blocks, head, alloc) = test.make_blocks(
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_tools/spec/blockchain_test.py", line 269, in make_blocks
    raise e
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_tools/spec/blockchain_test.py", line 266, in make_blocks
    verify_post_alloc(self.post, alloc)
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_tools/spec/base_test.py", line 77, in verify_post_alloc
    account.check_alloc(address, got_alloc[address])
  File "/home/wins/Ethereum/execution-spec-tests/src/ethereum_test_tools/common/types.py", line 427, in check_alloc
    expected_storage.must_be_equal(actual_storage)
[...]

PR #541, opened 2024-05-02
feat(pytest): add a plugin that removes tracebacks from test report failures summaries

danceratopz:
## 🗒️ Description
Adds a pytest plugin that modifies the "Failures" terminal summary to only show the exception name and message; it removes all traceback information for a clearer summary.

Setting `--tb=short` is close, but still contains a short traceback that is rarely helpful to determine a test failure. The traceback is unhelpful in our case as the underlying checks that trigger failures are implemented in EEST's library functions.

The stdout and stderr sections of a failing test will still be displayed in the summary.

The user can disable this plugin and achieve default traceback behavior by either:

1. Specifying the `--disable-exceptions-only` flag, or,
2. Specifying the `--tb` flag.

## Comparison

The examples below are use `-x` (exit upon first failure) - the difference between 2. and 3. is much larger with multiple failures. It's also much clearer in the HTML report (see #537).

### 1. Default behavior without this PR
Doesn't fit on screen; skip screenshot :smile: 
### 2. `--tb=short` (with or without this PR activated)
image
### 3. Default behavior with this PR
image

## 🔗 Related Issues
#77.

## ✅ Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -20,6 +20,7 @@ Test fixtures for use by clients are available for each release on the [Github r
 - ✨ Added EOF fixture format ([#512](https://github.com/ethereum/execution-spec-tests/pull/512)).
 - ✨ Verify filled EOF fixtures using `evmone-eofparse` during `fill` execution ([#519](https://github.com/ethereum/execution-spec-tests/pull/519)).
 - ✨ Added `--traces` support when running with Hyperledger Besu ([#511](https://github.com/ethereum/execution-spec-tests/pull/511)).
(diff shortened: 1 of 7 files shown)


danceratopz:
Abandoning this idea: 
1. It doesn't use `longrepr` from pytest's testreport correctly. This means that we'll lose pytest's assertion introspection capability which can is very helpful to compare two mismatching objects.
2. We need to have some traceback info. The last tb line was added in [7e21136](https://github.com/ethereum/execution-spec-tests/pull/541/commits/7e21136c672be7b8666fe672006cde55a49e1b00), but we probably need the last and first. In which case, in combination with 1. we can just settle for `--tb=short`.

PR #549, opened 2024-05-09
eof/t8n specs WIP

winsvega:
## 🗒️ Description
Specs for eof tool implementation,

## ✅ Checklist


docs/tool_specs/eof/eof_specs.md
```diff
@@ -0,0 +1,35 @@
...
+  --state.fork value (required)
+	   Sets the fork rules and all EIPs included in it for eof code validation.
+	   Current supported forks and EIPs: ["Prague"]
+
+Inputs:
+  --hex code (required)
+	   Direct input of the evm eof bytecode
+
+
+Outputs:
```
shemnon, line 29: This doesn't vibe with the tools that were written for shanghai - https://github.com/holiman/txparse/blob/main/README.md#eof-parser-eofparse - Each client already had one thing set up for this

Key changes needed::
* support a `stdin` mode where there are multiple inputs (one per line) and one output per input.
  * This is to counteract the startup penalty that half the clients have.
* Success has an output, not just terminating the validator
  * The output includes the parsed code sections.  This is usefull in detecting improper parsing on success.
  * Hencee, the success output should be required to match

marioevz:
Hey @winsvega, I revamped the specs and introduced the JSON schemas for all types that the tools are expected to parse or output, including examples for most JSON types.

Use `mkdocs serve` to see the changes and let me know what you think.

winsvega:
I am getting lost now. there was a nice description with examples why 600 lines of py code?

marioevz:
> the auto docs gen is ok, but I am not the type of a person who could read that.
> to me my initial version was much preferable. basically a short example in console on how to use it is the best type of doc for me.

Yes we now have auto-gen of these specs, I think it's important because with each fork we introduce new fields to the `Environment`, so it's easy to link implementers to this page for them to look at an example.

I really like the example parts of the new auto-gen doc, more than the parameters, I think the types are one of the most important aspects.

> this can also be hex, but not padded. "0x1", "0x2"

Yes I think it should only be hex, not decimal, more so when the other fields like `withdrawals` already have hex numbers in their fields.
If you're ok with this, I can update the environment class to not produce anything but unpadded hex numbers as you mention.

> another tricky part here that depending on the fork, this fields could be different.

This could definitely be added to the schema, I think we could have something like this:
```yaml
  currentBaseFee:
    anyOf:
    - pattern: ^([0-9]|[1-9][0-9]+)$
      title: number
      type: string
    - type: 'null'
    default: null
    title: Currentbasefee
    fork: London
```
Although this is not within the standard of the JSON schema: https://json-schema.org/understanding-json-schema/reference/annotations

marioevz:
@winsvega Updated `Environment` to use hex number exclusively in the latest commit, and if you try `mkdocs serve` again, the types and examples are automatically updated by the auto-gen:
image

winsvega:
could we keep my original description of eof code verification tool too? 

also about forks some t8n return errors if you provide cancun field on legacy fork. 
I don't think we can fully autogen all the docs. still would be nice to have written description too.

winsvega:
at this point this not my pr anymore

marioevz:
> at this point this not my pr anymore

I think it's still valuable to merge this, I'll take a look into rebasing it at some point to have it merged, but I agree that it morphed too much maybe, sorry about that.

winsvega:
@marioevz  lets finish this.

PR #550, opened 2024-05-09
more efExample tests and eof exceptions

winsvega:
## 🗒️ Description
add more eof exceptions and tests from efExample by Ori

## ✅ Checklist


src/ethereum_test_tools/eof/v1/__init__.py
```diff
@@ -209,6 +209,9 @@ def list_header(sections: List["Section"]) -> bytes:
         Creates the single code header for all code sections contained in
         the list.
         """
+        if sections[0].skip_header_listing:
```
winsvega, line 212: this is to allow types section to use skip_header_listing flag

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_example_valid_invalid.py
```diff
@@ -384,6 +391,874 @@
             EOFException.UNDEFINED_INSTRUCTION,
             id="jump_jumpdest_fails",
         ),
+        pytest.param(
```
chfast, line 394: We discussed with @marioevz and @danceratopz that this array with tests is not great. I don't mind some much the array but maybe it should be a separate test?

But what is bad that we add a ton of sophisticated tests to "example" category.

You should also add test cases to the EOF tracker file.

winsvega: I personally find it confusing when such an array is stored in another file, but here it is quite transparent

marioevz: I think the problem is that `tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_example_valid_invalid.py` contains too many unrelated test cases.

For example it tries to cover many of the deprecated opcodes but not all of them.

We need to break up `test_example_valid_invalid` into different coherent test functions testing similar things, it's too big, and it's not coherent with what it's testing (also no longer an example, it contains multiple valid tests).

src/ethereum_test_tools/exceptions/exceptions.py
```diff
@@ -247,6 +247,10 @@ class EOFException(ExceptionBase):
     """
     EOF container missing code section
     """
+    MISSING_DATA_SECTION = auto()
```
marioevz, line 250: I think this is already in the main branch, so rebasing should remove this.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_example_valid_invalid.py
```diff
@@ -384,6 +391,874 @@
...
+            EOFException.UNDEFINED_INSTRUCTION,
+            id="selfdestruct_fails",
+        ),
+        pytest.param(
+            # Check that that code that uses removed opcodes fails (Data 21)
+            Container(
+                name="EOF1I0015",
+                sections=[
+                    Section.Code(
+                        code=Op.SELFDESTRUCT(1) + Op.STOP,
```
marioevz, line 436: I don't understand the reason why we have two `SELFDESTRUCT` opcode tests here.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_example_valid_invalid.py
```diff
@@ -384,6 +391,874 @@
...
+                ],
+                auto_type_section=AutoSection.ONLY_BODY,
+            ),
+            "ef000102000100030400010000800001305000ef",
+            EOFException.MISSING_TYPE_HEADER,
+            id="bad_type_section_order_3",
+        ),
     ],
 )
 def test_example_valid_invalid(
```
marioevz, line 1264: I think we should break this function up into different ones to make it easier to find implemented tests, and we can create different functions that really test similar things. E.g. we could make a `test_deprecated_opcodes` functions and parameterize it to each opcode that is deprecated in EOF (see `V1_EOF_DEPRECATED_OPCODES` in tests/prague/eip7692_eof_v1/eip3540_eof_v1/opcodes.py, which actually needs an update), and another one for header failures, etc.

Also, I think there's definitely duplicates between this file and file `tests/prague/eip7692_eof_v1/eip3540_eof_v1/container.py`, which also should be cleaned up, but perhaps in a follow up PR.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_section_order.py
```diff
@@ -0,0 +1,220 @@
...
+    AutoSection,
+    Bytes,
+    Container,
+    EOFException,
+    Section,
+    SectionKind,
+)
+from ethereum_test_tools.eof.v1.constants import NON_RETURNING_SECTION
+
+from .spec import EOF_FORK_NAME
```
marioevz, line 22:
```suggestion
from .. import EOF_FORK_NAME
```

On latest commit in main, this is where the `EOF_FORK_NAME` is actually defined.

src/ethereum_test_tools/exceptions/exceptions.py
```diff
@@ -287,6 +291,10 @@ class EOFException(ExceptionBase):
     """
     EOF container header does not have types section first
     """
+    INVALID_CODE_SECTION_INDEX = auto()
```
marioevz, line 294: This is ok, but I think that now some tests in the `tests/prague/eip7692_eof_v1/eip6206_jumpf/test_jumpf_execution.py` should be updated because they no longer fit the `EOFException.UNDEFINED_EXCEPTION` definition (because their exception is now defined with this change).

marioevz (commented):
Some comments on the readability. I think we should make test cases easier to find in general, with better categorization, and splitting them in files where it makes sense, instead of having everything in one file/function.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_example_valid_invalid.py
winsvega: yes likely a copy paste, since the other test is actually testing data section

winsvega:
this tests will be reworked

winsvega:
I think this all has been implemented 2 times as of now.

PR #556, opened 2024-05-15
chore(fw): Utilize an EthRPC class within consume rlp

spencer-tb:
## 🗒️ Description

Adds the EthRPC class alongside the BaseRPC class. The latter are used within `tests_consume/test_via_rlp.py`, and will be extended for consume engine.

~~I added this within `src/pytest_plugins/consume/json_rpc.py` but open to moving this somewhere else.~~
Moved to: `src/ethereum_test_tools/rpc/rpc.py`

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/rpc/rpc.py
+++ b/src/ethereum_test_tools/rpc/rpc.py
@@ -1,0 +1,115 @@
+"""
+JSON-RPC methods and helper functions for EEST consume based hive simulators.
+"""
+
+from abc import ABC
+from typing import Any, Dict, List, Literal, Optional, Union
+
+import requests
+from tenacity import retry, stop_after_attempt, wait_exponential
+
+from ethereum_test_tools import Address
+
+BlockNumberType = Union[int, Literal["latest", "earliest", "pending"]]
+
+
+class BaseRPC(ABC):
+    """
+    Represents a base RPC class for every RPC call used within EEST based hive simulators.
+    """
+
+    def __init__(self, client_ip: str, port: int):
+        self.ip = client_ip
+        self.url = f"http://{client_ip}:{port}"
+
+    @retry(stop=stop_after_attempt(3), wait=wait_exponential(multiplier=1, max=10))
+    def post_request(
+        self, method: str, params: List[Any], extra_headers: Optional[Dict] = None
+    ) -> Dict:
+        """
+        Sends a JSON-RPC POST request to the client RPC server at port defined in the url.
+        """
+        payload = {
+            "jsonrpc": "2.0",
+            "method": method,
+            "params": params,
+            "id": 1,
+        }
+        base_header = {
+            "Content-Type": "application/json",
+        }
+        headers = base_header if extra_headers is None else {**base_header, **extra_headers}
+
+        response = requests.post(self.url, json=payload, headers=headers)
+        response.raise_for_status()
+        result = response.json().get("result")
+
+        if result is None or "error" in result:
+            error_info = "result is None; and therefore contains no error info"
+            error_code = None
+            if result is not None:
+                error_info = result["error"]
+                error_code = result["error"]["code"]
+            raise Exception(
+                f"Error calling JSON RPC {method}, code: {error_code}, " f"message: {error_info}"
+            )
+
+        return result
+
+
+class EthRPC(BaseRPC):
+    """
+    Represents an `eth_X` RPC class for every default ethereum RPC method used within EEST based
+    hive simulators.
+    """
+
+    def __init__(self, client_ip):
+        """
+        Initializes the EthRPC class with the http port 8545, which requires no authentication.
+        """
+        super().__init__(client_ip, port=8545)
+
+    BlockNumberType = Union[int, Literal["latest", "earliest", "pending"]]
+
+    def get_block_by_number(self, block_number: BlockNumberType = "latest", full_txs: bool = True):
+        """
+        `eth_getBlockByNumber`: Returns information about a block by block number.
+        """
+        block = hex(block_number) if isinstance(block_number, int) else block_number
+        return self.post_request("eth_getBlockByNumber", [block, full_txs])
+
+    def get_balance(self, address: str, block_number: BlockNumberType = "latest"):
+        """
+        `eth_getBalance`: Returns the balance of the account of given address.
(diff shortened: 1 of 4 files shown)


src/ethereum_test_tools/rpc/rpc.py
```diff
@@ -0,0 +1,118 @@
...
+from ethereum_test_tools import Address
+
+
+class BaseRPC:
+    """
+    Represents a base RPC class for every RPC call used within EEST based hive simulators.
+    """
+
+    def __init__(self, client):
+        self.client = client
```
danceratopz, line 21: This is of type `hive.client.Client`, I don't think this has any business in a generic RPC class :slightly_smiling_face:.

src/ethereum_test_tools/rpc/rpc.py
```diff
@@ -0,0 +1,118 @@
...
+        self.url = f"http://{client.ip}:8551"
+        self.jwt_secret = (
+            b"secretsecretsecretsecretsecretse"  # oh wow, guess its not a secret anymore
```
danceratopz, line 24: Us this `jwt_secret` value used by Hive? Or any reason for this default?

src/ethereum_test_tools/rpc/rpc.py
```diff
@@ -0,0 +1,118 @@
...
+        return result
+
+
+class EthRPC(BaseRPC):
+    """
+    Represents an `eth_X` RPC class for every default ethereum RPC method used within EEST based
+    hive simulators.
+    """
+
+    BlockNumberType = Union[int, Literal["latest", "earliest", "pending"]]
```
danceratopz, line 73: I think I would move this type definition out of the class and expose it via __init__ for easier use in other modules.

danceratopz (commented):
Hey, just a few comments! Happy to address these during the `gentest` refactor.

src/ethereum_test_tools/rpc/rpc.py
spencer-tb: Its required for the engine rpc commands not the eth rpc commands! 

I can remove it and and add it with then engine PR?

src/ethereum_test_tools/rpc/rpc.py
spencer-tb: Yeah exactly, its so we can instantiate an RPC "client" based on the hive client we are sending the RPC requests to.

I put it here originally so we can define both `eth_rpc = EthRPC(geth_client)` and `engine_rpc = EngineRPC(besu_client)`, to then use the respective rpc functions appropriately, `engine_rpc.new_payload(...)`.

src/ethereum_test_tools/rpc/rpc.py
spencer-tb: Added it to init here: [49352c5](https://github.com/ethereum/execution-spec-tests/pull/556/commits/49352c576871dc0c3978dc0901c81280ff5d6d8c)

src/ethereum_test_tools/rpc/rpc.py
spencer-tb: I removed the jwt auth. here: [32bfef2](https://github.com/ethereum/execution-spec-tests/pull/556/commits/32bfef2774a96fe6f7754abd9d29170b9dee16ee)

As its only specific to engine calls I can re-add it to a seperate PR.

src/ethereum_test_tools/rpc/rpc.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+
+    def __init__(self, client, port):
+        self.client = client
+        self.url = f"http://{client.ip}:{port}"
+
+    @retry(stop=stop_after_attempt(3), wait=wait_exponential(multiplier=1, max=10))
+    def post_request(self, method, params):
+        """
+        Sends a JSON-RPC POST request to the client RPC server at port 8551.
```
marioevz, line 27:
```suggestion
        Sends a JSON-RPC POST request to the client RPC server at port defined in the url.
```

src/ethereum_test_tools/rpc/rpc.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def storage_at_keys(
+        self, account: str, keys: List[str], block_number: BlockNumberType = "latest"
+    ) -> Dict:
+        """
+        Helper to retrieve the storage values for the specified keys at a given address and block
+        number.
+        """
+        if isinstance(block_number, int):
+            block_number
```
marioevz, line 108:
```suggestion
```

This seems like it should be deleted.

marioevz (approved):
LGTM! Two minor comments.

src/ethereum_test_tools/rpc/rpc.py
danceratopz: Yes, but the hive client is only used to get the `ip` address, as far as I can tell? In which case I would suggest providing the ip to the constructor instead of a hive client. Otherwise any user of this class has to instantiate a hive client first, which is a bit weird (e.g., `src/cli/gentest.py`).

src/ethereum_test_tools/rpc/rpc.py
danceratopz: You can leave it :) I just thought we could change the comment to say whey we use this default.

src/ethereum_test_tools/rpc/rpc.py
```diff
@@ -0,0 +1,111 @@
...
+
+import requests
+from tenacity import retry, stop_after_attempt, wait_exponential
+
+from ethereum_test_tools import Address
+
+BlockNumberType = Union[int, Literal["latest", "earliest", "pending"]]
+
+
+class BaseRPC:
```
danceratopz, line 15: We could inherit python's `abc.ABC` abstract base class library class to express/enforce that `BaseRPC` can never be instantiated. If we add abstract methods we can mark them as such using the `abc.abstractmethod` decorator. We don't widely use this yet, but perhaps we should.

```suggestion
from abc import ABC

class BaseRPC(ABC):
```
This will break tox ))

danceratopz (commented):
One more small comment!

danceratopz (approved):
LGTM! :rocket:

Issue #498, opened 2024-04-09
Prague Tracker


marioevz:
```[tasklist]
### EIPs
- [x] [EIP-2537: Precompile for BLS12-381 curve operations](https://eips.ethereum.org/EIPS/eip-2537)
- [x] [EIP-2935: Save historical block hashes in state](https://eips.ethereum.org/EIPS/eip-2935)
- [x] [EIP-3074: AUTH and AUTHCALL opcodes](https://eips.ethereum.org/EIPS/eip-3074)
- [x] [EIP-6110: Supply validator deposits on chain](https://eips.ethereum.org/EIPS/eip-6110)
- [x] [EIP-7002: Execution layer triggerable exits](https://eips.ethereum.org/EIPS/eip-7002)
- [x] [EIP-7251: Increase the MAX_EFFECTIVE_BALANCE](https://eips.ethereum.org/EIPS/eip-7251) *
```

* Impact depends on whether the EIP includes modifications to the Execution layer or not.

marioevz:
Devops Devnet-0 Specs: https://notes.ethereum.org/@ethpandaops/pectra-devnet-0

spencer-tb:
CIFO #1299

PR #564, opened 2024-05-17
new(tests): EIP-2935: Serve historical block hashes from state

marioevz:
## 🗒️ Description

Implement EIP-2935 tests for Prague.

## 🔗 Related Issues
#498

## ✅ Checklist


tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py
```diff
@@ -0,0 +1,175 @@
...
+        block_outside_window = current_block_number - Spec.HISTORY_SERVE_WINDOW - 1
+
+        # Check using BLOCKHASH, must be zero
+        code += Op.SSTORE(storage.store_next(1), Op.ISZERO(Op.BLOCKHASH(block_outside_window)))
+
+        # Check calling the HISTORY_STORAGE_ADDRESS, must be zero
+        code += Op.MSTORE(0, block_outside_window)
+        code += Op.POP(Op.CALL(Op.GAS, Spec.HISTORY_STORAGE_ADDRESS, 0, 0, 32, 0, 32))
+        code += Op.RETURNDATACOPY(0, 0, 32)
+        code += Op.SSTORE(storage.store_next(1), Op.ISZERO(Op.MLOAD(0)))
```
spencer-tb, line 135: I think we could make this code a reusuable pytest fixture in the future, if we decide to keep it post spec change.

spencer-tb (approved):
LGTM! Happy for you to merge this after tox is passing.

I'm aware that we need to update this based on these change:
https://github.com/ethereum/EIPs/pull/8577

I really liked the approach with the `contract.bin` file, but I guess we wont need it now. 

I tried to outline what the key changes below are from my understanding, I think we could add extra tests for some of these cases, on the off chance clients keep some of the original funcitionality:

1) Block hash storage: only update the ring buffer starting from the fork block and not including additional ancestors. Maybe add a test to check that additional ancestors are not included.
2) BLOCKHASH: before the fork check that the opcode returns the blockhash for blocks within the old blockhash window, zero otherwise. We could extend this to be 256 + 1 blocks before transitioning. After the fork check the same for the appropriate new windonw. I believe it should be gradual, so we could check inside an outside for each block as the ring buffer is filling. Up until > 8192 blocks. We could maybe do this all in one test.
3) BLOCKHASH gas costs,  as with the new update the opcode gas cost now remains the same. We could add a sanity check for this to make sure clients adhere to this.

spencer-tb (commented):
Oh and then the CHANGELOG too!

tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py
marioevz: Tried my best, but since we are not doing one check per parameterization, I reduced the code generation to a function, and it's a bit more readable.

PR #579, opened 2024-05-29
feat(fw/consume): write the test class & function docstrings to `_info["description"]` for use in hive reports

danceratopz:
## 🗒️ Description

This PR: 
- Adds a new field "`description`" in the `_info` section that contains the current test class and test function docstrings concatenated. 
- Uses this new field and sends it to the Hive Server when creating a new `HiveTest` in the `pytest_hive` plugin. This makes a test description available in HiveView.

Example:

```json
        "_info": {
            "hash": "0xd7409ee1b8337363780056f9703c8a11122191196155f1e1837b4bce19d1c884",
            "comment": "`execution-spec-tests` generated test",
            "filling-transition-tool": "evm version 1.14.1-unstable-d6e91e2e-20240507",
            "description": "Test class documentation:\n\n    Test contract creation via the CREATE/CREATE2 opcodes that have an initcode\n    that is on/over the max allowed limit.\n    \n\nTest function documentation:\n\n        Test contract creation via the CREATE/CREATE2 opcodes that have an\n        initcode that is on/over the max allowed limit.",
            "reference-spec": "https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3860.md",
            "reference-spec-version": "5f8151e19ad1c99da4bafd514ce0e8ab89783c8f"
        }
```

### TODO

- [ ] Clean-up duplicate info contained in class and function docstrings (as in the current example above) in the scope of this PR (they're aren't many test classes).
- [ ] Test in `hiveview`

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/spec/base/base_test.py
+++ b/src/ethereum_test_tools/spec/base/base_test.py
@@ -100,6 +100,7 @@ class BaseFixture(CamelModel):
     def fill_info(
         self,
         t8n: TransitionTool,
+        fixture_description: str,
         ref_spec: ReferenceSpec | None,
     ):
         """
@@ -108,6 +109,7 @@ class BaseFixture(CamelModel):
         if "comment" not in self.info:
             self.info["comment"] = "`execution-spec-tests` generated test"
         self.info["filling-transition-tool"] = t8n.version()
+        self.info["description"] = fixture_description
         if ref_spec is not None:
             ref_spec.write_info(self.info)
 
--- a/src/pytest_plugins/consume/simulator_common.py
+++ b/src/pytest_plugins/consume/simulator_common.py
@@ -36,3 +36,13 @@ def fixture(fixture_source: JsonSource, test_case: TestCase) -> Fixture:
         fixtures = BlockchainFixtures.from_file(Path(fixture_source) / test_case.json_path)
         fixture = fixtures[test_case.id]
     return fixture
+
+
+@pytest.fixture(scope="function")
+def fixture_description(fixture: Fixture) -> str:
+    """
+    Return the description of the current test case.
+    """
+    if "description" not in fixture.info:
+        return "No description field provided in the fixture's 'info' section."
+    return fixture.info["description"]
--- a/src/pytest_plugins/pytest_hive/pytest_hive.py
+++ b/src/pytest_plugins/pytest_hive/pytest_hive.py
@@ -107,11 +107,17 @@ def hive_test(request, test_suite: HiveTestSuite):
     """
     Propagate the pytest test case and its result to the hive server.
     """
+    try:
+        fixture_description = request.getfixturevalue("fixture_description")
+    except pytest.FixtureLookupError:
+        pytest.exit(
+            "Error: The 'fixture_description' fixture has not been defined by the simulator "
+            "or pytest plugin using this plugin!"
+        )
     test_parameter_string = request.node.nodeid.split("[")[-1].rstrip("]")  # test fixture name
     test: HiveTest = test_suite.start_test(
-        # TODO: pass test case documentation when available
         name=test_parameter_string,
-        description="TODO: This should come from the '_info' field.",
+        description=fixture_description,
     )
     yield test
     try:
--- a/src/pytest_plugins/test_filler/test_filler.py
+++ b/src/pytest_plugins/test_filler/test_filler.py
@@ -559,6 +559,24 @@ def node_to_test_info(node) -> TestInfo:
     )
 
 
+@pytest.fixture(scope="function")
+def fixture_description(request):
+    """Fixture to extract and combine docstrings from the test class and the test function."""
+    description_unavailable = (
+        "No description available - add a docstring to the python test class or function."
+    )
+    test_class_doc = f"Test class documentation:\n{request.cls.__doc__}" if request.cls else ""
+    test_function_doc = (
+        f"Test function documentation:\n{request.function.__doc__}"
+        if request.function.__doc__
+        else ""
+    )
+    if not test_class_doc and not test_function_doc:
+        return description_unavailable
+    combined_docstring = f"{test_class_doc}\n\n{test_function_doc}".strip()
+    return combined_docstring
+
+
 def base_test_parametrizer(cls: Type[BaseTest]):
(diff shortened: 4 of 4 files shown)


src/pytest_plugins/consume/simulator_common.py
```diff
@@ -36,3 +36,18 @@ def fixture(fixture_source: JsonSource, test_case: TestCase) -> Fixture:
...
+    """
+    Return the description of the current test case.
+    """
+    description = f"Test id: {test_case.id}"
+    if "url" in fixture.info:
+        description += f"\n\nTest source: {fixture.info['url']}"
+    if "description" not in fixture.info:
+        description += "\n\nNo description field provided in the fixture's 'info' section."
+    else:
+        description += f"\n\n{fixture.info['description']}"
```
spencer-tb, line 52: Personally I prefer having only the test ID and URL. It would be amazing to have the test descriptions but I dont find the tradeoff of adding it to the fixture to be worth it

I'm open to it if others like it though

src/pytest_plugins/pytest_hive/pytest_hive.py
```diff
@@ -107,11 +107,17 @@ def hive_test(request, test_suite: HiveTestSuite):
...
     Propagate the pytest test case and its result to the hive server.
     """
+    try:
+        fixture_description = request.getfixturevalue("fixture_description")
+    except pytest.FixtureLookupError:
+        pytest.exit(
+            "Error: The 'fixture_description' fixture has not been defined by the simulator "
+            "or pytest plugin using this plugin!"
+        )
     test_parameter_string = request.node.nodeid.split("[")[-1].rstrip("]")  # test fixture name
```
spencer-tb, line 117:
```suggestion
    test_parameter_string = request.node.nodeid  # test fixture name
```

I'm not sure if this is the correct suggestion, but for hive pyspec we have the full test name:
image

Currently in consume rlp we have:
image
for the same test - and from that alone I have no idea what the test is 🥲

spencer-tb (commented):
Thanks for fixing this! Added a couple comments :)

src/pytest_plugins/consume/simulator_common.py
spencer-tb: (will wait and see what the generated fixtures look like, maybe it will change my mind)

src/pytest_plugins/pytest_hive/pytest_hive.py
danceratopz: Ah ok, yes, but that's an existing issue that i must have chose to ignore - it's not directly related to what we're adding n this PR :-) Will take a look!

But with this PR, you will get the fixture **ID** in the "Description" for easy copy-paste. I.e., you can copy this ID pass it to `fill`. 

Perhaps we can add the consume test ID too, so that you pass it directly to `consume` - this is probably more helpful.

I think the goal would be to provide versions of all components and two command-lines in the description to allow easy local re-execution via either:
1. `consume rlp` (or `consume engine`).
2. `./hive`.

spencer-tb: > Perhaps we can add the consume test ID too, so that you pass it directly to consume - this is probably more helpful.

💯

spencer-tb: Would love a backwards compatible `-k` option for hive actually. Where running hive with `--sim.limit "/eip2935"` does the same as `consume -k eip2935` :D

danceratopz:
@spencer-tb Here's the size difference in fixture directories with and without "description" and "url" fields. Seems reasonable, imo. 
```
➜ du -hs fixtures-description/ fixtures-main
82M	fixtures-description/
80M	fixtures-main

execution-spec-tests on  main [$!?] via 🐍 310-est-gentest-rpc-refactor 
➜ du -s fixtures-description/ fixtures-main
83748	fixtures-description/
81096	fixtures-main
```

src/pytest_plugins/consume/simulator_common.py
danceratopz: https://github.com/ethereum/execution-spec-tests/pull/579#issuecomment-2139749086

marioevz:
This is awesome :grinning:

src/ethereum_test_tools/utility/versioning.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    Generate a permalink to a source file in Github.
+    """
+    base_url = "https://github.com"
+    username = "ethereum"
+    repository = "execution-spec-tests"
+    if line_number:
+        line_number = f"#L{line_number}"
+    if re.match(
+        r"^v[0-9]{1,2}\.[0-9]{1,3}\.[0-9]{1,3}(a[0-9]+|b[0-9]+|rc[0-9]+)?$",
```
marioevz, line 40: You're underestimating us getting to v100 :stuck_out_tongue:

src/ethereum_test_tools/utility/versioning.py
```diff
@@ -0,0 +1,51 @@
...
+        branch_or_commit_or_tag,
+    ):
+        return (
+            f"{base_url}/{username}/{repository}/tree/"
```
marioevz, line 44: Could also do:
```python
tree_or_blob = "tree" if re.match(...) else "blob"
...
f"{base_url}/{username}/{repository}/{tree_or_blob}/"
```

marioevz (commented):
Looking good so far! Some very subtle comments.

src/ethereum_test_tools/utility/versioning.py
danceratopz: lol, increased the runway here! At least for testing purposes :wink:

src/ethereum_test_tools/spec/base/base_test.py
```diff
@@ -108,6 +110,8 @@ def fill_info(
         if "comment" not in self.info:
             self.info["comment"] = "`execution-spec-tests` generated test"
```
marioevz, line 111: I think we could remove the "comment" ?

marioevz (approved):
LGTM!

You can apply or discard the comment I made.

PR #589, opened 2024-06-04
Add more validation cases to EOF tests tracker

gumb0:


gumb0:
@marioevz this part is ready, PTAL

marioevz (approved):
LGTM. I think we need to start coming back to the tracker and either retroactively fill or fill along with implementation.

PR #592, opened 2024-06-05
feat(tests): eof section order tests

winsvega:
## 🗒️ Description
Construct eof container with skipped sections combinations

Example: 
- take code section, make it appear only in container header and header types, but skip printing bytes in the body
- take code section, make it appear only in container body but skip printing bytes in the header

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/pull/550

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [ ] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [x] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.

PR #598, opened 2024-06-06
new(tests): EOF - EIP-3540: Migrate validation tests: EIP3540/validInvalidFiller.yml

chfast:
## 🗒️ Description

Migrate https://github.com/ethereum/tests/blob/develop/src/EOFTestsFiller/EIP3540/validInvalidFiller.yml.

## ✅ Checklist

### DIFF
--- a/converted-ethereum-tests.txt
+++ b/converted-ethereum-tests.txt
@@ -2,6 +2,9 @@
 GeneralStateTests/stCreate2/call_outsize_then_create2_successful_then_returndatasize.json
 GeneralStateTests/stCreate2/call_then_create2_successful_then_returndatasize.json
 
+([#598](https://github.com/ethereum/execution-spec-tests/pull/598))
+EOFTests/EIP3540/validInvalid.json
+
 EOFTests/efValidation/EOF1_eofcreate_valid_.json
 EOFTests/efValidation/EOF1_truncated_section_.json
 
--- a/tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
+++ b/tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
@@ -293,6 +293,64 @@ pytestmark = pytest.mark.valid_from(EOF_FORK_NAME)
             EOFException.MISSING_TERMINATOR,
             id="EOF1I3540_0049",
         ),
+        # TODO: Duplicated tests
+        # The following test cases are duplicates but added to confirm test coverage is retained.
+        pytest.param(
+            Container(
+                name="EOF1I3540_0001 (Invalid) No magic",
+                raw_bytes="ef",
+            ),
+            EOFException.INVALID_MAGIC,
+        ),
+        pytest.param(
+            Container(
+                name="EOF1I3540_0002 (Invalid) Invalid magic",
+                raw_bytes="ef010101000402000100010400000000000000fe",
+            ),
+            EOFException.INVALID_MAGIC,
+        ),
+        pytest.param(
+            Container(
+                name="EOF1I3540_0003",
+                raw_bytes="ef020101000402000100010400000000000000fe",
+            ),
+            EOFException.INVALID_MAGIC,
+        ),
+        pytest.param(
+            Container(
+                name="EOF1I3540_0004",
+                raw_bytes="efff0101000402000100010400000000000000fe",
+            ),
+            EOFException.INVALID_MAGIC,
+        ),
+        pytest.param(
+            Container(
+                name="EOF1I3540_0005 (Invalid) No version",
+                raw_bytes="ef00",
+            ),
+            EOFException.INVALID_VERSION,
+        ),
+        pytest.param(
+            Container(
+                name="EOF1I3540_0006 (Invalid) Invalid version",
+                raw_bytes="ef000001000402000100010400000000000000fe",
+            ),
+            EOFException.INVALID_VERSION,
+        ),
+        pytest.param(
+            Container(
+                name="EOF1I3540_0007",
+                raw_bytes="ef000201000402000100010400000000000000fe",
+            ),
+            EOFException.INVALID_VERSION,
+        ),
+        pytest.param(
+            Container(
+                name="EOF1I3540_0008",
+                raw_bytes="ef00ff01000402000100010400000000000000fe",
+            ),
+            EOFException.INVALID_VERSION,
+        ),
     ],
 )
 def test_migrated_valid_invalid(
--- a/tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
+++ b/tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
@@ -75,7 +75,10 @@ pytestmark = pytest.mark.valid_from(EOF_FORK_NAME)
         ),
         pytest.param(
             # Empty code section with non-empty data section
-            bytes.fromhex("ef000101000402000100000400020000000000aabb"),
+            Container(
+                sections=[Section.Code(code_outputs=0), Section.Data("aabb")],
+                expected_bytecode="ef000101000402000100000400020000000000aabb",
+            ),
             EOFException.ZERO_SECTION_SIZE,
             id="EOF1I3540_0012",
         ),
--- a/tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
+++ b/tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
@@ -1,0 +1,287 @@
+"""
+EOF validation tests for EIP-3540 migrated from
+ethereum/tests/src/EOFTestsFiller/EIP3540/validInvalidFiller.yml
+"""
+
+import pytest
+
+from ethereum_test_tools import EOFTestFiller
+from ethereum_test_tools import Opcodes as Op
+from ethereum_test_tools.eof.v1 import Container, EOFException, Section
+
+from .. import EOF_FORK_NAME
+
+REFERENCE_SPEC_GIT_PATH = "EIPS/eip-3540.md"
+REFERENCE_SPEC_VERSION = "8dcb0a8c1c0102c87224308028632cc986a61183"
+
+pytestmark = pytest.mark.valid_from(EOF_FORK_NAME)
+
+
+@pytest.mark.parametrize(
+    "eof_code,exception",
+    [
+        pytest.param(
+            # Deployed code without data section
+            Container(
+                name="EOF1V3540_0001",
+                sections=[
+                    Section.Code(code=Op.PUSH1[0] + Op.POP + Op.STOP, max_stack_height=1),
+                ],
+            ),
+            None,
+            id="EOF1V3540_0001",
+        ),
+        pytest.param(
+            # Deployed code with data section
+            Container(
+                name="EOF1V3540_0002",
+                sections=[
+                    Section.Code(code=Op.PUSH1[0] + Op.POP + Op.STOP, max_stack_height=1),
+                    Section.Data("aabbccdd"),
+                ],
+            ),
+            None,
+            id="EOF1V3540_0002",
+        ),
+        pytest.param(
+            # No data section contents (valid according to relaxed validation)
+            Container(
+                name="EOF1V3540_0003",
+                sections=[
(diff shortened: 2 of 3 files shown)


tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+            Container(
+                name="EOF1V3540_0004",
+                sections=[
+                    Section.Code(code=Op.INVALID),
+                    Section.Data("aa", custom_size=2),
+                ],
+            ),
+            None,
+            id="EOF1V3540_0004",
+        ),
```
marioevz, line 70: These fail on evmone after https://github.com/ethereum/evmone/pull/921.

Maybe embedding these in a subcontainer would be the way to go.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+            # Unknown section ID (after data section)
+            bytes.fromhex("ef00010100040200010001040000ff00010000000000fe"),
+            EOFException.MISSING_TERMINATOR,
+            id="EOF1I3540_0049",
+        ),
+    ],
+)
+def test_migrated_valid_invalid(
+    eof_test: EOFTestFiller,
+    eof_code: Container | bytes,
```
marioevz, line 279: We could accept `Container | str` here instead of bytes, and remove whitespaces and do the `bytes.fromhex` inside the function:
```python
if type(eof_code) is str:
    eof_code = bytes.fromhex(eof_code.replace(" ", ""))  # type: ignore
```

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+        pytest.param(
+            # No data section after code section size
+            bytes.fromhex("ef00010100040200010001"),
+            EOFException.MISSING_HEADERS_TERMINATOR,
+            id="EOF1I3540_0017",
+        ),
+        pytest.param(
+            # No data size
+            bytes.fromhex("ef0001010004020001000104"),
+            EOFException.MISSING_HEADERS_TERMINATOR,
```
marioevz, line 92: We can remove `"no_data_section_size"` from `tests/prague/eip7692_eof_v1/eip3540_eof_v1/container.py` (which is actually incorrect) in this same PR because this test supersedes it.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+            id="EOF1I3540_0018",
+        ),
+        pytest.param(
+            # Data size incomplete
+            bytes.fromhex("ef000101000402000100010400"),
+            EOFException.INCOMPLETE_SECTION_SIZE,
```
marioevz, line 98: Also remove `"data_section_size_incomplete"` from `tests/prague/eip7692_eof_v1/eip3540_eof_v1/container.py`.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+REFERENCE_SPEC_VERSION = "8dcb0a8c1c0102c87224308028632cc986a61183"
+
+pytestmark = pytest.mark.valid_from(EOF_FORK_NAME)
+
+
+@pytest.mark.parametrize(
+    "eof_code,exception",
+    [
+        pytest.param(
+            # Deployed code without data section
```
marioevz, line 25: I think it's better to have these comments as the ID, because the current string that is used, while it links to the original test, is not very descriptive.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+        ),
+        pytest.param(
+            # Code section contents incomplete
+            bytes.fromhex("ef0001010004020001002904000000000000027f"),
+            EOFException.INVALID_SECTION_BODIES_SIZE,
+            id="EOF1I3540_0026",
+        ),
+        pytest.param(
+            # Trailing bytes after code section
+            bytes.fromhex("ef000101000402000100010400000000000000feaabbcc"),
```
marioevz, line 145:
```suggestion
            "ef00 01 01 0004 02 0001 0001 04 0000 00 0080 0000 fe aabbcc",
```
It's currently failing with `EOFException.INVALID_FIRST_SECTION_TYPE`.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+            bytes.fromhex("ef000104000201000402000100010000000000feaabb"),
+            EOFException.MISSING_TYPE_HEADER,
+            id="EOF1I3540_0031",
+        ),
+        pytest.param(
+            # Data section preceding code section
+            bytes.fromhex("ef000101000404000202000100010000000000feaabb"),
+            EOFException.MISSING_CODE_HEADER,
+            id="EOF1I3540_0032",
+        ),
```
marioevz, line 172: These three cases are also being tested in https://github.com/ethereum/execution-spec-tests/pull/592.

The other PR is a file dedicated to section order testing and places the tests in a dedicated file too, so I think we can remove these tests from this file.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+            bytes.fromhex("ef00010100040400020000000000aabb"),
+            EOFException.MISSING_CODE_HEADER,
+            id="EOF1I3540_0033",
+        ),
+        pytest.param(
+            # No data section
+            bytes.fromhex("ef000101000402000100010000000000fe"),
+            EOFException.MISSING_DATA_SECTION,
+            id="EOF1I3540_0034",
+        ),
```
marioevz, line 184: These tests are also covered by #592

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+            bytes.fromhex("ef000101000402000100010400020000000000feaabbccdd"),
+            EOFException.INVALID_SECTION_BODIES_SIZE,
+            id="EOF1I3540_0035",
+        ),
+        pytest.param(
+            # Multiple data sections
+            bytes.fromhex("ef000101000402000100010400020400020000000000feaabbaabb"),
+            EOFException.MISSING_TERMINATOR,
+            id="EOF1I3540_0036",
+        ),
```
marioevz, line 196: This test case is repeated in `"multiple_data_sections"` case in `tests/prague/eip7692_eof_v1/eip3540_eof_v1/container.py`, although in this case we could use the other file approach because it also correctly fills the type section:
```python
Container(
        name="multiple_data_sections",
        sections=[
            Section.Code(code=Op.INVALID),
            Section.Data(data="0xAABB"),
            Section.Data(data="0xAABB"),
        ],
        validity_error=EOFException.MISSING_TERMINATOR,
    )
```
And this yields almost the same bytecode but with the correct type section body.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+        pytest.param(
+            # Multiple code and data sections
+            bytes.fromhex("ef000101000802000200010001040002040002000000000000000000fefeaabbaabb"),
+            EOFException.MISSING_TERMINATOR,
+            id="EOF1I3540_0037",
+        ),
```
marioevz, line 202: Same as previous comment:
```python
Container(
    name="multiple_code_and_data_sections",
    sections=[
        Section.Code(Op.INVALID),
        Section.Code(Op.INVALID),
        Section.Data(data="0xAABB"),
        Section.Data(data="0xAABB"),
    ],
    validity_error=EOFException.MISSING_TERMINATOR,
),
```

marioevz (commented):
Some comments. I think the end-goal should be to get rid of `tests/prague/eip7692_eof_v1/eip3540_eof_v1/container.py` and place the tests it contains into several better categorized tests.

converted-ethereum-tests.txt
```diff
@@ -0,0 +1,4 @@
+src/
```
winsvega, line 1: this must be a path to `EOFTests/EIP3540/validInvalid.json` to be parsed for the coverage script

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
```diff
@@ -0,0 +1,288 @@
...
+    [
+        pytest.param(
+            # Deployed code without data section
+            Container(
+                name="EOF1V3540_0001",
+                sections=[
+                    Section.Code(code=Op.PUSH1[0] + Op.POP + Op.STOP, max_stack_height=1),
+                ],
+            ),
+            None,
```
winsvega, line 32: I am getting exceptions here and exception names mismatch. 
Does this PR require most recent evmone build? 
if so to measure the coverage I must update the evmone docker image

converted-ethereum-tests.txt
chfast: ok, is the script public? this is not explain anywhere is the file is empty currently.

winsvega: its in https://github.com/ethereum/execution-spec-tests/blob/main/.github/workflows/coverage.yaml
and https://github.com/winsvega/evmtest_coverage/tree/main/dockerhub

the result shows here in github workflow

chfast:
I put it on hold because this is too fragile: it requires evmone, ethereum/tests and EEST to full agree.

chfast:
I have no idea where the CI failure is...

chfast:
Because the coverage comparison only considers the newly added python file removing duplicated tests is impractical. I'm not sure how do you want to handle this.

Moreover, the coverage report also points to the lost coverage in the "bytes to hex" helper functions. I don't know why exactly this happens but these are irrelevant. Quick fix is probably to exclude EVMC from the coverage.

image

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
chfast: I will convert everything into `Container` only. However, I noticed that a `pytest.param` with `Cointainer` type produces very long test names... is this expected?

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
winsvega: Is it the same file from test repo. Validinvalid?

chfast: You added these @winsvega.

chfast:
Should I push these tests further?

marioevz (approved):
Thanks!

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
marioevz: Not really, let me push a quick fix for this.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_migrated_valid_invalid.py
marioevz: Resolving conversations as we plan to remove "tests/prague/eip7692_eof_v1/eip3540_eof_v1/container.py" eventually.

PR #602, opened 2024-06-06
feat(fw): Add `with_all_contract_creating_tx_types` marker

marioevz:
## 🗒️ Description
- Makes the transaction class more permissible to always allow the creation of a valid transaction with the minimum number of set fields (in order to more easily parameterize the same test with multiple transaction types without having to worry about the specifics of each type).
- Fix for evmone when passing a contract-creating tx: Use `tx: ""` instead of `tx: null`.
- Add a `contract_creating_tx_types` to all forks to return all the supported transactions that can create contracts for the given fork.
- Add a new covariant marker called `with_all_contract_creating_tx_types` to automatically parameterize a test with all the fork supported transaction types that can create a contract.

~~TODO: Check 1-to-1 fixture generation~~ 1-to-1  fixture generation is verified.

## 🔗 Related Issues
None

## ✅ Checklist


docs/writing_tests/test_markers.md
```diff
@@ -0,0 +1,111 @@
...
+
+This marker is used to mark a test as expected to fail.
+
+```python
+import pytest
+
+@pytest.mark.xfail("EVM binary doesn't support this opcode")
+def test_something():
+    pass
+```
```
spencer-tb, line 111: Really nice docs :)

src/evm_transition_tool/evmone.py
```diff
@@ -37,3 +37,11 @@ def is_fork_supported(self, fork: Fork) -> bool:
         Currently, evmone-t8n provides no way to determine supported forks.
         """
         return True
+
+    @classmethod
+    def empty_string_to(cls) -> bool:
+        """
+        Evmone requires an empty string to be passed as an `to` field in the case of
+        contract-creating transactions.
```
spencer-tb, line 45:
```suggestion
        Evmone requires an empty string within the `to` field for
        contract-creating transactions.
```

src/evm_transition_tool/transition_tool.py
```diff
@@ -225,6 +225,13 @@ def detect_binary(cls, binary_output: str) -> bool:
 
         return cls.detect_binary_pattern.match(binary_output) is not None
 
+    @classmethod
+    def empty_string_to(cls) -> bool:
+        """
+        Returns True if the tool requires empty strings as addresses
```
spencer-tb, line 231:
```suggestion
        Returns True if the tool requires an empty string `to` field
        for contract creating transactions.
```

src/evm_transition_tool/transition_tool.py
```diff
@@ -311,6 +318,18 @@ class TransitionToolData:
...
+        def __post_init__(self):
+            """
+            Ensure that the `to` field of transactions is not None
+            """
+            if self.empty_string_to:
+                for tx in self.txs:
+                    if "to" not in tx:
+                        tx["to"] = ""
+                    elif tx["to"] is None:
+                        tx["to"] = ""
```
spencer-tb, line 332: I think we can do this:
```suggestion
                for tx in self.txs:
                    tx["to"] = tx.get("to") or ""
```

src/evm_transition_tool/transition_tool.py
```diff
@@ -571,7 +590,13 @@ def evaluate(
...
             reward = -1
         t8n_data = TransitionTool.TransitionToolData(
-            alloc=alloc, txs=txs, env=env, fork_name=fork_name, chain_id=chain_id, reward=reward
+            alloc=alloc,
+            txs=txs,
+            env=env,
+            fork_name=fork_name,
+            chain_id=chain_id,
+            reward=reward,
+            empty_string_to=self.empty_string_to(),
```
spencer-tb, line 599: A part of me feels like it would be better for the evmone t8n to handle this, as this maybe bloats `t8n_data` with a field that isn's "super" important in comparison to the others.

I think it could be worth exploring, but very happy with this as a workaround :D

spencer-tb (commented):
Really nice addition! I commented a few small potenetial tweaks if you want to add them.

And also a comment about checking whether evmone could handle the empty string case on there end. My naive-ity might be underestimating C++ though

src/evm_transition_tool/transition_tool.py
marioevz: I agree, both tools should expect the same type of input.

Filed an issue in the evmone repo for this: https://github.com/ethereum/evmone/issues/923

Should we merge as is and rollback once the behavior is standardized?

spencer-tb: Lets do it! :D

spencer-tb (approved):

PR #610, opened 2024-06-07
feat(fw,forks,tests): Add EVM code type marker

marioevz:
### `@pytest.mark.with_all_evm_code_types` marker
Adds the `with_all_evm_code_types` marker that can be added to a test as follows:
```python
@pytest.mark.with_all_evm_code_types
@pytest.mark.valid_from("Frontier")
def test_something_with_all_evm_code_types(evm_code_type: EVMCodeType):
    pass
```
and the test will be parameterized for parameter `evm_code_type` only with value `[EVMCodeType.LEGACY]` starting on fork Frontier, and eventually it will be parametrized with with values `[EVMCodeType.LEGACY, EVMCodeType.EOF_V1]` on the EOF activation fork.

Adding the `evm_code_type` parameter to the function signature is optional and can be done only if required.

In all calls to `pre.deploy_contract`, if the code parameter is `Bytecode` type, and `evm_code_type==EVMCodeType.EOF_V1`, the bytecode will be automatically wrapped in an EOF V1 container.

Code wrapping might fail in the following circumstances:

- The code contains invalid EOF V1 opcodes.
- The code does not end with a valid EOF V1 terminating opcode (such as `Op.STOP` or `Op.REVERT` or `Op.RETURN`).

In the case where the code wrapping fails, `evm_code_type` can be added as a parameter to the test and the bytecode can be dynamically modified to be compatible with the EOF V1 container.

### `@pytest.mark.with_all_call_opcodes` marker

This marker is used to automatically parameterize a test with all EVM call opcodes that are valid for the fork being tested.

```python
import pytest

@pytest.mark.with_all_call_opcodes
@pytest.mark.valid_from("Frontier")
def test_something_with_all_call_opcodes(pre: Alloc, call_opcode: Op):
    ...
```

In this example, the test will be parametrized for parameter `call_opcode` with values `[Op.CALL, Op.CALLCODE]` starting on fork Frontier, `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL]` on fork Homestead, `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL]` on fork Byzantium, and eventually it will be parametrized with with values `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL, Op.EXTCALL, Op.EXTSTATICCALL, Op.EXTDELEGATECALL]` on the EOF activation fork.

Parameter `evm_code_type` will also be parametrized with the correct EVM code type for the opcode under test.

### Code Generators `Switch` and `Conditional`

Code generators `Switch` and `Conditional` have been updated to take the `evm_code_type` and correctly generate EOF V1 code in case of `EVMCodeType.EOF_V1`.

### `--evm-code-type` parameter for `fill`

Parameter `--evm-code-type` to force all tests to be filled with an specific EVM code type and, while most tests would currently fail because of their incompatibility with EOF, it can serve the purpose of exploring which tests could be easily updated for EOF.

### `call_return_code` helper

This function returns the correct return code depending on the give opcode, whether a success is expected or not, and whether a revert is expected or not.

It is useful to automatically set the correct storage expectations depending on the type of call used, since all legacy `*CALL` opcodes and the EOF `EXT*CALL` opcodes return different values.

### Fork Covariant Processor Changes

Changes were required to `src/pytest_plugins/forks/forks.py` to allow markers to parametrize two or more parameters at the same time, depending on the fork. It was necessary because the new marker `with_all_call_opcodes` parametrizes `call_opcode` and `evm_code_type` at the same time.

### Example tests updated

The following tests were updated as an example of how an updated test would look like:
- tests/cancun/eip4844_blobs/test_point_evaluation_precompile.py
- tests/frontier/opcodes/test_dup.py

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/docs/writing_tests/test_markers.md
+++ b/docs/writing_tests/test_markers.md
@@ -53,7 +53,7 @@ import pytest
 
 @pytest.mark.with_all_tx_types
 @pytest.mark.valid_from("Berlin")
-def test_something_with_all_tx_types(tx_type):
+def test_something_with_all_tx_types(tx_type: int):
     pass
 ```
 
@@ -74,11 +74,49 @@ import pytest
 
 @pytest.mark.with_all_precompiles
 @pytest.mark.valid_from("Shanghai")
-def test_something_with_all_precompiles(precompile):
+def test_something_with_all_precompiles(precompile: int):
     pass
 ```
 
 In this example, the test will be parameterized for parameter `precompile` with values `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]` for fork Shanghai, but with values `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]` for fork Cancun (because of EIP-4844).
+
+### pytest.mark.with_all_evm_code_types
+
+This marker is used to automatically parameterize a test with all EVM code types that are valid for the fork being tested.
+
+```python
+import pytest
+
+@pytest.mark.with_all_evm_code_types
+@pytest.mark.valid_from("Frontier")
+def test_something_with_all_evm_code_types(pre: Alloc):
+    pass
+```
+
+In this example, the test will be parameterized for parameter `evm_code_type` only with value `[EVMCodeType.LEGACY]` starting on fork Frontier, and eventually it will be parametrized with with values `[EVMCodeType.LEGACY, EVMCodeType.EOF_V1]` on the EOF activation fork.
+
+In all calls to `pre.deploy_contract`, if the code parameter is `Bytecode` type, and `evm_code_type==EVMCodeType.EOF_V1`, the bytecode will be automatically wrapped in an EOF V1 container.
+
+Code wrapping might fail in the following circumstances:
+
+- The code contains invalid EOF V1 opcodes.
+- The code does not end with a valid EOF V1 terminating opcode (such as `Op.STOP` or `Op.REVERT` or `Op.RETURN`).
+
+In the case where the code wrapping fails, `evm_code_type` can be added as a parameter to the test and the bytecode can be dynamically modified to be compatible with the EOF V1 container.
+
+```python
+import pytest
+
+@pytest.mark.with_all_evm_code_types
+@pytest.mark.valid_from("Frontier")
+def test_something_with_all_evm_code_types(pre: Alloc, evm_code_type: EVMCodeType):
+    code = Op.SSTORE(1, 1)
+    if evm_code_type == EVMCodeType.EOF_V1:
+        # Modify the bytecode to be compatible with EOF V1 container
+        code += Op.STOP
+    pre.deploy_contract(code)
+    ...
+```
 
 ## Other Markers
 
--- a/docs/writing_tests/test_markers.md
+++ b/docs/writing_tests/test_markers.md
@@ -118,6 +118,23 @@ def test_something_with_all_evm_code_types(pre: Alloc, evm_code_type: EVMCodeTyp
     ...
 ```
 
+### pytest.mark.with_all_call_opcodes
+
+This marker is used to automatically parameterize a test with all EVM call opcodes that are valid for the fork being tested.
+
+```python
+import pytest
+
+@pytest.mark.with_all_call_opcodes
+@pytest.mark.valid_from("Frontier")
+def test_something_with_all_call_opcodes(pre: Alloc, call_opcode: Op):
+    ...
+```
+
+In this example, the test will be parametrized for parameter `call_opcode` with values `[Op.CALL, Op.CALLCODE]` starting on fork Frontier, `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL]` on fork Homestead, `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL]` on fork Byzantium, and eventually it will be parametrized with with values `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL, Op.EXTCALL, Op.EXTSTATICCALL, Op.EXTDELEGATECALL]` on the EOF activation fork.
+
+Parameter `evm_code_type` will also be parametrized with the correct EVM code type for the opcode under test.
+
 ## Other Markers
 
 ### pytest.mark.slow
--- a/src/ethereum_test_vm/helpers.py
+++ b/src/ethereum_test_vm/helpers.py
@@ -15,6 +15,6 @@ def call_return_code(opcode: Op, success: bool, *, revert: bool = False) -> int:
         if success:
             return 0
         if revert:
-            return 2
-        return 1
+            return 1
+        return 2
     raise ValueError(f"Not a call opcode: {opcode}")
--- a/src/ethereum_test_vm/helpers.py
+++ b/src/ethereum_test_vm/helpers.py
@@ -1,0 +1,20 @@
+"""
+Helper functions for the EVM.
+"""
+
+from .opcode import Opcodes as Op
+
+
+def call_return_code(opcode: Op, success: bool, *, revert: bool = False) -> int:
+    """
+    Returns the return code for a CALL operation.
+    """
+    if opcode in [Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL]:
+        return int(success)
+    elif opcode in [Op.EXTCALL, Op.EXTDELEGATECALL, Op.EXTSTATICCALL]:
+        if success:
+            return 0
+        if revert:
+            return 2
+        return 1
+    raise ValueError(f"Not a call opcode: {opcode}")
--- a/src/pytest_plugins/forks/forks.py
+++ b/src/pytest_plugins/forks/forks.py
@@ -173,7 +173,7 @@ fork_covariant_descriptors = [
     CovariantDescriptor(
         marker_name="with_all_evm_code_types",
         description="marks a test to be parametrized for all EVM code types at parameter named"
-        " evm_code_type of type EVMCodeType",
+        " `evm_code_type` of type `EVMCodeType`, such as `LEGACY` and `EOF_V1`",
         fork_attribute_name="evm_code_types",
         parameter_name="evm_code_type",
     ),
--- a/src/pytest_plugins/forks/forks.py
+++ b/src/pytest_plugins/forks/forks.py
@@ -81,18 +81,35 @@ class ForkParametrizer:
         """
         Return the parameter names for the test case.
         """
-        return ["fork"] + [p.name for p in self.fork_covariant_parameters]
+        parameter_names = ["fork"]
+        for p in self.fork_covariant_parameters:
+            if "," in p.name:
+                parameter_names.extend(p.name.split(","))
+            else:
+                parameter_names.append(p.name)
+        return parameter_names
 
     def get_parameter_values(self) -> List[Any]:
         """
         Return the parameter values for the test case.
         """
-        return [
-            pytest.param(*params, marks=[self.mark] if self.mark else [])
+        param_value_combinations = [
+            params
             for params in itertools.product(
                 [self.fork],
                 *[p.values for p in self.fork_covariant_parameters],
             )
+        ]
+        for i in range(len(param_value_combinations)):
+            # if the parameter is a tuple, we need to flatten it
+            param_value_combinations[i] = list(
+                itertools.chain.from_iterable(
+                    [v] if not isinstance(v, tuple) else v for v in param_value_combinations[i]
+                )
+            )
+        return [
+            pytest.param(*params, marks=[self.mark] if self.mark else [])
+            for params in param_value_combinations
         ]
 
 
@@ -159,6 +176,13 @@ fork_covariant_descriptors = [
         " evm_code_type of type EVMCodeType",
         fork_attribute_name="evm_code_types",
         parameter_name="evm_code_type",
+    ),
+    CovariantDescriptor(
+        marker_name="with_all_call_opcodes",
+        description="marks a test to be parametrized for all *CALL opcodes at parameter named"
+        " call_opcode, and also the appropriate EVM code type at parameter named evm_code_type",
+        fork_attribute_name="call_opcodes",
+        parameter_name="call_opcode,evm_code_type",
     ),
 ]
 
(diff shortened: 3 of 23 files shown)


tests/frontier/opcodes/test_dup.py
```diff
@@ -59,9 +62,20 @@ def test_dup(
...
+        code=account_code
+        if evm_code_type == EVMCodeType.LEGACY
+        else Container(
+            sections=[
+                Section.Code(
+                    account_code,
+                    max_stack_height=19,
+                ),
+            ]
+        )
```
shemnon, line 77: To make this easier to use I think we need a standard method for this, like `maybe_eof(<code>, evm_code_type)` 

But it's not that simple as spit out raw code for legacy and for EOF wrap it in a container with  the code in section 0. 
 We also need to validate there are no forbidden opcodes (like jumps or calls ;) ) and calculate the stack height.

marioevz: This is where `pre.deploy_contract` from #584 will potentially come into play.

With this standardized way of contract deployment, the pre-allocation could try to convert the code into EOF automatically in many of the tests, and give the tester the option to implement the container manually if not possible. I.e. you can add the marker and then run the test, which should fail if the EOF container could not be produced, and then you need to manually implement the container.

This is one example for one of the updated EOF tests in that PR: https://github.com/ethereum/execution-spec-tests/pull/584/files#diff-e89eb58de375df2e5b5e0472312229ce973d61cddac4955c037bfe63ecccac48

shemnon: That will work great with yul honestly.  Just re-compile with the eof target when EOF lands in solidity.

Perhaps if we had configs for the types to test. Default value is fork dependent, (legacy pre Prague, both post), fails code generation if it cannot meet the flags), maybe a flag to error and the default is a warning.

src/pytest_plugins/forks/forks.py
```diff
@@ -153,6 +170,20 @@ def add_values(self, metafunc: Metafunc, fork_parametrizer: ForkParametrizer) ->
         fork_attribute_name="precompiles",
         parameter_name="precompile",
     ),
+    CovariantDescriptor(
+        marker_name="with_all_evm_code_types",
+        description="marks a test to be parametrized for all EVM code types at parameter named"
+        " evm_code_type of type EVMCodeType",
+        fork_attribute_name="evm_code_types",
+        parameter_name="evm_code_type",
+    ),
```
winsvega, line 179: Such as plain evm, eof

src/pytest_plugins/forks/forks.py
```diff
@@ -153,6 +170,20 @@ def add_values(self, metafunc: Metafunc, fork_parametrizer: ForkParametrizer) ->
...
+    CovariantDescriptor(
+        marker_name="with_all_call_opcodes",
+        description="marks a test to be parametrized for all *CALL opcodes at parameter named"
+        " call_opcode, and also the appropriate EVM code type at parameter named evm_code_type",
```
winsvega, line 183: So call_opcode will be automatically passed as an argument to the test function?
Thats a little implicit. Is there an example how it will look?

src/pytest_plugins/forks/forks.py
```diff
@@ -153,6 +170,20 @@ def add_values(self, metafunc: Metafunc, fork_parametrizer: ForkParametrizer) ->
...
+        fork_attribute_name="call_opcodes",
+        parameter_name="call_opcode,evm_code_type",
```
winsvega, line 185: Also if I already have parametrisation with this name how the collision resolve?

docs/writing_tests/test_markers.md
```diff
@@ -74,12 +74,66 @@ import pytest
...
+```python
+import pytest
+
+@pytest.mark.with_all_call_opcodes
+@pytest.mark.valid_from("Frontier")
+def test_something_with_all_call_opcodes(pre: Alloc, call_opcode: Op):
+    ...
+```
+
+In this example, the test will be parametrized for parameter `call_opcode` with values `[Op.CALL, Op.CALLCODE]` starting on fork Frontier, `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL]` on fork Homestead, `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL]` on fork Byzantium, and eventually it will be parametrized with with values `[Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL, Op.EXTCALL, Op.EXTSTATICCALL, Op.EXTDELEGATECALL]` on the EOF activation fork.
```
winsvega, line 133: What about parameter numbers. It is automatically handled with the opcodes that skip value?

src/ethereum_test_vm/helpers.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    Returns the return code for a CALL operation.
+    """
+    if opcode in [Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL]:
+        return int(success)
+    elif opcode in [Op.EXTCALL, Op.EXTDELEGATECALL, Op.EXTSTATICCALL]:
+        if success:
+            return 0
+        if revert:
+            return 2
```
pdobacz, line 18: just scanning quickly, but noticed this: revert is code 1 not 2. 2 is for exceptional failure, i.e. oog or Op.INVALID

tests/frontier/opcodes/test_dup.py
marioevz: I added a `--evm-code-type` flag to `fill` and it's pretty interesting to see what tests contain code that is simply plug-and-play into EOF.

I'm currently analyzing the tests but if you pick up this branch and then do:
```bash
fill --fork=CancunEIP7692 --evm-code-type=eof_v1 --evm-bin=evmone-t8n ./tests/<some test>
```
you can start trying out tests from previous forks in EOF mode to see which ones we could mark as interesting to also fill in EOF containers.

So what we could do is:
- Use the `--evm-code-type` to manually find interesting pre-EOF tests that we should add the permanent markers to.
- Fix tests that are not EOF compatible but we would like to run them in EOF mode because they test interesting things
- Add either `@pytest.mark.with_all_evm_code_types` or `@pytest.mark.with_all_call_opcodes` to the appropriate tests.

marioevz: Tagging ipsilon for visibility @pdobacz @gumb0 @chfast

src/pytest_plugins/forks/forks.py
marioevz: Yes I added a test as an example to also give myself an idea of how easy would it be to add the marker to any test:
https://github.com/ethereum/execution-spec-tests/blob/a87bb09025d8472c9efacb67ccd839f003211736/tests/cancun/eip4844_blobs/test_point_evaluation_precompile.py#L488-L523

You can see the actual test fixture that consumes the `call_opcode` parameter to prepare the bytecode here:
https://github.com/ethereum/execution-spec-tests/blob/a87bb09025d8472c9efacb67ccd839f003211736/tests/cancun/eip4844_blobs/test_point_evaluation_precompile.py#L144-L173

And my general experience was that it is not so easy and careful thought has to be put when adding this marker.

Main pain points are that:
- The return code is not the same for EOF vs non-EOF call opcodes, but this is somewhat alleviated by this helper function here: https://github.com/ethereum/execution-spec-tests/blob/a87bb09025d8472c9efacb67ccd839f003211736/src/ethereum_test_vm/helpers.py#L8-L20
- The out-of-gas tests simply don't work if we expect to be able to pass an specific amount of gas to a subcall, because the EOF call opcodes don't support a gas parameter.

src/pytest_plugins/forks/forks.py
marioevz: You have to manually remove it before adding this marker to a test.

You can see an example of this with the 4844 pre-compile test, where it previously had this:
https://github.com/ethereum/execution-spec-tests/blob/28f376b5264133c1eefedbc5d8006ae557924a15/tests/cancun/eip4844_blobs/test_point_evaluation_precompile.py#L450-L458
And was replaced with:
https://github.com/ethereum/execution-spec-tests/blob/a87bb09025d8472c9efacb67ccd839f003211736/tests/cancun/eip4844_blobs/test_point_evaluation_precompile.py#L497

Also all instances of `call_type` had to be renamed to what the marker actually parametrizes: `call_opcode`.

docs/writing_tests/test_markers.md
marioevz: We need to update the test to use opcode keyword-arguments for this to not be a problem.

This test used an `if` branch to handle different argument count between call opcodes with and without value:
https://github.com/ethereum/execution-spec-tests/blob/28f376b5264133c1eefedbc5d8006ae557924a15/tests/cancun/eip4844_blobs/test_point_evaluation_precompile.py#L111-L136
But using keyword arguments this is simplified:
https://github.com/ethereum/execution-spec-tests/blob/a87bb09025d8472c9efacb67ccd839f003211736/tests/cancun/eip4844_blobs/test_point_evaluation_precompile.py#L150-L160

src/ethereum_test_vm/helpers.py
marioevz: Must have been too tired yesterday since I swear I read the opposite from the EIP :sweat_smile:

tests/frontier/opcodes/test_dup.py
pdobacz: I have been following along more or less. This is great stuff, I'll give it a spin right away, thank you!

pdobacz: I've quickly experimented with one test (MCOPY) here https://github.com/ethereum/execution-spec-tests/pull/726. I think it works OK. I'm not yet sure myself if it is the right direction, but posting here to gather thoughts and feedback. Let me know what you think @marioevz. I'll also experiment more with this tmr.

Overall (being conscious that the MCOPY change is actually quite involved) I think the marker works great! (I haven't done much of the `call_opcode` one yet, this one is more quirky)

(I also tried the 7702 tests, but that got stuck on 7702 being still in flux, so I left it for after 7702 stabilizes and implementations are more ready)

Also this gets me wondering: any remote chance we have sth similar for `ethereum/tests`? :grimacing:

marioevz: > (I also tried the 7702 tests, but that got stuck on 7702 being still in flux, so I left it for after 7702 stabilizes and implementations are more ready)

Maybe this one is due to evmone not supporting 7702 yet? It might be possible to do this with besu, since Prague has EOF activated.

> Also this gets me wondering: any remote chance we have sth similar for `ethereum/tests`? 😬

I think it would be great, but it would be extremely time consuming IMO. It might be easier to just start porting tests over. Dimitry has already been doing this for a bit now.

marioevz: I still think it would yield almost no usable tests because I have the feeling most of them have either jumps or use invalid EOF opcodes.

One way of trying out this theory could be to use `evm_bytes_to_python` to load each account contract in every `pre` of the filled fixtures from `ethereum/tests`, and see what percentage can be put into `Container.Code` and results in valid EOF code.

docs/writing_tests/test_markers.md
```diff
@@ -74,12 +74,67 @@ import pytest
...
+```
+
+In this example, the test will be parameterized for parameter `evm_code_type` only with value `[EVMCodeType.LEGACY]` starting on fork Frontier, and eventually it will be parametrized with with values `[EVMCodeType.LEGACY, EVMCodeType.EOF_V1]` on the EOF activation fork.
+
+In all calls to `pre.deploy_contract`, if the code parameter is `Bytecode` type, and `evm_code_type==EVMCodeType.EOF_V1`, the bytecode will be automatically wrapped in an EOF V1 container.
+
+Code wrapping might fail in the following circumstances:
+
+- The code contains invalid EOF V1 opcodes.
+- The code does not end with a valid EOF V1 terminating opcode (such as `Op.STOP` or `Op.REVERT` or `Op.RETURN`).
```
spencer-tb, line 103: > The code does not end with a valid EOF V1 terminating opcode (such as `Op.STOP` or `Op.REVERT` or `Op.RETURN`).

Could we make this automatic? I.e if a the code doesn't en with one of these we simply add an `Op.STOP`. Or will this be specific to the test?

spencer-tb (commented):
Love this PR. LGTM! 

My only real comment relates to a potential enstension to the `with_all_...` markers.

Instead of having multiple test markers, i.e:
- `pytest.mark.with_all_evm_code_types`
- `pytest.mark.with_all_call_opcodes`
- `pytest.mark.with_all_tx_types`
- `pytest.mark.with_all_precompiles`

Should we instead create a single "with all" marker:
- `pytest.mark.with_all(...)`
That accepts a list of potential input/s:
- `pytest.mark.with_all("tx_types", "evm_code_types")

marioevz:
> Should we instead create a single "with all" marker:
> 
>     * `pytest.mark.with_all(...)`
> 
> 
> That accepts a list of potential input/s:
> 
>     * `pytest.mark.with_all("tx_types", "evm_code_types")`

This is a really nice idea, although it would require some refactoring to the covariant fork parameter logic, so I think this could be its own PR. Also would need to make some modifications because, as it is right now, `with_all_evm_code_types` and `with_all_call_opcodes` are mutually exclusive.

docs/writing_tests/test_markers.md
marioevz: Check it out: https://github.com/ethereum/execution-spec-tests/pull/610/commits/765324acaab305cf08bcf314881daefe0342eb79

Added a `terminating` property to certain opcodes that propagates into the constructed bytecode, and then on https://github.com/ethereum/execution-spec-tests/pull/610/commits/515e63a527c794b39420bb7c5b0907e8021a577b we check for it to be true, otherwise we add a `Op.STOP`.

tests/frontier/opcodes/test_dup.py
pdobacz: > I still think it would yield almost no usable tests because I have the feeling most of them have either jumps or use invalid EOF opcodes.

Ok, that indeed is problematic (unless we treat those as usable tests for EOF validation - however those would be rather trivial cases of undefined opcode). I'll make a self note to do the experiment you describe. Thanks for the input!

docs/writing_tests/test_markers.md
spencer-tb: Perfect! :D

spencer-tb (approved):
LGTM!!

PR #614, opened 2024-06-10
new(tests): EIP-7069: Different RETURNDATACOPY oob

pdobacz:
## 🗒️ Description

Simple test case covering the updated out-of-bounds behavior of RETURNDATACOPY accross legacy and EOF contracts.

I'm not sure whether to place this new test in the existing `7069_extcall` subdir or in its own, please advise.

## 🔗 Related Issues

Related update to the spec https://github.com/ethereum/EIPs/pull/8617. Update to evmone https://github.com/ethereum/evmone/pull/909

## ✅ Checklist

### DIFF
--- a/tests/prague/eip7692_eof_v1/eip7069_extcall/helpers.py
+++ b/tests/prague/eip7692_eof_v1/eip7069_extcall/helpers.py
@@ -7,7 +7,16 @@ import itertools
 _slot = itertools.count()
 next(_slot)  # don't use slot 0
 slot_code_worked = next(_slot)
+slot_eof_target_call_status = next(_slot)
+slot_legacy_target_call_status = next(_slot)
+slot_eof_target_returndata = next(_slot)
+slot_eof_target_returndatasize = next(_slot)
+slot_legacy_target_returndatasize = next(_slot)
+
 slot_last_slot = next(_slot)
+
+"""Storage value indicating an abort"""
+value_exceptional_abort_canary = 0x1984
 
 """Storage values for common testing fields"""
 value_code_worked = 0x2015
--- a/tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
+++ b/tests/prague/eip7692_eof_v1/eip7069_extcall/test_address_space_extension.py
@@ -10,6 +10,7 @@ from ethereum_test_tools.eof.v1 import Container, Section
 from ethereum_test_tools.vm.opcode import Opcodes as Op
 
 from .. import EOF_FORK_NAME
+from .helpers import value_exceptional_abort_canary
 from .spec import CALL_SUCCESS, EXTCALL_REVERT, EXTCALL_SUCCESS
 
 REFERENCE_SPEC_GIT_PATH = "EIPS/eip-7069.md"
@@ -21,8 +22,6 @@ _slot = itertools.count(1)
 slot_top_level_call_status = next(_slot)
 slot_target_call_status = next(_slot)
 slot_target_returndata = next(_slot)
-
-value_exceptional_abort_canary = 0x1984
 
 
 @pytest.mark.parametrize(
--- a/tests/prague/eip7692_eof_v1/eip7069_extcall/test_returndataload.py
+++ b/tests/prague/eip7692_eof_v1/eip7069_extcall/test_returndataload.py
@@ -7,12 +7,23 @@ from typing import List
 import pytest
 
 from ethereum_test_tools import Account, Alloc, Environment, StateTestFiller, Transaction
+from ethereum_test_tools.common.types import Storage
 from ethereum_test_tools.eof.v1 import Container, Section
 from ethereum_test_tools.vm.opcode import Opcodes as Op
 
 from .. import EOF_FORK_NAME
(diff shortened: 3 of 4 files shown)


pdobacz:
I admit https://github.com/ethereum/execution-spec-tests/pull/595 went completely under my radar. But this PR here adds a test which I intended to test a specific edge case - that the RETURNDATACOPY  OOB behavior is correct regardless of the caller frame being legacy or eof, that is:
- eof calls legacy, which does OOB RETURNDATACOPY
- eof calls eof, which does OOB RETURNDATACOPY
- legacy calls legacy, which does OOB RETURNDATACOPY
- legacy calls eof, which does OOB RETURNDATACOPY

In other words, that the correctness of OOB RETURNDATACOPY execution doesn't depend on the outside frame being eof/legacy.

@marioevz @shemnon let me know how you see incorporating this edge case (maybe I'm missing that #595 does cover all 4 of these? I think it doesn't b/c entry point is always legacy)

marioevz:
> I think it doesn't b/c entry point is always legacy

It's true but then the legacy code makes a call to a contract that can be either EOF or legacy, and that is the one that contains the `RETURNDATACOPY` part under test, then it does the oob testing when parameter `size` is greater than `len(return_data)`.

pdobacz:
> > I think it doesn't b/c entry point is always legacy
> 
> It's true but then the legacy code makes a call to a contract that can be either EOF or legacy, and that is the one that contains the `RETURNDATACOPY` part under test, then it does the oob testing when parameter `size` is greater than `len(return_data)`.

Correct, but I specifically want to check if part under test is independent of the "outside", which is the caller context, i.e. RETURNDATACOPY logic correctly changes on the eof-legacy boundary, both ways. This is so far our only opcode where EOF and legacy behavior differs, so I want to give it a little extra attention. WDYT?

marioevz:
> > > I think it doesn't b/c entry point is always legacy
> > 
> > 
> > It's true but then the legacy code makes a call to a contract that can be either EOF or legacy, and that is the one that contains the `RETURNDATACOPY` part under test, then it does the oob testing when parameter `size` is greater than `len(return_data)`.
> 
> Correct, but I specifically want to check if part under test is independent of the "outside", which is the caller context, i.e. RETURNDATACOPY logic correctly changes on the eof-legacy boundary, both ways. This is so far our only opcode where EOF and legacy behavior differs, so I want to give it a little extra attention. WDYT?

Makes sense, seems to me though that we could add another `pytest.mark.parametrize` to https://github.com/ethereum/execution-spec-tests/blob/eee1625f21ec7f2bf78850e5e49192fc118ef2cf/tests/prague/eip7692_eof_v1/eip7069_extcall/test_returndataload.py#L74, where we modify the entry-point address calling opcode here: https://github.com/ethereum/execution-spec-tests/blob/eee1625f21ec7f2bf78850e5e49192fc118ef2cf/tests/prague/eip7692_eof_v1/eip7069_extcall/test_returndataload.py#L103 to make it either `EXT*CALL` or `*CALL`, which would make it EOF or legacy.

Although, it would create a big amount of tests, but it might be necessary.

shemnon:
> In other words, that the correctness of OOB RETURNDATACOPY execution doesn't depend on the outside frame being eof/legacy.

I'm not understanding how any rational implementation would have different handling of returndatacopy if the called contract was EOF or legacy.  It's a byte array in the frame context, not caring nor knowing how it was generated. This brings us into the realm of testing hypothetical implementation strategies.

marioevz:
> yea, I was wondering that this is better, once we have `test_returndataload.py` now, but the large number of tests was why I considered a separate one. What would be your preference (hypothetically, assuming we do want to have it)?

Yes I think having a separate function, even if it has much of the code in `test_returndatacopy_handling`, would be better since that function already produces 448 tests.

shemnon:
I can look into trimming the returndata test to representative values, the cartesian join gets big quick.

tests/prague/eip7692_eof_v1/eip7069_returndatacopy/test_returndatacopy_execution.py
```diff
@@ -0,0 +1,141 @@
...
+)
+code_legacy_oob_returndatacopy = Op.RETURNDATACOPY(0, 0, 1) + Op.RETURN(0, 1)
+
+
+@pytest.mark.parametrize("top_level", ("LegacyContract", "EOFContract"), ids=lambda x: x)
+def test_returndatacopy_oob(
+    state_test: StateTestFiller,
+    top_level: str,
+):
+    """Test RETURNDATACOPY correct oob behavior"""
```
marioevz, line 61: Could we expand a bit on this docstring?

I think it would be useful to explain a bit more the reason of this test as well as the reasoning behind each parameter.

pdobacz:
Now I'm also worried about the complexity of the resulting test function, if we combine them, not only # of generated tests. Let me make the first step of moving the new test into the existing file and aligning conventions, and then we can see if we like it or not.

pdobacz:
I reworked the test, adding it under the existing test file. I think it is better not to overload the existing function, it takes a long time to fill already, and that would double it.

shemnon (approved):

PR #636, opened 2024-06-19
new(tests): EIP-7610: init-collision tests

marioevz:
## 🗒️ Description
Adds an init-collision test where the target contract creation address already contains a non-zero storage value.

## 🔗 Related Issues
None

## ✅ Checklist


marioevz:
@winsvega we're getting a failure here on the coverage diff, maybe I included the path in an incorrect format?

winsvega:
The log it shows is from outdated version of the script. Is the branch rebased?

tests/paris/eip7610_create_collision/test_initcollision.py
```diff
@@ -0,0 +1,212 @@
...
+    initcode = Initcode(
+        deploy_code=deploy_code,
+        initcode_prefix=Op.SSTORE(0, 1) + Op.SSTORE(1, 0),
+    )
+
+    salt = 0x0
+    salt_param = [salt] if opcode == Op.CREATE2 else []
+    assert len(initcode) <= 32
+    contract_creator_code = (
+        Op.MSTORE(0, Op.PUSH32(bytes(initcode).ljust(32, b"\0")))
```
winsvega, line 104: ah so this is how we can do too instead of using transaction data? 
but shall not forget the assert

tests/paris/eip7610_create_collision/test_initcollision.py
```diff
@@ -0,0 +1,212 @@
...
+
+    deploy_code = Op.STOP
+
+    initcode = Initcode(
+        deploy_code=deploy_code,
+        initcode_prefix=Op.SSTORE(0, 1) + Op.SSTORE(1, 0),
+    )
+
+    salt = 0x0
+    salt_param = [salt] if opcode == Op.CREATE2 else []
```
winsvega, line 101: isn't it optional with default = 0 anyway ?

tests/paris/eip7610_create_collision/test_initcollision.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    env = Environment()
+
+    sender = pre.fund_eoa()
+
+    deploy_code = Op.STOP
+
+    initcode = Initcode(
+        deploy_code=deploy_code,
+        initcode_prefix=Op.SSTORE(0, 1) + Op.SSTORE(1, 0),
```
winsvega, line 97: in original test the existing contract has storage 1:1 
in init code it change to 1:0  then to 1:1  
the idea I see was to check storage refund logic in init code then collision happens. but it shall be reverted because of collision. then more refunds happens

winsvega (changes requested):
needs rebase to run on latest coverage script
I think it's too early to claim that original .json tests are converted as those had 
storage refund logic and reverts.

tests/paris/eip7610_create_collision/test_initcollision.py
```diff
@@ -0,0 +1,130 @@
...
+                storage={0x01: 0x01},
+            ),
+        },
+        tx=tx,
+    )
+
+
+@pytest.mark.parametrize("opcode", [Op.CREATE, Op.CREATE2])
+@pytest.mark.pre_alloc_modify  # We need to modify the pre-alloc to include the collision
+@pytest.mark.valid_from("Paris")
```
LouisTsai-Csie, line 76: I am not very familiar with porting legacy test cases now, so i am curious should we add the `ported_from` marker like it does in `test_init_collision_create_tx`?

tests/paris/eip7610_create_collision/test_initcollision.py
```diff
@@ -0,0 +1,130 @@
...
+        sender=pre.fund_eoa(),
+        to=contract_creator_address,
+        data=initcode,
+        gas_limit=2_000_000,
+    )
+
+    pre[created_contract_address] = Account(
+        storage={0x01: 0x01},
+        nonce=1,
+    )
```
LouisTsai-Csie, line 119: I noticed that two legacy tests are being deleted: `InitCollisionNonZeroNonceFiller.json` and `InitCollisionParisFiller.json`. The only difference between them is the account nonce. In the latter, the pre-allocated accounts have non-zero storage but a zero nonce, which I believe is testing this scenario:

> Before [EIP-161](https://eips.ethereum.org/EIPS/eip-161) was applied, a newly deployed contract’s nonce remained at zero. As a result, it was possible to create a contract with a zero nonce and zero code length but with non-empty storage, if slots were set in the constructor. There are currently 28 such contracts on Ethereum mainnet.

Consider this, I wonder if we should parametrize the nonce to cover both cases here. However, it may be unnecessary since this situation cannot happen after EIP-161.

LouisTsai-Csie (changes requested):
Thanks for the case. I’ve added some comments, they are more like questions than suggested changes. It’s fine not to apply the changes if they aren’t necessary.

Regarding the deleted legacy test, I noticed that two cases have not been ported:
- Collision with `REVERT`: the `initcode` contains a revert statement, but the contract creation itself leads to a collision.
- Collision with `OOG`: the `initcode` execution would run out of gas, but the contract creation itself also leads to a collision.

I’m not sure whether we want to skip these cases. In my personal view, we can’t really test both the `REVERT` and `OOG` behaviors here, since the collision should stop `initcode` execution immediately. That makes it impossible to execute the operation that leads to the `REVERT` or `OOG` outcomes.

If we still want to include these cases for consistency, we could consider parametrizing `code_suffix` for the `initcode`, for example:
- `Op.REVERT(0, 0)` for the `REVERT` case
- `Op.INVALID` for the `OOG` case

and then update `test_init_collision_create_opcode`:

```python
initcode = Initcode(
    deploy_code=Op.STOP,
    initcode_prefix=Op.SSTORE(0, 1) + Op.SSTORE(1, 0) + code_suffix,
)
```

In both cases, it still not breaks the invariants statement below: `assert len(initcode) <= 32`

I don’t have a strong opinion on this, if we aren’t missing coverage, we could ignore the changes.

Note: i review the CI script but i do not know what would be affected, it would be good if someone else could take a look

tests/paris/eip7610_create_collision/test_initcollision.py
marioevz: This was because EELS does not support empty accounts anymore and fails with the following error:
```
ethereum.exceptions.StateWithEmptyAccount: Empty account at 0x6295ee1b4f6dd65047762f924ecd367c17eabf8f.
```
What we could add though is having cases for non-empty code and non-zero balance, with zero nonce in both cases.

PR #642, opened 2024-06-20
new(tests): EIP-7251: Increase the MAX_EFFECTIVE_BALANCE (EL Consolidations) 

marioevz:
## 🗒️ Description
Implements tests for EIP-7251

## 🔗 Related Issues
#606

## ✅ Checklist


src/ethereum_test_tools/common/conversions.py
```diff
@@ -59,6 +59,8 @@ def to_fixed_size_bytes(input: FixedSizeBytesConvertible, size: int) -> bytes:
     Converts multiple types into fixed-size bytes.
     """
     if isinstance(input, int):
+        if input < 0:
+            return int.to_bytes(input, length=size, byteorder="big", signed=True)
         return int.to_bytes(input, length=size, byteorder="big")
```
spencer-tb, line 64:
```suggestion
        return int.to_bytes(input, length=size, byteorder="big", signed=input < 0)
```

Small hack for 3 lines -> 1 line?

tests/prague/eip7251_consolidations/helpers.py
```diff
@@ -0,0 +1,292 @@
...
+    """
+    Balance of the account that sends the transaction.
+    """
+    sender_account: EOA | None = None
+    """
+    Account that will send the transaction.
+    """
+    requests: List[ConsolidationRequest]
+    """
+    Withdrawal request to be included in the block.
```
spencer-tb, line 81:
```suggestion
    Consolidation requests to be included in the block.
```

tests/prague/eip7251_consolidations/spec.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    MAX_CONSOLIDATION_REQUESTS_PER_BLOCK = (
+        1  # Maximum number of consolidation requests that can be de-queued into a block
+    )
+    TARGET_CONSOLIDATION_REQUESTS_PER_BLOCK = 1
+    MIN_CONSOLIDATION_REQUEST_FEE = 1
+    CONSOLIDATION_REQUEST_FEE_UPDATE_FRACTION = 17
+    EXCESS_INHIBITOR = 1181
+
+    MAX_AMOUNT = 2**64 - 1
```
spencer-tb, line 52: I'm not sure if we use this anywhere? Maybe copied over from 7002


```suggestion
```

spencer-tb (commented):
Tests are looking good from my side. Helped my understanding of the EIP. These align really nicely with the 6110, 7002 & 7685 tests. 

Some additional coverage ideas:
- Should we add cases where the source and target public keys are in an invalid format? Assuming t8n rejects filling these cases then I assume we'd want to go the "overwriting tx rlp" route. Maybe we could do this for the other request eips.
- I'm not certain what would happen when we have a duplicate consolidation request. My assumption is that both would be valid on the EL side where the CL would reject one of the requests when adding it to the queue. Nonetheless we could add a duplicate consolidation request test.
- Maybe a test where the target and source keys are the same.
- These might exist here already but could we do recursive consolidation requests? My guess is where we use the same target public key.

marioevz:
>     * These might exist here already but could we do recursive consolidation requests? My guess is where we use the same target public key.

I don't understand how could they be recursive since the contracts cannot do calls themselves to other contracts, do you mean on a loop?

marioevz:
Added a modification that ensures the test keeps producing blocks until the consolidation/withdrawal requests queues are exhausted, in order to properly validate that the invalid requests are not included in a future block. This is done to all test cases.

spencer-tb:
> > ```
> > * These might exist here already but could we do recursive consolidation requests? My guess is where we use the same target public key.
> > ```
> 
> I don't understand how could they be recursive since the contracts cannot do calls themselves to other contracts, do you mean on a loop?

Yeah you are right, they can't be recursive!

tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
```diff
@@ -0,0 +1,376 @@
...
+        pytest.param(
+            [
+                single_withdrawal_from_eoa(0),
+                single_deposit_from_eoa(0),
+            ],
+            [
+                single_withdrawal(0).with_source_address(TestAddress),
+                single_deposit(0),
+            ],
+            # TODO: on the Engine API, the issue should be detected as an invalid block hash
```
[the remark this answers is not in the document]
marioevz: Yeah, let's leave it to the PR that actually checks for the exceptions, I can fix it then since in that PR we have the infra ready to check for exceptions.

tests/prague/eip7251_consolidations/test_consolidations.py
```diff
@@ -0,0 +1,769 @@
...
+        pytest.param(
+            [
+                [
+                    ConsolidationRequestTransaction(
+                        requests=[
+                            ConsolidationRequest(
+                                source_pubkey=0x01,
+                                target_pubkey=0x02,
+                                fee=Spec.get_fee(0),
+                                # TODO: Value obtained from trace minus one
```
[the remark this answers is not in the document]
marioevz: Fixed in latest commit :raised_hands:

PR #648, opened 2024-06-24
feat(vm): Opcode keyword arguments

marioevz:
## 🗒️ Description
Introduce keyword arguments to some of the opcodes:
- `Op.CREATE`
- `Op.CREATE2`
- `Op.EOFCREATE`
- `Op.CALL`
- `Op.DELEGATECALL`
- `Op.STATICCALL`
- `Op.CALLCODE`
- `Op.EXTCALL`
- `Op.EXTSTATICCALL`
- `Op.EXTDELEGATECALL`

And most of the keyword arguments also have defaults, but they cannot be mixed with positional arguments, so this is valid:
```python
Op.CALL(address=0x1234)
```
but this is not:
```python
Op.CALL(Op.GAS, address=0x1234)
```

This feature allows easier parametrization, as we no longer have to workaround the cases where an opcode has different number of parameters or in different order:
```python
@pytest.mark.parametrize("op", [Op.CREATE, Op.CREATE2])
def test_example(op: Op):
    op(offset=0, size=1234)
```

And in this case the `salt` parameter of `CREATE2` is simply passed as zero (and also `value` for both options).

At the moment we allow extra keyword arguments (that do not belong to opcode called) but this is up for discussion.

## ✅ Checklist


spencer-tb (approved):
LGTM!

PR #652, opened 2024-06-26
feat(fw): Add `is_state_test` flag to `Environment`

marioevz:
## 🗒️ Description

Adds a flag `isStateTest` into the environment information that is passed to the t8n tool which is only set to `true` when executing the state tests, and should inhibit any block-level operations in order to reduce the changes to the state that are produced by:
- The mining reward
- Block Withdrawals
- System operations

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/common/types.py
+++ b/src/ethereum_test_tools/common/types.py
@@ -917,6 +917,8 @@ class Environment(EnvironmentGeneric[Number]):
     withdrawals: List[Withdrawal] | None = Field(None)
     extra_data: Bytes = Field(Bytes(b"\x00"), exclude=True)
 
+    is_state_test: bool = Field(False)
+
     @computed_field  # type: ignore[misc]
     @cached_property
     def parent_hash(self) -> Hash | None:
--- a/src/ethereum_test_tools/spec/state/state_test.py
+++ b/src/ethereum_test_tools/spec/state/state_test.py
@@ -115,6 +115,7 @@ class StateTest(BaseTest):
         fork = fork.fork_at(self.env.number, self.env.timestamp)
 
         env = self.env.set_fork_requirements(fork)
+        env.is_state_test = True
         tx = self.tx.with_signature_and_sender(keep_secret_key=True)
         pre_alloc = Alloc.merge(
             Alloc.model_validate(fork.pre_allocation()),


marioevz:
Requires: https://github.com/ethereum/go-ethereum/pull/30074

chfast:
I like the change however, there are some other quirks around "block processing" in `t8n`, e.g. you can pass `--state.reward=-1` to signal the block reward should not be applied.

How will this accept state tests themselves? They are currently expected to process system calls, withdrawals and apply the block reward of 0 :imp:.

winsvega (commented):
I remember I offered this as a flag to t8n a year ago, but it was rejected. 
--statetest

But even better:

--mode=statetest [default blockchain]

Or in env file 

```
transitionMode: "statetest"
```

IMHO its more of a cmd flag than env option

winsvega:
> I like the change however, there are some other quirks around "block processing" in `t8n`, e.g. you can pass `--state.reward=-1` to signal the block reward should not be applied.
> 
> How will this accept state tests themselves? They are currently expected to process system calls, withdrawals and apply the block reward of 0 👿.

only block reward of 0 as coinbase touch. no withdrawals in state tests. 
lets make a list of what should be disabled in a pure state test and check with the teams if they can easily support that in t8n.

marioevz:
> I like the change however, there are some other quirks around "block processing" in `t8n`, e.g. you can pass `--state.reward=-1` to signal the block reward should not be applied.
> 
> How will this accept state tests themselves? They are currently expected to process system calls, withdrawals and apply the block reward of 0 👿.

I think `--state.reward` < 0 should be deprecated eventually, I don't think it can be done immediately because retesteth still sends this value for state tests, but `is_state_test=True` should be equivalent to `--state.reward=-1` and also remove the block processing.

marioevz:
> ```
> transitionMode: "statetest"
> ```
I like this one, maybe `transitionMode` could be `transaction` or `block`, where `transaction` is equivalent to `is_state_test=True`.

Wdyt @chfast ?

spencer-tb:
> I like this one, maybe `transitionMode` could be `transaction` or `block`, where `transaction` is equivalent to `is_state_test=True`.


Just to chime in! I'm a fan of having the mode in the env, `transaction`/`block` sounds good to me as options.

Would `executionLevel` or `executionContext` make more sense than `transitionMode`. 

```
 --executionContext <type>  Specifies the execution context for the state transition:
                           transaction - Executes individual transactions (equivalent to is_state_test=True)
                           block       - Executes full blocks
```

winsvega:
Ah wait. In state tests reward 0.
Reward -1 is used to calculate state root of genesis  
I never send txs with reward -1 so I don't think there is state mode currently supported.

chfast:
I don't care so much about the specific name so you can pick the one you like the best.

I have small preference for the env JSON over command line flag. The invocation is already not commnad-line friendly because there are 3 mandatory JSON files to be used. From my point of view the `t8n` is the same as `statetest` just the input is provided differently and in `statetest` there is the "expected" section. However, the `statetest` could be easily used without the "expected" section as some EVM fuzzers already do.

PR #665, opened 2024-07-03
feat(fw): support invalid containers in EOFStateTest

chfast:
## 🗒️ Description

For invalid EOF containers we also generate EIP-7698 create transaction but with the expectation to fail.

## ✅ Checklist


marioevz:
I like this idea, but I think this deploy success/fail logic should be done in the base `eof_test` instead of `eof_state_test`, because that one is aimed for container validation, whereas the second one is aimed to container execution.

We could generate a state test from every `eof_test` in which the container under test is the first container section of an initcode container in a deploy tx, and then simply verify that the container is deployed or not depending on the `expected_exception`.

chfast:
I only want to fix the pytest skip warning.

PR #668, opened 2024-07-03
fix(tests): EOF - EIP-3540: Organize code_validation.py tests

marioevz:
## 🗒️ Description
Update tests contained in `tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py`, and delete the file.

Implement dynamic stack properties after the data portion of the opcode has been provided.

## 🔗 Related Issues
None

## ✅ Checklist


tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,271 @@ def test_all_opcodes_in_container(
...
+)
+def test_all_invalid_terminating_opcodes(
+    eof_test: EOFTestFiller,
+    opcode: Opcode,
+):
+    """
+    Test all opcodes that are invalid as the last opcode in a container
+    """
+    if opcode.has_data_portion():
+        opcode_with_data_portion = opcode[0] if opcode != Op.CALLF else opcode[1]
```
chfast, line 154: It took me 20 mins to figure out what it does.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,271 @@ def test_all_opcodes_in_container(
...
+    else:
+        opcode_with_data_portion = opcode
+
+    bytecode = (Op.ORIGIN * opcode_with_data_portion.min_stack_height) + opcode_with_data_portion
```
chfast, line 158: Can we use something less fancy than `ORIGIN`?

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,271 @@ def test_all_opcodes_in_container(
...
+@pytest.mark.parametrize(
+    "opcode",
+    [op for op in list(Op) if op in halting_opcodes],
+)
+def test_all_unreachable_terminating_opcodes_after_stop(
+    eof_test: EOFTestFiller,
+    opcode: Opcode,
+):
+    """
+    Test all opcodes terminating opcodes after stop.
```
chfast, line 195:
```suggestion
    Test all terminating opcodes after stop.
```

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,271 @@ def test_all_opcodes_in_container(
...
+    """
+    match opcode:
+        case Op.STOP:
+            sections = [Section.Code(code=Op.STOP + Op.STOP)]
+        case Op.RETF:
+            sections = [
+                Section.Code(code=Op.STOP + Op.CALLF[1]),
```
chfast, line 202: This will fail validation on the first code section.
```suggestion
                Section.Code(code=Op.CALLF[1]),
```

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,271 @@ def test_all_opcodes_in_container(
...
+                        sections=[
+                            Section.Code(code=Op.STOP + Op.RETURNCONTRACT[0](0, 0)),
+                            Section.Container(Container.Code(code=Op.STOP)),
+                        ]
+                    )
+                ),
+            ]
+        case Op.RETURN | Op.REVERT | Op.INVALID:
+            sections = [
+                Section.Code(code=Op.PUSH1(0) + Op.PUSH1(0) + Op.STOP + opcode),
```
chfast, line 224:
```suggestion
                Section.Code(code=Op.PUSH0 + Op.PUSH0 + Op.STOP + opcode),
```

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,271 @@ def test_all_opcodes_in_container(
...
+        case _:
+            raise NotImplementedError(f"Opcode {opcode} is not implemented")
+
+    eof_test(
+        data=Container(
+            sections=sections,
+        ),
+        expect_exception=EOFException.UNREACHABLE_INSTRUCTIONS
+        if opcode != Op.RETURNCONTRACT
+        else EOFException.INCOMPATIBLE_CONTAINER_KIND,
```
chfast, line 235: I'm not sure there should be a special case for the `RETURNCONTRACT`.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,271 @@ def test_all_opcodes_in_container(
...
+    eof_test: EOFTestFiller,
+    opcode: Opcode,
+    exception: EOFException,
+):
+    """
+    Test stack overflow on all opcodes that push more items than they pop
+    """
+    opcode_count = ceiling_division(
+        MAX_OPERAND_STACK_HEIGHT + 1 - opcode.min_stack_height,
+        opcode.pushed_stack_items - opcode.popped_stack_items,
```
chfast, line 377: This may be over-generalized. So far "pushing" instruction increase the stack height only by 1. Maybe use assert instead for future proofing?

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
marioevz: It seems to me like the subcontainer that should be initcode it's being detected as runtime container because the `STOP` is found first before the `RETURNCONTRACT`.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
marioevz: `opcode.pushed_stack_items` is guaranteed to be higher than `opcode.popped_stack_items` by the `pytest.mark.parametrize` condition, or do you mean some other assertion?

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
marioevz: Added a comment here, and in some other confusing parts of the file for clarity.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
chfast: ok, I get it. I think the `REVERT` instead of `STOP` may prevent it, but I'm not sure this qualifies as the same test.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
chfast:
```suggestion
        assert opcode.pushed_stack_items - opcode.popped_stack_items == 1
        opcode_count = MAX_OPERAND_STACK_HEIGHT + 1 - opcode.min_stack_height
```

src/ethereum_test_tools/vm/opcode.py
```diff
@@ -321,6 +321,7 @@ class Opcode(Bytecode):
 
     data_portion_length: int
     data_portion_formatter: Optional[Callable[[Any], bytes]]
+    stack_properties_modifier: Optional[Callable[[Any], tuple[int, int, int, int]]]
```
winsvega, line 324: This place can have more detailed comments with description

winsvega (commented):
Hm this makes my existed test too complicated. 

The parametrisation gets conditional
The test logic becomes hidden somewhere in sub preparators that have even more ifs

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,280 @@ def test_all_opcodes_in_container(
...
+    Test all opcodes that are invalid as the last opcode in a container
+    """
+    if opcode.has_data_portion():
+        # Add the appropriate data portion to the opcode by using the get_item method.
+        # On the CALLF opcode we need to reference the second code section, hence the [1] index.
+        opcode_with_data_portion = opcode[0] if opcode != Op.CALLF else opcode[1]
+    else:
+        opcode_with_data_portion = opcode
+
+    bytecode = (Op.PUSH0 * opcode_with_data_portion.min_stack_height) + opcode_with_data_portion
```
winsvega, line 160: name collision with the fixture

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -45,6 +46,7 @@
     Op.RETURN,
     Op.REVERT,
     Op.INVALID,
+    Op.RETF,
```
marioevz, line 49: This might be incorrect.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,280 @@ def test_all_opcodes_in_container(
...
+        data=Container(
+            sections=sections,
+        ),
+        expect_exception=EOFException.MISSING_STOP_OPCODE,
+    )
+
+
+@pytest.mark.parametrize(
+    "opcode",
+    [op for op in list(Op) if op in halting_opcodes],
```
winsvega, line 190: why not just op in list(halting_opcodes) ?

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -129,3 +131,280 @@ def test_all_opcodes_in_container(
...
+        ),
+        expect_exception=EOFException.UNREACHABLE_INSTRUCTIONS
+        if opcode != Op.RETURNCONTRACT
+        else EOFException.INCOMPATIBLE_CONTAINER_KIND,
+    )
+
+
+@pytest.mark.parametrize(
+    "opcode",
+    [op for op in list(Op) if op in halting_opcodes and op != Op.STOP],
```
winsvega, line 243: again for op in list(Op) does not make sense here. you anyway limit it by condition of haltin_opcode.

winsvega:
can remote from this file now:
https://github.com/ethereum/execution-spec-tests/blob/main/tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_example_valid_invalid.py

lines 88-100
```
pytest.param(
            # Sections with unreachable code fail
```

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
winsvega: the comment is incorrect lets remove it as it now used by other tests.
the meaning of the array if fine. this instructions stop code execution. retf  is like return

marioevz: But halting IMO should only be the ones that stop execution in the current contract context and returns it to the caller contract (or ends the tx). So with that in mind, JUMPF and RETF only change the current code section context but are not really halting the contract execution.

winsvega: ah, halting I meant stopping the code in the current context, not the global execution neceserraly

marioevz: Updated the test for better categorization.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
marioevz: Did a workaround because using a set (`halting_opcodes` in this case) resulted in pytest getting confused in the test cases during collection (because a set is not guaranteed to be iterated in any order). So we now use `sorted`.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
marioevz: Applied on latest version, thanks.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
marioevz: Applied a fix by using `sorted` on the latest version.

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -98,34 +100,285 @@ def sections(
...
+    Test all opcodes that are invalid as the last opcode in a container
     """
-    if opcode == Op.RETURNCONTRACT:
-        eof_code = Container(sections=sections, kind=ContainerKind.INITCODE)
+    if opcode.has_data_portion():
+        # Add the appropriate data portion to the opcode by using the get_item method.
+        # On the CALLF opcode we need to reference the second code section, hence the [1] index.
+        opcode = opcode[0] if opcode != Op.CALLF else opcode[1]
+
+    bytecode = (Op.PUSH0 * opcode.min_stack_height) + opcode
```
winsvega, line 142: This looks much cleaner than `opcode with data portion`

But will it conflict with line 146 ?

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
```diff
@@ -98,34 +100,285 @@ def sections(
...
+                        Section.Code(code=Op.PUSH0 * (opcode.min_stack_height - 1) + opcode[0]),
+                        Section.Container(Container.Code(code=Op.STOP)),
+                    ]
+                )
+            ),
+        ]
     else:
-        eof_code = Container(sections=sections)
+        bytecode = Op.PUSH0 * (opcode.min_stack_height - 1)
+        if opcode.has_data_portion():
```
winsvega, line 321: Oh, this is even better. Instead of modifying opcode it makes bytecode which construction is much cleaner

winsvega (approved):
Whoa the tests are getting complicated, huh?

tests/prague/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py
marioevz: I thought about this conflict for a while because after appending the data portion with the square brackets, if we compare it against the opcode without the data portion, they are not equal, so `Op.OPCODE[0x123] != Op.OPCODE`, which is a bit annoying.

We could do some fancy method like `Op.OPCODE.is_descendant(Op.OPCODE[0x123])` or similar, which could return true if the opcode is the same but with the data portion added, but I'm not convinced it's super necessary.

So that's why on 146 we compare against `Op.CALLF[1]` and not against `Op.CALLF`.

PR #676, opened 2024-07-09
new(tests): EOF - EIP-7620: Add more tests for validating EOF subcontainer kinds

gumb0:
## 🗒️ Description
Test cases that check subcontainer validation in a deeply nested subcontainer, and in code sections other than section 0.

## ✅ Checklist

### DIFF
--- a/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
+++ b/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
@@ -1,10 +1,12 @@
 """
 EOF Subcontainer tests covering simple cases.
 """
+
 import pytest
 
 from ethereum_test_tools import Account, EOFException, EOFStateTestFiller, EOFTestFiller
 from ethereum_test_tools.eof.v1 import Container, ContainerKind, Section
+from ethereum_test_tools.eof.v1.constants import MAX_BYTECODE_SIZE
 from ethereum_test_tools.vm.opcode import Opcodes as Op
 
 from .. import EOF_FORK_NAME
@@ -18,6 +20,9 @@ pytestmark = pytest.mark.valid_from(EOF_FORK_NAME)
 eofcreate_code_section = Section.Code(
     code=Op.EOFCREATE[0](0, 0, 0, 0) + Op.SSTORE(slot_code_worked, value_code_worked) + Op.STOP,
     max_stack_height=4,
+)
+eofcreate_revert_code_section = Section.Code(
+    code=Op.EOFCREATE[0](0, 0, 0, 0) + Op.REVERT(0, 0),
 )
 returncontract_code_section = Section.Code(
     code=Op.SSTORE(slot_code_worked, value_code_worked) + Op.RETURNCONTRACT[0](0, 0),
@@ -90,9 +95,11 @@ def test_reverting_container(
                 zero_section,
                 revert_sub_container,
             ],
-            kind=ContainerKind.INITCODE
-            if zero_section == returncontract_code_section
-            else ContainerKind.RUNTIME,
+            kind=(
+                ContainerKind.INITCODE
+                if zero_section == returncontract_code_section
+                else ContainerKind.RUNTIME
+            ),
         ),
         container_post=Account(storage={slot_code_worked: value_code_worked}),
     )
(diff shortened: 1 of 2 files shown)


tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
```diff
@@ -218,6 +222,101 @@ def test_container_combos_invalid(
...
+
+    container = invalid_container
+    while len(container) < 0x6000:
+        container = Container(
+            sections=[
+                eofcreate_code_section,
+                Section.Container(container=container.copy()),
+            ],
+            kind=ContainerKind.INITCODE,
+        )
```
marioevz, line 277:
```suggestion
        container = Container(
            sections=[
                eofcreate_code_section,
                Section.Container(
                    container=Container(
                        sections=[
                            Section.Code(code=Op.RETURNCONTRACT[0](0, 0)),
                            Section.Container(container=container.copy()),
                        ],
                    )
                ),
            ],
            kind=ContainerKind.INITCODE,
        )
```



I think this pattern is invalid in itself because all nested containers do EOFCREATE but only the deepest container does return contract.

IMO we should have a valid combination in the parametrize to prove that the pattern works if the deepest container is correct.

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
```diff
@@ -218,6 +222,101 @@ def test_container_combos_invalid(
...
+    code_section: Section,
+    first_sub_container: Container,
+    container_kind: ContainerKind,
+):
+    """Test invalid subcontainer reference / opcode combos in a non-first code section"""
+    eof_test(
+        data=Container(
+            sections=[Section.Code(Op.JUMPF[i]) for i in range(1, 1024)]
+            + [code_section, first_sub_container],
+            kind=container_kind,
```
marioevz, line 319: Similar comment to previous one, we should include a valid option in the parametrize to prove that the pattern only fails in case of an incorrect sub-container.

marioevz (changes requested):
Thanks! Some comments.

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
gumb0: It's not completely invalid, because EOFCREATE inside EOFCREATE initcode is allowed. And so I skipped RETURNCONTRACTs to make it somewhat simpler. However such initcode cannot end with STOP, so this was invalid for this reason, and I added now a variant that does REVERT after EOFCREATE, this is valid.

I've added positive cases too, and discovered an issue with how `container.copy()` works

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
```diff
@@ -218,6 +229,168 @@ def test_container_combos_invalid(
...
+            first_sub_container,
+        ],
+        kind=ContainerKind.INITCODE,
+    )
+
+    container = valid_container
+    while len(container) < MAX_BYTECODE_SIZE:
+        container_copy = container.copy()
+
+        # container.copy() sets all fields and sections' custom_size becomes 0
```
gumb0, line 286: So in the container produced by `copy()` all fields are present in `container.model_fields_set`, then bytecode is produced it thinks that `custom_size` is set to `0`, and puts 0 section size into header.

I tried to fix `copy()` by filtering out the fields not in source `model_fields_set`, but this didn't work so easy, because it's a hierarchy of objects, it needs to do the filtering when doing recursive `copy()` of nested `Section` objects...
@marioevz any other thoughts how to address this?

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
gumb0: Added.

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
gumb0: Fixed in #679

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
```diff
@@ -19,6 +21,12 @@
     code=Op.EOFCREATE[0](0, 0, 0, 0) + Op.SSTORE(slot_code_worked, value_code_worked) + Op.STOP,
     max_stack_height=4,
 )
+eofcreate_revert_code_section = Section.Code(
+    code=Op.EOFCREATE[0](0, 0, 0, 0)
+    + Op.SSTORE(slot_code_worked, value_code_worked)
```
marioevz, line 26:
```suggestion
```
I think this line is a tiny bit misleading, could be removed to save bytecode length IMO.

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
```diff
@@ -19,6 +21,12 @@
...
+    + Op.REVERT(0, 0),
+    max_stack_height=4,
```
marioevz, line 28:
```suggestion
```
We could also take this opportunity to remove this type of line from these containers.

marioevz (approved):
LGTM, two small suggestions, plus we should rebase before merging. Thanks!

marioevz:
Also, could we add this PR to the changelog in the same format as other EOF PRs? Thanks.

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
gumb0: True, this is not useful, will remove.

PR #677, opened 2024-07-10
feat(fw): Refactor `Bytecode`, create `MSTORE` macro

marioevz:
## 🗒️ Description
This PR implements the following changes:
- Refactors `Bytecode` class to its own file `vm/bytecode.py`
- Trims `FixedSizeBytes` arguments on opcode calls to remove zeros and use smaller `PUSH` operations
- Implements `MSTORE` macro, which takes an arbitrary length `bytes` object and automatically creates the bytecode to place it in the EVM execution memory.

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/tests/test_vm.py
+++ b/src/ethereum_test_tools/tests/test_vm.py
@@ -79,8 +79,7 @@ from ..vm.opcode import Opcodes as Op
         ),
         pytest.param(
             Op.CALL(Op.GAS, Address(0x1234), 0, 0, 0, 0, 32),
-            b"\x60\x20\x60\x00\x60\x00\x60\x00\x60\x00\x73\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
-            + b"\x00\x00\x00\x00\x00\x00\x00\x00\x12\x34\x5A\xF1",
+            b"\x60\x20\x60\x00\x60\x00\x60\x00\x60\x00\x61\x12\x34\x5A\xF1",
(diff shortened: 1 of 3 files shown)


src/ethereum_test_tools/tests/test_vm.py
```diff
@@ -287,6 +284,49 @@
...
+        pytest.param(
+            Om.MSTORE(bytes(range(64))),
+            bytes(Op.MSTORE(0, bytes(range(32))) + Op.MSTORE(32, bytes(range(32, 64)))),
+            id="Om.MSTORE(bytes(range(64)))",
+        ),
+        pytest.param(
+            Om.MSTORE(bytes(range(33))),
+            bytes(
+                Op.MSTORE(0, bytes(range(32)))
+                + Op.MLOAD(32)
```
winsvega, line 306: Why it so complicated here?

Shouldn't it be ?
```
mstore(0, bytes(range(32)))
mstore(32, 0x33...00)
```

winsvega (approved):
It looks cool, but starting to look like we overtooling the tools for test creation. Just look how much logic it requires.

But yes it comes handy in test development

src/ethereum_test_tools/tests/test_vm.py
marioevz: The reason is that it tries to preserve the memory outside of the range we are storing because we can only write 32-byte chunks, so we always assume there's already some information in the mstore range + 32 bytes, so we first read the existing memory range, mask it, then perform an OR with the new data, and finally store.

PR #680, opened 2024-07-11
feat(coverage): check patched tests coverage [TEST]

winsvega:
## 🗒️ Description
Enable coverage script on any test .py files changes:

if no new lines in converted-ethereum-tests.txt, then seek for .py files with tests and compare the coverage
from before and after PR, otherwise it compares the coverage from test repo and .py produced tests as before.

## ✅ Checklist


.github/workflows/coverage.yaml
```diff
@@ -149,13 +151,32 @@ jobs:
...
+          # Use a while loop with a here-string to avoid subshell issues
+          while IFS= read -r line; do
             file=$(echo "$line" | cut -c 3-)
-            fill $file --until=Cancun --evm-bin evmone-t8n || true >> filloutput.log 2>&1
-            (fill $file --fork=CancunEIP7692 --evm-bin evmone-t8n -k eof_test || true) > >(tee -a filloutput.log filloutputEOF.log) 2>&1
-          done
+            if grep -q "def test_" "$file"; then
+              FOUND_TEST=true
+              fill "$file" --until=Cancun --evm-bin evmone-t8n || true >> filloutput.log 2>&1
+              (fill "$file" --fork=CancunEIP7692 --evm-bin evmone-t8n -k eof_test || true) > >(tee -a filloutput.log filloutputEOF.log) 2>&1
```
marioevz, line 164: I think if we start adding hard-coded stuff here it will become very difficult to maintain this file.

We should declare an array of fill parameter combinations to use somewhere before this.

.github/workflows/coverage.yaml
```diff
@@ -149,13 +151,32 @@ jobs:
...
           mkdir -p fixtures/eof_tests
-          echo "$files" | while read line; do
+
+          # Use a while loop with a here-string to avoid subshell issues
+          while IFS= read -r line; do
             file=$(echo "$line" | cut -c 3-)
-            fill $file --until=Cancun --evm-bin evmone-t8n || true >> filloutput.log 2>&1
-            (fill $file --fork=CancunEIP7692 --evm-bin evmone-t8n -k eof_test || true) > >(tee -a filloutput.log filloutputEOF.log) 2>&1
-          done
+            if grep -q "def test_" "$file"; then
```
marioevz, line 161: This is kinda risky, pytest also loads `class Test*` as tests, but even after adding that option too, it seems fragile still.

I looked up for an alternative and there is this pytest plugin that fills only modified tests: https://github.com/anapaulagomes/pytest-picked

I tried it out with `fill` and it seems to work out of the box. We could add it to the setup (`pytest-picked==0.5.0` in the `install_requires` of `setup.cfg`), use `fill --picked` to fill the modified tests and then maybe checkout the base branch and then do `fill --picked --mode=branch --parent-branch=origin/PR-${{ github.event.pull_request.number }}` to fill the original tests.

marioevz (commented):
Some comments, I think the current approach is a bit fragile IMO, but I made a suggestion on how we could improve it.

.github/workflows/coverage.yaml
winsvega: If you do that it will be same as before.
Where I just pass each .py to fill and it fills if there are tests. 

This condition here is only to skip the script if there are no actual tests introduced by the pr. Because previously it would error that no tests generated.


 Which is also unlikely situation as a pr that touchs tests folder will most often touch the actual test files.

So in short I think this logic could be rolled back and removed if you think its fragile.

.github/workflows/coverage.yaml
winsvega: I only want to run retesteth coverage via t8n for some time to be sure that t8n and evmone test runners cause the same code coverage. (Because all the time, teams pointed out that t8n is not as reliable as native test runners)

.github/workflows/coverage.yaml
winsvega: I have disables test detection of def test_  this is also safer so we won't skip the coverage script. 
new tests over no tests ==> green coverage

winsvega: also fill --picked could is nice, but I am already building a file array to print it in the script for debug information. I could just iterate over it. I make this array global now

.github/workflows/coverage.yaml
winsvega: I simplified the test logic now:
- the tests are not detected by script, if no tests it just fails (but it should be a rare case when we commit into test folder without touching the tests)
- the files are detected once and we can clearly see what .py files are executed by the script

PR #683, opened 2024-07-11
fix(cli): `RJUMPV` in `evm_bytes_to_python`

marioevz:
## 🗒️ Description
Small fix to `evm_bytes_to_python` to properly parse `RJUMPV` data portion.

We could make this a bit cleaner in the future by providing a function in `class Opcode` that could be overridden by the opcodes to provide custom logic on data-portion parsing.

Also the output now contains the data portion as an integer instead of an hex-formatted string:
```
"0xe30001"
```
before:
```
Op.CALLF("0x0001")
```
now:
```
Op.CALLF[0x1]
```

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/cli/tests/test_evm_bytes_to_python.py
+++ b/src/cli/tests/test_evm_bytes_to_python.py
@@ -10,11 +10,11 @@ from ..evm_bytes_to_python import process_evm_bytes
 
 basic_vector = [
     "0x60008080808061AAAA612d5ff1600055",
-    'Op.PUSH1("0x00") + Op.DUP1 + Op.DUP1 + Op.DUP1 + Op.DUP1 + Op.PUSH2("0xaaaa") + Op.PUSH2("0x2d5f") + Op.CALL + Op.PUSH1("0x00") + Op.SSTORE',  # noqa: E501
+    "Op.PUSH1[0x0] + Op.DUP1 + Op.DUP1 + Op.DUP1 + Op.DUP1 + Op.PUSH2[0xaaaa] + Op.PUSH2[0x2d5f] + Op.CALL + Op.PUSH1[0x0] + Op.SSTORE",  # noqa: E501
 ]
 complex_vector = [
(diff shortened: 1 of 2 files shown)


src/cli/tests/test_evm_bytes_to_python.py
```diff
@@ -10,11 +10,11 @@
 
 basic_vector = [
     "0x60008080808061AAAA612d5ff1600055",
-    'Op.PUSH1("0x00") + Op.DUP1 + Op.DUP1 + Op.DUP1 + Op.DUP1 + Op.PUSH2("0xaaaa") + Op.PUSH2("0x2d5f") + Op.CALL + Op.PUSH1("0x00") + Op.SSTORE',  # noqa: E501
+    "Op.PUSH1[0x0] + Op.DUP1 + Op.DUP1 + Op.DUP1 + Op.DUP1 + Op.PUSH2[0xaaaa] + Op.PUSH2[0x2d5f] + Op.CALL + Op.PUSH1[0x0] + Op.SSTORE",  # noqa: E501
```
winsvega, line 13: The result of push2(1) is still 610001 right?

Also what happens if push1(257)?

And one more thing that 0x0123 in python is really confusing as it is integer, but in evm we think of 0x0123 as bytes or hash. And thus makes me want to write in python push2(0x0001)

winsvega (approved):
Should add more unittests for the opcodes with leading zeros and overflows.

src/cli/tests/test_evm_bytes_to_python.py
marioevz: > The result of push2(1) is still 610001 right?

Yes that is correct, I don't think we could remove that feature

> Also what happens if push1(257)?

It will fail, if the number overflows the data-portion length of any opcode, it will throw an error.

> And one more thing that 0x0123 in python is really confusing as it is integer, but in evm we think of 0x0123 as bytes or hash. And thus makes me want to write in python push2(0x0001)

If you write `0x0123` it will work, but any linter will automatically change it to remove the leading zeros.

PR #687, opened 2024-07-17
chore(refactor): blockchain test hive fixture format (include `params`)

marioevz:
## 🗒️ Description
Refactors the blockchain hive fixture format to match the execution-apis format (see https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md#request).

Namely, the format now directly contains the `params` field mentioned in the specification, and should be possible to simply relay this field to the execution client during testing.

These changes have been integrated to work with `consume engine`. Additionally all references to "blockchain hive" or "blockchain_tests_hive" etc has been changed to "blockchain_engine" or "blockchain_tests_engine". As in the future we will have many more hive related fixture formats.

## 🔗 Related Issues
None

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: A PR with removal of converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been opened.
- x ] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [x] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.

PR #692, opened 2024-07-18
refactor(plugins): Move pre-alloc logic to a separate filler plugin, rename test_filler

marioevz:
## 🗒️ Description
Moves the `deploy_contract`, `fund_eoa` and `fund_address` to a pytest-plugin in `src/pytest_plugins/test_filler/pre_alloc.py`.

This allows a more modular approach to setting pre-allocation functionality directly in the pytest runner that is currently executing the tests.

For example, in the current `test_filler` (only test runner implementation) it will automatically generate a new pre-alloc on every single test, but in the case of combining multiple tests into a single fixture, it should be possible to modify this `pre` fixture to have a different scope and different functionality.

Also it fixes typing in `src/pytest_plugins/test_filler/test_filler.py`.

## 🔗 Related Issues
None

## ✅ Checklist


marioevz:
@spencer-tb @danceratopz  finally renamed `test_filler.py` -> `filler.py` :wink:

pytest-framework.ini
```diff
@@ -7,7 +7,6 @@ testpaths =
     src
 addopts = 
     -p pytester
-    --ignore=src/pytest_plugins/test_filler/test_filler.py
     --ignore=src/pytest_plugins/consume/direct/test_direct.py
     --ignore=src/pytest_plugins/consume/hive_simulators/engine/test_via_engine.py
     --ignore=src/pytest_plugins/consume/hive_simulators/rlp/test_via_rlp.py
```
spencer-tb, line 12: I actually wonder if we can just remove these too:
```suggestion
```

As we don't have the `__init__.py` files. Let me double check.

spencer-tb: Hmm nope :(

We could also just ignore all of `consume`:
```
--ignore=src/pytest_plugins/consume/
```

I don't see us adding any framework tests to consume any time soon!

spencer-tb (approved):
LGTM! Docs look all good too

I rebased onto https://github.com/ethereum/execution-spec-tests/pull/687 locally to see if there are any conflicts but it none at all!

pytest-framework.ini
marioevz: I think it's fine as is, it's only three extra lines, we can change it once we implement tests for consume IMO.

PR #701, opened 2024-07-23
chore(consume): fix `--input` when running from hive

spencer-tb:
## 🗒️ Description

This PR allows us to run consume commands using the following Dockerfile entrypoint from hive:
```dockerfile
ENTRYPOINT ["consume", "engine", "--input=https://.../execution-spec-tests/releases/.../fixtures.tar.gz", "-v"]
```

### Problem
The condition `if not sys.stdin.isatty():` checks if the standard input (stdin) is connected to a terminal. If it's not connected to a terminal, it implies that the input might be coming from a pipe or a file, which is why the command adds --input=stdin. 

In a Docker container, `sys.stdin.isatty()` returns False because Docker doesn't provide a terminal interface by default. This causes an issue where the input is incorrectly automatically set to stdin when running in Docker.

### Solution
To fix the latter we additionally check for `--input` within the consume command arguements. We do not supply this when piping.

## 🔗 Related Issues
N/A

## ✅ Checklist

### DIFF
--- a/whitelist.txt
+++ b/whitelist.txt
@@ -333,6 +333,7 @@ StateTest
 StateTests
 StateTestFiller
 staticcalled
+stdin
 stExample
 str
 streetsidesoftware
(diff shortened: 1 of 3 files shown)


whitelist.txt
```diff
@@ -185,7 +185,7 @@ HexNumber
 hexsha
 hexbytes
 homebrew
-html
+ht l
```
marioevz, line 188: Is this a typo? :eyes:

marioevz (approved):
LGTM, just the potential typo in the whitelist file.

whitelist.txt
spencer-tb: whoops! this keeps happening :P

PR #713, opened 2024-07-25
new(tests): EOF - EIP-7069: Call Gas Testing for EXT*CALL

shemnon:
## 🗒️ Description

Test call gas for EXT*CALL, using a legacy harness.

Signed-off-by: Danno Ferrin <danno@numisight.com>

## ✅ Checklist


pdobacz (commented):
I was opposed at first, but this looks very practical. Any ideas how to resolve the complaints I had?
- philosophically against the grain of removing gas introspection (not so important)
- ignores EOF-only EVMs, should such exist
- ignores post-legacy-EVM-ban worlds
- OOG testing - I suppose with some complication of your harness this could be done: if the harness stored the setup/teardown gas usage, added cold/warm gas to it and subtracted 1, or sth along these lines? For a healthy test though, this should probably be generated as a separate `state_test`, so 1 `gas_test` => 2 `state_tests`, is that even possible?

Are you proposing this more as a long-term or short-term approach?

shemnon:
How do we assert the gas within the test and durring a fill? If all we want is observed gas then we can get the same effect by filling the desired behavior and having the fixtures pass/fail based on total consumed gas.

What I want is a test that will fail only because of gas calculations so an implementor can look at the test and know it's a gas problem.

pdobacz:
> What I had in mind to not require gas introspection is to use a single contract, pre-calculate the gas and use that as the gas limit value of the transaction (plus the 21,000).

The approach from [MCOPY mem expansion test](https://github.com/ethereum/execution-spec-tests/blob/main/tests/cancun/eip5656_mcopy/test_mcopy_memory_expansion.py) correct? I initially preferred that to using GAS, but this here has certain appeal

marioevz:
Yes I think this approach is really nice too, I think we could plan on doing a decorator around this, even if it uses `Op.GAS`, as long as that opcode is available we can use it.

I have a branch with a decorator for the `MCOPY` type of gas test, https://github.com/ethereum/execution-spec-tests/pull/719, we could use that decorator as a template to create another decorator for this type of gas tests.

pdobacz:
@marioevz are we maybe closer to a decision whether we'll adopt this or #719 moving forward (or both?). Also, I'm wondering which approach would lend itself better to the EOF marker? (for an example MCOPY gas test using the EOF marker, without the decoration, is in #754).

Another question still up is the OOG testing part (see previous comment) - is such an extension to the harness here feasible?

@shemnon Re:

> We need to detect and fail on both overcharging and undercharging. A test with a strict gas limit only gets us overcharging.

In #719 both strict gas limit and strict gas limit - 1 are tested, former expected to work, latter expected to OOG, so undercharging is also covered.

marioevz:
> Also, I'm wondering which approach would lend itself better to the EOF marker? (for an example MCOPY gas test using the EOF marker, without the decoration, is in #754).

I'm taking a look at `gas_test` in this PR and it lends itself pretty well to the EOF decorator without that many changes.

`pre.deploy_contract` has the `evm_code_type` keyword argument which can force a deployment of a contract to be in legacy mode, even when the test has been parametrized to run in EOF mode, so we can use this to force `address_legacy_harness` to always be a legacy contract, regardless of the context, and `address_baseline`/`address_subject` can be dynamically deployed as EOF of Legacy depending on the context.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -0,0 +1,129 @@
...
+                + Op.GAS
+                + Op.SWAP1
+                + Op.SUB
+            )
+            # Store warm gas
+            + (Op.DUP2 + Op.SWAP1 + Op.SUB + Op.PUSH2(slot_warm_gas) + Op.SSTORE)
+            # store cold gas
+            + (Op.SWAP1 + Op.SUB + Op.PUSH2(slot_cold_gas) + Op.SSTORE)
+            + Op.STOP
+        )
```
marioevz, line 84:
```suggestion
        ),
        evm_code_type=EVMCodeType.LEGACY,  # Needs to be legacy to use GAS opcode
```
This forces the `address_legacy_harness` to always be legacy type, even if the test is parametrized to deploy EOF code.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -0,0 +1,129 @@
...
+    env = Environment()
+
+    sender = pre.fund_eoa(10**18)
+
+    address_baseline = pre.deploy_contract(
+        Container(sections=[Section.Code(code=setup_code + tear_down_code)])
+    )
+    address_subject = pre.deploy_contract(
+        Container(sections=[Section.Code(code=setup_code + subject_code + tear_down_code)])
+    )
```
marioevz, line 47:
```suggestion
    address_baseline = pre.deploy_contract(setup_code + tear_down_code)
    address_subject = pre.deploy_contract(setup_code + subject_code + tear_down_code)
```
`@pytest.mark.parametrize("evm_code_type", [EVMCodeType.EOF_V1])` now should automatically convert these into EOF if requested.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -0,0 +1,129 @@
...
+            # store cold gas
+            + (Op.SWAP1 + Op.SUB + Op.PUSH2(slot_cold_gas) + Op.SSTORE)
+            + Op.STOP
+        )
+    )
+
+    post = {
+        address_legacy_harness: Account(
+            storage={
+                slot_warm_gas: cold_gas if warm_gas is None else warm_gas,
```
marioevz, line 90:
```suggestion
                slot_warm_gas: warm_gas,
```
`warm_gas` is never `None` due to line 36.

marioevz (commented):
Looks good! We could move `gas_test` to `src/ethereum_test_tools/code/generators.py` and export it to use it anywhere, although I'm thinking of any other cases (other than *CALL opcode testing) where the cold-warm dynamic would come into play.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -0,0 +1,129 @@
...
+
+def gas_test(
+    state_test: StateTestFiller,
+    pre: Alloc,
+    setup_code: Bytecode,
+    subject_code: Bytecode,
+    tear_down_code: Bytecode,
+    cold_gas: int,
+    warm_gas: int | None = None,
+):
```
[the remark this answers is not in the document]
shemnon: done

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -0,0 +1,129 @@
...
+
+@pytest.mark.parametrize(
+    ["opcode", "pre_setup", "cold_gas", "warm_gas"],
+    [
+        pytest.param(Op.EXTCALL, Op.PUSH0, 2600, 100, id="EXTCALL"),
+        pytest.param(Op.EXTCALL, Op.PUSH1(1), 2600 + 9000, 100 + 9000, id="EXTCALL_with_value"),
+        pytest.param(Op.EXTDELEGATECALL, Op.NOOP, 2600, 100, id="EXTSTATICCALL"),
+        pytest.param(Op.EXTSTATICCALL, Op.NOOP, 2600, 100, id="EXTDELEGATECALL"),
+    ],
+)
```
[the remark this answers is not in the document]
shemnon: Added

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
shemnon: Added

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
shemnon: updated

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
shemnon: Yep, it was from an earlier draft where I didn't need to resolve warm_gas early.  Fixed.

shemnon:
I'm going to punt on moving it.  Mostly because of the storage slot allocation needed to store data(`slot_warm_gas` and `slot_cold_gas` values), I'm not sure what we want to introduce to the framework as a whole vs what is just a pattern for EOF tests.

marioevz (approved):
LGTM, thanks!

marioevz:
Pushed a commit to make tox happy, I'll merge in a bit.

PR #717, opened 2024-07-26
feat(consume): add reproducible consume/hive commands

spencer-tb:
## 🗒️ Description
Adds the consume and hive commands to re-run a test case within hiveview. This is to aid test case reproduction for client devs incase of a failure.

Requires a small change to `app-suite.js` within hive: https://github.com/ethereum/hive/commit/cdbd33641dd2ba94133360535662e7309c34d83c#
You can test using this branch: https://github.com/spencer-tb/hive/tree/prague/devnet-1

Gives us the following on hiveview:
Screenshot 2024-07-27 at 18 52 03

## 🔗 Related Issues
N/A

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/consume/consume.py
+++ b/src/pytest_plugins/consume/consume.py
@@ -70,7 +70,7 @@ def download_and_extract(url: str, base_directory: Path) -> Path:
     with tarfile.open(archive_path, "r:gz") as tar:
         tar.extractall(path=extract_to)
 
-    return extract_to
+    return extract_to / "fixtures"
 
 
 def pytest_addoption(parser):  # noqa: D103
@@ -164,9 +164,9 @@ def pytest_configure(config):  # noqa: D103
     index_file = input_source / ".meta" / "index.json"
     if not index_file.exists():
         rich.print(f"Generating index file [bold cyan]{index_file}[/]...")
-    generate_fixtures_index(
-        input_source, quiet_mode=False, force_flag=False, disable_infer_format=False
-    )
+        generate_fixtures_index(
+            input_source, quiet_mode=False, force_flag=False, disable_infer_format=False
+        )
     config.test_cases = TestCases.from_index_file(index_file)
 
     if config.option.collectonly:
--- a/src/pytest_plugins/consume/consume.py
+++ b/src/pytest_plugins/consume/consume.py
@@ -128,6 +128,12 @@ def pytest_configure(config):  # noqa: D103
         config.test_cases = TestCases.from_stream(sys.stdin)
         return
 
+    if input_source == "latest":  # TODO: align this with release tags
+        input_source = (
+            "https://github.com/ethereum/execution-spec-tests/releases/latest/download/"
+            "fixtures_develop.tar.gz"
+        )
+
     if is_url(input_source):
         cached_downloads_directory.mkdir(parents=True, exist_ok=True)
         input_source = download_and_extract(input_source, cached_downloads_directory)
--- a/src/pytest_plugins/consume/consume.py
+++ b/src/pytest_plugins/consume/consume.py
@@ -88,6 +88,15 @@ def pytest_addoption(parser):  # noqa: D103
         ),
     )
     consume_group.addoption(
+        "--latest",
+        action="store_true",
+        dest="latest_source",
+        default=False,
+        help=(
+            "The latest EEST development JSON test fixtures. Cannot be used alongside `--input`."
+        ),
+    )
+    consume_group.addoption(
         "--fork",
         action="store",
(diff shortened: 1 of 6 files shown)


src/pytest_plugins/consume/consume.py
```diff
@@ -128,6 +128,12 @@ def pytest_configure(config):  # noqa: D103
         config.test_cases = TestCases.from_stream(sys.stdin)
         return
 
+    if input_source == "latest":  # TODO: align this with release tags
+        input_source = (
+            "https://github.com/ethereum/execution-spec-tests/releases/latest/download/"
+            "fixtures_develop.tar.gz"
+        )
```
spencer-tb, line 135: I was thinking we could add an `--input-release` flag that accepts existing tags, i.e develop, eipxxx.

Currently we have the single confict where we might want to use a folder called "latest/" but it would get overridden by the logic here and get the latest fixture release instead. Although right now I don't see this every causing an issue.

marioevz: I think it's simply cleaner to just be able to specify an `--input-release` flag. We could add it in this PR.

src/pytest_plugins/consume/hive_simulators/conftest.py
```diff
@@ -28,12 +28,54 @@ def eth_rpc(client: Client) -> EthRPC:
...
+) -> str:
+    """
+    Command to run the test within hive.
+    """
+    return (
+        f"./hive --sim ethereum/{test_suite_name} "
+        f"--client-file configs/develop.yaml "
+        f"--client {client_type.name} "
+        f'--sim.limit "{test_case.id}" '
+        f"--docker.output "
```
marioevz, line 45: I almost never use this one, is it necessary?

src/pytest_plugins/consume/hive_simulators/conftest.py
```diff
@@ -28,12 +28,54 @@ def eth_rpc(client: Client) -> EthRPC:
...
+    client_type: ClientType,
+    test_case: TestCaseIndexFile | TestCaseStream,
+) -> List[str]:
+    """
+    Commands to run the test within EEST using a hive dev back-end.
+    """
+    hive_dev = (
+        f"./hive --dev --client-file configs/develop.yaml "
+        f"--client {client_type.name} "
+        f"--docker.output "
```
marioevz, line 62: Same here.

src/pytest_plugins/consume/hive_simulators/conftest.py
```diff
@@ -42,6 +84,15 @@ def fixture_description(
         description += "\n\nNo description field provided in the fixture's 'info' section."
     else:
         description += f"\n\n{blockchain_fixture.info['description']}"
+    description += (
+        f"\n\nCommand to reproduce entirely in hive:" f"\n<code>{hive_consume_command}</code>"
+    )
+    description += (
+        f"\n\nCommands to reproduce within EEST using a hive dev back-end:"
+        f"\n1. <code>{eest_consume_commands[0]}</code>"
+        f"\n2. <code>{eest_consume_commands[1]}</code>"
```
marioevz, line 93: Use `sum` here instead of hard-coding indexes.

marioevz (commented):
Looks good overall, some comments.

src/pytest_plugins/consume/hive_simulators/conftest.py
spencer-tb: You mean `--docker.output` or the entire command?

I tend to use the former *almost* all the time, and check the logs via the terminal over instead of the `workspace/`.

src/pytest_plugins/consume/hive_simulators/conftest.py
spencer-tb: Sounds good!

src/pytest_plugins/consume/consume.py
spencer-tb: Sounds good! Will look into adding this over the next couple days!

spencer-tb: I added a `--latest` flag for now.

Specifying by tag specifically was too complex if we additionally account for all the `*.tar.gz`'s for full releases like v3.0.0. 

Latest always points to `fixtures_develop.tar.gz`.

spencer-tb: > I think it's simply cleaner to just be able to specify an `--input-release` flag. We could add it in this PR.

Now I've re-read were you meaning using `--input-release` for urls and `--input` for local fixtures?

danceratopz:
Nice addition!

PR #719, opened 2024-07-29
feat(fw): Gas test generator

marioevz:
## 🗒️ Description
Creates a gas test generator based off the `MCOPY` memory copy tests.

It can be applied to *any test and the test writer only needs to provide a couple of fixtures for it to work.

*It only works on legacy EVM code at the moment and not EOF, but I have a couple of ideas of how to make it work.

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_tools/generators/state.py
+++ b/src/ethereum_test_tools/generators/state.py
@@ -1,0 +1,163 @@
+"""
+Includes state test generators to use in Ethereum tests.
+"""
+from enum import Enum
+from typing import Any, Callable, List
+
+import pytest
+
+from ethereum_test_base_types import Account
+from ethereum_test_specs import StateTestFiller
+from ethereum_test_types import Alloc, Environment, Transaction, eip_2028_transaction_data_cost
+from ethereum_test_vm import Bytecode
+from ethereum_test_vm import Opcodes as Op
+
+Decorator = Callable[[Callable[..., Any]], Callable[..., Any]]
+
+
+class GasTestType(str, Enum):
+    """
+    Enum to list the types of gas tests that can be generated.
+    """
+
+    EXACT_GAS = "exact_gas"
+    OOG = "oog"
+
+
+def bytecode_gas_test(
+    *,
+    gas_test_types: List[GasTestType] | GasTestType = list(GasTestType),
+    with_data: bool = False,
+) -> Decorator:
+    """
+    Used to generate a parametrized test that checks that specified gas is
+    exactly the same as the gas used by the EVM.
+
+    Automatically generates an exact-gas test, where the transaction is sent
+    with the exact amount of gas required, and an out-of-gas test, where the
+    transaction is sent with one less gas than required.
+
+    Body of the decorated test is ignored.
+
+    Required fixtures:
+      pre: Alloc - Automatically provided by filler.
+      state_test: StateTestFiller - Automatically provided by filler.
+      exact_gas_cost: int - Exact gas cost required to execute the bytecode.
+      bytecode: Bytecode, the bytecode to be tested.
+
+    Optional fixtures:
+      data: bytes - if `with_data` is True, is the transaction calldata.
+
+    Args:
+      gas_test_types: List[GasTestType] | GasTestType = list(GasTestType) - List of gas test types
+        to generate.
+      with_data: bool = False - If True, a `data` fixture needs to be defined and will be
+        included.
+    """
+    if type(gas_test_types) is GasTestType:
+        gas_test_types = [gas_test_types]
+
+    assert type(gas_test_types) is list
+
+    def generator(
+        pre: Alloc,
+        state_test: StateTestFiller,
+        exact_gas_cost: int,
+        gas_test_type: GasTestType,
+        bytecode: Bytecode,
+        data: bytes = b"",
+        env: Environment = Environment(),
+    ):
+        """
+        Generate a test case that can be further parametrized to verify gas comsumption of a
+        specific EVM bytecode.
+        """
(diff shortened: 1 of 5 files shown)


pdobacz:
> *It only works on legacy EVM code at the moment and not EOF, but I have a couple of ideas of how to make it work.

Why? I tried and just wrapped `test_code` in a container and it seems to just have worked. It would need to "know" whether it should generate legacy or EOF or both (either directly or by combining with #610). Were there any other issues?

marioevz:
> > *It only works on legacy EVM code at the moment and not EOF, but I have a couple of ideas of how to make it work.
> 
> Why? I tried and just wrapped `test_code` in a container and it seems to just have worked. It would need to "know" whether it should generate legacy or EOF or both (either directly or by combining with #610). Were there any other issues?

No other issues, I meant working automatically, and yes I agree #610 should solve this.

marioevz:
@pdobacz @shemnon I rebased to main and renamed this generator to `exact_gas_test` so it differentiates from the one in #713.

I also added the EOF marker `@pytest.mark.with_all_evm_code_types` to the MCOPY test and it works out of the box in conjunction with this generator to produce legacy and EOF tests automatically.

pdobacz:
@shemnon @marioevz I'm totally on the fence with the #713 vs #719, ever so slightly in favor of this #719, just because it already has OOG testing and easier integration with the EOF marker and no GAS opcode dependency. Or is the plan to have both? Maybe this is the way, and we just write both kinds for each case to test, serving two slightly different purposes.

Would other people on the testing team have a preference here?

src/ethereum_test_tools/generators/state.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+        }
+        state_test(
+            env=env,
+            pre=pre,
+            post=post,
+            tx=tx,
+        )
+
+    if with_data:
```
spencer-tb, line 105: I think this still works if we move the `with_data` logic witih the decorator:
```python
def decorator(func: Callable[..., Any], with_data: bool) -> Callable[..., Any]:
    @pytest.mark.parametrize(
        "gas_test_type",
        gas_test_types,
        ids=lambda x: x.value,
    )
    def generated_gas_test(
        pre: Alloc,
        state_test: StateTestFiller,
        exact_gas_cost: int,
        gas_test_type: GasTestType,
        bytecode: Bytecode,
        data: bytes = None,
    ):
        if with_data:
            generator(
                pre=pre,
                state_test=state_test,
                exact_gas_cost=exact_gas_cost,
                gas_test_type=gas_test_type,
                bytecode=bytecode,
                data=data,
            )
        else:
            generator(
                pre=pre,
                state_test=state_test,
                exact_gas_cost=exact_gas_cost,
                gas_test_type=gas_test_type,
                bytecode=bytecode,
            )

    generated_gas_test.__name__ = func.__name__
    generated_gas_test.__doc__ = func.__doc__

    return generated_gas_test
```

src/ethereum_test_tools/generators/state.py
```diff
@@ -0,0 +1,163 @@
...
+    Args:
+      gas_test_types: List[GasTestType] | GasTestType = list(GasTestType) - List of gas test types
+        to generate.
+      with_data: bool = False - If True, a `data` fixture needs to be defined and will be
+        included.
+    """
+    if type(gas_test_types) is GasTestType:
+        gas_test_types = [gas_test_types]
+
+    assert type(gas_test_types) is list
```
spencer-tb, line 60: Should we assert all the required fixtures?

spencer-tb (commented):
Really like this idea!! There are so many ways we can extend this for other types of state test generators.

The refactor in the test is very clean! My only concern is that we are not being verbose enough with respect how we define the test, as we don't know where the required fixtures are going. Maybe decorating the test function like so:
```python
@exact_gas_test(
    state_test: StateTestFiller,
    ...
    tx: Transaction,
    gas_test_types=GasTestType.OOG,
    with_data=True,
)
def test_mcopy_huge_memory_expansion():
    pass
```
This could be a me problem though, and maybe its more obvious when writing the test.

danceratopz (commented):
This is so nice! This automatic test generation is going to be so valuable.

However, I tend to prefer a pytest plugin approach instead of a library decorator that returns `@pytest.mark.parametrize`:
1. I'm not fond of moving the `state_test` call out of the test function (it's now directly called in the library helper). Apart from being unexpected, it means that the implementer can't modify any of the objects going to the state test, `post`, for example (as it's set in the library function, it can't be extended by the implementer).
2. I'm also not keen on moving test parametrization to custom decorators that aren't exposed to pytest as we don't get a mark for these custom decorators (this means we can't filter by this type of test, for example). I.e., I'd prefer something like `@pytest.mark.generate_gas_test()`. I suspect we'll also run into problems applying xfail to subsets of these parametrized tests (but we could add a custom marks keyword as suggested for the `with_all_...` markers).

I created a rough POC for a pytest plugin, it doesn't modify `post` yet, but I think it's possible:
https://github.com/ethereum/execution-spec-tests/pull/760

tests/cancun/eip5656_mcopy/test_mcopy_memory_expansion.py
```diff
@@ -57,222 +49,81 @@ def subcall_exact_cost(
...
     Perform MCOPY operations that expand the memory by huge amounts, and verify that it correctly
     runs out of gas.
     """
-    state_test(
-        env=env,
-        pre=pre,
-        post=post,
-        tx=tx,
-    )
+    pass
```
winsvega, line 129: so the actual test is where?

PR #720, opened 2024-07-29
chore(consume): update consume index file exception

spencer-tb:
## 🗒️ Description
Updates consume for the case where the index file is not found. It no longer re-generates this.

To me it makes sense to enforce this if the filler is now the index generator.

Open to not including this.


marioevz:
I think this might be unnecessary, it's just a roadblock to run tests IMO.

spencer-tb:
I don't disagree. Will close this.

PR #726, opened 2024-08-01
new(tests): EIP-5656/7692 - use new marker to EOF-ize MCOPY test

pdobacz:
This has some rough edges and is a big rework, but I wanted to see how far along I can get this. In a nutshell this reworks the MCOPY memory expansion test to not require the CALL/GAS combination (relies on the tx gas cost instead, which is understandably uglier). 

The upside is that (I think) it works and generates the MCOPY mem expansion tests in the EOF world.

Feedback and thoughts welcome

### DIFF
--- a/tests/cancun/eip5656_mcopy/test_mcopy_memory_expansion.py
+++ b/tests/cancun/eip5656_mcopy/test_mcopy_memory_expansion.py
@@ -4,13 +4,14 @@ abstract: Tests [EIP-5656: MCOPY - Memory copying instruction](https://eips.ethe
     that produce a memory expansion, and potentially an out-of-gas error.
 
 """  # noqa: E501
-from typing import Mapping, Tuple
+from typing import Mapping
 
 import pytest
 
 from ethereum_test_tools import Account, Address, Alloc, Bytecode, Environment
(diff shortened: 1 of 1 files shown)


tests/cancun/eip5656_mcopy/test_mcopy_memory_expansion.py
```diff
@@ -40,14 +42,16 @@ def callee_bytecode(dest: int, src: int, length: int) -> Bytecode:
...
-def subcall_exact_cost(
+def call_exact_cost(
     initial_memory: bytes,
     dest: int,
     length: int,
 ) -> int:
     """
     Returns the exact cost of the subcall, based on the initial memory and the length of the copy.
     """
+    intrinsic_cost = 21000 + len(initial_memory) * 16
```
marioevz, line 53:
```suggestion
    intrinsic_cost = 21000 + eip_2028_transaction_data_cost(initial_memory)
```
This will allow the data to be also zeros without issues (it can be imported from `ethereum_test_tools`).

marioevz (approved):
Looks good to me, it works surprisingly good if you take into account that gas introspection is being removed.

We can merge once #610 is merged.

PR #730, opened 2024-08-06
new(tests): EOF - EIP-7069: Add tests, part 2.

pdobacz:
Mainly inspired by https://github.com/ipsilon/eof/pull/138 and evmone state_transition tests.

### DIFF
--- a/tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py
+++ b/tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py
@@ -32,6 +32,7 @@ slot_returndata = next(_slot)
 slot_returndatasize = next(_slot)
 slot_caller = next(_slot)
 slot_returndatasize_before_clear = next(_slot)
+slot_max_depth = next(_slot)
 slot_last_slot = next(_slot)
 
 """Storage values for common testing fields"""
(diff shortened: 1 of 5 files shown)


pdobacz:
I pushed two more tests in the same vein. I think this is all from the "easy" ones (not involving choosing the right "gas" approach) from the https://github.com/ipsilon/eof/pull/138 list, so I won't push more to this PR, ready for review

tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py
```diff
@@ -644,3 +645,275 @@ def test_eof_calls_clear_return_buffer(
...
+    state_test(
+        env=env,
+        pre=pre,
+        post=post,
+        tx=tx,
+    )
+
+
+# `no_oog_gas` is minimum amount of gas_limit which makes the transaction not go oog.
+no_oog_gas = 45728
```
marioevz, line 709: I suggest putting something like this inside the function:
```python
# `no_oog_gas` is minimum amount of gas_limit which makes the transaction not go oog.
    push_operations = 3 + len(opcode.kwargs)  # type: ignore
    no_oog_gas = (
        21_000
        + 20_000  # SSTORE
        + 2_100  # SSTORE COLD_SLOAD_COST
        + push_operations * 3  # PUSH operations
        + 100  # WARM_STORAGE_READ_COST
        + 2500  # COLD_ACCOUNT_ACCESS - WARM_STORAGE_READ_COST
        + 4  # RJUMPI
        + 3  # EQ
    )
```

tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py
```diff
@@ -644,3 +645,275 @@ def test_eof_calls_clear_return_buffer(
...
+    deep_most_result_block = (
+        Op.MSTORE(32, Op.ADD(Op.CALLDATALOAD(0), 1)) + Op.MSTORE(64, Op.NOOP) + Op.RETURN(32, 64)
+    )
+    rjump_offset = len(returndatacopy_block)
+
+    callee_code = Container.Code(
+        # current stack depth in memory bytes 0-31
+        Op.MSTORE(0, Op.ADD(Op.CALLDATALOAD(0), 1))
+        # pass it along deeper as calldata
+        + opcode(address=callee_address, args_size=32)
```
marioevz, line 874:
```suggestion
        + opcode(address=Op.ADDRESS, args_size=32)
```
We could use `Op.ADDRESS` here to avoid hard-coding the address into the pre-alloc.

The only caveat is that we need to use `Op.EXTCALL` in the `calling_contract_address` instead of `opcode` because the trick doesn't work when the top level call is initiated with `Op.EXTDELEGATECALL`, but I think the test still accomplishes its purpose with this change.

marioevz (commented):
Looks good overall, some comments. Thanks!

tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py
```diff
@@ -644,3 +645,275 @@ def test_eof_calls_clear_return_buffer(
...
+        Container.Code(
+            Op.SSTORE(slot_call_result, Op.EXTSTATICCALL(address=failing_contract_address))
+            + Op.SSTORE(slot_code_worked, value_code_worked)
+            + Op.STOP
+        )
+    )
+    tx = Transaction(
+        sender=sender,
+        to=Address(calling_contract_address),
+        gas_limit=50000000,
```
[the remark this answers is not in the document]
marioevz: Reduced the gas limit and tested and the result still passed with this amount.

marioevz:
@pdobacz opened a PR to your branch with most of my suggestions already implemented, because I was modifying the code locally to make sure it worked: https://github.com/ipsilon/execution-spec-tests/pull/2

Let me know if these look ok to you.

pdobacz:
> @pdobacz opened a PR to your branch with most of my suggestions already implemented, because I was modifying the code locally to make sure it worked: [ipsilon#2](https://github.com/ipsilon/execution-spec-tests/pull/2)
> 
> Let me know if these look ok to you.

Looks great, thank you. I accidentally made a merge commit, hope that's ok (you were probably going to squash merge anyway?)

marioevz (approved):
Thanks!

PR #734, opened 2024-08-08
bug(consume): genesis block hash exception

spencer-tb:
## 🗒️ Description
Fixes a bug in the consume (hive) genesis mismatch exception. For cases where the hash mismatch was present the error would overwrite the incorrect hash to the correct hash showing no mismatch within the exception. This occurred as we cast the returned genesis block from clients into the `FixtureHeader` mode, which always recalculates the block hash.

Additionally adds a further check to highlight if fields in the recieved client genesis block are unexpected. Highlighting that an incorrect fork could be enabled.

image

## 🔗 Related Issues
N/A

## ✅ Checklist


marioevz (approved):
LGTM, thanks!

Issue #555, opened 2024-05-15
bug(pytest): help flags are broken for fill and consume pytest commands


spencer-tb:
Update `fill --help` to not require the inclusion of an evm bin in the path or with the `--evm-bin` flag.

Error example below:
<img width="1222" alt="Screenshot 2024-05-15 at 15 11 06" src="https://github.com/ethereum/execution-spec-tests/assets/60348173/f2313ee2-4c86-44ed-8770-4495cb1b2b06">

spencer-tb:
Note `--pytest-help` doesn't work for `consume`:

```
consume --pytest-help
```

winsvega:
@spencer-tb 
Is it fixed now?

spencer-tb:
Ah yes! Thanks

PR #745, opened 2024-08-13
feat(consume): update `--input` & fix `--help`

spencer-tb:
## 🗒️ Description
~~Adds the `--stable` flag to consume which runs the lastest stable release fixtures for the default hiveview: 
https://hivetests.ethdevops.io/ where we don't care about running any development fixtures.~~

From @danceratopz's suggestion.
Removes `--develop` and adds the following keywords for release tarballs for `--input`:
- `latest-stable-release`
- `latest-develop-release`

Fixes the `--help` command for consume, where `consume rlp/engine/hive/direct --help` works as intended again.

Additionally fixes `--help` for fill when an evm bin is not attached. See related issues.

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/555

## ✅ Checklist


danceratopz (commented):
Appreciate the shortcuts, I think they'll be useful!

I think I'd prefer to avoid adding new cli options to achieve it though and use dedicated keywords instead to `--input` instead. Something along the lines of:

```python
def pytest_addoption(parser):  # noqa: D103
    consume_group = parser.getgroup(
        "consume", "Arguments related to consuming fixtures via a client"
    )
    consume_group.addoption(
        "--input",
        action="store",
        dest="fixture_source",
        default=default_input_directory(),
        help=(
            "Specify the JSON test fixtures source. Can be a URL, a local directory, "
            "or one of the special keywords: 'stable', 'develop'. "
            f"Default: '{default_input_directory()}'."
        ),
    )
```
We could consider using `latest-stable-release` or `latest-develop-release` as keywords?

Then we'd have:
```
consume rlp -v --input=latest-stable-release
```

Bit wordy, but has clear intent. What do you think?

danceratopz:
> I think I'd prefer to avoid adding new cli options to achieve it though and use dedicated keywords to --input instead.

We already use `stdin` as a keyword to `--input`.

spencer-tb:
> We could consider using latest-stable-release or latest-develop-release as keywords?
> Then we'd have:
> ```
> consume rlp -v --input=latest-stable-release
> ```
> Bit wordy, but has clear intent. What do you think?

Yeah this much sounds better! 

The main reason I wanted to add these flags initially was due to the hive commands on hiveview actually. To not take so much space with the URLs! (small reason I know)

1} I was hoping a while ago to be able to add the release tag and fixture "tag" as input options. 

So we could have `--input=<release_tag|latest>-<fixture_tag>-release`, for example:
- `--input=latest-stable-release` equivalent to `--input=v3.0.0-stable-release`.
- `--input=eip7692@v1.0.7-eip7692-release` for the EOF v1.0.7 fixtures.

At this point we are may be going overkill! But at least if we include the former it covers all cases regarding releases.

I do still like URLs especially if pointing to releases outside of the main repo but then...
2} we could add git user names to the input!

I'll add your suggestion here! And leave the latter for now but if you like 1) or 2) let me know and I can create a different PR in the future :)

danceratopz:
> The main reason I wanted to add these flags initially was due to the hive commands on hiveview actually. To not take so much space with the URLs! (small reason I know)

Ah ok, makes sense! I really like #717. I think this helps readability over specifying the full URL.

I also pondered on 1), but then I thought that deferring to the full URL in such cases was ok. I didn't get as far as 2) :slightly_smiling_face: Yeah, we can think on it and add it/them later.

spencer-tb:
@danceratopz - I added your suggestion and then got frustrated with `--help` for consume.
So I ended up fixing that too and removing some redundancies!

One pytest question I have for you: should `consume --pytest-help` work?

danceratopz:
> One pytest question I have for you: should `consume --pytest-help` work?

No, I don't think that's required :-) `consume <sub-command> --pytest-help` is already more than enough :slightly_smiling_face:

src/pytest_plugins/consume/hive_simulators/conftest.py
```diff
@@ -55,7 +71,10 @@ def eest_consume_commands(
     Commands to run the test within EEST using a hive dev back-end.
     """
     hive_dev = f"./hive --dev --client-file configs/develop.yaml --client {client_type.name}"
-    consume = f'consume {test_suite_name.split("-")[-1]} -v --latest -k "{test_case.id}"'
+    consume = (
+        f'consume {test_suite_name.split("-")[-1]} -v --input "latest-develop-release" -k '
```
danceratopz, line 75: This is used to display the consume command in order to reproduce the test locally, right? Here `"latest-develop-release"` will be ambiguous and could point to a different release than the one executed in this instance.

src/pytest_plugins/consume/consume.py
```diff
@@ -133,20 +119,17 @@ def pytest_configure(config):  # noqa: D103
...
         return
 
-    if latest_flag:
-        release_base_url = "https://github.com/ethereum/execution-spec-tests/releases"
-        input_source = f"{release_base_url}/latest/download/fixtures_develop.tar.gz"
+    latest_base_url = "https://github.com/ethereum/execution-spec-tests/releases/latest/download"
+    if input_source == "latest-stable-release":
+        input_source = f"{latest_base_url}/fixtures_stable.tar.gz"
+    if input_source == "latest-develop-release":
+        input_source = f"{latest_base_url}/fixtures_develop.tar.gz"
```
danceratopz, line 132: I think we should use the Github API to determine which release is the latest and then define `input_source` to the the URL that points to a specific release (i.e., a URL containing `v[0-9]+\.[0-9]+\.[0-9]+`). This has a few benefits:

1. The pytest header will display the correct release version (this is a must imho, so that you can quickly see which release was actually being executed).
2. We can display the exact release in the consume commands displayed in hiveview to help reproduce tests.
3. The extracted fixtures will be extracted to the correct cache directory.

Another approach would be to grab the version from the `.meta/fixtures.ini` file, but I thinkI prefer resolving `latest`. If we're being very careful, we could check that the resolved version matches that of `.meta/fixtures.ini`, but it's a bit paranoid.

danceratopz: Btw, there's an implementation of resolving `latest` on the command-line in [our docs](https://ethereum.github.io/execution-spec-tests/v3.0.0/consuming_tests/#obtaining-the-most-recent-release-artifacts) - but we should add a pure python implementation here instead of relying on `jq` (and potentially update the docs).

danceratopz (commented):
I really appreciate the clean-up, also to the `test_help` plugin and thanks for the tests!

I added a couple of commits, as discussed async:
1. Added `-h`: `consume -h`.
2. Fixed the output of `consume --help` to add back the one line descriptions for each consume sub-command.

I noticed that `consume all` disappeared and that piping from fill to consume is broken - I'll create a separate PR(s) for those :slightly_smiling_face:.

With regards to the latest release keywords, I think `consume` needs to know the version of the latest release it's consuming so that it can be reported correctly to the user (and cached locally to a better directory name). I added a couple of comments below, what do you think?

PR #748, opened 2024-08-15
feat(tests): test all opcodes by fork

winsvega:
## 🗒️ Description
define lists of opcodes by fork and verify that they are actually supported only at the allowed forks

## 🔗 Related Issues
PR https://github.com/ethereum/tests/pull/1399

## ✅ Checklist

### DIFF
--- a/tests/frontier/opcodes/test_all_opcodes.py
+++ b/tests/frontier/opcodes/test_all_opcodes.py
@@ -103,5 +103,4 @@ def test_cover_revert(state_test: StateTestFiller, pre: Alloc):
         protected=False,
     )
 
-    post = {}
-    state_test(env=Environment(), pre=pre, post=post, tx=tx)
+    state_test(env=Environment(), pre=pre, post={}, tx=tx)
--- a/tests/frontier/opcodes/test_all_opcodes.py
+++ b/tests/frontier/opcodes/test_all_opcodes.py
@@ -1,6 +1,6 @@
 """
-Call every possible opcode to see if it is recognized by the evm
-If opcode is undefined the subcall will fail
+Call every possible opcode and test that the subcall is successful
+if the opcode is supported by the fork supports and fails otherwise.
 """
 
 from typing import Dict
@@ -23,8 +23,6 @@ from ethereum_test_tools.vm.opcode import UndefinedOpcodes
 
 REFERENCE_SPEC_GIT_PATH = "N/A"
 REFERENCE_SPEC_VERSION = "N/A"
-all_opcodes = set(Op)
-undefined_opcodes = set(UndefinedOpcodes)
 
 
 def prepare_stack(opcode: Opcode) -> Bytecode:
@@ -49,11 +47,15 @@ def prepare_suffix(opcode: Opcode) -> Bytecode:
 
 @pytest.mark.valid_from("Frontier")
 def test_all_opcodes(state_test: StateTestFiller, pre: Alloc, fork: Fork):
-    """Check if the opcode supported on given fork"""
+    """
+    Test each possible opcode on the fork with a single contract that
+    calls each opcode in succession. Check that each subcall passes
+    if the opcode is supported and fails otherwise.
+    """
     code_worked = 1000
 
     code_contract: Dict[Opcode, Address] = {}
-    for opcode in sorted(all_opcodes | undefined_opcodes):
+    for opcode in sorted(set(Op) | set(UndefinedOpcodes)):
         code_contract[opcode] = pre.deploy_contract(
             balance=10,
             code=prepare_stack(opcode) + opcode + prepare_suffix(opcode),
@@ -61,7 +63,7 @@ def test_all_opcodes(state_test: StateTestFiller, pre: Alloc, fork: Fork):
         )
 
     # EVM code to make the call and store the result
-    caller = pre.deploy_contract(
+    contract_address = pre.deploy_contract(
         code=sum(
             Op.SSTORE(
                 Op.PUSH1(opcode.int()),
@@ -74,7 +76,7 @@ def test_all_opcodes(state_test: StateTestFiller, pre: Alloc, fork: Fork):
     )
 
     post = {
-        caller: Account(
+        contract_address: Account(
             storage={**{opcode.int(): 1 for opcode in fork.valid_opcodes()}, code_worked: 1}
         ),
     }
@@ -82,7 +84,7 @@ def test_all_opcodes(state_test: StateTestFiller, pre: Alloc, fork: Fork):
     tx = Transaction(
         sender=pre.fund_eoa(),
         gas_limit=500_000_000,
-        to=caller,
+        to=contract_address,
         data=b"",
         value=0,
         protected=False,
--- a/tests/frontier/opcodes/test_all_opcodes.py
+++ b/tests/frontier/opcodes/test_all_opcodes.py
@@ -1,0 +1,107 @@
+"""
+Call every possible opcode to see if it is recognized by the evm
+If opcode is undefined the subcall will fail
+"""
+
+from typing import Dict
+
+import pytest
+
+from ethereum_test_forks import Fork
+from ethereum_test_tools import (
+    Account,
+    Address,
+    Alloc,
+    Bytecode,
+    Environment,
+    StateTestFiller,
+    Transaction,
+)
+from ethereum_test_tools.vm.opcode import Opcode
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+from ethereum_test_tools.vm.opcode import UndefinedOpcodes
+
+REFERENCE_SPEC_GIT_PATH = "N/A"
+REFERENCE_SPEC_VERSION = "N/A"
+all_opcodes = set(Op)
+undefined_opcodes = set(UndefinedOpcodes)
+
+
+def prepare_stack(opcode: Opcode) -> Bytecode:
+    """Prepare valid stack for opcode"""
+    if opcode == Op.CREATE:
+        return Op.MSTORE(0, 0x6001600155) + Op.PUSH1(5) + Op.PUSH1(27) + Op.PUSH1(5)
+    if opcode == Op.CREATE2:
+        return Op.MSTORE(0, 0x6001600155) + Op.PUSH1(1) + Op.PUSH1(5) + Op.PUSH1(27) + Op.PUSH1(5)
+    if opcode == Op.JUMPI:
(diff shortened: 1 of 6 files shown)


tests/bugs/test_bytecode.py
```diff
@@ -0,0 +1,52 @@
+"""
+A bytecode sequence that caused bug, calling to undefined opcode with filled stack
```
marioevz, line 2: I think the name of the file is too generic, `test_undefined_opcode.py` might be a bit more clear.

tests/bugs/test_bytecode.py
```diff
@@ -0,0 +1,52 @@
...
+        code=Op.PUSH8(0xFFFFFFFFFFFFFFFF)
+        + Op.PUSH1(0x01)
+        + Op.PUSH1(0x00)
+        + Op.PUSH1(0x00)
+        + Opcode(0xFB),
+        storage={},
+    )
+
+    caller = pre.deploy_contract(
+        code=Op.SSTORE(0, Op.CALL(0, code_contract, 0, 0, 0, 0, 0)) + Op.SSTORE(code_worked, 1)
```
marioevz, line 36:
```suggestion
        code=Op.SSTORE(0, Op.CALL(address=code_contract)) + Op.SSTORE(code_worked, 1)
```

tests/bugs/test_bytecode.py
```diff
@@ -0,0 +1,52 @@
...
+        caller: Account(storage={0: 0, code_worked: 1}),
+    }
+
+    tx = Transaction(
+        sender=pre.fund_eoa(),
+        gas_limit=120000,
+        to=caller,
+        data=b"",
+        value=0,
+        protected=False,
```
marioevz, line 49: Is `protected` necessary in this case? It's been causing some trouble when executing these tests in live networks.

marioevz (commented):
I like the idea of having fork-agnostic tests, we might just need to define the correct folder name. Maybe `generic`? Wdyt @danceratopz @spencer-tb

tests/bugs/test_bytecode.py
winsvega: Oh cool, because there is actually a mistake here. Call passed 0 gas.

tests/bugs/test_bytecode.py
winsvega: We actually should have this one covered in undefined opcode tests. Where we fill up the stack and then call all possible opcodes. Verifying if there was evm exception or no. (Undefined opcodes cause exception) + the same must now be done in eof context.

spencer-tb:
> I like the idea of having fork-agnostic tests, we might just need to define the correct folder name. Maybe `generic`? Wdyt @danceratopz @spencer-tb

I like `generic`, but also `universal` and `crossfork`!

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -0,0 +1,170 @@
...
+from ethereum_test_tools.vm.opcode import Opcode
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "N/A"
+REFERENCE_SPEC_VERSION = "N/A"
+all_opcodes = set(Op)
+
+
+opcodes_by_fork: Dict[Fork, List[Opcode]] = {}
+opcodes_by_fork[Frontier] = [
```
winsvega, line 34: can I make this a global list that can be imported from ethereum_test_tools ?

winsvega:
I put def test_cover_revert(state_test: StateTestFiller, pre: Alloc):
there to prove that no coverage lost. but is it ok if the coverage is red, and it merged with explanation?

here the coverage is red because in original test there were some failing transactions due to invalid stack for opcodes, and create opcodes didn't work as intended, also transaction was running as creatte instead I use call contracts here.

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -0,0 +1,106 @@
...
+    Transaction,
+)
+from ethereum_test_tools.vm.opcode import Opcode
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+from ethereum_test_tools.vm.opcode import UndefinedOpcodes
+
+REFERENCE_SPEC_GIT_PATH = "N/A"
+REFERENCE_SPEC_VERSION = "N/A"
+all_opcodes = set(Op)
+undefined_opcodes = set(UndefinedOpcodes)
```
danceratopz, line 27: I'd avoid these global defs and just define the sets below (they're only used in one place).
```suggestion
```

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -0,0 +1,106 @@
...
+    return Op.STOP
+
+
+@pytest.mark.valid_from("Frontier")
+def test_all_opcodes(state_test: StateTestFiller, pre: Alloc, fork: Fork):
+    """Check if the opcode supported on given fork"""
+    code_worked = 1000
+
+    code_contract: Dict[Opcode, Address] = {}
+    for opcode in sorted(all_opcodes | undefined_opcodes):
```
danceratopz, line 56: To avoid the global defs above:
```suggestion
    for opcode in sorted(set(Op) | set(UndefinedOpcodes)):
```
Added a framework test to check that `set(Op) | set(UndefinedOpcodes)` does indeed create opcodes in the range 0,255: bfc84a525.

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -0,0 +1,106 @@
...
+        code_contract[opcode] = pre.deploy_contract(
+            balance=10,
+            code=prepare_stack(opcode) + opcode + prepare_suffix(opcode),
+            storage={},
+        )
+
+    # EVM code to make the call and store the result
+    caller = pre.deploy_contract(
```
danceratopz, line 64: I know you meant "opcode caller", but on first view `caller` sounded like an EOA to me. How about?
```suggestion
    contract_address = pre.deploy_contract(
```
But feel free to leave as is.

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -0,0 +1,106 @@
...
+                Op.CALL(1_000_000, opcode_address, 0, 0, 0, 0, 0),
+            )
+            for opcode, opcode_address in code_contract.items()
+        )
+        + Op.SSTORE(code_worked, 1)
+        + Op.STOP,
+    )
+
+    post = {
+        caller: Account(
```
danceratopz, line 77: If `caller` is changed above:
```suggestion
        contract_address: Account(
```

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -0,0 +1,106 @@
+"""
+Call every possible opcode to see if it is recognized by the evm
+If opcode is undefined the subcall will fail
```
danceratopz, line 3:
```suggestion
Call every possible opcode and test that the subcall is successful
if the opcode is supported by the fork supports and fails otherwise.
```

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -0,0 +1,106 @@
...
+def prepare_suffix(opcode: Opcode) -> Bytecode:
+    """Prepare after opcode instructions"""
+    if opcode == Op.JUMPI or opcode == Op.JUMP:
+        return Op.JUMPDEST
+    return Op.STOP
+
+
+@pytest.mark.valid_from("Frontier")
+def test_all_opcodes(state_test: StateTestFiller, pre: Alloc, fork: Fork):
+    """Check if the opcode supported on given fork"""
```
danceratopz, line 52:
```suggestion
    """
    Test each possible opcode on the fork with a single contract that 
    calls each opcode in succession. Check that each subcall passes
    if the opcode is supported and fails otherwise.
    """
```

danceratopz (approved):
LGTM - nice test!

Added a few minor suggestions below, mainly just style and clearer docstrings.

Also added a framwork test to check that we cover the full possible opcode range: https://github.com/ethereum/execution-spec-tests/pull/748/commits/bfc84a525620ebfa8cd2aca9274df9c8a9958a64

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -0,0 +1,106 @@
...
+    code_worked = 1000
+
+    code_contract: Dict[Opcode, Address] = {}
+    for opcode in sorted(all_opcodes | undefined_opcodes):
+        code_contract[opcode] = pre.deploy_contract(
```
marioevz, line 57: Might be getting ahead of myself, but this code does not handle EOF opcodes, and it's possible that we need more information inside of each opcode for this to be seamless.

We could add a new property to [`Bytecode` class](https://github.com/ethereum/execution-spec-tests/blob/14f9cbf32d3e69e9ca18d9f8c0ef80d526277f7b/src/ethereum_test_vm/bytecode.py#L9) like `supported_evm_code_types` which could contain a set of [`EVMCodeType`](https://github.com/ethereum/execution-spec-tests/blob/14f9cbf32d3e69e9ca18d9f8c0ef80d526277f7b/src/ethereum_test_vm/evm_types.py#L9C7-L9C18) and e.g. `Op.PUSH1.supported_evm_code_types == {EVMCodeType.LEGACY, EVMCodeType.EOF_V1}`, `Op.EXTCALL.supported_evm_code_types == {EVMCodeType.EOF_V1}` and `Op.CALL.supported_evm_code_types == {EVMCodeType.LEGACY}`.

Then in this case we read that property and deploy a different contract per each evm code type.

We could leave this for a follow-up PR because it seems out-of-scope in my opinion.

marioevz (approved):
One single comment, but I think it could be left for a follow up PR, and we could merge this one as is (after Dan's comments). Thanks!

tests/frontier/opcodes/test_all_opcodes.py
winsvega: All opcodes in eof already covered in eof tests.

This is frontier model test. And it also need to ensure that eof opcodes rejected in legacy code

danceratopz: I think this is would be a useful addition to `Opcodes`, regardless of whether we apply it in this test or not. Created a follow-up issue #770.

danceratopz (approved):
LGTM! Added a follow-up ticket for @marioevz's idea to extend the Opcode (or Bytecode) class with `supported_evm_code_types` #770.

PR #739, opened 2024-08-10
feat(fw): New `with_all` markers for precompiles, create opcode + other fixes

marioevz:
## 🗒️ Description

This PR implements features useful for EIP-7702 tests.

### `with_all_system_contracts` marker
Adds `with_all_system_contracts` marker and `system_contracts` method to forks, which helps parametrize tests with all contracts that affect the chain on a system level.

### `with_all_create_opcodes` marker
Adds `with_all_create_opcodes` marker and `create_opcodes` method to forks, to parametrize tests with all contract creating opcodes that a given fork supports, along with the appropriate `EVMCodeType` (Legacy or EOF V1).

### Lambda function parameter for `with_all_*` markers
Allow a lambda function filter to be passed as a single parameter to any `with_all_*` marker in order to filter-out parametrization values.

```python
import pytest

@pytest.mark.with_all_tx_types(lambda tx_type: tx_type != 3)
@pytest.mark.valid_from("London")
def test_something_with_all_tx_types(tx_type: int):
    pass
```

In this example, the test will be parametrized with all transaction types except for Type-3.

This is preferred to using `pytest.mark.parametrize` and selecting the values directly because in future forks new types will be added, and even when the marked test fails due to the new transaction not being immediately compatible, it will prompt a proper test update.

### `parametrize_with_defaults` decorator function

`pytest.mark.parametrize` replacement that allows to specify parameter names with default
    values for each parameter.

Example usage:
```python
@parametrize_with_defaults(
    parameter_1=0,  # Default value for parameter_1 is zero
    parameter_2='default',  # Default value for parameter_2 is 'default' string
    cases=[
        dict(parameter_1=1, id='test_1'),
        dict(parameter_2='custom', id='test_2'),
    ],
)
def test(parameter_1, parameter_2):
    ...
```

### Other Fixes
- Fix `compute_create_address` to not use the `sender.nonce` if `nonce` argument is provided.
- Implement EOF V1 `Container.Init` that returns a simple init container which deploys the specified `deployed_container`.
- Add `stop` boolean parameter to `CodeGasMeasure` in order to skip adding `Op.STOP` at the end of the returned bytecode.
- Export `Bytes` from `ethereum_test_tools`.
- Fix decimal values printing on `Storage.KeyValueMismatch`.
- Allow `storage` to be set on `pre.fund_eoa`

## 🔗 Related Issues
None

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
[...]

PR #754, opened 2024-08-19
new(tests): EIP-5656/7692 - use new marker to EOF-ize MCOPY test (2)

pdobacz:
This is https://github.com/ethereum/execution-spec-tests/pull/726 rebased after the EVM code type marker got merged to `main` and cleaned up. I also completed the set of EOF-MCOPY tests by applying the parameter in the remainder of existing 5656 tests.

I ran it through evmone intentionally breaking the MCOPY implementation in various ways, and the failures appear to be "in-sync" with ones which manifested in the old version of the test, as well as make sense in general.

### DIFF
--- a/tests/cancun/eip5656_mcopy/test_mcopy_contexts.py
+++ b/tests/cancun/eip5656_mcopy/test_mcopy_contexts.py
@@ -26,7 +26,7 @@ def initial_memory_length() -> int:  # noqa: D103
 @pytest.fixture
 def callee_bytecode(
     initial_memory_length: int,
-    opcode: Op,
+    call_opcode: Op,
 ) -> Bytecode:
     """
     Callee simply performs mcopy operations that should not have any effect on the
@@ -44,7 +44,7 @@ def callee_bytecode(
     bytecode += Op.MCOPY(0x00, initial_memory_length * 2, 1)
     bytecode += Op.MCOPY(initial_memory_length * 2, 0x00, 1)
 
-    if opcode != Op.STATICCALL:
+    if call_opcode != Op.STATICCALL and call_opcode != Op.EXTSTATICCALL:
         # Simple sstore to make sure we actually ran the code
         bytecode += Op.SSTORE(200_000, 1)
 
@@ -58,7 +58,7 @@ def callee_bytecode(
 def initial_memory(
     callee_bytecode: Bytecode,
     initial_memory_length: int,
-    opcode: Op,
+    call_opcode: Op,
 ) -> bytes:
     """
     Initial memory for the test.
@@ -67,7 +67,7 @@ def initial_memory(
 
     ret = bytes(list(islice(cycle(range(0x01, 0x100)), initial_memory_length)))
 
-    if opcode in [Op.CREATE, Op.CREATE2]:
+    if call_opcode in [Op.CREATE, Op.CREATE2]:
         # We also need to put the callee_bytecode as initcode in memory for create operations
         ret = bytes(callee_bytecode) + ret[len(callee_bytecode) :]
 
@@ -85,7 +85,7 @@ def caller_bytecode(
     initial_memory: bytes,
     callee_address: Address,
     callee_bytecode: Bytecode,
-    opcode: Op,
+    call_opcode: Op,
     caller_storage: Storage,
 ) -> Bytecode:
     """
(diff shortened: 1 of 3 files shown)


tests/cancun/eip5656_mcopy/test_mcopy_contexts.py
```diff
@@ -189,7 +175,7 @@ def test_no_memory_corruption_on_upper_call_stack_levels(
 
 
 @pytest.mark.parametrize(
-    "opcode",
+    "call_opcode",
```
pdobacz, line 178: this is clearly abusing the `call_opcode` parameter, is there pytest magic which could make it more elegant (combine `call_opcode` with `[CREATE, CREATE2]` to become `opcode`)?

marioevz: I think you need another marker for this (sorry about that), it's in an open PR (#739) and it's called `with_all_create_opcodes` so it parametrizes `create_opcode` with `[Op.CREATE, Op.CREATE2, Op.EOFCREATE]`.

marioevz: I also think it might not be as seamless to use this marker as is right now, it might need to require to move this test to another file to make use of different fixtures perhaps.

pdobacz: Yup, I have seen it coming, which is also why I didn't want to overthink this for now. Would you like me to do this moving/splitting in this PR or leave it for after when that other marker merges and "stabilizes"?

pdobacz:
@marioevz I have a question pertaining more to the evm code type marker in general, but following this MCOPY case as an example - how would such EOF-ized fixtures be organized and released? Right now they just land in the legacy MCOPY fixtures.

marioevz:
> @marioevz I have a question pertaining more to the evm code type marker in general, but following this MCOPY case as an example - how would such EOF-ized fixtures be organized and released? Right now they just land in the legacy MCOPY fixtures.

As of right now, they should go into the Cancun fork folder so `fixtures/state_tests/cancun/eip5656_mcopy/mcopy/mcopy_on_empty_memory.json` for example, but I get what you mean that now we have tests for EOF outside of `fixtures/state_tests/prague/eip7692_eof_v1/*`, and there is not much that we can do regarding the output file path, but what I've been thinking we could do is dump the test markers to the resulting fixture.

E.g. JSON fixture file `fixtures/state_tests/cancun/eip5656_mcopy/mcopy/mcopy_on_empty_memory.json` could contain the following fixtures:
```json
{
    "tests/cancun/eip5656_mcopy/test_mcopy.py::test_mcopy_on_empty_memory[fork_CancunEIP7692-evm_code_type_LEGACY-state_test-empty_memory-length_0-src_32-dest_32]": {
        ...
        "_info": {
            ...
            "markers": "evm_code_type=LEGACY",
            ...
        }
    },
    "tests/cancun/eip5656_mcopy/test_mcopy.py::test_mcopy_on_empty_memory[fork_CancunEIP7692-evm_code_type_EOF_V1-state_test-empty_memory-length_0-src_32-dest_32]": {
        ...
        "_info": {
            ...
            "markers": "evm_code_type=EOF_V1",
            ...
        }
    },
```
And this way test consumers can filter-out tests based on this information.

cc @danceratopz @spencer-tb @winsvega

winsvega:
Yes, another approach could be to separate

fixtures/state_tests/*
fixtures/state_tests_eof/*

So each test is filled in both modes. But stored to separate folders.

spencer-tb:
I think I like the folder approach better, until we have an `EOF_V2`!

marioevz:
@spencer-tb @winsvega But this is not a different fixture format, so it kinda breaks the consistency that each of the subfolders at the first level are all fixture formats, and I feel that it's not justified.

spencer-tb:
Ah I understand now, we now have state/blockchain/eof test fixture formats. Yeah I'd like to keep the fixture format consistency!

We could even just add it as a root field in the fixtures. So instead of placing it in the `_info` section we add another `evm_code` root field.

chfast:
The separate directory is not strictly needed but maybe good addition. However, you can dump everything into the same dir and wait if anyone complains...

pdobacz:
@marioevz any steps needed from me to move forward with this?

marioevz (approved):
Thanks!

marioevz:
I created an issue to discuss how to handle automatically-generated EOF tests from other EIPs here #767

PR #755, opened 2024-08-19
new(tests): EIP-7069 - EXTCALL with balance and other

pdobacz:
I noticed there was no EXTCALL test with a _successful_ value transfer, so I added one to complete the case from tracker.md


pdobacz:
quick update: I found some more low hanging fruit in this department, so I went ahead and pushed some more + a tracker.md update. Sorry about the noise

tests/prague/eip7692_eof_v1/tracker.md
```diff
@@ -375,48 +375,47 @@
...
-- [ ] OOG after EXTSTATICCALL (evmone-tests: state_tests/state_transition/eof_calls/extstaticcall_then_oog.json)
-- [ ] REVERT inside EXTCALL (evmone-tests: state_tests/state_transition/eof_calls/extcall_callee_revert.json)
-- [ ] REVERT inside EXTDELEGATECALL (evmone-tests: state_tests/state_transition/eof_calls/extdelegatecall_callee_revert.json)
-- [ ] REVERT inside EXTSTATICCALL (evmone-tests: state_tests/state_transition/eof_calls/extstaticcall_callee_revert.json)
-- [ ] EXTCALL with input (evmone-tests: state_tests/state_transition/eof_calls/extcall_input.json)
-- [ ] EXTDELEGATECALL with input (evmone-tests: state_tests/state_transition/eof_calls/extdelegatecall_input.json)
-- [ ] EXTSTATICCALL with input (evmone-tests: state_tests/state_transition/eof_calls/extstaticcall_input.json)
-- [ ] EXTCALL with just enough gas for MIN_RETAINED_GAS and MIN_CALLEE_GAS (evmone-tests: state_tests/state_transition/eof_calls/extcall_with_value_enough_gas.json)
-- [ ] EXTCALL with not enough gas for MIN_CALLEE_GAS (evmone-tests: state_tests/state_transition/eof_calls/extcall_with_value_enough_gasextcall_with_value_low_gas.json)
```
pdobacz, line 411: note the removed line here - I don't see such a test anymore in evmone tests

tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py
```diff
@@ -811,7 +825,8 @@ def test_eof_calls_no_balance(
             Op.SSTORE(slot_call_result, Op.EXTCALL(address=noop_callee_address, value=1))
```
marioevz, line 825:
```suggestion
            Op.SSTORE(slot_call_result, Op.EXTCALL(address=noop_callee_address, value=balance + 1))
```
I suggest we instead use `balance + 1` here to test:
- Zero balance, non-zero value transfer
- Non-zero balance, not enough value transfer

And the transfer fails in both cases, instead of testing valid value transfer, which I think should already be covered elsewhere.

tests/prague/eip7692_eof_v1/tracker.md
marioevz: I feel like we should not remove this and eventually implement them (or mark them as such if they already are implemented) even if there is no evmone equivalent test, wdyt?

marioevz (commented):
Looks good! A few comments.

tests/prague/eip7692_eof_v1/tracker.md
pdobacz: ah, hold on, it is just a copy pasta error in the path, fixed now (and we have such test already, I just need to parametrize with value, should be easily done in this PR). Thanks for flagging, I was going to fast here and at first thought it was just a renamed test.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py
pdobacz: I think they aren't, successful EXTCALL with value was the case I wanted to add here. Or have I missed them somewhere?

That being said, adding a balance=1 value=2 case makes a lot of sense. I have generalized the test a bit, and I've also updated it name to reflect this properly, PTAL.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_calls.py
```diff
@@ -796,22 +814,27 @@ def test_eof_calls_min_callee_gas(
     )
 
 
-def test_eof_calls_no_balance(
+@pytest.mark.parametrize("balance", [0, 1, 2, 2**256 - 1])
```
marioevz, line 817:
```suggestion
@pytest.mark.parametrize(
    "balance", [0, 1, 2, pytest.param(2**256 - 1, marks=pytest.mark.pre_alloc_modify)]
)
```
Adding a marker just to make sure we don't try to run this test in live networks due to the required balance amount.

marioevz (approved):
LGTM, thanks for the changes!

PR #761, opened 2024-08-23
verkle: fix balance overflow

jsign:
The current EIP-6800 defines the balance as a value of [up to 128 bits](https://eips.ethereum.org/EIPS/eip-6800#header-values). We had a test which defined a pre-state with a balance of $10^{40} > 128$ bits.


spencer-tb (approved):
LGTM! :D

PR #762, opened 2024-08-23
feat(plugins/forks): Fix covariant markers, add `selector`/`marks` parameters

marioevz:
## 🗒️ Description
- Fix case when the fork method returns a list or tuple and the marker selector only allows a single parameter.
- Disallow positional parameters on all `with_all` markers
- Fix parametrization when combining `with_all` markers that share a common parameter, e.g. `with_all_call_opcodes` and `with_all_create_opcodes` share `evm_code_type`, therefore when used together, only test cases where the `evm_code_type` is compatible with each marker (e.g `Op.CALL + Op.CREATE` or `Op.EXTCALL + Op.EOFCREATE`, but not `Op.EXTCALL + Op.CREATE`)
- Add support for markers by allowing `marks` and `marks_selector` keyword parameter in all `with_all` markers, this method has the disadvantage that the same set of markers needs to be used in all test cases where the marker is applied.

## 🔗 Related Issues
None

## ✅ Checklist


danceratopz (approved):
Hey LGTM! Few small comments below. It's pretty heavy stuff :laughing:  but :100: worth the overhead!

I like it (and think we can merge this as-is), but I'm not entirely sure we have the right interface for applying marks yet (these don't seem to used anywhere...yet). 

For example, it's entirely possible to want to apply multiple marks to multiple parameters and, if I understand correctly, it's currently only possible to apply multiple marks to *one* of the parameters?

What I mean is something like either:
1. ```python
    @pytest.mark.with_all_tx_types(
                marks=[[pytest.mark.xfail], [pytest.mark.skip]],
                marks_selector=[lambda tx_type: tx_type == 1, lambda tx_type: tx_type == 2],
            )
    ```
    or,
2. ```python
    def tx_types_marks(tx_type: int):
        if tx_type == 1:
            return [pytest.mark.xfail, pytest.mark.slow]
        if tx_type == 2:
            return [pytest.mark.skipif(...)]
        return []
    @pytest.mark.with_all_tx_types(marks=tx_types_marks)
    ```
    i.e., Allow the user to define a function that returns the marks to be applied for each param. 

But neither of these solve the case for marking `with_all_txtypes` and `with_all_precompiles` and wanting to mark `if tx_type == 2 and precompile == 1`? :laughing: 

I pushed some additions to the tests, which I mainly made to help me get into the topic. In particular, I added a couple of test cases for `marks` and `marks_selector` cf f17a38351.

Also, not necessary for this PR, but what do you think about moving these fork helper functions `get_from_until_fork_set`, `get_forks_with_no_parents`, `get_forks_with_no_descendants` and `get_last_descendants` from `pytest_plugins.forks.forks` to `ethereum_test_forks.helpers`? Or do you consider them too specific?

marioevz:
>     2. ```python
>         def tx_types_marks(tx_type: int):
>             if tx_type == 1:
>                 return [pytest.mark.xfail, pytest.mark.slow]
>             if tx_type == 2:
>                 return [pytest.mark.skipif(...)]
>             return []
>         @pytest.mark.with_all_tx_types(marks=tx_types_marks)
>        ```
>        i.e., Allow the user to define a function that returns the marks to be applied for each param.

I really like number 2, I think it's better than the selector implemented in this PR. I'm gonna give this a try and see if I can implement it in this PR.

> But neither of these solve the case for marking `with_all_txtypes` and `with_all_precompiles` and wanting to mark `if tx_type == 2 and precompile == 1`? 😆

We'll think about something when we need this :P

> Also, not necessary for this PR, but what do you think about moving these fork helper functions `get_from_until_fork_set`, `get_forks_with_no_parents`, `get_forks_with_no_descendants` and `get_last_descendants` from `pytest_plugins.forks.forks` to `ethereum_test_forks.helpers`? Or do you consider them too specific?

Sounds good, will do it in this PR :+1:

PR #764, opened 2024-08-26
chore: simplify python project config

danceratopz:
## 🗒️ Description

This PR consolidates the project's config into `pyproject.toml` (and `.flake8` - `flake8` doesn't support `pyproject.toml`; this config file can be removed later if we replace `flake8` with `ruff`).

This brings EEST more inline with modern Python project standards: "Starting with [PEP 621](https://peps.python.org/pep-0621/), the Python community selected `pyproject.toml` as a standard way of specifying project metadata.", cf [setuptools.pypa](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html).

This is a first step so that EEST can:
1. Use a python package manager such as uv, poetry or pdm.
2. Distribute its python packages as a wheel.

This PR does some additional clean-up of the config in `pyproject.toml`:
- Removes the explicit definition of packages to include in the wheel: Previously auto-discovery was enabled in `setup.py` to discover sub-modules (see 20fa624f42c8a86bc39ad106aa8a068ba38efae6) but packages to include were additionally manually listed in `setup.cfg`. Now, auto-discovery is adequate with the use of `exclude` in `pyproject.toml`.
- Removes explicit inclusion of `"py.typed"` as package data; these are added by default in newer versions of setuptools (https://github.com/pypa/setuptools/pull/4021).

## 🔗 Related Issues
None.

## ✅ Checklist


.flake8
```diff
@@ -0,0 +1,34 @@
...
+per-file-ignore =
+    tests/evm_transition_tool/test_evaluate.py:E501
+
+extend-exclude =
+    setup.py
+    src/evm_transition_tool/tests/
+    src/ethereum_test_tools/tests/
+    src/ethereum_test_forks/tests/
+
+# vim: set ft=dosini:
```
spencer-tb, line 34: Do we need these modelines?

spencer-tb (commented):
Very much in favor of this! :D

.flake8
danceratopz: Well spotted, thanks. That snuck over from `setup.cfg`, lightclient added them in the initial setup (99bfc0b30f2fd235ffa349606605ce9a9683c6fb). I don't have anything against these (limited to a few places, at least) if they help people's editing experience, but as it was intended for `.cfg`, I think it's save to remove them here. Done in bc50330b0.

PR #771, opened 2024-08-28
new(tests): EOF - EIP-7069 - expand EXT*CALL gas testing

pdobacz:
## 🗒️ Description

Expand the gas testing mechanism from `test_gas.py` to handle also oog conditions. Then leverage it to test several new conditions for EXT*CALL (mem expansion and account creation). CC @shemnon 

## 🔗 Related Issues

Cases taken from coverage gaps report in https://github.com/ipsilon/eof/pull/138

## ✅ Checklist


tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -50,25 +57,29 @@ def gas_test(
...
+        Container.Code(setup_code + subject_code + tear_down_code)
+    )
+    # GAS, POP, CALL, 7 times PUSH1 - instructions charged for at every gas run
+    gas_single_gas_run = 2 + 2 + 100 + 7 * 3
     address_legacy_harness = pre.deploy_contract(
         code=(
             # warm subject and baseline without executing
             (Op.BALANCE(address_subject) + Op.POP + Op.BALANCE(address_baseline) + Op.POP)
-            # cold gas run
+            # Baseline gas run
```
shemnon, line 70: Why swap the cold and baseline order?

pdobacz: It seemed cleaner to keep the baseline at the "bottom of the stack" and then "fetch it" using DUPn 3 times, from varying depths

shemnon: We need to be careful then and ensure setup and teardown have no cold/warm differences.  As long as it's pushes that will hold.  But we need to keep an eye on memory expansion.

pdobacz: yeah, I thought about this a bit, but figured if baseline (setup & teardown) has any side effects affecting gas, the test is broken no matter the order, because you subtract the baseline from cold+baseline and warm+basline gas sums, so if any of the "baselines" has different gas cost, the results are off.

Memory expansion is local to the execution frame, so its warmness shouldn't persist between the `Op.CALL`s, but indeed setup/teardown have a limited set of operations they can do.

shemnon (approved):

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -50,25 +57,29 @@ def gas_test(
 
     sender = pre.fund_eoa(10**18)
```
marioevz, line 58:
```suggestion
    sender = pre.fund_eoa()
```
Can we take the opportunity to remove this hard-coded value?

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
marioevz: I think it would be a good idea to add these considerations to this function's docstring, just so that anyone using it can have this in mind when designing the setup/teardown.

tests/prague/eip7692_eof_v1/tracker.md
```diff
@@ -379,25 +379,25 @@
 - [x] EXTDELEGATECALL from EOF to legacy fails (evmone-tests: state_tests/state_transition/eof_calls/eof1_extdelegatecall_legacy.json)
 - [ ] EXTSTATICCALL forwards static mode (evmone-tests: state_tests/state_transition/eof_calls/extdelegatecall_static.json)
 - [x] EXTCALL with value success (evmone-tests: state_tests/state_transition/eof_calls/extcall_with_value.json)
-- [ ] EXTCALL with value from EXTSTATICCALL (evmone-tests: state_tests/state_transition/eof_calls/extcall_static_with_value.json)
+- [x] EXTCALL with value from EXTSTATICCALL (evmone-tests: state_tests/state_transition/eof_calls/extcall_static_with_value.json)
```
marioevz, line 382: I don't like having paths on external references in this file when the test is marked as implemented, because it has happened in the past that this link becomes stale if the other repo is updated.

Also this seems to suggest that the tests marked as [x] will not be implemented in this repo.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -50,25 +57,29 @@ def gas_test(
...
     address_legacy_harness = pre.deploy_contract(
         code=(
             # warm subject and baseline without executing
             (Op.BALANCE(address_subject) + Op.POP + Op.BALANCE(address_baseline) + Op.POP)
-            # cold gas run
+            # Baseline gas run
             + (
                 Op.GAS
-                + Op.CALL(address=address_subject, gas=500_000)
+                + Op.CALL(address=address_baseline, gas=500_000)
```
marioevz, line 73:
```suggestion
                + Op.CALL(address=address_baseline, gas=call_gas)
```
I think the call gas value should be a parameter (with a default) to this function.

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -97,22 +128,46 @@ def gas_test(
...
-    ["opcode", "pre_setup", "cold_gas", "warm_gas"],
+    ["opcode", "pre_setup", "cold_gas", "warm_gas", "new_account"],
     [
-        pytest.param(Op.EXTCALL, Op.PUSH0, 2600, 100, id="EXTCALL"),
-        pytest.param(Op.EXTCALL, Op.PUSH1(1), 2600 + 9000, 100 + 9000, id="EXTCALL_with_value"),
-        pytest.param(Op.EXTDELEGATECALL, Op.NOOP, 2600, 100, id="EXTSTATICCALL"),
-        pytest.param(Op.EXTSTATICCALL, Op.NOOP, 2600, 100, id="EXTDELEGATECALL"),
+        pytest.param(Op.EXTCALL, Op.PUSH0, 2600, 100, False, id="EXTCALL"),
+        pytest.param(
+            Op.EXTCALL, Op.PUSH1(1), 2600 + 9000, 100 + 9000, False, id="EXTCALL_with_value"
```
marioevz, line 147:
```suggestion
            Op.EXTCALL, Op.PUSH1(1), COLD_ACCOUNT_ACCESS_GAS + CALL_WITH_VALUE_GAS, WARM_ACCOUNT_ACCESS_GAS + CALL_WITH_VALUE_GAS, False, id="EXTCALL_with_value"
```
It would be nice if these values were spelt out in the form of named constants. It just makes it easier to follow what the test is trying to test exactly.

marioevz (commented):
Looks really good overall, just a couple of minor comments. Thanks!

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
pdobacz: I have a different proposal, the `gas` for these 3 calls is irrelevant, and actually if not enough would interfere with the idea of the test - gas should never run out.

How about `Op.GAS` is sent here, this would require only minimal adjustments to the harness, would this make sense @shemnon ?

In the same vein I'd set Transaction's gas_limit to be the block's gas_limit, which is a huge quantity by default.

pdobacz: see 8d233acd

marioevz (approved):
Thanks!

tests/prague/eip7692_eof_v1/tracker.md
pdobacz: @marioevz whoopsie, I think I totally missed this comment, sorry.

So I guess in order to avoid stale links, I should remove the evmone-tests links for the items I gave an [x]?

Though I'm not sure I share your concerns, hopefully we're done migrating at one point and we can just drop this file completely (before those links go stale)? I thought the tests with [x] mean just that they're taken care of in EEST.

pdobacz: Hey @marioevz I'm not sure how to proceed here. The items marked as [x] in this PR are not the only ones marked [x] and with the link left, so I don't want to make things inconsistent in the file. Maybe we should decide how we want to proceed here and potentially fix in bulk in a separate PR?

pdobacz: ah ok, I thouht you've meant remove altogether. OK I'll put the EEST links in then

pdobacz: @marioevz See commit `Update tracker.md with implemented EEST tests links`

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -101,22 +140,93 @@ def gas_test(
...
+            True,
+            id="EXTCALL_with_value_new_acc",
+        ),
+        pytest.param(
+            Op.EXTDELEGATECALL,
+            Op.NOOP,
+            COLD_ACCOUNT_ACCESS_GAS,
+            WARM_ACCOUNT_ACCESS_GAS,
+            True,
+            id="EXTSTATICCALL_new_acc",
```
danceratopz, line 211:
```suggestion
            id="EXTDELEGATECALL_new_acc",
```

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -101,22 +140,93 @@ def gas_test(
...
+            False,
+            id="EXTCALL_with_value",
+        ),
+        pytest.param(
+            Op.EXTDELEGATECALL,
+            Op.NOOP,
+            COLD_ACCOUNT_ACCESS_GAS,
+            WARM_ACCOUNT_ACCESS_GAS,
+            False,
+            id="EXTSTATICCALL",
```
danceratopz, line 179:
```suggestion
            id="EXTDELEGATECALL",
```

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -101,22 +140,93 @@ def gas_test(
...
+        ),
+        pytest.param(
+            Op.EXTSTATICCALL,
+            Op.NOOP,
+            COLD_ACCOUNT_ACCESS_GAS,
+            WARM_ACCOUNT_ACCESS_GAS,
+            False,
+            id="EXTDELEGATECALL",
```
danceratopz, line 187:
```suggestion
            id="EXTSTATICCALL",
```

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -101,22 +140,93 @@ def gas_test(
...
+            True,
+            id="EXTSTATICCALL_new_acc",
+        ),
+        pytest.param(
+            Op.EXTSTATICCALL,
+            Op.NOOP,
+            COLD_ACCOUNT_ACCESS_GAS,
+            WARM_ACCOUNT_ACCESS_GAS,
+            True,
+            id="EXTDELEGATECALL_new_acc",
```
danceratopz, line 219:
```suggestion
            id="EXTSTATICCALL_new_acc",
```

tests/prague/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -126,18 +236,23 @@ def test_ext_calls_gas(
...
     cold_gas: int,
-    warm_gas: int | None,
+    warm_gas: int,
+    new_account: bool,
+    mem_expansion_size: int,
+    mem_expansion_extra_gas: int,
 ):
-    """Tests 4 variations of EXT*CALL gas, both warm and cold"""
-    address_target = pre.deploy_contract(Container(sections=[Section.Code(code=Op.STOP)]))
+    """Tests variations of EXT*CALL gas, both warm and cold"""
```
danceratopz, line 244:
```suggestion
    """Tests variations of EXT*CALL gas, both warm and cold, without and with mem expansions"""
```

danceratopz (commented):
LGTM, but did the test ids get mixed up?

danceratopz (approved):
Thanks! LGTM!

PR #618, opened 2024-06-11
fix(ci): add solc-select to fill stage of build fixtures

spencer-tb:
## 🗒️ Description
Fixes the new `solc-select` dependency required to build fixtures, for both stable/develop fixture releases as well and feature (EOF/Verkle) fixture releases.

## 🔗 Related Issues
This was an oversight from the following PR: https://github.com/ethereum/execution-spec-tests/pull/604

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md) - no addition.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [x] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.

PR #604, opened 2024-06-07
feat(ci): use solc-select for tox

spencer-tb:
## 🗒️ Description
Adds a step toward deprecating the requirement of solc for EEST (using solc-select instead).

Tox passes after merging this into my origin main, check recent commits:
https://github.com/spencer-tb/execution-spec-tests

I personally have not used an "installed solc" for the past 3 months, and instead used `solc-select`.
Due to the latter I've kept these changes trailing in my local repo for running tox.

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/pull/461 - planning to split the latter into smaller PRs, this PR being one of them.

Future PRs will include the deprecation of solc completely in within EEST in favor of solc-select.

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [x] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.

PR #772, opened 2024-08-28
feat(fill): add solc plugin and automatic installation of solc

danceratopz:
## 🗒️ Description
- Adds a new plugin `solc` (most logic is refactored from `fill` to this plugin).
- Adds a new command-line flag `--solc-version=x.y.z` to `fill`
- Adds logic to install the version specified by the flag via `solc-select` automatically when `fill` is called.
- If `--solc-bin` is specified, `solc-select` is not used; old behavior remains.

The only difference is that by default (if neither of the solc flags are specified), the first solc in the path is no longer taken; rather the default version of solc as currently specified by the plugin is used via `solc-select` (0.8.24).

`solc-select` has proven to be reliable if configured correctly. Unfortunately, the config is lost when switching venv; this avoids this pitfall.

## 🔗 Related Issues
#604 #618 

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/filler/tests/test_solc.py
+++ b/src/pytest_plugins/filler/tests/test_solc.py
@@ -1,0 +1,201 @@
+"""
+Tests for the solc plugin.
+"""
+
+import os
+import shutil
+import tempfile
+from pathlib import Path
+
+import pytest
+import solc_select.constants  # type: ignore
+import solc_select.solc_select  # type: ignore
+
+pytestmark = [pytest.mark.run_in_serial]
+
+
+@pytest.fixture(scope="module")
+def create_clean_solc_select_environment(request):
+    """
+    Setup: Copies solc artifacts to a temporary location before starting tests
+    Teardown: Restores the artifacts after tests are done.
+    """
+    try:
+        test_session_solc_version, _ = solc_select.solc_select.current_version()
+    except Exception as e:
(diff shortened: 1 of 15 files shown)


src/pytest_plugins/filler/tests/test_solc.py
```diff
@@ -0,0 +1,201 @@
...
+        pytester.makeconftest(
+            f"""
+            import pytest
+            from ethereum_test_tools.code import Solc
+
+            @pytest.fixture(autouse=True)
+            def check_solc_version(request):
+                assert request.config.getoption("solc_version") == "{solc_version}"
+                #solc_bin_version = Solc(request.config.getoption("solc_bin")).version
+                #assert request.config.getoption("solc_version") == solc_bin_version
```
marioevz, line 83:
```suggestion
```
Should these be removed?

marioevz (approved):
LGTM, just one comment. Thanks!

src/pytest_plugins/filler/tests/test_solc.py
danceratopz: Thanks for catching this! I've added back a check that the `Solc` instance we create from the `solc_bin` fixture reports the expected version. This is the actual test :slightly_smiling_face:. cf c3b5f2a58.

PR #777, opened 2024-08-29
chore: use uv for package management

danceratopz:
## 🗒️ Description

Adds a `uv.lock` file and updates docs and workflow to use [`uv`](https://github.com/astral-sh/uv):
1. `uv` replaces `pip` (much faster).
2. `uv` replaces `python -m venv ./venv` / `virtualenv` / `pyenv` (venv creation much faster and more integrated).
3. `uv` manages Python installations in Github workflows (faster and ensures consistent environments across different CI environments, even when [executing workflows locally with nectos/act](https://ethereum.github.io/execution-spec-tests/main/dev/test_actions_locally/)).
4. `tox-uv` replaces `tox` (but user still calls `tox` as usual).

### Why?

The main motivation is to:
1. Manage third-party packages via a lock file ([uv.lock](https://docs.astral.sh/uv/concepts/projects/#project-lockfile)) for reproducible installations, ensuring consistency across different environments and machines.

`uv` in particular (in comparison to `poetry`/`pdm`) can help improve the developer experience:
1. It's significantly faster than traditional tools like `pip` and `virtualenv`, cf [benchmarks.md](https://github.com/astral-sh/uv/blob/main/BENCHMARKS.md).
2. Removes the need to manually create/manage venvs (`uv` automatically checks and manages deps in`./venv` when necessary, see [the project environments doc](https://docs.astral.sh/uv/concepts/projects/#project-environments)).
3. Can run [scripts](https://docs.astral.sh/uv/concepts/projects/#running-scripts) and external [tools](https://docs.astral.sh/uv/concepts/tools/) in isolated environments automatically, simplifying workflows.
4. Can be installed without Python available (it's implemented in rust) and can be used to [manage/install Python versions](https://docs.astral.sh/uv/concepts/python-versions/#installing-a-python-version) (particularly useful bootstrapping and for maintaining consistent Python versions in CI environments).

### Trying out `uv` 

**This branch is not required to experiment working with `uv` in other branches.**

To get started, it's prob best to clean-up any old venvs and disable pyenv, if enabled, but not essential. Then install `uv` and `tox-uv` in the best place for your system:
```
pip install uv tox-uv
```
On Ubuntu (as pip is not executed as root), these will be installed in `~/.local`.

Then, in your execution-spec-tests checkout, run:
``` 
`uv sync --all-extras`
```
This will generate a `uv.lock` file and will create and install `ethereum-execution-spec-tests` in `./.venv`.

All commands (including `fill` and `consume`) will only be available via `uv run ...`. It is recommended to work this way as `uv`  checks for dependency or package changes when executing commands and using `uv run` will always trigger a re-install of `ethereum-execution-spec-tests`. This is very valuable and ensures that your local environment is always up-to-date when switching branches. `uv` is extremely quick so the checks, respectively re-installation, aren't bothersome. A huge upside is never having to create a new virtual environment when switching branches.

To avoid `uv run fill`/`uv run consume`, these aliases can be used:
```
echo -e "alias fill='uv run fill'\nalias config='uv run config'\n" >> ~/.config/fish/config.fish
```
It's also possible to source `uv`'s venv: `source .venv/bin/activate.fish` to get direct access to the `fill` and `consume` access points, but then there will be no automatic dependency checks.

`tox` can simply be called as previously. 

### More information

- Announcement of `uv`'s new capabilities: https://astral.sh/blog/uv-unified-python-packaging
- Armin Ronacher's thoughts on Python packaging and `uv`: https://lucumr.pocoo.org/2024/8/21/harvest-season/ (previous maintainer of `rye`, which inspired `uv`) 
- Docs: https://docs.astral.sh/uv/

### Other Changes

1. The new script in `.github/scripts/get_release_props.py` makes use of `uv`'s [script dependency](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies) feature. It always installs/checks dependencies and runs in its own isolated environment.
    ```console
    uv run .github/scripts/get_release_props.py develop
    # ->
    # evm-type=develop
    # fill-params=--until=Prague
    # solc=0.8.21
    ```
### Potential Issues / Todos

- [x] `tox run-parallel` doesn't seem to work reliably (still debugging; not considered a blocker), cf 385ea80. Even without running the tox envs in parallel, >= 385ea80 is slightly faster than non-uv actions (~7m 0s vs ~8m 30s).
- [ ] Do we still support all required platforms (e.g. ARM for Guillaume?).
- [ ] Should we consolidate `docs`, `test`, `lint` extras to the developer deps section (for simplicity).

### Other Caveats.

1. `uv` is maintained by [astral.sh](https://astral.sh/about) which is a venture-backed company. Non issue, imo, `uv` uses an Apache 2.0 license.
2. `uv` is new. Initially launched mid-Feb 2024, it's newer features are from mid-Aug 2024. It's seen fast adoption and appears to have taken many learnings (and more?) from [`rye`](https://github.com/astral-sh/rye) (rye was previously maintained by Armin Ronacher, who handed over stewardship of rye to astral in 2024).

## 🔗 Related Issues
#764 was a preparatory step for this PR.

## ✅ Checklist


.github/workflows/coverage.yaml
```diff
@@ -151,6 +154,8 @@ jobs:
           mkdir -p fixtures/eof_tests
           echo "$files" | while read line; do
             file=$(echo "$line" | cut -c 3-)
+            uv run fill $file --until=Cancun --evm-bin evmone-t8n --solc-version=0.8.25 || true >> filloutput.log 2>&1
+            (uv run fill $file --fork=CancunEIP7692 --evm-bin evmone-t8n --solc-version=0.8.25 -k eof_test || true) > >(tee -a filloutput.log filloutputEOF.log) 2>&1
             fill $file --until=Cancun --evm-bin evmone-t8n --solc-version=0.8.25 || true >> filloutput.log 2>&1
             (fill $file --fork=CancunEIP7692 --evm-bin evmone-t8n --solc-version=0.8.25 -k eof_test || true) > >(tee -a filloutput.log filloutputEOF.log) 2>&1
```
marioevz, line 160:
```suggestion
```
Merge issue.

marioevz (approved):
LGTM. I think inevitably there will be issues on some setups that we need to address once we find them.

.github/workflows/coverage.yaml
danceratopz: Aaaaah!! Thanks!

PR #781, opened 2024-09-02
feat(plugins): Features plug-in, EIP-7702 mode

marioevz:
## 🗒️ Description
Introduces a new pytest plugin called "features" which allows enabling a specific feature globally for all tests.

### EIP-7702 Mode

Fills all tests with each deployed contract being a set-code-delegated account instead of an actual contract, i.e. the behavior of `pre.deploy_contract` changes to the following:

1) Deploy the requested code at some location `A`
2) Generate an EOA account `B`
3) Set storage of `B` to the requested initial storage (In case of normal allocation during filling, the storage is simply set in the dictionary, but if the test was to be executed on a live network, there would need to be a first authorization transaction to a contract which only purpose is to set the storage)
4) Create an authorization to set-code of `B` to `A`
5) Return address `B`

### EOF V1 Mode

Fill all tests with each deployed contract (with unspecified EVM code type) set to EOF V1 code type, i.e. `pre.deploy_contract` now first checks if the parameter `evm_code_type` is unset, and if the provided deployment code is not already `ethereum_test_tools.eof.v1.Container`, and if so then wraps the code in an EOF container before setting/deploying the code.

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/filler/pre_alloc.py
+++ b/src/pytest_plugins/filler/pre_alloc.py
@@ -39,15 +39,10 @@ def pytest_addoption(parser: pytest.Parser):
     """
     Adds command-line options to pytest.
     """
-    pre_alloc_group = parser.getgroup("pre_alloc", "Arguments defining pre-allocation behavior.")
-
-    pre_alloc_group.addoption(
-        "--strict-alloc",
-        action="store_true",
-        dest="strict_alloc",
-        default=False,
-        help=("[DEBUG ONLY] Disallows deploying a contract in a predefined address."),
+    pre_alloc_group = parser.getgroup(
+        "pre_alloc", "Arguments defining pre-allocation behavior during test filling"
     )
+
     pre_alloc_group.addoption(
         "--ca-start",
         "--contract-address-start",
@@ -67,13 +62,11 @@ def pytest_addoption(parser: pytest.Parser):
         help="The address increment value to each deployed contract by a test.",
     )
     pre_alloc_group.addoption(
-        "--evm-code-type",
-        action="store",
-        dest="evm_code_type",
-        default=None,
-        type=EVMCodeType,
-        choices=list(EVMCodeType),
-        help="Type of EVM code to deploy in each test by default.",
+        "--strict-alloc",
+        action="store_true",
+        dest="strict_alloc",
+        default=False,
+        help=("[DEBUG ONLY] Disallows deploying a contract in a predefined address."),
     )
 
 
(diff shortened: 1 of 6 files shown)


src/pytest_plugins/filler/pre_alloc.py
```diff
@@ -86,6 +79,9 @@ class AllocMode(IntEnum):
     STRICT = 1
 
 
+SET_CODE_DELEGATION_DESIGNATION = b"\xef\x01\x00"
```
danceratopz, line 82:
```suggestion
SET_CODE_DELEGATION_DESIGNATION = bytes.fromhex("ef0100")
```

danceratopz (commented):
Very powerful! Few small comments below.

Maybe I missed this somewhere: For `EOF_V1` mode, could we add some checks and potentially `pytest.fail()` if it tries to write invalid EOF bytecode to the container (or simply defer to legacy EVM code deployment)? Or do we have state_tests with intentionally invalid EOF bytecode? cf `code_pre_processor` in https://github.com/ethereum/execution-spec-tests/pull/781/files#diff-e8af78ece9137088fbc9a58dc792341b4881687217d5375010daf85a98d5cd6dR124

What's the intent with regards to handling fork support of features? Should we check that the current fork supports the feature and only enable it, if so?

Btw, I really wish we could add a marker to mark all tests that do call `pre.deploy_contract()`, but I don't think it's possible?

marioevz:
> What's the intent with regards to handling fork support of features? Should we check that the current fork supports the feature and only enable it, if so?

I don't see the flag as being used either automatically by the CI or to fill tests for a release just yet. Right now if you enable the flag you should expect failures because not all tests are compatible, and I think that should be ok. Purpose as I see it is to analyze the errors and check why the incompatible tests are incompatible, fill the ones that can be filled and run those against all clients.

> Maybe I missed this somewhere: For `EOF_V1` mode, could we add some checks and potentially `pytest.fail()` if it tries to write invalid EOF bytecode to the container (or simply defer to legacy EVM code deployment)? Or do we have state_tests with intentionally invalid EOF bytecode? cf `code_pre_processor` in https://github.com/ethereum/execution-spec-tests/pull/781/files#diff-e8af78ece9137088fbc9a58dc792341b4881687217d5375010daf85a98d5cd6dR124

I feel like it's not straightforward to check whether any code is valid EOF code, and the result could be just a different failure? (Is the evm tool failing  != `pytest.fail()`?)

> Btw, I really wish we could add a marker to mark all tests that do call `pre.deploy_contract()`, but I don't think it's possible?

I've been thinking that we could run the test functions during collection with a mock `state_test` or `blockchain_test`, and a mock `pre` that only gathers info and never runs the transition tool for example, but for other purposes such as automatically marking the opcodes used in each test.

PR #785, opened 2024-09-04
new(tests): EOF - EIP-7620 EOFCREATE gas testing

pdobacz:
## 🗒️ Description

EOFCREATE gas testing using the "Op.GAS harness" used in #771.

~I'm pushing early to signal the [fix(fw): max stack height calculation in `__add__`](https://github.com/ethereum/execution-spec-tests/commit/acee5dab6abac2c8293d3951b5a8613ecf0e0729). The new bytecodes used in the new tests didn't have the correct autocalculated max stack height. After some digging, I've noticed that the algorithm might not handle all cases well, and I am proposing an alternative. I've left all my overly verbose comments in the code to illustrate the train of thought, as this has been quite hard to grasp the original code intentions (maybe there has been an easier fix hidden from my sight there).~ EDIT: the max stack height fix split out to https://github.com/ethereum/execution-spec-tests/pull/810, merge 810 first

Apart from that the EOFCREATE test reveals some difficulties related to the harness, and how the harness needs to be generalized to test such instructions. However, since the difficulties are related to satisfying the requirements of subject code (the EOFCREATE in question here has many), I suppose these difficulties would be present in all gas testing approaches.

## 🔗 Related Issues

Addresses the last remaining item off of the list https://github.com/ipsilon/eof/pull/138.

## ✅ Checklist

### DIFF
--- a/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_gas.py
+++ b/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_gas.py
@@ -37,19 +37,6 @@ pytestmark = pytest.mark.valid_from(EOF_FORK_NAME)
 EOFCREATE_GAS = 32000
 
 
-def make_initcode(runtime: Container):
-    """
-    Wraps a runtime container into a minimal initcontainer
-    """
-    return Container(
-        name="Initcode Subcontainer",
-        sections=[
-            Section.Code(code=Op.RETURNCONTRACT[0](0, 0)),
-            Section.Container(container=runtime),
-        ],
-    )
-
-
 def make_factory(initcode: Container):
     """
     Wraps an initcontainer into a minimal runtime container
@@ -106,7 +93,7 @@ def make_factory(initcode: Container):
             id="smallest_code",
         ),
         pytest.param(
-            make_initcode(aborting_container),
+            Container.Init(aborting_container),
             smallest_initcode_subcontainer_gas,
             aborting_container,
             id="aborting_runtime",
@@ -121,13 +108,13 @@ def make_factory(initcode: Container):
             id="expensively_reverting_initcode",
         ),
         pytest.param(
-            make_initcode(big_runtime_subcontainer),
+            Container.Init(big_runtime_subcontainer),
             smallest_initcode_subcontainer_gas,
             big_runtime_subcontainer,
             id="big_runtime",
         ),
         pytest.param(
-            make_initcode(make_factory(smallest_initcode_subcontainer)),
+            Container.Init(make_factory(smallest_initcode_subcontainer)),
             smallest_initcode_subcontainer_gas,
             make_factory(smallest_initcode_subcontainer),
             id="nested_initcode",
--- a/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_gas.py
+++ b/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_gas.py
@@ -7,6 +7,7 @@ import pytest
 from ethereum_test_tools import Alloc, Environment, StateTestFiller, compute_eofcreate_address
 from ethereum_test_tools.eof.v1 import Container, Section
 from ethereum_test_tools.vm.opcode import Opcodes as Op
+from ethereum_test_types.helpers import cost_memory_bytes
 
 from .. import EOF_FORK_NAME
 from ..gas_test import gas_test
@@ -92,13 +93,8 @@ def make_factory(initcode: Container):
     ],
 )
 @pytest.mark.parametrize(
-    ["mem_expansion_size", "mem_expansion_extra_gas"],
-    [
-        pytest.param(0, 0, id="no_mem_expansion"),
-        pytest.param(1, 3, id="1byte_mem_expansion"),
-        pytest.param(32, 3, id="1word_mem_expansion"),
-        pytest.param(33, 6, id="33bytes_mem_expansion"),
-    ],
+    "mem_expansion_bytes",
+    [0, 1, 32, 33],
 )
 @pytest.mark.parametrize(
     ["initcode", "initcode_execution_cost", "runtime"],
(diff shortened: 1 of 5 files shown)


tests/prague/eip7692_eof_v1/gas_test.py
pdobacz, line 1: I have made a rudimentary refactor of the `gas_test` utility into this separate file. Not sure about placement or conventions, please advise if this is acceptable and if not, where to best put it. The `ethereum_test_types.eof.v1` perhaps?

src/ethereum_test_vm/bytecode.py
```diff
@@ -148,12 +148,35 @@ def __add__(self, other: "Bytecode | int | None") -> "Bytecode":
         a_min, a_max = self.min_stack_height, self.max_stack_height
         b_pop, b_push = other.popped_stack_items, other.pushed_stack_items
         b_min, b_max = other.min_stack_height, other.max_stack_height
-        a_out = a_min - a_pop + a_push
 
-        c_pop = max(0, a_pop + (b_pop - a_push))
-        c_push = max(0, a_push + b_push - b_pop)
-        c_min = a_min if a_out >= b_min else (b_min - a_out) + a_min
-        c_max = max(a_max + max(0, b_min - a_out), b_max + max(0, a_out - b_min))
+        # NOTE: "_pop" is understood as the number of elements required by an instruction or
```
marioevz, line 152: I think this should definitely be split into its own PR, and also it would be great if we could add the test cases that this new approach covers and the previous implementation didn't (here: https://github.com/ethereum/execution-spec-tests/blob/061e41cce851e5d4d7de1d0612c84349f78b492e/src/ethereum_test_vm/tests/test_vm.py#L361).

marioevz: Just a tip: if you want to execute these framework tests specifically, you can simply run:
```bash
pytest -c pytest-framework.ini ./src/ethereum_test_vm/tests/test_vm.py::test_bytecode_properties
```

tests/prague/eip7692_eof_v1/gas_test.py
marioevz: I think the placement here is fine, only because this generator is very specific (at the moment at least) to EOF because it forces the use of containers for the contracts under test.

We have `src/ethereum_test_tools/code/generators.py` but it's not exactly the same since the classes in that file only generate `Bytecode` instances but this one uses `pre` to deploy many contracts.

marioevz (commented):
It looks good, but I think we should merge the changes to `Bytecode` first as a separate PR.

src/ethereum_test_vm/bytecode.py
pdobacz: Done: https://github.com/ethereum/execution-spec-tests/pull/810

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_gas.py
```diff
@@ -0,0 +1,208 @@
...
+    initcode_execution_cost: int,
+    runtime: Container,
+):
+    """Tests variations of EOFCREATE gas"""
+    initcode_hashing_cost = 6 * ((len(initcode) + 31) // 32)
+    deployed_code_cost = 200 * len(runtime) if runtime else 0
+
+    sender = pre.fund_eoa()
+
+    salt_addresses = [compute_eofcreate_address(sender, i, initcode) for i in range(10)]
```
marioevz, line 177:
```suggestion
    salt_addresses = [compute_eofcreate_address(sender, i, initcode) for i in range(2)]
```
I think this can be reduced.

marioevz (commented):
Looks really good, just a few comments.

One concern I have is that it might be too many tests, we have an utility to parametrize in a more conservative manner by using defaults in some of the parameters and only create a limited amount of vectors, instead of creating the product of all values of all parameters: https://ethereum.github.io/execution-spec-tests/main/writing_tests/writing_a_new_test/#the-extend_with_defaults-utility

It might be worth exploring using this utility here.

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_gas.py
pdobacz: Thanks for flagging. I started digging at this, and it turns out I incorrectly took this from the EIP-7069 test. I'll push a fix

pdobacz: Ah, on the number of required salt_addresses - it is actually 4 - as many as the number of times the counter may be bumped. This is related to the number of calls to the `address_subject` in `gas_test.py`. It is very ugly to couple this test with the specifics of `gas_test.py`, but I don't really have an idea on how to fix this (without making a terrible mess).

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_gas.py
```diff
@@ -0,0 +1,208 @@
...
+        subject_code=Op.EOFCREATE[0],
+        tear_down_code=Op.STOP,
+        cold_gas=cold_gas
+        + mem_expansion_extra_gas
+        + initcode_hashing_cost
+        + initcode_execution_cost
+        + deployed_code_cost,
+        subject_subcontainer=initcode,
+        sender=sender,
+        subject_balance=10**18,
```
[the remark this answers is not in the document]
pdobacz: as above, `value*4`, but agreed and thx :)

pdobacz:
> One concern I have is that it might be too many tests, we have an utility to parametrize in a more conservative manner by using 

Ah, missed this comment, let me take a look before I ask for re-review, sorry

marioevz:
> But this is a bit impacting the test with implementation details. Is the fill time, test run time or JSON files size the limiting factor here?

Yes I guess it's a bit limiting, the main issues is the run time, but we are planning to mitigate this by composing a large test fixture with all state tests in it, so the amount of tests will be a bit less important.

I think we can roll this back, just wanted for us to give it some consideration.

PR #792, opened 2024-09-06
feat(transition_tool): Support EELS resolver in daemon mode

petertdavies:
## 🗒️ Description
Implement the EELS daemon support and fix various bugs and deficiencies found during testing EELS daemon support. Pre-Prague forks can be filled by installing the [EELS resolver](https://github.com/petertdavies/ethereum-spec-evm-resolver) and running `fill --evm-bin=ethereum-spec-evm-resolver`.

Currently Prague support is broken due to issues in EELS.

## 🔗 Related Issues
Most of the code in this PR was implemented in the draft PR #453.
Resolves #264.

Follow-up issues:
- #818
- #820

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [ ] All: Considered squashing commits to improve commit history.
- [ ] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.

PR #793, opened 2024-09-06
bug(fw): fix unsupported forks in forks plugin

spencer-tb:
## 🗒️ Description
Fixes how we use unsupported forks when filling tests, such that they are skipped appropriately, when a transition tool does not support a fork we try to fill for.

We were using a `filter()` which is an **iterator**, when determining the unsupported forks:
```python
config.unsupported_forks = filter(lambda fork: not t8n.is_fork_supported(fork), fork_set)
```

This essentially meant that on the 2nd call to unsupported forks, it would always be empty. This is called everytime we parameterize a test case.

The solution was to wrap the `filter()` in a `set()`, as sets are concrete collections that can be accessed multiple times.

## 🔗 Related Issues
N/A

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/forks/forks.py
+++ b/src/pytest_plugins/forks/forks.py
@@ -170,14 +170,19 @@ class CovariantDescriptor:
     def process_value(
         values: Any | List[Any] | Tuple[Any],
         selector: FunctionType,
-        marks: None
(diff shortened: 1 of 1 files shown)


spencer-tb:
Close in favour of #792!

PR #795, opened 2024-09-12
feat(exceptions,specs): class to verify exception strings

winsvega:
## 🗒️ Description
Extent exception mapper with base class so that any t8n can define exception string maps for any exception types

I put transaction exception validation into exception mapper for convinience atm.
Example exception message:
```
src/ethereum_test_exceptions/exception_mapper.py:84: in check_transaction
    raise Exception(
E   Exception: TransactionException (pos=1, nonce=0x1)
E   Error: exception mismatch:
E    want = 'Undefined' (TransactionException.TYPE_3_TX_INVALID_BLOB_VERSIONED_HASH),
E    got  = 'blob 0 has invalid hash version' (UndefinedException.UNDEFINED_EXCEPTION)
E    No message defined for TransactionException.TYPE_3_TX_INVALID_BLOB_VERSIONED_HASH, please add it to GethExceptionMapper
E    No exception defined for error message got, please add it to GethExceptionMapper
```

## ✅ Checklist


src/ethereum_test_specs/base.py
```diff
@@ -46,6 +49,13 @@ def verify_transactions(txs: List[Transaction], result: Result) -> List[int]:
             raise Exception(f"tx expected to fail succeeded: pos={i}, nonce={tx.nonce}")
         elif not tx.error and error:
             raise Exception(f"tx unexpectedly failed: {error}")
+        elif isinstance(tx.error, TransactionException) and error:
```
winsvega, line 52: perhaps this whole part should be inside ExceptionMapper. since we compare x and y and have to cover cases
expected y,  got x == y
expected y, got x != y
expected 0, got x 
expected 0, got 0

src/ethereum_test_exceptions/geth_exceptions.py
```diff
@@ -0,0 +1,140 @@
+"""
+Geth exceptions ENUM -> str mapper
+"""
+
+from .exception_mapper import ExceptionMapper, ExceptionMessage
+from .exceptions import EOFException, TransactionException
+
+
+class GethExceptionMapper(ExceptionMapper):
```
marioevz, line 9: These are very directly related to the Geth T8N implementation, it might be a good idea to simply move it into `src/evm_transition_tool/geth.py`.

Same for the `evmone_exceptions.py`.

src/ethereum_test_exceptions/exception_mapper.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def message_to_exception(self, exception_string: str) -> ExceptionBase:
+        """Takes a string and tries to find matching exception"""
+        # TODO inform tester where to add the missing exception if get uses default
+        exception = self.exception_to_message_map.inverse.get(
+            exception_string, UndefinedException.UNDEFINED_EXCEPTION
+        )
+        return exception
+
+    def check_transaction(self, tx_error_message: str | None, tx, tx_pos: int):
```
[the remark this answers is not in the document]
marioevz: After looking through all the changes, I think what we could do is have a subclass called `TransitionToolExceptionMapper` that is specifically dedicated to map exceptions from the transition tool, and this can be used to subclass specific `GethTransaitionToolExceptionMapper` and so on.

Another approach could be that we don't need a subclass `GethTransaitionToolExceptionMapper` and only `TransitionToolExceptionMapper`, and then in `src/evm_transition_tool/geth.py` we create an instance of `TransitionToolExceptionMapper` that contains the map.

marioevz (commented):
I like the `ExceptionMapper` idea, I think it might be too t8n-focused but it could be used in many other places, e.g. exceptions from the Engine API when we run the tests in hive.

I think this PR can start with `TransitionToolExceptionMapper` defined in `src/evm_transition_tool` to make `ExceptionMapper` a bit more generic to open the door to future PRs to implement other subclasses.

src/ethereum_test_exceptions/geth_exceptions.py
winsvega: Yes. This can also be a base exception mapper.
Then geth t8n defines geth exception mapper for t8n.
And geth consume defines exception mapper for consume driver.

src/ethereum_test_exceptions/exception_mapper.py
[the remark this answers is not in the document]
winsvega: removed from exception mapper info test_specs

winsvega:
still need to setup other clients for this. but I refactored the suggestions. can check the code. 
now I need to fill the exception strings for other tools.

src/ethereum_test_specs/helpers.py
```diff
@@ -0,0 +1,94 @@
+"""
+Helper functions
+"""
+
+from dataclasses import dataclass
+from typing import Dict, List
+
+from ethereum_test_exceptions import ExceptionMapper, TransactionException, UndefinedException
```
marioevz, line 8:
```suggestion
from ethereum_test_exceptions import ExceptionMapper, UndefinedException
```
Unused import.

src/ethereum_test_specs/helpers.py
```diff
@@ -0,0 +1,94 @@
...
+            return  # will never happen but removes python logic check
+        else:
+            expected_exception = info.tx.error
+
+        expected_error_msg = exception_mapper.exception_to_message(expected_exception)
+        error_exception = exception_mapper.message_to_exception(info.error_message)
+        exception_mapper_name = exception_mapper.__class__.__name__
+
+        define_message_hint = (
+            f"No message defined for {expected_exception}, please add it to {exception_mapper_name}"
```
marioevz, line 57:
```suggestion
            f"No message defined for {expected_exception}, "
            f"please add it to {exception_mapper_name}"
```

src/ethereum_test_specs/helpers.py
```diff
@@ -0,0 +1,94 @@
...
+            if expected_error_msg == "Undefined"
+            else ""
+        )
+        define_exception_hint = (
+            f"No exception defined for error message got, "
+            f"please add it to {exception_mapper.__class__.__name__}"
```
marioevz, line 63:
```suggestion
            "No exception defined for error message got, "
            f"please add it to {exception_mapper_name}"
```

src/ethereum_test_specs/helpers.py
```diff
@@ -0,0 +1,94 @@
...
+            if error_exception == UndefinedException.UNDEFINED_EXCEPTION
+            else ""
+        )
+
+        if expected_error_msg not in info.error_message:
+            raise Exception(
+                f"{exception_info}"
+                f"Error: exception mismatch:\n want = '{expected_error_msg}' ({expected_exception}),"
```
marioevz, line 71:
```suggestion
                f"Error: exception mismatch:\n want = '{expected_error_msg}' "
                f"({expected_exception}),"
```

src/ethereum_test_exceptions/exception_mapper.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @property
+    @abstractmethod
+    def _mapping_data(self):
+        """This method should be overridden in the subclass to provide mapping data."""
+        pass
+
+    def exception_to_message(self, exception: ExceptionBase) -> str:
+        """Takes an exception and returns a formatted string."""
+        message = self.exception_to_message_map.get(exception, "Undefined")
```
marioevz, line 48:
```suggestion
    def exception_to_message(self, exception: ExceptionBase) -> str | None:
        """Takes an exception and returns a formatted string."""
        message = self.exception_to_message_map.get(exception, None)
```
I think we should return `None` here and then handle this case in `verify_transaction_exception`, because with `"Undefined"` the exception we get on failure is:
```
E               Exception: TransactionException (pos=1, nonce=0x5)
E               Error: exception mismatch:
E                want = 'Undefined' (TransactionException.PRIORITY_GREATER_THAN_MAX_FEE_PER_GAS),
E                got  = 'Failed transaction: ' (TransactionException.TYPE_4_TX_CONTRACT_CREATION)
E                No message defined for TransactionException.PRIORITY_GREATER_THAN_MAX_FEE_PER_GAS, please add it to ExecutionSpecsExceptionMapper
```
which is confusing.

src/evm_transition_tool/execution_specs.py
```diff
@@ -102,3 +110,137 @@ def is_fork_supported(self, fork: Fork) -> bool:
...
         return (fork.transition_tool_name() + "\n") in self.help_string
+
+
+class ExecutionSpecsExceptionMapper(ExceptionMapper):
+    """
+    Translate between EEST exceptions and error strings returned by execution specs t8n.
+    """
+
+    @property
+    def _mapping_data(self):
```
marioevz, line 121: Since this is now the default t8n tool to fill tests, I think this list should be complete and correctly populated from this PR. Currently tox is failing because it's not finding the expected exception messages, and I think there are going to be issues when running `fill` like this.

marioevz (commented):
Hey it's looking great overall, but I think we need to address the comments, mainly because the default t8n to fill tests has become the `execution_specs` t8n, so I think the exception mapper should be complete from the start.

src/ethereum_test_specs/helpers.py
winsvega: but why, that was requested to print a hint to where this missing bits shall be added

src/ethereum_test_exceptions/exception_mapper.py
winsvega: this is actually not confusing. it tells me that ExecutionSpecsExceptionMapper is missing string for `TransactionException.PRIORITY_GREATER_THAN_MAX_FEE_PER_GAS`

src/evm_transition_tool/execution_specs.py
winsvega: the fill works, but the error messages are not provided by exec specs t8n itself. it only prints Failed transaction: 
also if you managed to hit undefined exception, it must be defined yes.

src/ethereum_test_exceptions/exception_mapper.py
marioevz: What I meant is that it's looking for the string `'Undefined'` in the error message returned by the transition tool, which I think should not be the case. It should stop processing before at a prior point in the code because it's never going to find `'Undefined'` in the message (or maybe it will and that's a bit worse because it's a false positive).

src/evm_transition_tool/execution_specs.py
marioevz: But if you run tox at the moment (for the framework with `uvx --with=tox-uv tox -e framework`) it will show a failing test and I think it's related to the exceptions.

src/ethereum_test_specs/helpers.py
winsvega: can just use # macro to disable the error?

src/ethereum_test_exceptions/exception_mapper.py
winsvega: I see, I added the None return and replaced checks against "Undefined" string to checks against None

src/evm_transition_tool/execution_specs.py
winsvega: now it is green

src/ethereum_test_specs/blockchain.py
```diff
@@ -440,7 +441,9 @@ def generate_block_data(
         )
 
         try:
-            rejected_txs = verify_transactions(txs, transition_tool_output.result)
+            rejected_txs = verify_transactions(
+                t8n.exception_mapper, txs, transition_tool_output.result
+            )
```
danceratopz, line 446: I think we'll also need to add `verify_transactions()` to `src/ethereum_test_specs/state.py`.

src/ethereum_clis/clis/execution_specs.py
```diff
@@ -102,3 +110,141 @@ def is_fork_supported(self, fork: Fork) -> bool:
...
+                TransactionException.TYPE_3_TX_MAX_BLOB_GAS_ALLOWANCE_EXCEEDED,
+                "iled transaction: ",
+            ),
+            ExceptionMessage(
+                TransactionException.INSUFFICIENT_MAX_FEE_PER_BLOB_GAS,
+                "led transaction: ",
+            ),
+            ExceptionMessage(
+                TransactionException.INSUFFICIENT_MAX_FEE_PER_GAS,
+                "ed transaction: ",
```
danceratopz, line 141: These strings look very strange and deserve an explanation :slightly_smiling_face: 

I think this is because EELS doesn't yet have unique exceptions and you're just trying to ensure that the bidict has unique keys (in the reverse direction), right? And still correctly catch expected exceptions?

How about adding a ticket with a list of the missing exceptions to execution-specs for easy tracking and linking it here. I think the EELS Team still needs to know the list of exceptions they have to improve exceptions strings for, too?

src/ethereum_test_specs/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+):
+    """Verify transaction exception"""
+    exception_info = f"TransactionException (pos={info.transaction_ind}, nonce={info.tx.nonce})\n"
+    expected_error: bool = info.tx.error is not None or (
+        isinstance(info.tx.error, list) and len(info.tx.error) != 0
+    )
+
+    # info.tx.error is expected error code defined in .py test
+    if expected_error and not info.t8n_error_message:
+        raise Exception(f"{exception_info} Error: tx expected to fail succeeded")
```
danceratopz, line 33: You could define a new Exception class (or classes) to help with these exceptions that handle the relevant info instead of using a `TransactionExceptionInfo` class. I think this would be less surprising to the reader and consistent with how we implement this elsewhere.

For example, this Exception class (see also its base class):
https://github.com/ethereum/execution-spec-tests/blob/5d8d350470f45948c7813fdba21eb8c54a035b67/src/ethereum_test_specs/eof.py#L72-L85

and usage of it here:
https://github.com/ethereum/execution-spec-tests/blob/5d8d350470f45948c7813fdba21eb8c54a035b67/src/ethereum_test_specs/eof.py#L241-L272

src/ethereum_test_specs/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+        if expected_error_msg is None or expected_error_msg not in info.t8n_error_message:
+            raise Exception(
+                f"{exception_info}"
+                f"Error: exception mismatch:"
+                f"\n want = '{expected_error_msg}' ({expected_exception}),"
+                f"\n got  = '{info.t8n_error_message}' ({t8n_error_exception})"
+                f"\n {define_message_hint}"
+                f"\n {define_exception_hint}"
```
danceratopz, line 80: I would also add an Exception class for this and use a mutli-line f-string for the message and apply `textwrap.dedent()` to it. There's examples of `dedent()` usage within EEST.

danceratopz (commented):
This looks fantastic, looking forward to using these exceptions in `consume` in due course! 

A few small comments below.

src/ethereum_test_specs/blockchain.py
winsvega: yes this is so far the blockchain version of tests only

src/ethereum_clis/clis/execution_specs.py
winsvega: the issue is opened https://github.com/ethereum/execution-specs/issues/1021

src/ethereum_test_specs/helpers.py
winsvega: I think TransactionExceptionInfo  still needed. because it has info from blocks but there is also info from state test

src/ethereum_test_specs/blockchain.py
winsvega: fixed

src/ethereum_test_specs/helpers.py
winsvega: done

src/ethereum_test_specs/helpers.py
winsvega: looks like it's ok without dedent. but should keep it in mind

marioevz:
Fixed tox. It turns out there was an invalid transaction in the framework tests.

marioevz (approved):
LGTM, thanks!

PR #796, opened 2024-09-12
chore(fw): add chain id to all fixtures

spencer-tb:
## 🗒️ Description
Adds the `chain_id` as a field for all fixture types. Set to 1 by default unless overwritten in the state/blockchain test environment.

### Blockchain/Engine Test Example Json
```json
{
    "tests/shanghai/eip3855_push0/test_push0.py::test_push0_contracts[fork_Shanghai-blockchain_test]": {
        "network": "Shanghai",
        "chainId": "0x01",  # root of the test fixture
        ...
```

### State Test Example Json
```json
{
    "tests/shanghai/eip3855_push0/test_push0.py::test_push0_contracts[fork_Shanghai-state_test]": {
        "chainId": "0x01",  # within the state test env
        "env": {
            "currentCoinbase": "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
            ...
```

## 🔗 Related Issues
N/A

## ✅ Checklist


src/ethereum_test_fixtures/blockchain.py
```diff
@@ -483,6 +483,8 @@ class FixtureCommon(BaseFixture):
     """
 
     fork: str = Field(..., alias="network")
+    chain_id: ZeroPaddedHexNumber = Field(1, alias="chainId")  # type: ignore
```
marioevz, line 486: I think it should not be necessary to add the alias here.

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -483,6 +483,8 @@ class FixtureCommon(BaseFixture):
...
+
```
marioevz, line 487:
```suggestion
```
Extra line.

src/ethereum_test_specs/blockchain.py
```diff
@@ -316,7 +316,6 @@ class BlockchainTest(BaseTest):
     blocks: List[Block]
     genesis_environment: Environment = Field(default_factory=Environment)
     verify_sync: bool = False
-    chain_id: int = 1
```
marioevz, line 319: Wait we already had this in the blockchain test?
I think it's safer to reuse this, because it's not really part of the environment, but more of the chain configuration.
Same for the state test, it should be a property of `state_test`, not environment.

marioevz: Also, the value has to be copied over on https://github.com/ethereum/execution-spec-tests/blob/061e41cce851e5d4d7de1d0612c84349f78b492e/src/ethereum_test_specs/state.py#L95

marioevz (commented):
Thanks, some comments.

src/ethereum_test_specs/blockchain.py
spencer-tb: Ah yes this makes more sense! For some reason I thought it would be easier to parameterize the chain id this way but definitely not.

1) What about moving it up to BaseTest (on the assumption that all tests required a chain id, which I feel is safe)?

2) This has me thinking about moving `pre, post, env` from State/BlockchainTest into BaseTest as well. This would work for now but would force future tests to include these fields. A future TransactionTest might not want these fields.

src/ethereum_test_fixtures/blockchain.py
spencer-tb: Nice catch!

src/ethereum_test_specs/blockchain.py
marioevz: Yes I think it might be a good idea, there might be some issues I'm not seeing right now but it should be straight-forwards.

spencer-tb: I completed 1) here (moving it up to base test). Skipping 2) as transaction tests will make it redundant.

winsvega:
So when generating a test with chain_id  will all the transactions be of that same chain_id ? 
Also we should allow to have tx with different chain id for test purposes

winsvega:
surpassed by https://github.com/ethereum/execution-spec-tests/pull/1131

PR #798, opened 2024-09-13
new(tests): TSTORE: ensure transient storage is cleared after transactions

jochem-brouwer:
## 🗒️ Description
This PR introduces 2 tests, which test that the transient storage is cleared after transactions. EthereumJS did clear storage after "normal" transactions, but did not clear the transient storage after a transaction created a contract. If the deployed contract address was then called, then the transient storage was not empty. See: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3625. This was not caught by these tests, so I have added two tests, one which does the create-tx and then call-contract, and the other calls the same contract twice.

The tests are very repetitive, so I would love some suggestions how to make those more compact (we could also move these two tests into one single test, where we use 3 transactions instead of two, writing to different storage slots and different transient storage slots).

The good/bad cases of these tests have been tested using EthereumJS (with and without the bug) using our new t8ntool :smile: 

## 🔗 Related Issues
Test produced by a bug in EthereumJS: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3625

## ✅ Checklist

### DIFF
--- a/tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
+++ b/tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
@@ -51,8 +51,6 @@ def test_tstore_clear_after_deployment_tx(
     else:
         code = Initcode(deploy_code=deploy_code, initcode_prefix=init_code)
 
-    code = Initcode(deploy_code=deploy_code, initcode_prefix=init_code)
-
     sender = pre.fund_eoa()
 
     deployment_tx = Transaction(
--- a/tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
+++ b/tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
@@ -1,0 +1,96 @@
+"""
+Ethereum Transient Storage EIP Tests
+https://eips.ethereum.org/EIPS/eip-1153
+"""
+
+import pytest
+
+from ethereum_test_tools import (
+    Account,
+    Alloc,
+    Block,
+    BlockchainTestFiller,
+    Environment,
+    Initcode,
+    Transaction,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+from .spec import ref_spec_1153
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_1153.git_path
+REFERENCE_SPEC_VERSION = ref_spec_1153.version
+
+
+@pytest.mark.valid_from("Cancun")
+def test_tstore_clear_after_deployment_tx(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+):
+    """
+    This test first creates a contract, which TSTOREs a value 1 in slot 1.
+    After creating the contract, a new tx will call this contract, storing TLOAD(1) into slot 1.
+    The transient storage should be cleared after creating the contract (at tx-level), so
+    the storage should stay empty.
+    """
+    env = Environment()
+
+    init_code = Op.TSTORE(1, 1)
+    deploy_code = Op.SSTORE(1, Op.TLOAD(1))
+
+    code = Initcode(deploy_code=deploy_code, initcode_prefix=init_code)
+
+    sender = pre.fund_eoa()
+
+    deployment_tx = Transaction(
+        gas_limit=100000,
+        data=code,
+        to=None,
+        sender=sender,
+    )
+
+    address = deployment_tx.created_contract
+
+    invoke_contract_tx = Transaction(gas_limit=100000, to=address, sender=sender)
+
+    txs = [deployment_tx, invoke_contract_tx]
+
+    post = {
+        address: Account(storage={0x01: 0x00}),
+    }
+
+    blockchain_test(genesis_environment=env, pre=pre, post=post, blocks=[Block(txs=txs)])
+
+
+@pytest.mark.valid_from("Cancun")
+def test_tstore_clear_after_tx(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+):
+    """
+    This test first SSTOREs the TLOAD value of key 1 in slot 1. Then, it TSTOREs 1 in slot 1.
+    The second tx will re-call the contract. The storage should stay empty,
+    because the transient storage is cleared after the transaction.
+    """
+    env = Environment()
+
+    contract_code = Op.SSTORE(1, Op.TLOAD(1)) + Op.TSTORE(1, 1)
+    account = pre.deploy_contract(contract_code)
+
+    sender = pre.fund_eoa()
+
+    poke_tstore_tx = Transaction(
+        gas_limit=100000,
+        to=account,
+        sender=sender,
+    )
+
+    re_poke_tstore_tx = Transaction(gas_limit=100000, to=account, sender=sender)
+
+    txs = [poke_tstore_tx, re_poke_tstore_tx]
+
+    post = {
+        account: Account(storage={0x01: 0x00}),
+    }
+
+    blockchain_test(genesis_environment=env, pre=pre, post=post, blocks=[Block(txs=txs)])
--- a/tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
+++ b/tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
@@ -3,6 +3,8 @@ Ethereum Transient Storage EIP Tests
 https://eips.ethereum.org/EIPS/eip-1153
 """
 
+from typing import Optional
+
 import pytest
 
 from ethereum_test_tools import (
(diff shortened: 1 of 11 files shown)


tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
```diff
@@ -0,0 +1,96 @@
...
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_1153.git_path
+REFERENCE_SPEC_VERSION = ref_spec_1153.version
+
+
+@pytest.mark.valid_from("Cancun")
+def test_tstore_clear_after_deployment_tx(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+):
```
marioevz, line 29:
```suggestion
@pytest.mark.valid_from("Cancun")
@pytest.mark.with_all_evm_code_types
def test_tstore_clear_after_deployment_tx(
    blockchain_test: BlockchainTestFiller,
    pre: Alloc,
    evm_code_type: EVMCodeType,
):
```

I feel like we could expand this test a little bit to support EOF by simply adding this marker, and then in the test doing something like:

```python
if evm_code_type == EVMCodeType.EOF_V1:
  code = Container.Init(deploy_container=Container.Code(Op.SSTORE(1, Op.TLOAD(1)) + Op.STOP), initcode_prefix=Op.TSTORE(1, 1))
else:
  code = Initcode(deploy_code=deploy_code, initcode_prefix=init_code)
```
As of now it will not run in EOF until the test is filled in a fork where EOF is active.

marioevz (commented):
Looks good! Just one comment, thanks!

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
jochem-brouwer: Have added!

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
```diff
@@ -0,0 +1,118 @@
...
+    code = None
+
+    if evm_code_type == EVMCodeType.EOF_V1:
+        code = Container.Init(
+            deploy_container=Container.Code(deploy_code), initcode_prefix=init_code
+        )
+    else:
+        code = Initcode(deploy_code=deploy_code, initcode_prefix=init_code)
+
+    code = Initcode(deploy_code=deploy_code, initcode_prefix=init_code)
```
holgerd77, line 54: Doesn’t this line invalidate the previous conditional association?

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
```diff
@@ -0,0 +1,118 @@
...
+    deployment_tx = Transaction(
+        gas_limit=100000,
+        data=code,
+        to=None,
+        sender=sender,
+    )
+
+    address = deployment_tx.created_contract
+
+    invoke_contract_tx = Transaction(gas_limit=100000, to=address, sender=sender)
```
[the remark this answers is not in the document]
jochem-brouwer: Yes, this sends a transaction without any calldata (empty calldata) to the contract we just deployed.

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
jochem-brouwer: Woops, it does!! I should resolve this

jochem-brouwer: Should be resolved

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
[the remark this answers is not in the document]
jochem-brouwer: ("invocation" would mean the same as "call a contract" here, I might have messed up some grammar/spelling/definitions or whatever this would be called)

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
```diff
@@ -0,0 +1,118 @@
...
+
+    runtime_code = Op.SSTORE(1, Op.TLOAD(1)) + Op.TSTORE(1, 1)
+
+    code: Optional[Container | Bytecode] = None
+    if evm_code_type == EVMCodeType.EOF_V1:
+        code = Container.Code(runtime_code)
+    else:
+        code = runtime_code
+
+    account = pre.deploy_contract(code)
```
marioevz, line 100:
```suggestion
    code = Op.SSTORE(1, Op.TLOAD(1)) + Op.TSTORE(1, 1) + Op.STOP
    account = pre.deploy_contract(code)
```

The logic here shouldn't be necessary for this case, pre.deploy_contract changes its behavior when we use pytest.mark.with_all_evm_code_types to read fixture evm_code_type and automatically wrap the deployed contract within an EOF container (it also should automatically add the STOP).

Might need to remove Bytecode import after this change.

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    This test first creates a contract, which TSTOREs a value 1 in slot 1.
+    After creating the contract, a new tx will call this contract, storing TLOAD(1) into slot 1.
+    The transient storage should be cleared after creating the contract (at tx-level), so
+    the storage should stay empty.
+    """
+    env = Environment()
+
+    init_code = Op.TSTORE(1, 1)
+    deploy_code = Op.SSTORE(1, Op.TLOAD(1))
```
marioevz, line 46:
```suggestion
    deploy_code = Op.SSTORE(1, Op.TLOAD(1)) + Op.STOP
```

marioevz (approved):
LGTM, thanks!

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
jochem-brouwer: I was about to comment "we don't need `Op.STOP`", but this to make the EOF container valid - right?

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
jochem-brouwer: Ok, so I can remove `Op.STOP` and then just remove the `evm_code_type` switch logic?

I think I actually copied this logic from another test, so this logic (not automatically switching to EOF) is somewhere else (unless that is necessary for that test). I actually took this from the `test_code_set_txs.py` (EIP 7702), but looking there again I think that logic needs to be there because of the contract creations? :thinking: 

Would the logic still be necessary in `test_tstore_clear_after_deployment_tx` since that deploys a contract?

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
marioevz: Yes that is correct, only because of EOF.

tests/cancun/eip1153_tstore/test_tstorage_clear_after_tx.py
marioevz: > Ok, so I can remove `Op.STOP` and then just remove the `evm_code_type` switch logic?

Yes, `Op.STOP` is optional because `pre.deploy_contract` detects that the code is not `Container`, it's `Bytecode`, so before automatically wrapping, it checks for the `terminating` flag of the bytecode, and if it's false, it appends the `Op.STOP`.

> I think I actually copied this logic from another test, so this logic (not automatically switching to EOF) is somewhere else (unless that is necessary for that test). I actually took this from the `test_code_set_txs.py` (EIP 7702), but looking there again I think that logic needs to be there because of the contract creations? 🤔
> 
> Would the logic still be necessary in `test_tstore_clear_after_deployment_tx` since that deploys a contract?

Yes I think that's the reason,  if we are dealing with initcode, it needs the switch because we don't handle the `evm_code_type` flag automatically inside the `Initcode` class as we do with `pre.deploy_contract`.

jochem-brouwer: Have removed the EOF code switch logic in this specific test.

jochem-brouwer: I will merge once the CI passes :)

jochem-brouwer: Side-question, why does this PR stay approved even since I just pushed a commit? Shouldn't this require a re-review? I could now put arbitrary code here and merge it to `main`? :thinking:

jochem-brouwer: Oh wait, nvm, I cannot merge it myself

PR #799, opened 2024-09-13
new(test): EIP-7702 + EIP-1153: test that TransientStorage stays at correct address

jochem-brouwer:
## 🗒️ Description
This PR tests interaction between EIP-7702 (Set EOA account code) and EIP-1153 (transient storage). 

Diagram zonder titel drawio

The call chain contract will call into the EOA, which will then SSTORE in slot 1 the TLOADed value of transient slot 2, and then TSTORE the value 3 in transient slot 2. Then, it will return to the call chain contract and it will now call into the account which the EOA is delegated to and thus re-run the same code. The transient storage should be empty, thus the storage should stay empty.

This is parametrized by:

- It will perform both CALL and STATICCALL from the chain call contract EDIT: I should remove STATICCALL since we cannot SSTORE
- The contract which is called first (EOA or the code contract) is also switched

TODO: cleanup

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/tests/prague/eip7702_set_code_tx/test_set_code_txs.py
+++ b/tests/prague/eip7702_set_code_tx/test_set_code_txs.py
@@ -378,12 +378,9 @@ def test_set_code_to_tstore_reentry(
     )
 
 
-@pytest.mark.parametrize(
-    "call_opcode",
-    [
-        Op.CALL,
-        Op.STATICCALL,
-    ],
+@pytest.mark.with_all_call_opcodes(
+    selector=lambda call_opcode: call_opcode
+    not in [Op.DELEGATECALL, Op.CALLCODE, Op.STATICCALL, Op.EXTDELEGATECALL, Op.EXTSTATICCALL]
 )
 @pytest.mark.parametrize("call_eoa_first", [True, False])
 def test_set_code_to_tstore_available_at_correct_address(
@@ -410,13 +407,8 @@ def test_set_code_to_tstore_available_at_correct_address(
     set_code_to_address = pre.deploy_contract(tstore_check_code)
 
     def make_call(call_type: Op, call_eoa: bool) -> Bytecode:
-
         call_target = auth_signer if call_eoa else set_code_to_address
-        if call_type == Op.STATICCALL:
-            return call_type(Op.GAS(), call_target, 0, 0, 0, 0)
-        else:
-            # call_type = Op.CALL
-            return call_type(Op.GAS(), call_target, 0, 0, 0, 0, 0)
+        return call_type(address=call_target)
 
     chain_code = make_call(call_type=call_opcode, call_eoa=call_eoa_first) + make_call(
         call_type=call_opcode, call_eoa=not call_eoa_first
--- a/tests/prague/eip7702_set_code_tx/test_set_code_txs.py
+++ b/tests/prague/eip7702_set_code_tx/test_set_code_txs.py
@@ -379,6 +379,81 @@ def test_set_code_to_tstore_reentry(
 
 
 @pytest.mark.parametrize(
+    "call_opcode",
+    [
+        Op.CALL,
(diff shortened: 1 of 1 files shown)


tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -378,6 +378,81 @@ def test_set_code_to_tstore_reentry(
...
+    set_code_to_address = pre.deploy_contract(tstore_check_code)
+
+    def make_call(call_type: Op, call_eoa: bool) -> Bytecode:
+
+        call_target = auth_signer if call_eoa else set_code_to_address
+        if call_type == Op.STATICCALL:
+            return call_type(Op.GAS(), call_target, 0, 0, 0, 0)
+        else:
+            # call_type = Op.CALL
+            return call_type(Op.GAS(), call_target, 0, 0, 0, 0, 0)
```
marioevz, line 419:
```suggestion
        return call_type(address=call_target)
```
This is equivalent.

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -378,6 +378,81 @@ def test_set_code_to_tstore_reentry(
     )
 
 
+@pytest.mark.parametrize(
+    "call_opcode",
+    [
+        Op.CALL,
+        Op.STATICCALL,
+    ],
+)
```
marioevz, line 387:
```suggestion
@pytest.mark.with_all_call_opcodes(selector=lambda call_opcode: call_opcode not in [Op.DELEGATECALL, Op.CALLCODE, Op.EXTDELEGATECALL])
```

Should be equivalent and should automatically fill the test with EOF call opcodes when they are enabled.

marioevz (commented):
Thanks! Some suggestions.

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
jochem-brouwer: Sorry this param should be removed, in this context anything besides CALL makes no sense, since STATICCALL cannot SSTORE and all other opcodes will run in the context (i.e. where to write any store to) of the call chain contract address, which is not the purpose of this test (will remove this `call_opcode` thing).

Nevertheless, a good note to know for the future to select "all but not" :smile:

jochem-brouwer: I have added this, it now should select CALL and EXTCALL. Should I also add rules for when it runs in EOF mode? Because I think it now will deploy legacy code when it expects an EOF container? (For EXTCALL?)

marioevz: > Nevertheless, a good note to know for the future to select "all but not" 😄

Yes and the reasoning behind this (instead of directly selecting the opcodes) is that the test might fail on newer forks where another `*CALL*` opcode gets added, but this is a good thing because it prompts to update the tests.

marioevz: > I have added this, it now should select CALL and EXTCALL. Should I also add rules for when it runs in EOF mode? Because I think it now will deploy legacy code when it expects an EOF container? (For EXTCALL?)

I just tried it with the `PragueEIP7692` fork and filling with besu (supports 7702+EOF) and it worked flawlessly as is right now, so no changes needed :100:

marioevz: Just to elaborate a bit further this is because `pre.deploy_contract` automatically detects (1) the code being non-terminating `Bytecode` and appending the `Op.STOP`, and (2) the `evm_code_type` fixture automatically modifying the behavior of the contract deployment.

marioevz (approved):
Awesome, thanks!

PR #807, opened 2024-09-16
feat(hint): post state key hint [IDEA]

winsvega:
## 🗒️ Description
Add a hint dictionary for exceptions when generating the test. 
Before:
```
E   ethereum_test_base_types.composite_types.Storage.KeyValueMismatch: incorrect value in address 0x0000000000000000000000000000000000001200 (runner_contract) for key 0x0000000000000000000000000000000000000000000000000000000000000001: want 0x3 (dec:3), got 0x2 (dec:2)
```

After
```
src/ethereum_test_base_types/composite_types.py:261: in must_be_equal
    raise Storage.KeyValueMismatch(
E   ethereum_test_base_types.composite_types.Storage.KeyValueMismatch: incorrect value in address 0x0000000000000000000000000000000000001200 (runner_contract) for key SIMPLE_CALL: want 0x3 (dec:3), got 0x2 (dec:2)
```

Possible improvements:
make storage keyvalue a class like Address so it can have hint label? 
pass hint map for each address, each storage to convert sotrage keys into labels?

## 🔗 Related Issues
NONE

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_base_types/composite_types.py
+++ b/src/ethereum_test_base_types/composite_types.py
@@ -2,7 +2,7 @@
 Base composite types for Ethereum test cases.
 """
 from dataclasses import dataclass
-from typing import Any, ClassVar, Dict, SupportsBytes, Type, TypeAlias
+from typing import Any, ClassVar, Dict, Optional, SupportsBytes, Type, TypeAlias
 
 from pydantic import Field, PrivateAttr, RootModel, TypeAdapter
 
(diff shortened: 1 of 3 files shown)


src/ethereum_test_base_types/composite_types.py
```diff
@@ -92,22 +92,37 @@ class KeyValueMismatch(Exception):
...
 
         def __str__(self):
             """Print exception string"""
             label_str = ""
             if self.address.label is not None:
                 label_str = f" ({self.address.label})"
+
+            key = Hash(self.key)
+            if self.hint is not None:
+                key = self.hint[self.key]
```
winsvega, line 121: need to revert to default value of Hash(self.key) if value is not found

marioevz:
This is a nice idea, we could even end up exporting the address tags and storage key hints to the fixtures in order for client consumers to also use this metadata.

One thing that might be a bit cumbersome is that the tester now has to create and maintain the `post_hint` dictionary, and my fear is that with time it could even get out of sync (e.g. you add a new key to your test for another verification and increment all keys but forget to update the `post_hint`).

I have an idea from the top of my head on how we could make it automatic:

Add the key hint dictionary to the [`Storage`](https://github.com/ethereum/execution-spec-tests/blob/c0a381c2c7f8bcfabb122e14f2ad8a68deeaa3ad/src/ethereum_test_base_types/composite_types.py#L19) class, and allow users to pass the hint as a parameter to [`store_next`](https://github.com/ethereum/execution-spec-tests/blob/c0a381c2c7f8bcfabb122e14f2ad8a68deeaa3ad/src/ethereum_test_base_types/composite_types.py#L180). Caveat is that not all tests use this pattern, but it could be a way to encourage it.

We could also add some magic to the [`__setitem__`](https://github.com/ethereum/execution-spec-tests/blob/c0a381c2c7f8bcfabb122e14f2ad8a68deeaa3ad/src/ethereum_test_base_types/composite_types.py#L125) method like we do in the [`pre.deploy_contract`](https://github.com/ethereum/execution-spec-tests/blob/c0a381c2c7f8bcfabb122e14f2ad8a68deeaa3ad/src/pytest_plugins/filler/pre_alloc.py#L179-L189) function to get some information of the context where the key is being set and automatically obtain a hint in some of the cases.

winsvega:
surpassed by 917

Issue #184, opened 2023-07-01
Template scenarious for 'X' actions


winsvega:
about the x.  here is the idea. 
x is a set of actions that we want to test. Y is a vector of post conditions for F(x)=Y. we need a template that will take x
and execute it in different context (create tests). For each context i we need to verify that Fi(x)=Yi (transition for set of actions x on a template i makes post Yi)
The contexts are templates (but not limited to):
```
X is in transaction deployment code (create transaction)
X is in CREATE init code 
X is in CREATE2 init code 

X is in the contract code (being called by transaction)   and [OOG, REVERT, SELFDESTRUCT, NOTHING] happens
X is in the contract code (being called by CALL) and [OOG, REVERT, SELFDESTRUCT, NOTHING] happens
X is in the contract code (being called by CALLCODE) and [OOG, REVERT, SELFDESTRUCT, NOTHING] happens
X is in the contract code (being called by DELEGATECALL) and [OOG, REVERT, SELFDESTRUCT, NOTHING] happens
X is in the contract code (being called by STATICCALL) and [OOG, REVERT, SELFDESTRUCT, NOTHING] happens
```

winsvega:
Not just descriptions but actual test coverage.

For example x is a call to a newly defined opcode. Need to check its behavior by a template in all this cases.

winsvega:
Ori did some work in that direction:
https://github.com/ethereum/tests/blob/develop/src/Templates/DiffPlaces/templateGen.js

winsvega:
I hope I can merge it some day. it is almost ready. 
the Idea I got while implementing this is that I can create a generator that can be used as a fixture. 
so in the test it can take   pre, bytecode  and deploy contracts that simulate test scenario for a given bytecode. 

but first the original test scenarios PR

marioevz:
Fixed by #808

PR #808, opened 2024-09-16
feat(test): test opcode programs in different scenarios

winsvega:
## 🗒️ Description
Conversion of opcode diff places tests by ori. 
A test defines series of test scenarios that are run on each parametrized opcode sequence. 
Then we check if the sequence worked as expected in a given scenario. 

I think this is a powerful method to template test any new given opcode. 
we already have pre defined scenarios. then we just add one more parameter with what we want to test, and it will be covered on all the cases automatically. 

Cases can be like: 
callcode->staticcall-> [opcode]
create2-> [opcode]
[opcode] -> revert

check it out. so we can define opcode programs and scenarios. then the test will put each opcode program in each scenario and verify that it's result is the same (perhaps result will be complex depending on context and fork) 

the idea is so far to have a template test and then we can easily just add opcode programms and it will be run in all crazy combinations. 
likce call delegate call suicide revert    and so on 

This is still WIP. 

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/184

## ✅ Checklist


chfast (commented):
Can you give it any better name? The "opcode diff places" has no meaning.

winsvega:
it translates as opcode in different (logical) places

tests/frontier/scenarios/programs/all_frontier_opcodes.py
```diff
@@ -0,0 +1,81 @@
+"""
+Define a program for scenario test that executes all frontier opcodes and entangles it's result
+"""
+import pytest
+
+from ethereum_test_tools import Bytecode
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+
+def make_all_opcode_program() -> Bytecode:
```
marioevz, line 10: I think this could be a fixture and then it could be that we add opcodes as they become active through the forks:
```suggestion
def operation_result(sender: EOA, fork: Fork) -> Tuple[Bytecode, int]:
```
The return value would become:
```python
return (code, result)
```
where result, as of now, is `0x5831F0D814F4B8434FFDCE4DD24B00D8B7A3F67F8C316EC51A4D9EEC535AD4A`.

But then you can check for opcodes in `fork.valid_opcodes()`, and start appending code at the end (or rather before the `Op.RETURN(0, 32)`) and mutate the expected result:

```python
    result = 0x5831F0D814F4B8434FFDCE4DD24B00D8B7A3F67F8C316EC51A4D9EEC535AD4A  # As of Frontier
    if Op.PUSH0 in fork.valid_opcodes():
        code += Op.MSTORE(0, Op.SUB(Op.MLOAD(0), Op.PUSH0))
        result -= 0
    code += Op.RETURN(0, 32)
    return (code, result)
```

The `sender` parameter you can use it to predict the value of `Op.ORIGIN`.

One last thing is that you would have to move this function to either `test_scenarios.py` or a `conftest.py` file (so it would be picked up by any `test_*.py` file in the `scenarios` folder.

marioevz (commented):
A few suggestions after a quick review. I haven't checked files in the `./scenarios/ scenarios/` folder, but will do on the re-review.

tests/frontier/scenarios/programs/all_frontier_opcodes.py
winsvega: I made a return parameter system with examples how it will look like with verification of 
```
address
caller
origin
selfbalance
```
opcodes

now it gets tricky as each scenario must compute this opcodes for test validation

tests/frontier/scenarios/test_scenarios.py
```diff
@@ -0,0 +1,261 @@
...
+        program_basefee,
+        program_blobhash,
+        program_blobbasefee,
+        program_tload,
+        program_mcopy,
+        program_push0,
+        program_suicide,
+    ],
+)
+def test_scenarios(
```
marioevz, line 185:
```suggestion
@pytest.mark.execute(pytest.mark.skip(reason="gas usage"))
def test_scenarios(
```
This test has to be skipped in execute mode because it uses 500 million gas.

tests/frontier/scenarios/test_scenarios.py
```diff
@@ -0,0 +1,261 @@
...
+            result_slot: 0xFFFF,
+        },
+    )
+
+    tx = Transaction(
+        sender=tx_origin,
+        gas_limit=500_000_000,
+        gas_price=tx_gasprice,
+        to=runner_contract,
+        data=b"0x11223344",
```
marioevz, line 248: I think this is what you meant:
```suggestion
        data=b"\x11\x22\x33\x44",
```
The current one creates a byte array with the ascii equivalent of the "0x11223344" string.

tests/frontier/scenarios/programs/invalid_opcodes.py
```diff
@@ -0,0 +1,48 @@
...
+invalid_opcode_ranges = [
+    range(0x0C, 0x10),
+    range(0x1E, 0x20),
+    range(0x21, 0x30),
+    range(0x4B, 0x50),
+    range(0xA5, 0xF0),
+    range(0xF6, 0xFA),
+    range(0xFB, 0xFD),
+    range(0xFE, 0xFF),
+]
```
marioevz, line 19: I feel like this should not be a static list, and rather should be derived from the fork properties.

marioevz (commented):
More comments.

I feel like the programs should be dynamic based on the fork, and the approach could be improved in the sense that the program could be a `pytest.fixture` that takes the fork as input and then resolves to the actual bytecode.

See the comment in the `invalid_opcodes.py` file regarding the list of opcodes being invalid, since that is going to be changed frequently depending on the fork, it's going to be a pain to maintain, but we have the tools in pytest to do the maintenance more automatic and seamless.

winsvega:
> `pytest.fixture`

can you explain this part a little. 
currently the program is just a bytecode that is deployed and executed inside a dynamic address that is created by scenario. depending on scenario this address or rather execution context is prepared and executed. with this we achieve following model: define bytecodeX and execute it in different contextsY.

tests/frontier/scenarios/programs/invalid_opcodes.py
winsvega: yes this can be built depending on fork

tests/frontier/scenarios/test_scenarios.py
winsvega: is there a handy constructor like  BYTES("0x11223344")  ?

tests/frontier/scenarios/test_scenarios.py
winsvega: should not be that much. I need to correct

tests/frontier/scenarios/test_scenarios.py
```diff
@@ -0,0 +1,261 @@
...
+
+    call_combinations = scenarios_create_combinations(combinations_input)
+    for combination in call_combinations:
+        if not debug.scenario_name or combination.name == debug.scenario_name:
+            scenarios_list.append(combination)
+
+    revert_combinations = scenarios_revert_combinations(combinations_input)
+    for combination in revert_combinations:
+        if not debug.scenario_name or combination.name == debug.scenario_name:
+            scenarios_list.append(combination)
```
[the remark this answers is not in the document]
winsvega: here generate_X  is dynamic. 
so ideally we iterate over array of functions and call each function with environment conditions. 
and when we want to add new scenario generator it adds a function into that list

tests/frontier/scenarios/scenarios/call_combinations.py
```diff
@@ -0,0 +1,301 @@
...
+        """Definition of values we use to put in contract balances and call."""
+
+        root_call_value = 1
+        first_call_value = 3
+        second_call_value = 5
+
+        root_contract_balance = 105
+        scenario_contract_balance = 107
+        sub_contract_balance = 111
+        program_selfbalance = 113
```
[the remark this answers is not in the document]
winsvega: here I calculate manually how much the selfbalance is supposed to be in this context. and pass it as expected result to verifier. if program return different value while checking selfbalance there will be an error
this is equvivalent to post state defenition

tests/frontier/scenarios/test_scenarios.py
danceratopz: `bytes.fromhex()`, example here: https://github.com/ethereum/execution-spec-tests/pull/1067#discussion_r1910121660.

winsvega: I see here is confusion
`bytes.fromhex("11223344"),`
prefix 0x is not allowed, can mistake with dec

tests/frontier/scenarios/programs/invalid_opcodes.py
winsvega: ah, but the programs does not know about fork. I would have to refactor all programs to take fork as an argument.

tests/frontier/scenarios/test_scenarios.py
winsvega: can make a test per scenario. right now all scenarios combined in one test for each program. 
that would be a lot of test files though

tests/frontier/scenarios/programs/invalid_opcodes.py
winsvega: hm I can make this a predeployed contract in scenarios if you concern so much. but making programs a fixtures that accept fork is rather difficult design decision. programs are intended as byte sequences that are to be deployed in different scenarios and tested

winsvega: fixed invalid program

winsvega:
Since it is unreasonable to have all scenarios in one transaction (millions of gas and huge debug trace) 
I now put 1 scenario into each separate block. making it a blockchain test.
(it is impossible to generate 1 state test per each scenario as pytest collect fixtures at compile time and scenarios depend on fork dynamically)

so here what it does.
for each user created byte sequence (program) 
puts it into user created scenarios  (like call -> callcode -> program, or create -> init code program, or call->program, suicide) 
and verifies the expected result.  

so if we test something like selfbalance == actual selfbalance, we create bytecode
``
Op.MSTORE(0, Op.SELFBALANCE) + Op.RETURN(0, 32),
``
and the result
```
ProgramResult(result=ScenarioExpectOpcode.SELFBALANCE, from_fork=Istanbul),
```

then scenarios will make sure that this bytecode return value is equal to scenario expected selfbalance (which is calculated at scenario code) for each scenario situation.  like when we call  delegatecall->program,   delegatecall->call->program
this way we automatically test selfbalance opcode in all call combinations. 
and inside create/create2 init code  (this are basic examples of possible scenarios) 

then we can add a scenario of eof code.  and it will verify program value inside eof code. 
automatically for each program. 

this was one of the first ethereum tests. a simple verification that a bytecode like add(1,1)   really returns 2 in all contexts.

tests/frontier/scenarios/common.py
```diff
@@ -0,0 +1,302 @@
...
+        if res.result == ScenarioExpectOpcode.GASLIMIT:
+            return exec_env.gaslimit
+        if res.result == ScenarioExpectOpcode.SELFBALANCE:
+            return int(env.selfbalance)
+    else:
+        return res.result
+    return 0
+
+
+def replace_special_calls_in_operation(
```
marioevz, line 183: I think we should not use patterns like this in my opinion, we are basically making code introspection/morphing code we produced ourselves.

We should generate the correct code from the start.

Perhaps make the `operation` a callable code generator or just use fixtures.

tests/frontier/scenarios/common.py
```diff
@@ -0,0 +1,302 @@
...
+
+    name: str
+    code: Address
+    env: ScenarioEnvironment
+    halts: bool = False
+
+
+def translate_result(
+    res: ProgramResult, env: ScenarioEnvironment, exec_env: ExecutionEnvironment
+) -> int:
```
marioevz, line 142:
```suggestion
) -> int | Address:
```
So there won't be any need to transform the addresses to `int`.

tests/frontier/scenarios/common.py
```diff
@@ -0,0 +1,302 @@
...
+
+    """
+
+    name: str
+    code: Address
+    env: ScenarioEnvironment
+    halts: bool = False
+
+
+def translate_result(
```
marioevz, line 140: This can be a method of  `ProgramResult`:
```python
@dataclass
class ProgramResult:
    ...
    def translate_result(
        self, env: "ScenarioEnvironment", exec_env: "ExecutionEnvironment"
    ) -> int | Address:
        """
        Translate expected program result code into concrete value,
        given the scenario evm environment and test execution environment.
        """
```

tests/frontier/scenarios/common.py
```diff
@@ -0,0 +1,302 @@
...
+                if_true=Op.RETURN(0, 0),
+                if_false=Op.JUMP(5),
+            ),
+            if_false=Op.MSTORE(32, Op.SUB(Op.MLOAD(32), 1)) + Op.JUMP(14),
+        )
+    )
+    return gas_hash_address
+
+
+def make_invalid_opcode_contract(pre: Alloc, fork: Fork) -> Address:
```
marioevz, line 262: I think we can make this its own standalone test in a different .py file and just parametrize a test function with all invalid opcodes, instead merging everything into a single bytecode.

tests/frontier/scenarios/common.py
```diff
@@ -0,0 +1,302 @@
...
+            )
+            operation = Bytecode(
+                bytes_or_byte_code_base=to_bytes(new_operation_hex),
+                popped_stack_items=0,
+                pushed_stack_items=0,
+            )
+    return operation
+
+
+def make_gas_hash_contract(pre: Alloc) -> Address:
```
marioevz, line 236: This function can be a fixture, we should create `conftest.py` and put many of the functions there as fixtures.

tests/frontier/scenarios/common.py
```diff
@@ -0,0 +1,302 @@
...
+    So that if we can't check exact value in expect section,
+    we at least could spend unique gas amount.
+    """
+    gas_hash_address = pre.deploy_contract(
+        code=Op.MSTORE(0, 0)
+        + Op.JUMPDEST
+        + Op.CALLDATACOPY(63, Op.MLOAD(0), 1)
+        + Op.JUMPDEST
+        + Conditional(
+            condition=Op.EQ(Op.MLOAD(32), 0),
```
marioevz, line 249:
```suggestion
            condition=Op.ISZERO(Op.MLOAD(32)),
```

tests/frontier/scenarios/test_scenarios.py
```diff
@@ -0,0 +1,258 @@
...
+            if hasattr(debug_program, "hex") and operation.hex() != debug_program.hex():
+                continue
+
+        post_storage = Storage()
+        result_slot = post_storage.store_next(1, hint=f"runner result {scenario.name}")
+
+        tx_max_gas = (
+            7_000_000
+            if hasattr(program_invalid.values[0], "hex")
+            and operation.hex() == program_invalid.values[0].hex()
```
marioevz, line 208: Just a sidenote, generally to the code in the PR, there is no need to convert everything to strings and work using strings, bytes can be used almost everywhere and should be faster.

tests/frontier/scenarios/test_scenarios.py
```diff
@@ -0,0 +1,258 @@
...
+            debug_program = debug.test_param.values[0]
+            if hasattr(debug_program, "hex") and operation.hex() != debug_program.hex():
+                continue
+
+        post_storage = Storage()
+        result_slot = post_storage.store_next(1, hint=f"runner result {scenario.name}")
+
+        tx_max_gas = (
+            7_000_000
+            if hasattr(program_invalid.values[0], "hex")
```
marioevz, line 207: This is a side effect of using `pytest.param` to define the programs, it could be that they require its own class definition:
```
class Program:
    gas: int | None = None  # None means default
    ...
```

marioevz (changes requested):
I think there are a lot of anti-patterns used here, each program can be a separate python test file to make them simpler and more atomic.

The problem with putting so many verifications into a single test is that it's going to be difficult to client developers to debug in the case they find issues with one of these tests.

tests/frontier/scenarios/common.py
winsvega: This is because scenario program does not have a constructor, so it uses cistom system contacts

tests/frontier/scenarios/common.py
winsvega: it is now a possible import for test code generators (ScenarioTestProgram class)

winsvega:
Missing coverage 3 lines are `Prague evm init` and state test loader check for a storage of pre.account having value of x:0.

marioevz (approved):
I've rebased and deleted the static test files that this PR supersedes.

I also disabled this test in `executed` because it requires accurate value of `Op.GASPRICE` which is unpredictable in devnets (we need a way to pass this to the tests somehow perhaps via a fixture).

marioevz:
Verified the coverage files:
- `case EVMC_PRAGUE` is running the case in Prague, which the coverage script does not fill for.
- `if (is_zero(value))` is not covered because the new tests do not pre-set the storage of the contracts it deploys.

winsvega:
> I've rebased and deleted the static test files that this PR supersedes.
> 
> I also disabled this test in `executed` because it requires accurate value of `Op.GASPRICE` which is unpredictable in devnets (we need a way to pass this to the tests somehow perhaps via a fixture).

I am aming ro make it execute compatible.
I have gashash contract to solve that. 
If gasprice is unpredictable I call gashash contract with the value of op.gasprice that shall produce unique gas used so that others running the test can agree on gas used value without tests knowing which value is it

PR #810, opened 2024-09-16
 fix(fw): max stack height calculation in __add__ 

pdobacz:
## 🗒️ Description

Split out of #785 with a unit test added.

I've noticed that the algorithm might not handle all cases well, and I am proposing an alternative. I've left all my overly verbose comments in the code to illustrate the train of thought, as this has been quite hard to grasp the original code intentions (maybe there has been an easier fix hidden from my sight there).

## 🔗 Related Issues

n/a

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_vm/tests/test_vm.py
+++ b/src/ethereum_test_vm/tests/test_vm.py
@@ -383,6 +383,21 @@ def test_macros():
         pytest.param(
             Op.POP(Op.CALL(1, 2, 3, 4, 5, 6, 7)), 0, 0, 7, 0, id="POP(CALL(1, 2, 3, 4, 5, 6, 7))"
         ),
+        pytest.param(
+            Op.PUSH0 * 2 + Op.PUSH0 + Op.ADD + Op.PUSH0 + Op.POP * 2, 0, 1, 3, 0, id="parens1"
+        ),
+        pytest.param(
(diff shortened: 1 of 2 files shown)


src/ethereum_test_vm/tests/test_vm.py
```diff
@@ -383,6 +383,21 @@ def test_macros():
         pytest.param(
             Op.POP(Op.CALL(1, 2, 3, 4, 5, 6, 7)), 0, 0, 7, 0, id="POP(CALL(1, 2, 3, 4, 5, 6, 7))"
         ),
+        pytest.param(
```
pdobacz, line 386: I added 5 cases, all the latter 4 ones fail because of the parenthesis. I'm not sure exactly which one of them is the most minimal, but I think it's reasonable to expect all of them to yield the same result, so let's test it.

I tried but couldn't reduce it further by taking away the instructions, but also didn't spend too much time on it, I think this is minimal enough.

Also, the original problem I encountered was actually mismatching and incorrect max stack height. It turns out that only the last two of these parenthesisations lead to max height also being invalid

marioevz (approved):
This is great, thanks so much for implementing this. I think the concept I was missing is the stack-bottom, it makes it easier to follow this implementation.

I verified the resulting fixtures and they are exactly the same, so it's safe to merge.

PR #812, opened 2024-09-16
new(tests): EOF - EIP-7620: Dangling data in subcontainer test

shemnon:
Signed-off-by: Danno Ferrin <danno@numisight.com>

## 🗒️ Description
Test dangling data inside a subcontainer

## ✅ Checklist

### DIFF
--- a/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
+++ b/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
@@ -758,27 +758,6 @@ def test_migrated_eofcreate(eof_test: EOFTestFiller, container: Container):
     eof_test(data=container, expect_exception=container.validity_error)
 
 
-def test_dangling_subcontainer_bytes(
-    eof_test: EOFTestFiller,
-):
-    """EOF Subcontainer test with subcontainer containing dangling bytes."""
-    eof_test(
-        data=Container(
-            sections=[
-                returncontract_code_section,
-                Section.Container(
-                    custom_size=len(stop_sub_container.data),
-                    container=Container(
-                        raw_bytes=stop_sub_container.data + b"\x99",
-                    ),
-                ),
-            ],
-            kind=ContainerKind.INITCODE,
-        ),
-        expect_exception=EOFException.INVALID_SECTION_BODIES_SIZE,
-    )
-
-
 def test_dangling_initcode_subcontainer_bytes(
     eof_test: EOFTestFiller,
 ):
--- a/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
+++ b/tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
@@ -756,3 +756,66 @@ def test_migrated_eofcreate(eof_test: EOFTestFiller, container: Container):
     Tests migrated from EOFTests/efValidation/EOF1_eofcreate_valid_.json.
     """
     eof_test(data=container, expect_exception=container.validity_error)
+
+
+def test_dangling_subcontainer_bytes(
+    eof_test: EOFTestFiller,
+):
+    """EOF Subcontainer test with subcontainer containing dangling bytes."""
+    eof_test(
+        data=Container(
+            sections=[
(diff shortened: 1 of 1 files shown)


shemnon:
Covers this fuzzing find - https://github.com/ipsilon/eof/issues/146#issuecomment-2353549917

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
```diff
@@ -756,3 +756,65 @@ def test_migrated_eofcreate(eof_test: EOFTestFiller, container: Container):
...
+                    ),
+                ),
+            ],
+            kind=ContainerKind.INITCODE,
+        ),
+        expect_exception=EOFException.INVALID_SECTION_BODIES_SIZE,
+    )
+
+
+def test_dangling_initcode_subcontainer_bytes(
```
pdobacz, line 782: I can't find a difference vs the one above

shemnon: cmd+D one too many times.  I'll delete the first.

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
```diff
@@ -756,3 +756,44 @@ def test_migrated_eofcreate(eof_test: EOFTestFiller, container: Container):
...
+    eof_test: EOFTestFiller,
+):
+    """Initcode mode EOF Subcontainer test with subcontainer containing dangling bytes."""
+    eof_test(
+        data=Container(
+            sections=[
+                returncontract_code_section,
+                Section.Container(
+                    custom_size=len(stop_sub_container.data),
```
chfast, line 770: I think this is different from the issue found by the fuzzer: there the section size (21, 0x15) matches the content, so the error is only in the subcontainer itself. So maybe parametrize this detail?

chfast (commented):
Have you checked if geth is failing these tests? I believe it only should fail the initcode variant.

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
shemnon: You're right, fixed.

What would be parameterized?  I'm afraid if we parameterize it the failure we are probing for will be missed.

shemnon:
Interestingly, geth only fails the (corrected) initcode form of the test, and correctly marks the runtime mode of the test as invalid.

chfast:
> Interestingly, geth only fails the (corrected) initcode form of the test, and correctly marks the runtime mode of the test as invalid.

This agrees with the fuzzer. I believe this is because it passes "initcode" mode to subcontainer validation and maybe this incorrectly allows additional bytes (as in create transaction).

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
chfast: param value 1: as you have written it originally
param value 2: as the fuzzer fount it (current version)

chfast: I.e. with and without `custom_size`.

shemnon:
Here's the fix, explains the initcode only failure - https://github.com/MariusVanDerWijden/go-ethereum/pull/56/commits/4392a1c4e6d4ccc98f4e773a7d8a009b8ccd143e

I haven't dug in deeply but I think this ties in with Geth's handling of create transactions dangling data becoming calldata, so adding the toplevel flag is all that's needed

tests/prague/eip7692_eof_v1/eip7620_eof_create/test_subcontainer_validation.py
shemnon: Parametrizing it that way just turns the custom size EOF into another basic container test, which exists as `test_simple_create_from_deployed` already.

chfast (approved):

marioevz (approved):
LGTM, thanks!

PR #801, opened 2024-09-16
feat(docs): re-write test gen doc flow with pytest plugin

danceratopz:
## 🗒️ Description

This is a rewrite of the automatic test case doc generation as a pytest plugin (as apposed to a standalone script). A preview is available at https://ethereum.github.io/execution-spec-tests/pr-801-preview.

### Doc Improvements
1. Test module pages: Added a table that provides and overview of its test functions, e.g., [tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g1add](https://ethereum.github.io/execution-spec-tests/pr-801-preview/tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g1add/).
2. Added test function pages which include a table which tries to give an overview of its parametrized test cases, e.g., [tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g1add::test_valid](https://ethereum.github.io/execution-spec-tests/pr-801-preview/tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g1add/test_valid/).
3. Added static html files for the tables of the function parameters overview, as they often don't play well within mkdocs styling. 
4. Markdown pages: Added a header with a path and ref link for the markdown file, e.g., [tests/prague/eip7692_eof_v1/tracker.md](https://ethereum.github.io/execution-spec-tests/pr-801-preview/tests/prague/eip7692_eof_v1/tracker/)

### Technical Improvements

1. Doc build fails if no function docstring is available or the test type (e.g. `state_test`) can not be determined (both used in the function overview tables on test module pages).
2. Using a pytest plugin allows us to lever all the usual pytest flags for test item filtering and fork specification (great for debugging) and to introspect test items to easily grab test metadata.
3. [jinja2](https://jinja.palletsprojects.com/en/3.1.x/) is used for templating, which helps separate the content from its formatting. Jinja is a very powerful templating library which helps save a lot of boilerplate code. 

## 🔗 Related Issues
Perhaps solved, plz check whether the styling is acceptable, e.g., [test_bls12_g1add/test_valid.html](https://ethereum.github.io/execution-spec-tests/pr-801-preview/tests/prague/eip7702_set_code_tx/test_gas/test_gas_cost.html).
- #804.

Current issues:
- #803.

Ideas for future improvements:
- #802 (not a problem now, but we can only gen docs using one dev fork).
- #805.
- #813.
- #814.
- #815.
- #816.

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
[...]

PR #829, opened 2024-09-23
fix(docs,ci): add build-evm-base to docs deploy workflows

danceratopz:
## 🗒️ Description

**#792 would also fix this; we can disregard this PR if #792 gets merged first/quickly**

#801 re-implements the "Test Case Reference" generation script as a pytest plugin (based on top of `fill`). As the forks plugin (and others; didn't check) require a `t8n` tool in the path, we need to provide one, even if it's not strictly necessary to build docs.

This is a tricky subject. I'm not a fan of requiring a t8n tool to collect tests/build doc, but if we use t8n properties in test markers or other places, it's not easy to avoid.

## 🔗 Related Issues
#801 

## ✅ Checklist
- [x] ~~All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~ Skipped

### DIFF
--- a/.github/workflows/docs_main.yaml
+++ b/.github/workflows/docs_main.yaml
@@ -16,6 +16,11 @@ jobs:
         with:
           fetch-depth: 0
           ssh-key: ${{secrets.GH_ACTIONS_DEPLOY_KEY}}
+      - name: Build transition tool
+        uses: ./.github/actions/build-evm-base
+        with:
+          id: evm-builder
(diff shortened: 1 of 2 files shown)


marioevz (approved):
LGTM, I think we might need to abstract away the evm-bin part in the plugins, it should be possible to remove this requirement.

PR #833, opened 2024-09-24
new(tests): EOF - EIP-5450: RJUMP* vs CALLF tests

pdobacz:
## 🗒️ Description

Adds a first batch of tests checking stack validation of these two kinds of operations RJUMP-ing and CALLF-ing (and RETF-ing ofc). I'm still missing JUMPF, but I want to go step by step as this has taken a while to prepare anyway.

First remark: I've decided to start off with an "automated" approach here, but as you'll see this has several complexities/inelegancies involved. There always is an alternative to just manually type out some interesting scenarios by hand, which we must bear in mind.

The problematic parts are:
  - it would be quite hard for all the automatically generated codes to come with a 100% accurate max_stack_height, so I introduce a flag `no_expecations_on_validity` to `eof_test`. This is probably the biggest controversy, and I'm also still unsold on this, let's discuss more in the diff. The nice part is that somewhat automatically we get to test invalid and valid codes en masse, without spending too much time on figuring out the exact cases.
  - it initially generated many many more tests, I've managed to trim this down aggressively to 400, however I took a - potentially unacceptable - shortcut and trimmed the "uninteresting" cases by `pytest.skip`. Without the `pytest.skip` it goes up to 800, all of the extra cases are ones which never successfully validate. All that being said, 400 is still a lot of cases, and I'd like to add JUMPF and recursive calls to this in the future iteration
  - it is generally complex and has an air of "throw mud at wall, see if it sticks" and maybe gets into the court of what fuzzing should accomplish. Again, the "good part" of this is that we're not effectively re-writting the stack validation routines in EEST

## 🔗 Related Issues

NA

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_specs/eof.py
+++ b/src/ethereum_test_specs/eof.py
@@ -142,6 +142,7 @@ class EOFTest(BaseTest):
     data: Bytes
     expect_exception: EOFExceptionInstanceOrList | None = None
     container_kind: ContainerKind | None = None
+    no_expectations_on_validity: bool = False
 
     supported_fixture_formats: ClassVar[List[FixtureFormats]] = [
         FixtureFormats.EOF_TEST,
@@ -228,7 +229,17 @@ class EOFTest(BaseTest):
             if vector.container_kind == ContainerKind.INITCODE:
                 args.append("--initcode")
             result = eof_parse.run(*args, input=str(vector.code))
-            self.verify_result(result, expected_result, vector.code)
+            if self.no_expectations_on_validity:
+                parser = EvmoneExceptionMapper()
+                actual_message = result.stdout.strip()
+                if "OK" in actual_message:
+                    expected_result.valid = True
+                else:
+                    expected_result.valid = False
+                    actual_exception = parser.message_to_exception(actual_message)
+                    expected_result.exception = actual_exception
+            else:
+                self.verify_result(result, expected_result, vector.code)
 
         return fixture
 
--- a/tests/prague/eip7692_eof_v1/eip5450_stack/__init__.py
+++ b/tests/prague/eip7692_eof_v1/eip5450_stack/__init__.py
@@ -1,0 +1,3 @@
+"""
+EOF tests for EIP-5450 stack validation
+"""
--- a/tests/prague/eip7692_eof_v1/eip5450_stack/test_code_validation.py
+++ b/tests/prague/eip7692_eof_v1/eip5450_stack/test_code_validation.py
@@ -1,0 +1,270 @@
+"""
+Code validation of CALLF, JUMPF, RETF opcodes in conjunction with static relative jumps
+"""
+
+import itertools
+from enum import Enum, auto, unique
+from typing import Tuple
+
+import pytest
+
+from ethereum_test_tools import EOFTestFiller
+from ethereum_test_tools.eof.v1 import Container, Section
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+from ethereum_test_vm.bytecode import Bytecode
+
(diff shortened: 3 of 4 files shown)


src/ethereum_test_specs/eof.py
```diff
@@ -228,7 +229,17 @@ def make_eof_test_fixture(
             if vector.container_kind == ContainerKind.INITCODE:
                 args.append("--initcode")
             result = eof_parse.run(*args, input=str(vector.code))
-            self.verify_result(result, expected_result, vector.code)
+            if self.no_expectations_on_validity:
```
pdobacz, line 232: I have an alternative/extension to this flag. I think it would be good that either:
1/ `no_expectations_on_validity` flag is combined with `expect_exception`, so that either the code is valid OR the exception is one of the expected ones. Then ALL of the `expect_exception` ones would make it into the fixture, so that clients are not restricted to a particular expected error. In the case of the RJUMP*/CALLF tests those expected exceptions would be the stack related ones
2/ remove the new flag and allow `None` to be an element of the `expect_exception` list thereby admitting that the code may validate successfully OR throw one of the exceptions in the list. This has the disadvantage of requiring the support of this by the clients test runners

The motivation behind would be to safe guard against the test suddenly generating code which is EOF-invalid for an unrelated cause (say, undefined instruction or whatever), making the whole thing less likely to break silently.

I wonder what be your preference (or neither option is OK or even the whole concept is not acceptable)?

I have an implementation of (1/), but I haven't yet fully investigated its consequences.

pdobacz: After sleeping on this I realized (1/) is both the better approach and necessary, [`Close the catalogue of potential exceptions`](https://github.com/ethereum/execution-spec-tests/pull/833/commits/e90423631387d382e370858d70da12ae9340b207) handles that, PTAL

pdobacz:
I pushed some updates after incorporating the addition from https://github.com/ipsilon/execution-spec-tests/pull/3. Unfortunately, I've noticed that we should expand the tests by a little, I went to far while trimming and cut out some cases without which some RJUMP* snippets never validated. Now we're at 768 cases mark :grimacing:.

src/ethereum_test_specs/eof.py
marioevz: I still think that (1/) is too loose in the sense that we could miss detecting an important change and the tests passing inadvertently, that's my main concern.

I think we should analyze how hard it would be to isolate the valid containers from the invalid ones in the parametrization.

pdobacz: Thanks for taking a look! Yea, I did some thinking about this and it seemed unfeasible / too complex. But, actually you gave me an idea here - maybe I've been trying too hard to make the produced code valid. What I could instead focus on is generate the code as it is and add some heuristics which could narrow down the catalogue even further. Like exception A can only happen if condition X is met, exception B if condition Y, etc. That would also isolate valid from invalid, like you suggest.

Maybe those 3 remaining exceptions can be "predicted" easily enough, lemme try that

pdobacz: @marioevz I pushed the version which "predicts" validity based on some traits of the code. I didn't yet think about refactoring those traits into some kind of class maybe, to avoid the massive ifs (also not sure about my python-fu here), but here it is, LMK if this looks better now.

pdobacz: @marioevz Did you have a chance to look at the changes? :point_up:

marioevz (approved):
Awesome, thanks for the changes, it's cleaner IMO.

This might need a rebase because now EOF tests are in the `osaka` folder.

chfast (approved):

PR #842, opened 2024-09-26
feat(docs): include all parameters in test parameter datatables

danceratopz:
## 🗒️ Description
The test function parameter tables introduced in #801 try to only show "relevant" test function parameters. It's a nice idea, but I think this obfuscates the fact that we always parametrize by fork and potentially fixture type, so I think it's better to be transparent about that but limit the number of entries displayed by using additional select elements with reasonable defaults.

Preview available at [pr-842-preview](https://ethereum.github.io/execution-spec-tests/pr-842-preview/).

Example:
image

## 🔗 Related Issues
- Follows-up from:
-  #801
Includes:
- #853 
- #871 

Done:
- #813 
- Generate test function parameters for all forks by running `fill` with `--until={GENERATE_UNTIL_FORK}` instead of `--fork=...`.
- Specify a target fork that can be used to display a default set of parameters (`--gen-docs-target-fork={TARGET_FORK}`.
- Tooltip upon hover on Test Case ID entry.
- Copy-paste the full test id from the table via left-click on Test Case ID entry.
- #814 (not yet, we could add a hidden table column to display make entries searchable by full nodide).
- Fix following bug (by using datables API in select elements in order to correctly update in the table entries?):
    - The selectors work and update the rows in the table, but:
        - If I search in the "Search" box, I get output like "
        Showing 1 to 20 of 20 entries (filtered from 48 total entries)", and if there's no need to paginate, the values are shown on the first page.
        - If I filter using fork or fixture type, the rows are filtered, but the output doesn't change (it stays at "Showing 1 to 20 of 20 entries" and the pagination doesn't update; in the worst case the table is empty and you have to click through the pages until you find where the values are in the table (two example screenshots attached for the same selector values).
- Arrange select elements more nicely, apply styling.

 

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [ ] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
[...]

PR #853, opened 2024-10-01
🐞 fix: Datatable custom filter and other goodies

raxhvl:
## 🗒️ Description
Fixes:
-  Issues with custom fork and test filters.
-  Resolved a bug causing the datatable to break when changing entries per page.

Enhancements:

-  Upgraded jQuery from 3.5.1 to 3.7.1.
-  Upgraded DataTables from 1.10.21 to 2.1.7.
-  Migrated to minified scripts / css where possible for faster page load.
-  Added a plugin for searchable dropdowns.

image

## 🔗 Related Issues
#842 

## ✅ Checklist
- [x] ~~All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.~~ **skipped**

### DIFF
--- a/docs/javascripts/site.js
+++ b/docs/javascripts/site.js
@@ -17,6 +17,9 @@ document$.subscribe(() => {
     // Listen for copy event
     listenForClipboardCopy();
   }
+
+  // Setup up select 2
+  $(`select${FILTER_INPUT_SELECTOR}`).select2();
 });
 
 const initDataTable = () => {
--- a/docs/javascripts/site.js
+++ b/docs/javascripts/site.js
@@ -1,5 +1,6 @@
 // Config
 const FILTER_INPUT_SELECTOR = ".custom_dt_filter";
+const FILTER_SEARCH_SELECTOR = "#custom_dt_search";
 let table;
 
 // This script is used both within mkdocs and in standalone html files.
@@ -19,6 +20,8 @@ document$.subscribe(() => {
   if (table) {
     // Listen for changes to filters
     $(FILTER_INPUT_SELECTOR).on("change", filterRows);
+
+    $(FILTER_SEARCH_SELECTOR).on("input", filterRows);
 
     // Apply preselected filters (if present) on page load
     filterRows();
@@ -55,9 +58,12 @@ const initDataTable = () => {
 const filterRows = () => {
   table
     .rows()
-    .search(function (a, b, index) {
+    .search(function (rowContent, b, index) {
       const row = $(table.row(index).node());
       let match = true;
+
+      const searchKeyword = $(FILTER_SEARCH_SELECTOR).val();
+      const searchHit = rowContent.includes(searchKeyword);
 
       for (let filter of $(FILTER_INPUT_SELECTOR)) {
         // Filter is ignored if set to all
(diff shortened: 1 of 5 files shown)


danceratopz:
Hey @raxhvl, thanks for this! We had some issues with CI yesterday due to some upstream changes. Let me sort that out in [docs/include-all-parameters-in-test-parameter-datatables](https://github.com/ethereum/execution-spec-tests/tree/docs/include-all-parameters-in-test-parameter-datatables).

raxhvl:
Just noticed that filters doesn't play nice with the search. Looking into it.

docs/javascripts/site.js
```diff
@@ -1,92 +1,101 @@
...
-}
+// Config
+const FILTER_INPUT_SELECTOR = ".custom_dt_filter";
+let table;
 
-// Function to attach event listeners to the selectors
-function attachFilterListeners() {
+// Subscribe to when a page is loaded in mkdocs.
+// see: https://github.com/squidfunk/mkdocs-material/issues/5816#issuecomment-1667654560
+document$.subscribe(() => {
```
danceratopz, line 7: This breaks the data table initialization in the standalone pages as `site.js` also gets included there:
https://github.com/raxhvl/execution-spec-tests/blob/0c2549d6fc42a59b6a338e3f8821da722ce95827/docs/templates/function.html.j2#L24

danceratopz (commented):
Thanks, this looks excellent so far! Much cleaner handling of the data table initialization. I really like the select styling. Just one small comment regarding the standalone html pages.

raxhvl:
@danceratopz I think you will have to merge main again. Now search works within filter which is nice. Plus small updates to the UI. I'm done unless you have feedback.

docs/templates/function_parameter_datatable.html.j2
```diff
@@ -1,22 +1,47 @@
...
-    {% endfor %}
-</select>
-
-<table id="test_table" class="display nowrap">
+<div class="control_panel">
+    <div class="panel_row filters">
+    <div class="filter_wrapper">
+        <label for="fork_selector">
+         <span class="twemoji">
+            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
```
danceratopz, line 6: Here's how this renders on the standalone page (Firefox):
image

danceratopz (changes requested):
Thanks @raxhvl! The tables within the mkdocs pages look wonderful - love it!

Just one comment on the standalone pages. If there's no easy fix, we could consider splitting `function_parameter_datatable.html.j2` into a base template that would contain almost the logic and two derived pages, one for mkdocs, one for standalone. The only difference between the two would be whether to include the SVG, or not.

raxhvl:
> Thanks @raxhvl! The tables within the mkdocs pages look wonderful - love it!
> 
> Just one comment on the standalone pages. If there's no easy fix, we could consider splitting `function_parameter_datatable.html.j2` into a base template that would contain almost all the logic and add two derived templates, one for mkdocs, one for standalone. The only difference between the two would be whether to include the SVG, or not.

It was a small fix - I copied over relevant css code from mkdocs to custom.css so it applies at both places now. Also, the standalone page was too wide and close to the browser boundary; I made it a little more compact - let me know if that looks okay

PR #854, opened 2024-10-01
fix(specs): Track block fork in edge cases

etan-status:
There are a few situations where a block object loses track of the originating fork, which show up when implementing SSZ EIPs (EIP-6404). Fix those instances by setting the fork where it was forgotten.


src/ethereum_test_specs/blockchain.py
```diff
@@ -482,6 +483,7 @@ def generate_block_data(
             # Modify any parameter specified in the `rlp_modifier` after
             # transition tool processing.
             header = block.rlp_modifier.apply(header)
+            header.fork = fork  # Deleted during `apply` because `exclude=True`
```
danceratopz, line 486:
```suggestion
            header.fork = fork  # Deleted during `apply` because `exclude_none=True`
```

danceratopz (approved):
LGTM, thanks for this! I think the comment below should just be `exclude_none` (the keyword argument to `model_dump()`?

src/ethereum_test_specs/blockchain.py
etan-status: https://github.com/ethereum/execution-spec-tests/blob/6e4c3a1119039bd79058b56c4b28564fb2ee090c/src/ethereum_test_fixtures/blockchain.py#L122

it's actually full `exclude=True` in this instance.

danceratopz: Got it, thanks!

PR #889, opened 2024-10-10
new(tests): EOF - EIP-4750: Stack  validation in CALLF

shemnon:
## 🗒️ Description

Add a test to cover cases where stack would not overflow but CALLF stack
reservation rules would cause a revert.

## ✅ Checklist

### DIFF
--- a/tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
+++ b/tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
@@ -386,24 +386,24 @@ def test_callf_max_stack(
         code=Container(
             sections=[
                 Section.Code(
-                    code=Op.PUSH0 * 4 # fill the stack up a little bit
+                    code=Op.PUSH0 * 4  # fill the stack up a little bit
                     + Op.PUSH2(stack_height)
                     + Op.CALLF[1]
                     + Op.SSTORE(slot_code_worked, value_code_worked)
                     + Op.RETURN(0, 0),
-                    max_stack_height=7
+                    max_stack_height=7,
                 ),
                 Section.Code(
-                    Op.PUSH1(1)     # arg, 1
-                    + Op.SWAP1      # 1, arg
-                    + Op.SUB        # arg-1,
-                    + Op.DUP1       # arg-1, arg-1
-                    + Op.CALLF[2]   # arg-1, arg-1
-                    + Op.ISZERO     # jump?, arg-1,
+                    Op.PUSH1(1)  # arg, 1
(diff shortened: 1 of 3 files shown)


tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
```diff
@@ -252,3 +257,102 @@ def test_callf_with_inputs_stack_overflow(
...
+                ),
+            }
+        )
+    }
+
+    tx = Transaction(
+        to=contract_address,
+        gas_limit=10_000_000,
+        gas_price=10,
+        protected=False,
```
pdobacz, line 355: gas_price and protected can be left out

tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
```diff
@@ -252,3 +257,102 @@ def test_callf_with_inputs_stack_overflow(
         ),
         container_post=Account(storage={slot_code_worked: 0}),
     )
+
+
+@pytest.mark.parametrize(
+    ("stack_height", "inputs", "outputs", "failure"),
+    (
+        pytest.param(1020, 1, 3, False, id="1020"),
+        pytest.param(1021, 1, 3, True, id="1021"),
```
pdobacz, line 266:
```suggestion
        pytest.param(1020, 1, 3, False, id="success"),
        pytest.param(1021, 1, 3, True, id="failure"),
```
if we just have these two now (do I understand correctly, the intention is that `1022` would not validate anymore? Or are there other combinations of params which would validate and make sense to test?

pdobacz (approved):
LGTM, minor comments. Do we need to also update `eof_tracker.md`? I'm not sure but maybe `Max stack size (1024) in CALLF-ed function` was covering this case, please double check though, I didn't analyze it

tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
shemnon: The essential test is pass below, and fail above.  I was considering messing with the input/output levels, but determined it was not essential to the test.  I took input and output of the params, and just have stack size and wether we expect failure.  Seemed to communicate intent better than embedding what a failure is into the text code.

shemnon:
Added another tests to check CALLF when stack is at max ("rule #4") - by EIP this should revert, but there are ways it would succeed without the check, so we do this.

pdobacz (approved):

marioevz (approved):
LGTM, thanks!

Issue #863, opened 2024-10-02
🔀(gentest): use jinja2 templates to structure python output


danceratopz:
[Jinja2](https://jinja.palletsprojects.com/en/3.1.x/) is a very powerful templating library that we started using in the doc flow for the generated "Test Case Reference" output pages, see #801 and `./docs/templates`.

I think that `.py.j2` templates could be an excellent way to write the output from `gentest`. This will be particularly helpful, when/if we create `blockchain_tests` with an arbitrary number of blocks, each containing an arbitrary number of transactions. We could have a template for a `state_test` and one for `blockchain_test`. Jinja2 also allows base templates.

raxhvl:
Hey @danceratopz I'm starting with this.

PR #892, opened 2024-10-14
✨ feat(all): add application-wide configuration manager

### LINKED ISSUE #891: ✨(gentest): allow custom headers in `gentest` config file and add an example config file
1. Instead of using specific fields like `client_id` and `secret`, I propose we change the config file structure to include a single key called rpc_headers. This allows users to define a flexible number of headers as needed.

```
{
    "remote_nodes": [
        {
            "name": "mainnet_archive",
            "node_url": "https://abc.com",
            "rpc_headers": {
                "Authorization": "Bearer your_token_here",
                "Another-Header": "value"
            }
        }
    ]
}
```
2. Create a `config.json.example` file that serves as a template for new users. This will help them quickly understand how to structure their configuration.

### LINKED ISSUE #893: ✨ feat(gentest): Migrate `gentest` specific configration from JSON to YAML
Currently `gentest` config is a JSON file. JSON configurations are verbose and does not support comments. At some point it could be benificial to [migrate to YAML.](https://github.com/ethereum/execution-spec-tests/pull/892#pullrequestreview-2366234308)

raxhvl:
# 🗒️ Description

This PR introduces a package to manage environment and application configurations.
```console
.
├── src
│   └── 📁 config  [Application wide environment and configurations]
│       ├── 📄 __init__.py
│       ├── 📄 app.py [Configurations for application framework]
│       ├── 📄 docs.py [Configurations for documentation]
│       └── 📄 env.py  [Exposes `env.yaml` to the application]
└── 📄 env.yaml [Environment file (git ignored)]
```

## Environment Configurations

Application-wide [environment configuration](https://www.12factor.net/config), which varies across staging, production, and development environments are read from `env.yaml` in the project root.

This file will not be tracked by git, making it safe for storing local secrets.

To get started, run the command `env_init` cli to initialize your environment configuration.

#### 1. Generate env file

Run the  `env_init`  cli tool.

```console
➜ uv run env_init
Env file created:  execution-spec-tests/env.yaml
```

which should generate an `env.yaml` in the project root.

```yaml
remote_nodes:
  - name: mainnet_archive
    # Replace with your Ethereum RPC node URL
    node_url: http://example.com
    # Optional: Headers for RPC requests
    rpc_headers:
      client-secret: <secret>
```

#### 2. Import `EnvConfig`

```console
from config import EnvConfig
EnvConfig().remote_nodes[0].name
'mainnet_archive'
```

## Application configuration

Application configuration are pydantic classes.

```console
from config import DocsConfig
DocsConfig().TARGET_FORK
'Prague'
```

## 🔗 Related Issues
closes #891, closes #893

## ✅ Checklist


raxhvl:
@danceratopz  let me know if this needs a changelog entry.

config.json.example
```diff
@@ -0,0 +1,12 @@
+{
```
danceratopz, line 1: How about, instead of adding an example file to the project root directory, we maintain the template (perhaps as a Python object of type `List[Dict]`) within `gentest.py` (for now) that we dump to `~/eest/gentest.json` upon request? This could be achieved via `gentest --init-config`? This seems even lower overhead for new users to get started?

src/cli/gentest.py
```diff
@@ -284,10 +269,7 @@ def __init__(self, node_config: Config.RemoteNode):
         Initialize the RequestManager with specific client config.
         """
         self.node_url = node_config.node_url
-        headers = {
-            "CF-Access-Client-Id": node_config.client_id,
-            "CF-Access-Client-Secret": node_config.secret,
-        }
+        headers = node_config.rpc_headers
```
danceratopz, line 272: This will raise an error if the user doesn't add and `rpc_headers` entry to their config (as it's optional). We could check if this entry is present and set it to an empty dictionary `{}` if not.

src/cli/gentest.py
```diff
@@ -70,7 +56,7 @@
     "-c",
     envvar="GENTEST_CONFIG_FILE",
     type=click.File("r"),
-    default=os.path.expanduser("~/.eest/gentest"),
+    default=Path.home() / ".eest" / "gentest" / "config.json",
```
danceratopz, line 59:
```suggestion
    default=Path.home() / ".eest" / "gentest.json",
```

danceratopz (commented):
Good suggestions thanks! I made a further suggestion to help people get started with config below.

As we discussed via DM, we should probably move away from a JSON config at some point. Unless we go with a pydantic-settings setup for gentest's config, but that might be overkill :)

config.json.example
raxhvl: We can leverage Jijna for templating config too. I can either do it in this PR or as part of the #863.

raxhvl:
@danceratopz Let me know where should I document this so the team is aware.

src/config/env.py
```diff
@@ -0,0 +1,105 @@
+"""
+env.py
```
danceratopz, line 2: The first line should be a single-line top-level description of the module (preferably < 80 chars). See: https://google.github.io/styleguide/pyguide.html#382-modules

We're not too strict, but trying to get better :sweat_smile:

src/config/env.py
```diff
@@ -0,0 +1,105 @@
...
+
+import yaml
+from pydantic import BaseModel, HttpUrl, ValidationError
+
+ENV_PATH = Path(__file__).resolve().parent.parent.parent / "env.yaml"
+
+
+def create_default_config():
+    """Creates a default configuration file if it doesn't exist."""
+    default_config = """
```
danceratopz, line 32: You can use `textwrap.dedent()` here to maintain consistent indentation in the python source and avoid the indentation in the str variable. You can find some examples within EEST.

src/config/env.py
```diff
@@ -0,0 +1,105 @@
...
+# This file contains environment-specific configurations for the application.
+# It is git-ignored to keep local secrets safe.
+
+# The configuration stores Ethereum RPC node details.
+remote_nodes:
```
danceratopz, line 37: Ah, of course, if you're adding comments I can see why templating would be helpful. 

We could use a combination of python objects holding the default values and a jinja2 template. We can add this later though!

src/config/env.py
```diff
@@ -0,0 +1,105 @@
...
+  - name: mainnet_archive
+    # Replace with your Ethereum RPC node URL
+    node_url: http://example.com
+    # Optional: Headers for RPC requests
+    rpc_headers:
+      client-secret: <secret>
+"""
+    with ENV_PATH.open("w") as file:
+        file.write(default_config.strip())
+        print(f"Env file created: {ENV_PATH}")
```
danceratopz, line 47: I would suggest checking whether the default file exists and avoid clobbering it, if so.

src/config/env.py
```diff
@@ -0,0 +1,105 @@
...
+    Attributes:
+    - config (Config): An instance of the Config class containing validated settings.
+
+    Raises:
+    - ValueError: If the configuration is invalid.
+    """
+
+    def __init__(self):
+        if not ENV_PATH.exists():
+            create_default_config()
```
danceratopz, line 89: I'm not sure how useful this is? (If the config file doesn't contain valid secrets for gentest, for example). Perhaps it's better to raise an exception with a hint tellling the user to run `uv run env_init`? Then they're aware of the file. Perhaps we can customise the exception message to `FileNotFoundError` in order to provide this hint, I'm not sure?

https://docs.python.org/3/library/exceptions.html#FileNotFoundError

src/config/env.py
```diff
@@ -0,0 +1,105 @@
...
+    - node_url (HttpUrl): The URL for the remote node, validated as a proper URL.
+    - rpc_headers (Dict[str, str]): A dictionary of optional RPC headers, defaults to empty dict.
+    """
+
+    name: str
+    node_url: HttpUrl
+    rpc_headers: Dict[str, str] = {}
+
+
+class Config(BaseModel):
```
danceratopz, line 65: Just FYI: We could consider using pydantic's `BaseSettings` class in the future.
https://docs.pydantic.dev/latest/concepts/pydantic_settings/#settings-management

I'm not quite sure how pydantic would work with the yaml though? Pydantic is cable of creating a schema, which could be loaded in VS code to help users edit config files. I'm not sure this would work with yaml, but it does look like we could use pydantic's `YamlConfigSettingsSource`. A schema is overkill for now though imo :-)

danceratopz (commented):
Thanks for this, really like how this is shaping up! :comet:  A few nits below!

danceratopz:
> @danceratopz Let me know where should I document this so the team is aware.

As this config will be limited to `gentest` (for starters), you could re-use `gentest`'s docstrings in the HTML page for `gentest` under here:
https://ethereum.github.io/execution-spec-tests/main/library/cli/
which gets generated from:
https://github.com/ethereum/execution-spec-tests/blob/619e7d8340802f457f5f4e1ada0bff3826a1c27c/src/cli/evm_bytes.py
and here:
https://github.com/ethereum/execution-spec-tests/blob/619e7d8340802f457f5f4e1ada0bff3826a1c27c/docs/library/cli/evm_bytes.md

But feel free to leave this until to you start work on `gentest` proper, it's ok for me if we can skip it for now.

If you're very motivated, you could add a short explainer under the developer docs:
https://ethereum.github.io/execution-spec-tests/main/dev/

src/config/env.py
raxhvl: This is outdated. We moved to template file.

src/config/env.py
raxhvl: I reviewed it I believe for our usecase this has diminishing returns. `pydantic-settings` will try merge environment the current runtime or an .env file.  We don't need either either. This also takes performance hit and the possibility of collusion should be there be an environment variable with the same name available in a user's runtime.

https://github.com/pydantic/pydantic-settings/blob/a72fa7383f4436148db01595b2993b846c91d3fc/pydantic_settings/main.py#L326-L333

raxhvl:
I have made the following updates

1. Default configs: Now using a template. Also Default config is itself a Config model so it will also be validated
https://github.com/ethereum/execution-spec-tests/blob/7a5df435d4063f2936c09d89454ec39aa116b7e9/src/config/env.py#L89-L97

2. Overriding exiting environment: Show a friendly message
```console
❯ uv run env_init
🚧 The configuration file 'execution-spec-tests/env.yaml' already exists. Please update it manually if needed.
```
3.  Nudge user if the `env.yaml` does not exist
```console
FileNotFoundError: The configuration file 'execution-spec-tests/env.yaml' does not exist. Run `uv run env_int` to create it.
```

If this looks okay then I will start documenting it.

config.json.example
danceratopz, line 1: We can delete this now.

src/config/templates/env.yaml.j2
```diff
@@ -0,0 +1,16 @@
...
+  - name: {{ node.name }}
+    # Replace with your Ethereum RPC node URL
+    node_url: {{ node.node_url }}
+    # Optional: Headers for RPC requests
+    rpc_headers:
+      {% for key, value in node.rpc_headers.items() %}
+      {{ key }}: {{ value }}
+      {% endfor %}
+{% endfor %}
+
```
danceratopz, line 16: Remove trailing whitespace:
```suggestion

```

danceratopz (commented):
> I have made the following updates
> ....
> If this looks okay then I will start documenting it.

Thanks for addressing the nits. Looks great, thanks! Ready for doc.

src/config/templates/env.yaml.j2
danceratopz: I really like jinja2, btw :slightly_smiling_face:

danceratopz (approved):
This is great! 🌟 Thanks a lot for this!

PR #896, opened 2024-10-15
new(cli): Introduce eofwrap tool

pdobacz:
## 🗒️ Description

There are many general EVM tests in `ethereum/tests` which provide coverage for opcodes and patterns which should behave independently from whether they run as legacy or as EOF code. Think `ADDMOD`, `SSTORE`, `SHL` etc... These likely will eventually be migrated manually to EEST, and then the existing tools will be there to wrap them in EOF containers. However, this is a long term task, and we would like to have the coverage early, in preparation for the EOF fork.

It turns out many of `ethereum/tests` state tests (here using their blockchain_test format, read on to see why) can be opportunistically (that is if they "wrap" easily - good, if not - skip) wrapped into EOF containers and executed and the same expectations on post state will apply.

This PR introduces a new CLI tool `eofwrap` accomplishes this. I've used it on v14.1 release `ethereum/tests` (subdir `BlockchainTests/GeneralStateTests/`). The `metrics.json` contents is:

```
{
    "files_generated": 1108,
    "files_skipped": 1656,
    "fixtures_generated": 11110,
    "fixtures_cant_wrap": 7973,
    "fixtures_cant_generate": 262,
    "accounts_wrapped": 53344,
    "unique_accounts_wrapped": 1165,
    "accounts_invalid_eof": 102192,
    "accounts_cant_generate": 855,
    "validation_errors": {
        "err: undefined_instruction": 89507,
        "err: stack_underflow": 12346,
        "err: no_terminating_instruction": 2,
        "err: truncated_instruction": 106,
        "err: invalid_non_returning_flag": 1,
        "err: invalid_container_section_index": 2,
        "err: unreachable_instructions": 203,
        "err: invalid_code_section_index": 1,
        "err: container_size_above_limit": 12,
        "err: max_stack_height_above_limit": 12
    },
    "generation_errors": {
        "unexpected balance for account...": 223,
        "incorrect value in address 0x0...": 9,
        "Account missing from allocatio...": 19,
        "unexpected nonce for account 0...": 9,
        "incorrect value in address 0xc...": 2
    }
}
```

If one collects the traces with `evmone`, but modifying it to _only_ trace EOF code execution, then after some sorting/counting (`cat trace.log | grep -o '\"opName\"\:\".*\"' | sort | uniq -c`), one can get the following summary of the opcodes executed in EOF:

```
 305517 "opName":"ADD"
    131 "opName":"ADDMOD"
     15 "opName":"ADDRESS"
    105 "opName":"AND"
    185 "opName":"BALANCE"
     24 "opName":"BASEFEE"
     23 "opName":"BLOBBASEFEE"
     34 "opName":"BLOBHASH"
      1 "opName":"BLOCKHASH"
    114 "opName":"BYTE"
  17335 "opName":"CALLDATACOPY"
 151069 "opName":"CALLDATALOAD"
     28 "opName":"CALLDATASIZE"
     30 "opName":"CALLER"
     14 "opName":"CALLVALUE"
      2 "opName":"CHAINID"
     64 "opName":"COINBASE"
      1 "opName":"DATASIZE"
    159 "opName":"DIV"
     51 "opName":"DUP1"
      3 "opName":"DUP10"
      2 "opName":"DUP11"
      2 "opName":"DUP12"
      2 "opName":"DUP13"
      2 "opName":"DUP14"
      2 "opName":"DUP15"
      2 "opName":"DUP16"
     30 "opName":"DUP2"
     22 "opName":"DUP3"
     19 "opName":"DUP4"
     22 "opName":"DUP5"
     17 "opName":"DUP6"
     19 "opName":"DUP7"
     17 "opName":"DUP8"
      2 "opName":"DUP9"
    116 "opName":"EQ"
    883 "opName":"EXP"
     45 "opName":"GASLIMIT"
     32 "opName":"GASPRICE"
    111 "opName":"GT"
     34 "opName":"INVALID"
    105 "opName":"ISZERO"
1193406 "opName":"JUMPDEST"
    265 "opName":"KECCAK256"
     40 "opName":"LOG0"
     40 "opName":"LOG1"
     35 "opName":"LOG2"
     36 "opName":"LOG3"
     31 "opName":"LOG4"
    107 "opName":"LT"
     22 "opName":"MCOPY"
 205043 "opName":"MLOAD"
    112 "opName":"MOD"
     60 "opName":"MSIZE"
 259944 "opName":"MSTORE"
    180 "opName":"MSTORE8"
    551 "opName":"MUL"
    161 "opName":"MULMOD"
    112 "opName":"NOT"
     61 "opName":"NUMBER"
    105 "opName":"OR"
      3 "opName":"ORIGIN"
    693 "opName":"POP"
     30 "opName":"PREVRANDAO"
     26 "opName":"PUSH0"
1708938 "opName":"PUSH1"
      5 "opName":"PUSH10"
      7 "opName":"PUSH11"
      3 "opName":"PUSH12"
     18 "opName":"PUSH13"
     24 "opName":"PUSH14"
      9 "opName":"PUSH15"
    282 "opName":"PUSH16"
      5 "opName":"PUSH17"
     11 "opName":"PUSH18"
      1 "opName":"PUSH19"
 169477 "opName":"PUSH2"
    159 "opName":"PUSH20"
      4 "opName":"PUSH21"
      2 "opName":"PUSH22"
      1 "opName":"PUSH23"
      1 "opName":"PUSH24"
      4 "opName":"PUSH25"
      1 "opName":"PUSH26"
      3 "opName":"PUSH27"
      3 "opName":"PUSH28"
      3 "opName":"PUSH29"
     66 "opName":"PUSH3"
     12 "opName":"PUSH30"
     11 "opName":"PUSH31"
   2280 "opName":"PUSH32"
    341 "opName":"PUSH4"
     18 "opName":"PUSH5"
      8 "opName":"PUSH6"
   1156 "opName":"PUSH7"
    145 "opName":"PUSH8"
      8 "opName":"PUSH9"
 151214 "opName":"RETURN"
      6 "opName":"RETURNDATACOPY"
      2 "opName":"RETURNDATASIZE"
   3664 "opName":"REVERT"
    115 "opName":"SAR"
    119 "opName":"SDIV"
     31 "opName":"SELFBALANCE"
    106 "opName":"SGT"
    232 "opName":"SHL"
    108 "opName":"SHR"
     22 "opName":"SIGNEXTEND"
 100951 "opName":"SLOAD"
    103 "opName":"SLT"
    115 "opName":"SMOD"
 343150 "opName":"SSTORE"
 193260 "opName":"STOP"
    279 "opName":"SUB"
     10 "opName":"SWAP1"
      2 "opName":"SWAP10"
      2 "opName":"SWAP11"
      2 "opName":"SWAP12"
      2 "opName":"SWAP13"
      2 "opName":"SWAP14"
      2 "opName":"SWAP15"
      2 "opName":"SWAP16"
      7 "opName":"SWAP2"
      4 "opName":"SWAP3"
      7 "opName":"SWAP4"
      4 "opName":"SWAP5"
      4 "opName":"SWAP6"
      4 "opName":"SWAP7"
      3 "opName":"SWAP8"
      4 "opName":"SWAP9"
     50 "opName":"TIMESTAMP"
     22 "opName":"TLOAD"
     22 "opName":"TSTORE"
    110 "opName":"XOR"
```

### Open questions

1. Right now `eofwrap` takes an opportunistic approach - if the test wraps _something_ and passes, it's worthwhile to wrap and run it. For example, there are tests where it is not the most important code under test being EOF-wrapped, but some auxiliary contract which merely interacts with the important one. Opportunistic here means we still keep this test, to potentially test some edge cases of interactions between legacy and EOF contexts. Other aspect is that there are test categories where many of the fixtures are unwrappable, but several remain. Opportunistic means then that we nevertheless keep the ones which wrapped.
   The opposite side of the spectrum is to be more "systematic", i.e. filter out more test categories where the test is "weaker" and include tests more coarsely, only where we're certain EOF opcodes are executed.
   For now we err on the side of opportunistic, since this is intended to be a low-cost, stop-gap solution, and postpone the systematic effort until proper migration to EEST.
2. Not sure how would those JSON files be published. One idea is to include an invocation of such `eofwrap` in the `eip7692` prerelease builds, and just include the JSON files in the resulting directory structure. This maximizes the chance that the EVM implementations will actually run these tests, but might come off as messy.
3. Why use blockchain test format even though those are actually state tests? It includes the detailed post state assertions, which makes it a lot easier to generate meaningful tests (credit to @chfast for this idea)

## 🔗 Related Issues

NA

## ✅ Checklist


src/cli/eofwrap.py
```diff
@@ -0,0 +1,370 @@
...
+
+    return opcode_at == op
+
+
+def wrap_code(account_code: Bytes) -> Container:
+    """
+    Wraps `account_code` into a simplest EOF container, applying some simple heuristics in
+    order to obtain a valid code section termination.
+    """
+    assert len(account_code) > 0
```
marioevz, line 324: I have the feeling that doing something like we do here would yield better results maybe:
https://github.com/ethereum/execution-spec-tests/blob/0900aa8ef592f968c119f11cdcd934eb2d360693/src/cli/evm_bytes.py#L68-L92

The stack would be automatically calculated for you and also you could check the `terminating` attribute of `Bytecode` to decide whether to add a `STOP` or not.

src/cli/eofwrap.py
```diff
@@ -0,0 +1,370 @@
...
+                        account.code = Bytes(wrapped)
+                        wrapped_at_least_one_account = True
+                        self.metrics[self.ACCOUNTS_WRAPPED] += 1
+                        fixture_eof_codes.append(to_hex(account.code))
+
+                        # wrap the same account in post state the same way
+                        if fixture.post_state and fixture.post_state.root[address]:
+                            fixture.post_state.root[address].code = Bytes(wrapped)  # type: ignore
+                    else:
+                        self.metrics[self.ACCOUNTS_INVALID_EOF] += 1
```
marioevz, line 181: Maybe even these negative ones are valuable, if they could be wrapped as `eof_test` with an invalid container for negative testing.

src/cli/eofwrap.py
```diff
@@ -0,0 +1,370 @@
...
+            pre=pre.root,
+            post=fixture.post_state.root if fixture.post_state else {},
+            blocks=[],
+            tag="wrapped test",
+        )
+
+        for fixture_block in fixture.blocks:
+            if isinstance(fixture_block, FixtureBlock):
+                header = fixture_block.header
+                block = Block(
```
marioevz, line 228: I think this should focus on the transactions part of the fixture, because, for example, overriding the state root with the one in the original fixture is almost always going to generate an invalid block.

src/cli/eofwrap.py
```diff
@@ -0,0 +1,370 @@
...
+                    requests_root=header.requests_root,
+                )
+                assert not fixture_block.ommers
+                assert not fixture_block.withdrawals
+                assert not fixture_block.deposit_requests
+                assert not fixture_block.withdrawal_requests
+                assert not fixture_block.consolidation_requests
+
+                for fixture_tx in fixture_block.txs:
+                    tx = Transaction(
```
marioevz, line 258: I think `Transaction(**fixture_tx,model_dump())` should work with minimal modifications.

src/cli/eofwrap.py
```diff
@@ -0,0 +1,370 @@
...
+
+        if len(out_fixtures) == 0:
+            self.metrics[self.FILES_SKIPPED] += 1
+            return
+
+        os.makedirs(os.path.dirname(out_path), exist_ok=True)
+        out_fixtures.collect_into_file(Path(out_path))
+        self.metrics[self.FILES_GENERATED] += 1
+
+    def _wrap_fixture(self, fixture: BlockchainFixture, traces: bool):
```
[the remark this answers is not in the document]
marioevz: I don't think there is sadly, because the `spec` to `fixture` conversion has always been seen as a one-way street. But see comments in review, it might help this become a bit more stable.

marioevz (commented):
Looks good! I think we could apply this in a post-fill script.

I think @danceratopz might have some comments, and I see one regarding how this could be a pytest plugin, and I agree :smile:

src/cli/eofwrap.py
pdobacz: This crossed my mind, but then I had doubts. Comparing to the positive ones, the resulting coverage would be very erratic, and at the same time basic, as all the code would always be in 0th container in 0th section etc.

Also, if in the future we'd revisit this and add more heuristics to "fix" some of those negatives, we'd take that coverage away :).

Lastly it's 100k cases for undefined instruction and stack underflow alone, for example the cases where `012345` is used as placeholder "code".

I'll keep it at the back of my head for a follow-up PR though or maybe as an addition to the fuzzing corpus :thinking: ...

pdobacz: > Lastly it's 100k cases for undefined instruction and stack underflow alone, for example the cases where 012345 is used as placeholder "code".

of course, if we did this, we'd first de-dup it, so that particular part makes no sense :joy:

marioevz: Makes sense :+1:

danceratopz (commented):
Nice idea! I agree with Mario that we could make a pytest plugin out of this :slightly_smiling_face:  However, I think we can merge this as-is and then consider whether we do this later.

To explain a little how that could be helpful, you can try out `consume direct`, which is a small wrapper framework to help execute fixtures against client `statetest` and `blocktest`  commands. Currently, only `geth` is supported, but we're currently refactoring this code and plan to add other clients soon. Here's an example:
```
uv run consume direct --input=../../ethereum/tests/BlockchainTests/GeneralStateTests/Cancun/ --evm-bin=~/bin/evm
```
All the usual pytest flags are available. For example, tests can be filtered as for `fill` with `-k`. The `consume` command also generates an index file of all the discovered test fixtures - this can be used for subsequent runs, speeding up test discovery.

Of course, if this is more a "single shot" kinda tool, then this CLI could already be sufficient!  If we do decide to refactor this to be a pytest plugin, we'd be very happy to take the implementation over or help out. More than likely, we would try to generalize it a bit further in this case.

Anyway, one small comment about this PR below.

src/cli/tests/test_evm_bytes.py
```diff
@@ -16,6 +16,19 @@
     "0x7fa0a1a2a3a4a5a6a7a8a9aaabacadaeafb0b1b2b3b4b5b6b7b8b9babbbcbdbebf5f527fc0c1c2c3c4c5c6c7c8c9cacbcccdcecfd0d1d2d3d4d5d6d7d8d9dadbdcdddedf6020527fe0e1e2e3e4e5e6e7e8e9eaebecedeeeff0f1f2f3f4f5f6f7f8f9fafbfcfdfeff60405260786040356020355f35608a565b5f515f55602051600155604051600255005b5e56",  # noqa: E501
     "Op.PUSH32[0xa0a1a2a3a4a5a6a7a8a9aaabacadaeafb0b1b2b3b4b5b6b7b8b9babbbcbdbebf] + Op.PUSH0 + Op.MSTORE + Op.PUSH32[0xc0c1c2c3c4c5c6c7c8c9cacbcccdcecfd0d1d2d3d4d5d6d7d8d9dadbdcdddedf] + Op.PUSH1[0x20] + Op.MSTORE + Op.PUSH32[0xe0e1e2e3e4e5e6e7e8e9eaebecedeeeff0f1f2f3f4f5f6f7f8f9fafbfcfdfeff] + Op.PUSH1[0x40] + Op.MSTORE + Op.PUSH1[0x78] + Op.PUSH1[0x40] + Op.CALLDATALOAD + Op.PUSH1[0x20] + Op.CALLDATALOAD + Op.PUSH0 + Op.CALLDATALOAD + Op.PUSH1[0x8a] + Op.JUMP + Op.JUMPDEST + Op.PUSH0 + Op.MLOAD + Op.PUSH0 + Op.SSTORE + Op.PUSH1[0x20] + Op.MLOAD + Op.PUSH1[0x1] + Op.SSTORE + Op.PUSH1[0x40] + Op.MLOAD + Op.PUSH1[0x2] + Op.SSTORE + Op.STOP + Op.JUMPDEST + Op.MCOPY + Op.JUMP",  # noqa: E501
 ]
+rjump_vector = [
+    "0xe0ffff",
```
chfast, line 20: remember about endianness: use a test vector with different bytes (e.g. `-2`)

src/cli/eofwrap.py
pdobacz: `fix(cli): rework eofwrap using process_evm_bytes` accomplishes that, thanks for the suggestions. It required a bit  of fiddling and adding a parameter, but finally all JSON files went through.

src/cli/tests/test_evm_bytes.py
pdobacz: thx, fixed!

pdobacz:
@marioevz @danceratopz Thank you for your comments, I think I've addressed all except the plugin - would be better to leave for a separate PR indeed. I see the benefits, but not sure if I know where to start, and also it is indeed more of a one-off tool (hopefully temporary, until migration, but I have some level of fear that it will be "permanently-temporary", the migration is a lot to take on).

Do we want to tackle including of these tests in the `eip7962` prerelease at this PR? I thought I've seen a comment on this somewhere, but I cannot find it anymore. How can we approach it?

src/cli/eofwrap.py
```diff
@@ -0,0 +1,339 @@
...
+    order to obtain a valid code section termination.
+    """
+    assert len(account_code) > 0
+
+    opcodes = process_evm_bytes(account_code, allow_unknown=True)
+
+    if not opcodes[-1].opcode.terminating:
+        opcodes.append(OpcodeWithOperands(opcode=Op.STOP))
+
+    while len(opcodes) > 1 and opcodes[-2].opcode.terminating and opcodes[-1].opcode.terminating:
```
marioevz, line 320:
```suggestion
    while len(opcodes) > 1 and opcodes[-2].terminating and opcodes[-1].terminating:
```
I think it would be a bit cleaner to add:
```python
@property
def terminating(self) -> bool:
    """Whether the opcode is terminating or not"""
    return self.opcode.terminating
```
To the `OpcodeWithOperands` class in `evm_bytes.py`.

src/cli/eofwrap.py
```diff
@@ -0,0 +1,339 @@
...
+        opcodes.pop()
+
+    bytecode = Bytecode()
+
+    for opcode in opcodes:
+        if opcode.operands:
+            # opcode.opcode[*opcode.operands] crashes `black` formatter and doesn't work.
+            bytecode += opcode.opcode.__getitem__(*opcode.operands)
+        else:
+            bytecode += opcode.opcode
```
marioevz, line 330:
```suggestion
        bytecode += opcode.bytecode()
```
We could add this as a method to `OpcodeWithOperands`:
```python
@property
def bytecode(self) -> Bytecode:
    """Opcode as bytecode with its operands if any."""
    return self.opcode.__getitem__(*self.operands) if self.operands else self.opcode
```

src/cli/evm_bytes.py
```diff
@@ -57,40 +58,57 @@ def format_assembly(self) -> str:
...
-        opcode: Op
+        opcode: Opcode
         for op in Op:
             if not isinstance(op, Macro) and op.int() == opcode_byte:
                 opcode = op
                 break
         else:
-            raise ValueError(f"Unknown opcode: {opcode_byte}")
+            if allow_unknown:
+                opcode = Opcode(opcode_byte)
```
marioevz, line 78: I'm a bit skeptical of this tbh. Which tests in `ethereum/tests` require this?

marioevz (commented):
Thanks for the changes, it's looking very good. A couple of comments.

src/cli/evm_bytes.py
pdobacz: Very many (order of 20K codes). They never wrap successfully, but they fail on the `process_evm_bytes` step, where it was a bit inconvenient to handle failure. I can undo this flag and catch the `Unknown opcode` error in `eofwrap.py` instead if you prefer that.

pdobacz: Did so in `Revert support for unknown opcodes in evm_bytes.py`, PTAL. I'm still running the wrapping to see if metrics and all are fine

pdobacz: @marioevz If this looks good now then we can merge or do you guys have any more comments?

Then to follow-up with the including in the release - should I open an issue to track this? (I can help with this, but I'd need pointers on where to start)

marioevz (approved):
Looks great! Many thanks for the refactor in `evm_bytes.py`.

Issue #866, opened 2024-10-02
✨(gentest): make `gentest` great


danceratopz:
The `gentest` CLI is a really nice tool that can generate and write a Python (pytest) test module to disk given a type 0 transaction hash on a live network. This generates a `blockchain_test` from the mainnet transaction [`0xa41f...06be`](https://etherscan.io/tx/0xa41f343be7a150b740e5c939fa4d89f3a2850dbe21715df96b612fc20d1906be):

```bash
uv run gentest -c config.json 0xa41f343be7a150b740e5c939fa4d89f3a2850dbe21715df96b612fc20d1906be tests/paris/test_0xa41f.py
```

Currently, due to its limited transaction type support (type 0 only) it's of limited use. This meta ticket is to track progress of making `gentest` a easy-to-use CLI tool that can generate test cases for specific issues observed on devnets, and potentially, to allow retroactive benchmarking of an EVM's transaction processing using real mainnet data.
 
```[tasklist]
### Tasks
- [ ] #859 
- [ ] #860
- [ ] #864
- [ ] #863
- [ ] #861 
- [ ] #862 
- [ ] #865
```

danceratopz:
@raxhvl said he's happy to pick this up! :relaxed:

Thanks @raxhvl! Above I tried to outline the most obvious tasks for `gentest`. Please see these as suggestions - if you have any other ideas or inputs, just go for it! I'm also happy to jump in and help out if need be.

raxhvl:
Thanks @danceratopz I will comment on individual issues as I start working on them.

danceratopz:
The `testinit` command has now become `et make test` - the tasks related to `make test` are now tracked here: #973

chfast:
Hey there, I've just become aware of this source of execution data: https://ethpandaops.io/data/xatu/schema/canonical_execution_/#canonical_execution_storage_reads
It may be easier to access these instead of going through a node RPC.

raxhvl:
Thanks @chfast!

danceratopz:
> Hey there, I've just become aware of this source of execution data: https://ethpandaops.io/data/xatu/schema/canonical_execution_/#canonical_execution_storage_reads It may be easier to access these instead of going through a node RPC.

Thanks @chfast, this is a great idea! I think we'll have to go with RPC support for now, as xatu is unlikely to add devnet support to the database. But it is definitely something we can consider down the line once `gentest` is more feature complete.

chfast:
[...]

PR #901, opened 2024-10-18
♻️ refactor(gentest,base_types): Improved architecture, implement EthereumTestBaseModel, EthereumTestRootModel

### LINKED ISSUE #859: ✨(gentest): ensure the generated python test module passes tox checks
Currently, if a test module is generated via:

```bash
uv run gentest -c config.json 0xa41f343be7a150b740e5c939fa4d89f3a2850dbe21715df96b612fc20d1906be tests/paris/test_0xa41f.py
```

the generated test module `tests/paris/test_0xa41f.py` doesn't pass the `flake8` and `mypy` checks when running `tox` (`uvx --with=tox-uv tox`).

For example:
```
uv run mypy tests/paris/test_0xa41f.py
tests/paris/test_0xa41f.py:67: error: Need type annotation for "post" (hint: "post: Dict[<type>, <type>] = ...")  [var-annotated]
```

We should ensure that no manual processing or fixes are necessary for tox (and CI) to pass. It should just work :tm:.

raxhvl:
## 🗒️ Description

This PR ensures data from gentest  to generated tests reaches in a type-safe manner.
 
The Gentest module generates a test script in three steps:

1. The CLI entrypoint function captures the user input required for generating the test.
2. This input is used to generate additional context required for creating tests.
3. Finally, using this context, a formatted pytest source code is generated and written to disk.

The process of generating context and rendering templates will share the same logic across different kinds of tests. These are created as modules: (`cli`, `test_context_provider`, and `source_code_generator`).

```mermaid
graph TD;
    A[CLI Entry Point] -->|Captures User Input| B[Test Context Provider]
    B -->|Generates Context| C[Source Code Generator]
    C -->|Generates Formatted Pytest Code| D[Disk]
```

I am using the **same data type** when passing data from the provider to the generated source code. This keeps the behavior predictable; Pydantic will raise errors if the RPC call returns unexpected values even **before** we run the test. I'm leveraging `__repr__` values to print the data into the source code.

Finally, the code is formatted using the same formatter (Black) used elsewhere. The existing formatting configurations are applied, so any change in configuration should automatically reflect in the files generated as well. Formatting would become essential when reading contract bytecode, especially with the plans to convert bytecode to python objects (see #862).

This effort makes writing new kinds of gentests easier, requiring only the creation of new test providers and templates.

## 🔗 Related Issues
closes #859 

## ✅ Checklist


src/cli/gentest/source_code_generator.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns:
+        str: The formatted pytest source code.
+    """
+    template = template_env.get_template(template_path)
+    rendered_template = template.render(provider.get_context())
+    # return rendered_template
+    return format_code(rendered_template)
+
+
+def format_code(code: str) -> str:
```
danceratopz, line 45: Great idea.

src/cli/gentest/source_code_generator.py
```diff
@@ -0,0 +1,83 @@
...
+
+import jinja2
+
+from .test_providers import Provider
+
+template_loader = jinja2.PackageLoader("cli.gentest")
+template_env = jinja2.Environment(loader=template_loader, keep_trailing_newline=True)
+
+# This filter maps python objects to string
+template_env.filters["stringify"] = lambda input: repr(input)
```
danceratopz, line 20: I wondered whether we should call the filter `repr` to be explicit about it's implementation, but I don't have a strong opinion.
```suggestion
template_env.filters["repr"] = lambda input: repr(input)
```

src/cli/gentest/templates/blockchain_test/transaction.py.j2
```diff
@@ -1,5 +1,5 @@
 """
-gentest autogenerated test with debug_traceCall of tx.hash
+Gentest autogenerated test with `debug_traceCall` of tx.hash:
```
danceratopz, line 2: Would be great to get the network source (or chainid?) here. Just a thought - doesn't need to be done in the scope of this PR.
```suggestion
Gentest autogenerated test from `tx.hash`:
```

And, modify the docstring to have a pythonic one-liner with an abbreviated `tx.hash` and `chainid`. The full tx.hash can be then provided below with link.

src/cli/gentest/cli.py
```diff
@@ -9,16 +9,12 @@
...
-from .request_manager import RPCRequest
+from .source_code_generator import get_test_source
 from .test_providers import BlockchainTestProvider
 
-template_loader = jinja2.PackageLoader("cli.gentest")
-template_env = jinja2.Environment(loader=template_loader, keep_trailing_newline=True)
-
 
 @click.command()
 @click.argument("transaction_hash")
```
danceratopz, line 20: Would be nice to extend the interface to allow specification of other chains (especially testnets - we could test with sepolia, for example):
```suggestion
@click.argument("transaction_hash")
@click.argument("chainid)
```

Can be added in a separate PR! :slightly_smiling_face:

src/cli/gentest/cli.py
```diff
@@ -9,16 +9,12 @@
...
 
-from .request_manager import RPCRequest
+from .source_code_generator import get_test_source
 from .test_providers import BlockchainTestProvider
 
-template_loader = jinja2.PackageLoader("cli.gentest")
-template_env = jinja2.Environment(loader=template_loader, keep_trailing_newline=True)
-
 
 @click.command()
```
danceratopz, line 19: I'm wondering how you'd like to design the interface once you implement different types of `blockchain_test`, for example, from multiple txs. Would you infer the type from the arguments (allow a list of tx, for example)? Or add sub-commands to gentext, like `gentest fromtx 0x....`?

Again, no need to answer in any depth here. Or implement it in the scope of this PR. Just thoughts for the process ))

raxhvl: I have couple of ideas there. I will document it in the PR for #865.

src/cli/gentest/cli.py
raxhvl: I think that would be great feature. I have created #908. Could you add it to #866.

src/cli/gentest/templates/blockchain_test/transaction.py.j2
raxhvl: [replied below](https://github.com/ethereum/execution-spec-tests/pull/901#discussion_r1811822240)

src/cli/gentest/source_code_generator.py
raxhvl: I named it "stringify" to decouple it from `repr`, allowing for flexibility in adopting different formats in the future. This way, if we choose to implement a more complex output method, the filter name will still be relevant.

src/cli/gentest/cli.py
danceratopz: Done :slightly_smiling_face:

src/cli/gentest/source_code_generator.py
danceratopz: Ok! :+1:

raxhvl:
I think the output format of the generated test could be made better. Which is why the draft is still in PR.

image

We should ensure addresses is represented as hex and large numbers like v,r,s to be perhaps literal hex numbers. So that it is consistent with manually written tests. How about:

| Field   | Representation in Tests |
|---------|:-----------------------:|
| Address |        hex string       |
| Hash    |        hex string       |
| Number  | hex number literal      |
| Code    | Opcode class?           |

I think the best way to achieve this is to override `__repr__`  in individual base types. Perhaps like this:

```diff
class HexNumber(Number):
    """
    Class that helps represent an hexadecimal numbers in tests.
    """

+   def __repr__(self):
+       return f"0x{123}"

    def __str__(self) -> str:
        """
        Returns the string representation of the number.
        """
        return self.hex()

```

I would like you to weigh in on the effect of this across the codebase.

danceratopz:
> I think the output format of the generated test could be made more human readable. Which is why the PR is still in draft.
> 
> image
> 
> For example, we should ensure addresses are represented as hex and large numbers like v,r,s to be perhaps literal hex numbers. So that it is consistent with manually written tests. How about:
> Field 	Representation in Tests
> Address 	hex string
> Hash 	hex string
> Number 	hex number literal
> Code 	Opcode class?
> 
> I think the best way to achieve this is to override `__repr__` in individual base types. Perhaps like this:
> 
> ```diff
> class HexNumber(Number):
>     """
>     Class that helps represent an hexadecimal numbers in tests.
>     """
> 
> +   def __repr__(self):
> +       return f"0x{123}"
> 
>     def __str__(self) -> str:
>         """
>         Returns the string representation of the number.
>         """
>         return self.hex()
> ```
> 
> I would like you to weigh in on the effect of this across the codebase.

I'll set off with a broader topic, so we have the full picture. Bear with me! 

All right, so getting nice representations would also really help us with the Test Case Reference docs, see the bytes objects in the test case parameters tables, for example here: https://ethereum.github.io/execution-spec-tests/main/tests/prague/eip6110_deposits/test_deposits/test_deposit.html

For the objects defined and only used in the test cases (which aren't part of EEST's types libraries), adding a `__repr__` would help and seems like the way to go.

For the objects that we do define in our test libraries, we get nice representations (of most objects, at least) when we write them as string to the fixture JSON files. And, instead of just adding `__repr__` to all these classes, perhaps we can re-use the pydantic magic. I will try to get back to this later today, but if you'd like to play around in the meantime, try this:

Get some fixtures:
```bash
uv run fill tests/istanbul/
```

Then play around in `ipython` (https://ethereum.github.io/execution-spec-tests/main/dev/interactive_usage/)
```bash
uvx  --with-editable . ipython
```
for example:
```python
import rich

from ethereum_test_fixtures.file import Fixtures
from ethereum_test_base_types import to_json

fixtures = Fixtures.from_file("fixtures/state_tests/istanbul/eip1344_chainid/chainid/chainid.json", fixture_format=None)
fixtures.keys()
fixtures['tests/istanbul/eip1344_chainid/test_chainid.py::test_chainid[fork_Istanbul-state_test]']
fixtures['tests/istanbul/eip1344_chainid/test_chainid.py::test_chainid[fork_Istanbul-state_test]'].env

rich.print(fixtures['tests/istanbul/eip1344_chainid/test_chainid.py::test_chainid[fork_Shanghai-state_test]'].env)

to_json(fixtures['tests/istanbul/eip1344_chainid/test_chainid.py::test_chainid[fork_Shanghai-state_test]'].env.fee_recipient)
```

`to_json` uses pydantic's `model_dump()`: 
https://github.com/ethereum/execution-spec-tests/blob/7758f37e3dca5f7b8b7ab0ea5136671ed4246623/src/ethereum_test_base_types/json.py#L10-L21

Just be aware that fill serializes from the test **spec** format (e.g. `StateTest` src/ethereum_test_specs/state.py to (state) `Fixture` src/ethereum_test_fixtures/state.py) and when we load the fixture from file above we're not creating a `StateTest`, we're creating a state test `Fixture` object. Quite a lot of magic happens when we serialize, for example, removing objects that are `None`; don't belong in state test fixture, but do in a blockchain test fixture.

I can take a deeper look myself and happy to discuss to brainstorm, but if you have time and you're interested, have play around!

danceratopz:
> I get your point. The `to_json` method respects Pydantic attributes (like alias) and filters out `None` values, while your overrides of `__str__` produce nicely formatted JSON.
> 
> However, `to_json` outputs JSON, not valid Python code like `__repr__`. We need a way to combine the best of both without duplicating effort.
> 
> Let me know if my understanding is correct. Meanwhile, let me think this through.

Yup, that's right. 

Yes, unfortunately pydantic doesn't deserialize to Python. Calling `to_json` on each individual field (the only way I can think of doing it at the mo) makes little sense. So, as we'll only need `__repr__()` for these base classes (I think), it'll have minimal impact to add them. 

Perhaps @marioevz can chime in, if he has a better idea? :pray: 

Sorry for the detour and that I couldn't give you a better answer earlier @raxhvl.

raxhvl:
I have been marinating on this. I believe your intuition is correct: We should implement a unified serialization pipeline. This approach not only eliminates duplicate (tedious) efforts but also ensures that the framework produces consistent data outputs, whether for fixtures or gentests. Fundamentally, gentest should be as close to manual tests as possible.

Our goal is to serialize a Pydantic model to python code. After trying out few things, I have a solution.

We can override `__repr__args__` in the root class, which feeds into Pydantic's `__repr__`. This will ensure that JSON serialization and `__repr__` are consistent.

Here is a full example:


```python
from pydantic import BaseModel, Field
from typing import List


class RecipeBaseModel(BaseModel):
   
    def __repr_args__(self):
       # Skip tags
        attrs_names = self.model_dump(mode="json", exclude={"tags"}).keys()
        attrs = ((s, getattr(self, s)) for s in attrs_names)
        return [(a, v) for a, v in attrs if v is not None]


# Example child model
class Ingredient(RecipeBaseModel):
    name: str
    quantity: str
    tags: List[str]


# Example nested child model
class Recipe(RecipeBaseModel):
    title: str
    ingredients: List[Ingredient]


# Example usage
ingredient1 = Ingredient(name="Tomato", quantity="2 cups", tags=["vegetable", "fresh"])
ingredient2 = Ingredient(name="Basil", quantity="1 bunch", tags=["herb", "fresh"])
recipe = Recipe(title="Tomato Basil Salad", ingredients=[ingredient1, ingredient2])

print(repr(ingredient1))  # Ingredient(name='Tomato', quantity='2 cups')
print(
    repr(recipe)
)  # Recipe(title='Tomato Basil Salad', ingredients=[Ingredient(name='Tomato', quantity='2 cups'), Ingredient(name='Basil', quantity='1 bunch')])

```

I know this will take some effort to implement correctly. But I think incentive to unify serialization will go down as the code diverges because of the refactoring involved.

Let me know if we should pursue this or override `__repr__` everywhere for now.

danceratopz:
> I have been marinating on this. I believe your intuition is correct: We should implement a unified serialization pipeline. This approach not only eliminates duplicate (tedious) efforts but also ensures that the framework produces consistent data outputs, whether for fixtures or gentests. Fundamentally, gentest should be as close to manual tests as possible.
> 
> Our goal is to serialize a Pydantic model to python code. After trying out few things, I have a solution.
> 
> We can override `__repr__args__` in the root class, which feeds into Pydantic's `__repr__`. This will ensure that JSON serialization and `__repr__` are consistent.
> 
> Here is a full example:
> 
> ```python
> from pydantic import BaseModel, Field
> from typing import List
> 
> 
> class RecipeBaseModel(BaseModel):
>    
>     def __repr_args__(self):
>        # Skip tags
>         attrs_names = self.model_dump(mode="json", exclude={"tags"}).keys()
>         attrs = ((s, getattr(self, s)) for s in attrs_names)
>         return [(a, v) for a, v in attrs if v is not None]
> 
> 
> # Example child model
> class Ingredient(RecipeBaseModel):
>     name: str
>     quantity: str
>     tags: List[str]
> 
> 
> # Example nested child model
> class Recipe(RecipeBaseModel):
>     title: str
>     ingredients: List[Ingredient]
> 
> 
> # Example usage
> ingredient1 = Ingredient(name="Tomato", quantity="2 cups", tags=["vegetable", "fresh"])
> ingredient2 = Ingredient(name="Basil", quantity="1 bunch", tags=["herb", "fresh"])
> recipe = Recipe(title="Tomato Basil Salad", ingredients=[ingredient1, ingredient2])
> 
> print(repr(ingredient1))  # Ingredient(name='Tomato', quantity='2 cups')
> print(
>     repr(recipe)
> )  # Recipe(title='Tomato Basil Salad', ingredients=[Ingredient(name='Tomato', quantity='2 cups'), Ingredient(name='Basil', quantity='1 bunch')])
> ```
> 
> I know this will take some effort to implement correctly. But I think incentive to unify serialization will go down as the code diverges because of the refactoring involved.
> 
> Let me know if we should pursue this or override `__repr__` everywhere for now.

Thanks @raxhvl for taking the time to look so deeply into this! This looks like a great approach to unify our serialization and think we can start adding `__repr__args__` to our types (or to a new class that the types inherit from). 

If I understand correctly, the code paths for serializing the JSON fixture files would remain unchanged? But now we use  `json.dumps()` within the `__repr_args__` method  to achieve compatibility? In this case, this doesn't seem like a huge change to the codebase (and certainly not a breaking change). But perhaps I've misunderstood how intrusive this is. If I'm right, you could try adding these methods upfront to achieve your goal for `gentest` now which would allow us to better evaluate the change.

raxhvl:
I have added a new base class.

```mermaid
classDiagram
   BaseModel <|-- EthereumTestBaseModel
   EthereumTestBaseModel <|-- CopyValidateModel
   CopyValidateModel <|-- CamelModel

   class EthereumTestBaseModel{
    serialize()
    __repr_args__()
   }
```

The serialize method is a thin wrapper over `model_dump` which `__repr_args__`. Ideally I would like to change this:

https://github.com/ethereum/execution-spec-tests/blob/cfbfea9821e505a13227da85eb09f7d1642dc7f0/src/ethereum_test_base_types/json.py#L19

to `input.serialize()` so that serialization is at one place but I'm not sure if all classes does infact inherrit `EthereumTestBaseModel`. So the code is now duplicated at two places.


The `__repr__` logic is working fine. It ignores `None` fields, supports custom representation if required and falls back to `__str__` representation.

raxhvl:
- [ ] Migrated `to_json` to use `serialize`
- [ ] Rebased on top of main

>re: `black` to `ruff` migration.

The code uses cli to format code as long as `ruff` has a cli interface to format. The migration is trivial (<2 lines of code)

It would be nice if Mario reviews this once we are done.

src/cli/gentest/templates/blockchain_test/transaction.py.j2
```diff
@@ -8,37 +8,36 @@ from typing import Dict
...
 
 @pytest.fixture
 def env():  # noqa: D103
-    return Environment(
-{{ environment_kwargs }}
-    )
-
+    return {{ environment  | stringify }}
 
 @pytest.mark.valid_from("Paris")
```
marioevz, line 28: Not a requirement right now but it would be interesting that, if we have the fork schedule for the network from which we are obtaining the transaction, we can put the correct fork in here.

It would not work for devnets for example, but at least for mainnet and public testnets it's doable.

src/ethereum_test_base_types/pydantic.py
```diff
@@ -1,17 +1,82 @@
...
+
+        # Convert field values based on their type.
+        # This ensures consistency between JSON and Python object representations.
+        # Should a custom `__repr__` be needed for a specific type, it can added in the
+        # match statement below.
+        # Otherwise, the default string representation is used.
+        repr_attrs = []
+        for a, v in attrs:
+            match v:
+                case list() | dict() | BaseModel():
```
marioevz, line 70:
```suggestion
                case list() | dict() | BaseModel() | None:
```
We have an edge case in the `Transaction` class type where we force the `to` field to appear in the dumped model even when equal to `None`, because the transition tool expects this value when the transaction is contract-creating: https://github.com/ethereum/execution-spec-tests/blob/b78eb773b1bde5a333286f99dd96e7da8aac5a7f/src/ethereum_test_types/types.py#L670-L682

With this change, the field now is correctly printed (without it the field appeared as `to='None'`).

I think we should add some unit tests for the `__repr__` method.

marioevz (commented):
Amazing work so far, thanks!

Just a single comment (the other comment about the fork name is not really something that we need to do right now).

raxhvl:
All models are now using the same serialization logic. We have two types of custom models: one for `BaseModel` and one for `RootModel`. The customization for both of these models is handled through a shared mixin.

I have tested this thoroughly, but please review it with care as this change impacts **EVERY Pydantic model** in the codebase.

```mermaid
classDiagram
   BaseModel <|-- EthereumTestBaseModel
   RootModel <|-- EthereumTestRootModel
   ModelCustomizationsMixin <|-- EthereumTestBaseModel
   ModelCustomizationsMixin <|-- EthereumTestRootModel

   class ModelCustomizationsMixin{
    serialize()
    __repr_args__()
   }
```

src/cli/gentest/templates/blockchain_test/transaction.py.j2
raxhvl: Hey Mario - I don't quite follow this. Could you elaborate on a new issue? I will take it up separately. This PR has become quite big.

src/ethereum_test_base_types/pydantic.py
raxhvl: Done. Also linked you comment for future reference since `exclude_none` + `None` in match case could be could appear contradictory.

danceratopz (approved):
Really impressive, great addition! Thanks again @raxhvl for taking the time to find such an elegant solution for this!

danceratopz:
> I have tested this thoroughly, but please review it with care as this change impacts **EVERY Pydantic model** in the codebase. I'm especially unsure of [EthereumTestRootModel](https://github.com/ethereum/execution-spec-tests/blob/1fabe172b2388730b0ec4773e0d4fb7b21358dd0/src/ethereum_test_base_types/pydantic.py#L25-L30) override.

Hey @raxhvl, in regards to this, it looks good to me, I've asked @marioevz if he'd also like to take a look. I diff'd the fixtures locally and got an exact match.

src/cli/gentest/templates/blockchain_test/transaction.py.j2
marioevz: Basically, right now every test starts from the `Paris` fork, but it could be that the fork that was active when the transaction was executed was a newer fork, and therefore the transaction could execute differently or not even work in `Paris` because, e.g., some opcode that it uses was not enabled until after `Paris`.

marioevz (approved):
Many thanks for putting so much thought and care into this!

PR #911, opened 2024-10-23
new(tests): Explicit test for EXTDELEGATECALL value cost

pdobacz:
## 🗒️ Description

There has been divergence in EXTDELEGATECALL implementations, whereby effectively non-zero value call (non-zero value coming from an outer EXTCALL!) was charging an additional 9000. C.f. EIP update for background: https://github.com/ethereum/EIPs/pull/8990, in this PR we give extra emphasis that the additional cost, and other value-related rules, are only relevant to EXTCALL.

Unfortunately, due to the 63/64th rule on calls, I had to complicate `gas_test.py`  an additional bit :/. 

## 🔗 Related Issues
NA, reported on discord. We did have a test which covered this, but without an explicit post state assertion.

## ✅ Checklist


tests/osaka/eip7692_eof_v1/eip7069_extcall/test_gas.py
```diff
@@ -140,3 +140,36 @@ def test_ext_calls_gas(
...
+    value: int,
+):
+    """
+    Test that EXTDELEGATECALL call doesn't charge for value transfer, even if the outer call
+    transfered value.
+    """
+    noop_callee_address = pre.deploy_contract(Container.Code(Op.STOP))
+
+    extdelegatecall_contract_address = pre.deploy_contract(
+        Container.Code(Op.EXTDELEGATECALL(address=noop_callee_address) + Op.STOP)
```
marioevz, line 159: Should we test `Op.EXTSTATICCALL` too in the same way?

marioevz (commented):
Small comment, thanks!

tests/osaka/eip7692_eof_v1/eip7069_extcall/test_gas.py
pdobacz: We could, I don't think there would be any obstacles, except the "intent" of the test stems from EXTDELEGATECALL preserving value of the outer call. I can add EXTSTATICCALL and make a comment that it is just an addition for good measure. And actually, we can throw in EXTCALL without value arg set too, all 3 should have same effect (no 9000 charged). I think this makes sense, with an appropriate comment, one sec...

tests/osaka/eip7692_eof_v1/gas_test.py
```diff
@@ -104,27 +105,33 @@ def gas_test(
...
+                    + Op.SSTORE(
+                        slot_sanity_call_result,
+                        Op.CALL(
+                            gas=Op.ADD(warm_gas - gas_single_gas_run, Op.DUP7),
+                            address=address_subject,
+                        ),
+                    )
+                    + Op.STOP
+                )
+                if not skip_oog_test
```
shemnon, line 132: Style question:  Would 
```
Op.STOP if skip_oog_test
else <rest of code>
```
Be a better construction?  Do we want the longer clause first or the clause triggered by the flag first?
Alternately, could the flag be `out_of_gas_testing: bool = True` instead of `skip_oog_test` to appease both?

shemnon (approved):

tests/osaka/eip7692_eof_v1/gas_test.py
marioevz: +1 for the `out_of_gas_testing` flag.

PR #913, opened 2024-10-24
refactor(tests): EOF - EIP-4750: parametrize CALLF execution tests

chfast:
## 🗒️ Description

Use pytest parametrize to combine very similar tests for `CALLF` execution.

## ✅ Checklist

### DIFF
--- a/tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
+++ b/tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
@@ -22,12 +22,11 @@ REFERENCE_SPEC_VERSION = "14400434e1199c57d912082127b1d22643788d11"
 pytestmark = pytest.mark.valid_from(EOF_FORK_NAME)
 
 
-def test_callf_stack_size_1024(
-    eof_state_test: EOFStateTestFiller,
-):
-    """Test stack reaching 1024 items in called function"""
-    eof_state_test(
-        data=Container(
+@pytest.mark.parametrize(
+    "container",
+    (
+        Container(
+            name="no_inputs",
             sections=[
                 Section.Code(
                     code=Op.PUSH0 * 1023
@@ -45,16 +44,8 @@ def test_callf_stack_size_1024(
                 ),
             ],
         ),
-        container_post=Account(storage={slot_code_worked: value_code_worked}),
-    )
-
-
-def test_callf_with_inputs_stack_size_1024(
-    eof_state_test: EOFStateTestFiller,
-):
-    """Test stack reaching 1024 items in called function with inputs"""
-    eof_state_test(
-        data=Container(
+        Container(
+            name="with_inputs",
             sections=[
                 Section.Code(
                     code=Op.PUSH0 * 1023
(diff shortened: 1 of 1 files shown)


tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
```diff
@@ -144,90 +119,80 @@ def test_callf_stack_size_1024_at_push(
...
-):
-    """Test stack overflowing 1024 items in called function"""
-    eof_state_test(
-        data=Container(
+        Container(
+            name="nested_with_inputs_at_push0",
             sections=[
                 Section.Code(
-                    code=Op.PUSH0 * 1023
+                    code=Op.PUSH0 * 1022
```
chfast, line 126: This diff is because there was one test for `stack_size_max` after `stack_overflow`.

pdobacz (approved):

tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
```diff
@@ -144,90 +119,80 @@ def test_callf_stack_size_1024_at_push(
...
                     + Op.CALLF[1]
-                    + Op.POP * 1023
+                    + Op.POP * 1022
                     + Op.SSTORE(slot_code_worked, value_code_worked)
```
winsvega, line 129: there is Storage iterator class from Mario, check it out

tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
```diff
@@ -255,6 +220,13 @@ def test_callf_with_inputs_stack_overflow(
                 ),
             ],
         ),
+    ),
+    ids=lambda x: x.name,
+)
+def test_callf_operand_stack_overflow(eof_state_test: EOFStateTestFiller, container: Container):
+    """Test stack overflowing 1024 items in called function"""
```
winsvega, line 227: @marioevz  is there a possibility to catch this from evm?  that execution reverted exactly because of evm error stack overflow.  it's a separate kind of exception. because it is not transaction rejection.

winsvega (approved):
looks ok. however some comments for Mario.

tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
marioevz: I think we could achieve this by adding a field to the [`Transaction` class](https://github.com/ethereum/execution-spec-tests/blob/2939e0ac62f6efacb909cb778229ab77d07d866b/src/ethereum_test_types/types.py#L700), something like `expected_status`, and then compare it to the value in the receipt that we get from the transition tool. The verification could be done in the [`verify_transactions`](https://github.com/ethereum/execution-spec-tests/blob/2939e0ac62f6efacb909cb778229ab77d07d866b/src/ethereum_test_specs/helpers.py#L136) function.

One thing I noticed while checking this value before writing this comment is that Geth's evm tool does return `"0x0"` for failed transactions and `"0x1"` for successful ones, but `evmone-t8n` seems to always return `"0x1"`.

We need to do this in a separate PR though.

tests/osaka/eip7692_eof_v1/eip4750_functions/test_callf_execution.py
marioevz: I've never used it at the global scope either, so it's fine to leave these as is for the time being.

marioevz (approved):
Thanks! We can merge now and follow up on Dimitry's suggestion on a separate PR.

PR #933, opened 2024-10-31
feat(fixtures,specs,tests): Transaction tests, EIP-7702 convert invalid tx tests

marioevz:
### Transaction Test Format
Introduce the Transaction test type to EEST.

TODOs:
- [x] Correctly calculate the intrinsic gas of the transactions when valid
- [x] Documentation
- [ ] Migrate some tests from `ethereum/tests`

At the moment the spec only generates the transaction fixture type, and automatically generating blockchain and engine blockchain tests could be added in the future.

### Fixture format type in `_info`

A new info key `fixture_format` is added to the `_info` dictionary in the generated fixtures.

This is done in order to better discriminate between fixture formats, since the addition of the transaction test format was causing issues in `checkfixtures` command.

### Convert EIP-77022 Invalid Transaction Tests

This PR also converts all tests in `tests/prague/eip7702_set_code_tx/test_set_code_txs.py` to a transaction test, in order to not produce any errors during the parsing of the state tests.

I'm open to leave this out to a different PR to keep the commit granularity.

### Fix `Transaction(to=Address(0)).to == None`

Fixes a bug where a transaction with a destination address meant to be the zero address, ends up as a contract creating transaction instead.

## 🔗 Related Issues
None

## ✅ Checklist


winsvega:
Look at transaction test filler
It has no pre state.

We used to have t9n in geth to ask the client for transaction test validity

Now we have python implementation of transaction
So we can generate tests no problem.

Basically its like difficulty test. Lets say we test data field.
So we put all other tx fields to be of valid value. 

Then we iterate by forks by tx types.
And we iterate data by vector of tests.

Data is empty - tx valid
Data is 00 - valid
Data is 32 random bytes - valid
Data is max_tx_data sise - valid
Max size+1 - invalid
Data size  is uint64 +/- 1
Data size is uint128 +/-1
Data size is uint256 +/-1


The exception here is.
Intrinsic tx validation. So tx with gaslimit 0 is a valid structure. But it will be rejected as invalid because we validate intrinsic tx gaslimit rule

In this tests we don't care about pre or post state and out of funds exceptions or out of gas exceptions

marioevz:
> Look at transaction test filler It has no pre state.

Regarding this, I added the `pre` with the `execute` mode in mind:

We can test sending these faulty transactions to the `eth_sendRawTransaction` endpoint, but if the account that is recovered from the signature has no funds, we could get a false positive even if a buggy client processes the faulty transaction. Therefore we can use `pre.fund_eoa` to ensure the origin account always has funds in it.

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -37,6 +36,7 @@
     Storage,
     Transaction,
     TransactionException,
+    TransactionTestFiller,
```
winsvega, line 39: hm why this file is affected?  
I think we shall separate the transaction test fillers into separate folder because those are not converted into hive or bc format.

src/ethereum_test_specs/tests/fixtures/chainid_shanghai_blockchain_test_engine.json
```diff
@@ -1,7 +1,8 @@
 {
     "000/my_chain_id_test/Shanghai": {
         "_info": {
-            "hash": "0xfeded8f82a93725388c2436c76ea328cf2008dab43000de76306b3ed95de63b7"
+            "hash": "0xfeded8f82a93725388c2436c76ea328cf2008dab43000de76306b3ed95de63b7",
+            "fixture_format": "blockchain_test_engine"
```
winsvega, line 5: thats a good field.

winsvega (commented):
This will read existing .json files as well? 

also can be a good idea to export fields in readable format in the generated fixture json, not just tx bytes. 

the purpose of this test is to test rlp decode of txbytes. where txbytes can be ANYTHING
also the test pass if tx gasLimit * gasPrice  < 2**256, so a transaction with gasLimit * gasPrice == 2**255  is a valid one. (read from rlp works) 
so the pre defenition with this check will not work.

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
spencer-tb: Should we not have them in the bc or bc_engine format?
For the latter we can see if client correctly reject invalid txs via engine new payload

src/ethereum_test_specs/transaction.py
```diff
@@ -0,0 +1,108 @@
...
+from ethereum_clis import TransitionTool
+from ethereum_test_execution import BaseExecute, ExecuteFormat, TransactionPost
+from ethereum_test_fixtures import BaseFixture, FixtureFormat, TransactionFixture
+from ethereum_test_fixtures.transaction import Fixture, FixtureResult
+from ethereum_test_forks import Fork
+from ethereum_test_types import Alloc, Transaction
+
+from .base import BaseTest
+
+TARGET_BLOB_GAS_PER_BLOCK = 393216
```
danceratopz, line 18: I think this can be removed.

danceratopz (approved):
Really nice! Already looks really good, happy to get this merged and iterate further in subsequent PRs.

docs/writing_tests/types_of_tests.md
```diff
@@ -51,6 +52,21 @@ def test_blob_type_tx_pre_fork(
...
+Test correct transaction rejection/acceptance of a serialized transaction (currently RLP only).
+
+### Use cases
+
+- Verify that a badly formatted transaction is correctly rejected by the client.
+- Verify that a transaction with an invalid value in one of its fields is correctly rejected by the client.
+
+!!! info
+
+    Using the `execute` command, transaction tests can be sent to clients in a live network using the `eth_sendRawTransaction` endpoint.
```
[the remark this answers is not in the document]
marioevz: Actually it should work out of the box.

src/ethereum_test_specs/transaction.py
marioevz: Done!

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
marioevz: > Should we not have them in the bc or bc_engine format? For the latter we can see if client correctly reject invalid txs via engine new payload

I tried to add this but we need to update some error messages that we receive from the execution-specs, so we can leave this out for a follow-up PR.

PR #935, opened 2024-11-04
refactor(ethereum_clis): move `TransitionTool.verify_fixture()` to `FixtureConsumer`

danceratopz:
## 🗒️ Description
This PR refactors `ethereum_clis` and the `consume direct` interface so that it can be extended to other clients.

Previously, the `GethTransitionTool` class implemented fixture consumption and referred to it as "fixture verification". The reason is historical: the first "consumer" was implemented in `fill`.

This PR:
- Adds `GethStatetest` and `GethBlocktest` interface consumer classes to the `evm statetest`, respectively `evm_blocktest` commands.
- Adds a `GethFixtureConsumer` class which applies the appropriate consumer class for the fixture to run against the client.

Note, that as we now instantiate the `GethStatetest` and `GethBlocktest` classes from within `GethFixtureConsumer` we don't currently use any of the binary detection/automatic subclass instantiation provided by `EthereumCLI` that is used for the `TransitionTool`.

## 🔗 Related Issues
None.

## ✅ Checklist


src/ethereum_clis/clis/geth.py
```diff
@@ -287,3 +157,207 @@ def _mapping_data(self):
...
+    """
+    go-ethereum `evm` base class.
+    """
+
+    default_binary = Path("evm")
+    detect_binary_pattern = compile(r"^evm(.exe)? version\b")
+    binary: Path
+    cached_version: Optional[str] = None
+
+    def __init__(self, binary: Path, trace: bool = False, exception_mapper=GethExceptionMapper()):
```
winsvega, line 202: so we can actually override exception mapper for consume wrappers right?

src/ethereum_clis/clis/geth.py
```diff
@@ -287,3 +157,207 @@ def _mapping_data(self):
...
+                "consume_direct_returncode.txt": result.returncode,
+                "consume_direct_stdout.txt": result.stdout,
+                "consume_direct_stderr.txt": result.stderr,
+                "consume_direct.sh+x": consume_direct_script,
+            },
+        )
+        shutil.copyfile(fixture_path, debug_fixture_path)
+
+
+class GethTransitionTool(GethEvm, TransitionTool):
```
winsvega, line 245: I feel It is not necessarily GethEvm, it just happened to be inside gethEvm
for evmone it will be a standalone class?  it is a bit confusing

marioevz:
Summary of the latest commit:
- Rebased on top of main to get ruff
- Deleted `src/ethereum_clis/blocktest.py` and `src/ethereum_clis/statetest.py` in favor of a making a generic `FixtureConsumerTool` that can be specified to consume many different types of Fixtures, without having to create a new base class for every new fixture type.
- `FixtureConsumer` (in `ethereum_test_fixtures`) is now an abstract class and `FixtureConsumerTool` (in `ethereum_clis`) is an implementation that can automatically detect different fixture consumers binaries.

I'm thinking we can go all the way in this PR and accept multiple `--evm-bin` values in consume direct and let `FixtureConsumerTool` detect each, and then consume all fixtures with each different consumer automatically.

marioevz (approved):
This one was long overdue to be merged, sorry for the delay!

PR #932, opened 2024-10-31
new(tests): EOF validation of opcodes

chfast:
## 🗒️ Description

Add missing opcode tests:
- invalid opcode placed after a STOP instruction,
- PUSH opcodes with truncated immediate bytes.

Mark rest of the "opcode tests" as done by providing links to tests.

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [ ] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [ ] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: A PR with removal of converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been opened: https://github.com/ethereum/tests/pull/1412.
- [ ] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [ ] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.

PR #961, opened 2024-11-20
fix(fw): allow Bytecode(Opcode)

pdobacz:
## 🗒️ Description

Without this, I cannot fill eof tests with:

```
$ fill  --fork Osaka tests/osaka/eip7692_eof_v1/
...
___________________________________________________________ test_invalid_opcodes_after_stop[fork_Osaka-eof_test-terminating_opcode_STOP-opcode_OPCODE_0C] ___________________________________________________________
[gw3] linux -- Python 3.10.12 /home/user/sources/ethereum/execution-spec-tests/.venv/bin/python3
tests/osaka/eip7692_eof_v1/eip3540_eof_v1/test_all_opcodes_in_container.py:141: in test_invalid_opcodes_after_stop
    terminating_code = Bytecode(terminating_opcode)
src/ethereum_test_vm/bytecode.py:92: in __new__
    raise TypeError("Bytecode constructor '__new__' didn't return an instance!")
E   TypeError: Bytecode constructor '__new__' didn't return an instance!
```

I think #932 introduced it.

I'm not sure if this is the correct way to Python this. Alternative I have is to do `Bytecode() + some_opcode` instead of `Bytecode(some_opcode)` in the said test

## 🔗 Related Issues

na

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_vm/bytecode.py
+++ b/src/ethereum_test_vm/bytecode.py
@@ -57,7 +57,7 @@ class Bytecode:
             instance._name_ = name
             return instance
 
-        if type(bytes_or_byte_code_base) is Bytecode:
+        if isinstance(bytes_or_byte_code_base, Bytecode):
             # Required because Enum class calls the base class with the instantiated object as
             # parameter.
             obj = super().__new__(cls)


danceratopz (approved):
Thanks for this! :+1: 

Just for clarity using `isinstance()` is generally considered more idiomatic Python. `isinstance()` will also check against subclasses of `Bytecode`; `type()` does not.

danceratopz:
> I'm not sure if this is the correct way to Python this. Alternative I have is to do Bytecode() + some_opcode instead of Bytecode(some_opcode) in the said test

Definitely `Bytecode(some_opcode)` is preferred and we should apply your fix in this PR. 

The alternative seems to fix the error because `+` is overloaded in `Bytecode` which results in `Bytecode() + some_opcode` returning the expected type:
https://github.com/ipsilon/execution-spec-tests/blob/bf37249b38548ae672f822ee45c18866fc1975f6/src/ethereum_test_tools/vm/opcode.py#L155-L158

I still don't understand why `Bytecode(some_opcode)` doesn't work in this context - I didn't look into it further though, hope that's ok! Happy to look if you'd like more clarification.

Issue #940, opened 2024-11-05
impose test format verification after fill


winsvega:
looks like evmone-t8n  generated genesis with "baseFeePerGas": "0x00",  on frontier network

so we need to check the generated test format with consume. like I did with retesteth. if blockchain test is of network frontier.  the genesis must not have  other fork fields 
and vice versa. this is not being checked. I am getting test with genesis header with baseFeePerGas when using evmonet8n

danceratopz:
Currently evmone's `t8n` returns invalid block header fields for Frontier (to be double-checked by @winsvega) - this is blocking #808.

winsvega:
https://github.com/ethereum/evmone/issues/1073

winsvega:
@marioevz if you think it can be redundant perhaps it should stay at least in consume. 
I won't trust unit tests alone I think it is good to have a json reader validation scheme to yell an error if we detect malformed fixture. and this code better be separate from our internal filling logic to not be affected by it or for the case when we import generated json and don't know who has generated it.

marioevz:
> [@marioevz](https://github.com/marioevz) if you think it can be redundant perhaps it should stay at least in consume. I won't trust unit tests alone I think it is good to have a json reader validation scheme to yell an error if we detect malformed fixture. and this code better be separate from our internal filling logic to not be affected by it or for the case when we import generated json and don't know who has generated it.

I think [`FixtureHeader`](https://github.com/ethereum/execution-spec-tests/blob/9586dd9c8fd08a22a17d54253f26e594028fc5e2/src/ethereum_test_fixtures/blockchain.py#L74) is already our validator scheme, it's [broken at the moment](https://github.com/ethereum/execution-spec-tests/pull/968#issuecomment-2640023527) but the fix is simple in my opinion.

This class can be the source of truth, I don't see why we have to create more structures since this one can already verify.

PR #968, opened 2024-11-22
feat(tests): json schema verification

winsvega:
## 🗒️ Description
Pydanticfy the json schemas so we verify the quality of generated test fixtures

This is an MVP to catch what I think is the cause of bug

```
src/ethereum_test_fixtures/verify_format.py:28: in verify_blockchain_fixture_json
    raise Exception(
E   Exception: Error in generated blockchain test json (tests/frontier/opcodes/test_dup.py::test_dup[fork_Frontier-evm_code_type_LEGACY-blockchain_test-DUP1])[{"type":"extra_forbidden","loc":["baseFeePerGas"],"msg":"Extra inputs are not permitted","input":"0x00","url":"https://errors.pydantic.dev/2.9/v/extra_forbidden"}]
```

pyspecs with evmone-t8n export extra field on all forks which should not be there. perhaps this extra field leads to wrong block init in its' t8n when we blindly feeding the env section. and to a wrong state hash in my test scenarios. 

so the issue is 
evmone-t8n return 
```
"currentBaseFee": "0x0",
```
on all forks. and we export it to the final test. when this field is only in blockheader in exported test  since London

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/940

## ✅ Checklist


src/ethereum_test_fixtures/schemas/blockchain/headers.py
```diff
@@ -0,0 +1,190 @@
+"""
+Define genesisHeader schema for filled .json tests
+"""
+
+from typing import Any, Optional
+
+from pydantic import BaseModel, Field, model_validator
+from rlp import decode as rlp_decode
```
danceratopz, line 8: To fix the mypy issues, we should be able to extend [`./stubs/`](https://github.com/ethereum/execution-spec-tests/tree/2d361623337f300d36b067f99c418873cec7b4fa/stubs) folder with the functions used here. 

`rlp` - https://pypi.org/project/rlp/ doesn't appear to support typing yet, ethereum/pyrlp#143

winsvega: I added one, check if it is how it supposed to be. 
but still it does not detect in vscode where rlp is supposed to be imported from.

but strangely all works

marioevz:
A similar verification is already there, but this is the issue:
https://github.com/ethereum/execution-spec-tests/blob/83942b0011e32625ec8e74598673b1cce7b3db16/src/ethereum_test_fixtures/blockchain.py#L148-L153

If a header field is fork-dependent, it only checks that it is indeed there when required, but the opposite check is missing (field is there when it shouldn't).

I think adding this extra verification below these lines should in theory do the trick, and we could add a unit test just to double check.

PR #975, opened 2024-11-27
✨ feat(tests): PUSH*

### LINKED ISSUE #974: ✨feat(tests): `PUSH*` opcodes
Add tests for `PUSH*` opcodes. This should be done by porting [`pushFiller.yml`](https://github.com/ethereum/tests/blob/4f65a0a7cbecf4442415c226c65e089acaaf6a8b/src/GeneralStateTestsFiller/VMTests/vmTests/pushFiller.yml) into this  repository as part of #972.

raxhvl:
## 🗒️ Description
Introduces tests for `PUSH*` opcodes ported from `ethereum/tests`.

## 🔗 Related Issues
closes #974 

## ✅ Checklist

### DIFF
--- a/converted-ethereum-tests.txt
+++ b/converted-ethereum-tests.txt
@@ -57,3 +57,6 @@ GeneralStateTests/Cancun/stEIP1153-transientStorage/16_tloadGas.json
 GeneralStateTests/Cancun/stEIP1153-transientStorage/18_tloadAfterStore.json
 
 GeneralStateTests/Cancun/stEIP1153-transientStorage/20_oogUndoesTransientStoreInCall.json
+
+([#975](https://github.com/ethereum/execution-spec-tests/pull/975))
+src/GeneralStateTestsFiller/VMTests/vmTests/pushFiller.yml
--- a/converted-ethereum-tests.txt
+++ b/converted-ethereum-tests.txt
@@ -59,4 +59,4 @@ GeneralStateTests/Cancun/stEIP1153-transientStorage/18_tloadAfterStore.json
 GeneralStateTests/Cancun/stEIP1153-transientStorage/20_oogUndoesTransientStoreInCall.json
 
 ([#975](https://github.com/ethereum/execution-spec-tests/pull/975))
-src/GeneralStateTestsFiller/VMTests/vmTests/pushFiller.yml
+GeneralStateTests/VMTests/vmTests/push.json
--- a/tests/frontier/opcodes/test_push.py
+++ b/tests/frontier/opcodes/test_push.py
@@ -1,0 +1,62 @@
+"""
+A State test for the set of `PUSH*` opcodes.
+Ported from: https://github.com/ethereum/tests/blob/4f65a0a7cbecf4442415c226c65e089acaaf6a8b/src/GeneralStateTestsFiller/VMTests/vmTests/pushFiller.yml # noqa: E501
+"""
+
+import pytest
+
+from ethereum_test_forks import Frontier, Homestead
+from ethereum_test_tools import Account, Alloc, Environment
+from ethereum_test_tools import Opcodes as Op
+from ethereum_test_tools import StateTestFiller, Transaction
+
+
+@pytest.mark.parametrize(
+    "push_opcode",
+    [getattr(Op, f"PUSH{i}") for i in range(1, 33)],  # Dynamically parametrize PUSH opcodes
+    ids=lambda op: str(op),
+)
+@pytest.mark.valid_from("Frontier")
+def test_push(state_test: StateTestFiller, fork: str, push_opcode: Op, pre: Alloc):
+    """
+    The set of `PUSH*` opcodes pushes data onto the stack.
+
+    In this test, we ensure that the set of `PUSH*` opcodes writes
+    a portion of an excerpt from the Ethereum yellow paper to
+    storage.
+    """
+    # Input used to test the `PUSH*` opcode.
+    ethereum_state_machine = b"Ethereum is as a transaction-based state machine."
+
+    # Determine the size of the data to be pushed by the `PUSH*` opcode,
+    # and trim the input to an appropriate excerpt.
+    input_size = int(str(push_opcode)[4:])
+    excerpt = ethereum_state_machine[0:input_size]
+
+    env = Environment()
+
+    """
+     **               Bytecode explanation              **
+     +---------------------------------------------------+
+     | Bytecode      | Stack        | Storage            |
+     |---------------------------------------------------|
+     | PUSH* excerpt | excerpt      |                    |
+     | PUSH1 0       | 0 excerpt    |                    |
+     | SSTORE        |              | [0]: excerpt       |
+     +---------------------------------------------------+
+    """
+
+    contract_code = push_opcode(excerpt) + Op.PUSH1(0) + Op.SSTORE
+    contract = pre.deploy_contract(contract_code)
+
+    tx = Transaction(
+        sender=pre.fund_eoa(),
+        to=contract,
+        gas_limit=500_000,
+        protected=False if fork in [Frontier, Homestead] else True,
+    )
+
+    post = {}
+    post[contract] = Account(storage={0: int.from_bytes(excerpt)})
+
+    state_test(env=env, pre=pre, post=post, tx=tx)
--- a/tests/frontier/opcodes/test_push.py
+++ b/tests/frontier/opcodes/test_push.py
@@ -57,6 +57,6 @@ def test_push(state_test: StateTestFiller, fork: str, push_opcode: Op, pre: Allo
     )
 
     post = {}
-    post[contract] = Account(storage={0: int.from_bytes(excerpt)})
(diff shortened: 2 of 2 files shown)


raxhvl:
I don't understand how the contents of [converted-ethereum-tests.txt](https://github.com/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt) is ordered. Do I just add this at the end of the file?

tests/frontier/opcodes/test_push.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+
+    contract_code = push_opcode(excerpt) + Op.PUSH1(0) + Op.SSTORE
+    contract = pre.deploy_contract(contract_code)
+
+    tx = Transaction(
+        sender=pre.fund_eoa(),
+        to=contract,
+        gas_limit=500_000,
+        protected=False if fork in [Frontier, Homestead] else True,
```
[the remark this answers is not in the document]
danceratopz: Good observation, a possible solution could to make `tx` a pytest fixture (as `pre` is right now) which would allow automatic setting of the `protected` field based on the value of `fork` (which is also a pytest fixture). `tx` would have to be specified in the function arguments, as is `pre` right now.

I would like to add this functionality anyway, I think it's a possible solution that would allow generalization of transactions for other chains, unlocking EEST for L2 EVM testing.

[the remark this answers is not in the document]
raxhvl: Thats a nice solution. I'm noting things that feels against the grain as I port more tests. Should we create an issue for this?

raxhvl:
> This would even allow us to create a dedicated section in the docs, for example, that lists tests ported from ethereum/tests.
Yep! This would be really nice.

> Open to suggestions on this. I think the only source of ported tests would be ethereum/tests? But we could think about adding other metadata tags such as "fuzz", "gentest", "bug", etc.

Perhaps we can add a generic "meta" marker that lets you add arbitrary metadata?

converted-ethereum-tests.txt
```diff
@@ -47,3 +47,6 @@ GeneralStateTests/Cancun/stEIP1153-transientStorage/16_tloadGas.json
 GeneralStateTests/Cancun/stEIP1153-transientStorage/18_tloadAfterStore.json
 
 GeneralStateTests/Cancun/stEIP1153-transientStorage/20_oogUndoesTransientStoreInCall.json
+
+([#975](https://github.com/ethereum/execution-spec-tests/pull/975))
+src/GeneralStateTestsFiller/VMTests/vmTests/pushFiller.yml
```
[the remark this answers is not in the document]
winsvega: yes as you can see from other lines it expects the path to (old) filled test .json
it will be picked to run the coverage against .py generated test fixtures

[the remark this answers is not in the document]
danceratopz: Thanks!

danceratopz:
@raxhvl: Sometimes it can be tricky to get coverage parity, please reach out to @winsvega if you can't see anything obvious and need some pointers.

An example, although not applicable here, is if you port a YAML test with Yul Code to a Python test using `Opcodes`. In this case `solc` can optimize stack items increasing the "coverage" of the YAML-based test:
https://github.com/ethereum/execution-spec-tests/blob/5cf1f24b39ddef259470f58413337b81bda081a6/tests/homestead/coverage/test_coverage.py#L16-L29

raxhvl:
Thanks @danceratopz 

Let me give it a shot. I'm sure this will take some time, but I hope to pick up some learnings for future reference. If I get to a standstill, I will reach out to @winsvega.

raxhvl:
I spent last week trying to understand how coverage works, and why its failing here.

**TL;DR:** The current test uses fewer opcodes to achieve the same result.

##  Previous strategy 
The base test is parameterized using calldata. Here is a simplified pseudocode of how it works:
```python
for calldata in [0..1f]:
    call(
        gas=gas(), address=add(0x1000,calldata(4)), value=0, argOffset=0, argSize=0,
        returnOffset=0, returnSize=0
    )
```
In this setup, the each calldata is mapped to a contract address, which focuses on a single `PUSH` opcode. The `to` address then routes each test case to the appropriate contract.

## Current strategy

In the current approach, a unique contract is deployed for each PUSH opcode at the to address, eliminating the need for routing. Parameterization is handled directly by pytest.

## Why coverage fails

The absence of a routing mechanism in the current test causes coverage failures. The simplified structure of the test means that opcodes like `CALL`, `ADD`, and `CALLDATALOAD` are no longer required. For example, the following coverage failure occurs due to the missing calldataload call:

image

This is inherently because of the difference in strategy test strategies: parameterization using **EVM** vs parameterization using **pytest**.

## Larger picture
The base test employs a relatively complex strategy. Two areas of complexity include:

1. The calldata includes a function selector that is never used.
2. The calldata computes the destination contract address dynamically using an offset, which could be simplified by hardcoding the address.

Moreover, we should aim to test with a minimal set of opcodes—in this case, `PUSH*` and `SSTORE`—which would simplify the tests, make them faster, and avoid any potential unintended side effects.

With pytest now handling parameterization, the complex routing logic is no longer necessary.

cc: @winsvega @danceratopz  @marioevz

winsvega:
Yes, we have already some basic coverage that is applied by default. In test_all_opcodes.py.

But the goal of coverage here is to check that at least no lines lost after converting the test. 

Reading the coverage report already helped me to find out a few bugs where I relied that code works, but it didn't

tests/frontier/opcodes/test_push.py
```diff
@@ -0,0 +1,106 @@
...
+def test_push(state_test: StateTestFiller, fork: str, push_opcode: Op, pre: Alloc):
+    """
+    The set of `PUSH*` opcodes pushes data onto the stack.
+
+    In this test, we ensure that the set of `PUSH*` opcodes writes
+    a portion of an excerpt from the Ethereum yellow paper to
+    storage.
+    """
+    # Input used to test the `PUSH*` opcode.
+    ethereum_state_machine = b"Ethereum is as a transaction-based state machine."
```
marioevz, line 28: Nice.

tests/frontier/opcodes/test_push.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Determine the size of the data to be pushed by the `PUSH*` opcode,
+    # and trim the input to an appropriate excerpt.
+    input_size = int(str(push_opcode)[4:])
```
marioevz, line 32: `push_opcode.data_portion_length` should give you the size.

tests/frontier/opcodes/test_push.py
```diff
@@ -0,0 +1,106 @@
...
+from ethereum_test_tools import Opcodes as Op
+
+
+@pytest.mark.parametrize(
+    "push_opcode",
+    [getattr(Op, f"PUSH{i}") for i in range(1, 33)],  # Dynamically parametrize PUSH opcodes
+    ids=lambda op: str(op),
+)
+@pytest.mark.valid_from("Frontier")
+def test_push(state_test: StateTestFiller, fork: str, push_opcode: Op, pre: Alloc):
```
marioevz, line 19:
```suggestion
def test_push(state_test: StateTestFiller, fork: Fork, push_opcode: Op, pre: Alloc):
```
Then must import `Fork` from `ethereum_test_forks`.

tests/frontier/opcodes/test_push.py
```diff
@@ -0,0 +1,106 @@
...
+
+    post = {}
+    post[contract] = Account(storage={0: int.from_bytes(excerpt, byteorder="big")})
+
+    state_test(env=env, pre=pre, post=post, tx=tx)
+
+
+@pytest.mark.parametrize("stack_height", range(1024, 1026))
+@pytest.mark.valid_from("Frontier")
+def test_stack_overflow(state_test: StateTestFiller, fork: str, pre: Alloc, stack_height: int):
```
marioevz, line 66:
```suggestion
def test_stack_overflow(state_test: StateTestFiller, fork: Fork, pre: Alloc, stack_height: int):
```

tests/frontier/opcodes/test_push.py
```diff
@@ -0,0 +1,106 @@
...
+        protected=False if fork in [Frontier, Homestead] else True,
+    )
+
+    post = {}
+    post[contract] = Account(storage={0: int.from_bytes(excerpt, byteorder="big")})
+
+    state_test(env=env, pre=pre, post=post, tx=tx)
+
+
+@pytest.mark.parametrize("stack_height", range(1024, 1026))
```
marioevz, line 64: We should also parametrize by opcode for each `PUSH*` opcode, just as in the previous test.

marioevz (commented):
A few minor comments but overall looks pretty good, thanks!

tests/frontier/opcodes/test_push.py
raxhvl: Sweet!

raxhvl:
All done @marioevz!

marioevz (approved):
Awesome, thanks so much!

Issue #972, opened 2024-11-25
tracker(static tests): port ethereum/tests test cases


danceratopz:
This is a top-level issue for tracking the manual porting of ethereum/tests test cases to execution-spec-tests:

> **Contributions to manually port individual test cases are very welcome!**

**If you're already familiar with ethereum/tests and execution-spec-tests, jump to the issue tracker below!**

#### Background
[ethereum/tests](https://github.com/ethereum/tests) was the repository used to define EVM test cases from [Frontier](https://ethereum.org/en/history/#frontier) up to and including [The Merge](https://ethereum.org/en/history/#paris). The test cases are defined as YAML (and sometimes JSON) files in the [./src/ sub-directory](https://github.com/ethereum/tests/tree/develop/src) and the JSON "test fixtures" (fully-populated tests that can be executed against clients) are generated using [ethereum/retesteth](https://github.com/ethereum/retesteth). These JSON artifacts are regenerated when required and added to the repository, for example, in the [./GeneralStateTests/ sub-directory](https://github.com/ethereum/tests/tree/develop/GeneralStateTests).

From [Shanghai](https://ethereum.org/en/history/#shapella) on, new test cases (in particular for new features included in hard-forks) have been defined in Python in this repository (execution-spec-tests). The existing test cases are still important to execute against clients, however, and porting ethereum/tests to execution-spec-tests ensures that the test cases can be continually maintained and filled (generated) for newer forks. It also has the advantage that client teams will only need to obtain test fixture releases from one source.

#### Automated Port to Python
Although there will be an effort to automate ("transpile") the remaining test cases from YAML (respectively JSON) to Python; contributions to manually port individual test cases can be very beneficial and are welcome! 

The benefits of manual porting are:
1. Reduce the number of test cases by combing multiple YAML (resp. JSON) test cases in one Python test function (via parametrization).
2. Potentially improve coverage by parametrizing the Python version.
3. Code/docstrings are likely higher quality than an automated conversion.
4. The test can be better [organized within the `./tests` folder](https://ethereum.github.io/execution-spec-tests/main/writing_tests/adding_a_new_test/) in execution-spec-tests by the fork and the EIP in which the functionality was introduced.

#### Process
1. Pick a single or a set of YAML test cases in ethereum/tests to port and create an issue in this repository - optional, but definitely recommended if porting tests from the tracker below to avoid duplicate work.
2. Create a PR with the ported tests:
[...]

PR #977, opened 2024-11-28
feat(tests): STOP- assignment- Add Tests for STOP Opcode and Blob Data Handling in Shard Blob Transactions for EIP-4844

jcqln2:
## 🗒️ Description
This test covers:
- Transactions containing blob data with KZG commitments and proofs
- Validation of STOP opcode (0x00) in transaction data
- Execution simulation to verify that transactions with the STOP opcode halt as expected

## 🔗 Related Issues
- https://github.com/ethereum/execution-spec-tests/issues/972

### DIFF
--- a/tests/cancun/eip4844_blobs/eip4844_stopblob.py
+++ b/tests/cancun/eip4844_blobs/eip4844_stopblob.py
@@ -1,0 +1,158 @@
+"""
+abstract: Tests the `STOP` opcode and blob data handling in transactions for [EIP-4844: Shard Blob Transactions](https://eips.ethereum.org/EIPS/eip-4844) during the Cancun fork.
+    This test ensures proper inclusion of blob data, verifies the functionality of the `STOP` opcode, and validates transaction execution
+    behavior
+"""  # noqa: E501
+from typing import List
+
+import pytest
+
+from ethereum_test_tools import (
+    Block,
+    BlockchainTestFiller,
+    Transaction,
+    Alloc,
+    Environment,
+)
+
+from .common import Blob
+from .spec import Spec, SpecHelpers, ref_spec_4844
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_4844.git_path
+REFERENCE_SPEC_VERSION = ref_spec_4844.version
+
+
+@pytest.fixture
+def txs_blobs() -> List[List[Blob]]:
+    """
+    Default blob data for transactions
+    Each transaction has one blob
+    """
+    ZERO_COMMITMENT_PLACEHOLDER = b"\x00" * 48  # 48-byte zero array placeholder
+    return [
+        [
+            Blob(
+                blob=bytes(
+                    Spec.FIELD_ELEMENTS_PER_BLOB * SpecHelpers.BYTES_PER_FIELD_ELEMENT
+                ),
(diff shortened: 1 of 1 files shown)


danceratopz (commented):
Thanks for the PR @jcqln2! I appreciate your ambition in tackling a complex task like testing an opcode within a blob transaction!

While your `test test_stop_opcode_with_transaction` fills correctly and passes when executed against clients, it seems that it's not testing the opcode execution as intended. The reason is that the bytecode provided in the calldata (`tx.data`) isn't executed when sent to an EOA address. In Ethereum, only contract accounts execute code upon receiving a transaction.

In your current implementation, the transaction sends calldata to an EOA (`to="0x000000000000000000000000000000000000dead"`), but EOAs do not execute code. Although a contract creation transaction would execute the initcode that is set as part of `tx.data`, as per the [EIP-4844 specification](https://github.com/ethereum/EIPs/blob/70471d02d48a81ca963407abe9c48706059dc8e8/EIPS/eip-4844.md?plain=1#L107), type 3 transactions (blob transactions) cannot be contract creation transactions (where `to=None`). This means we need an alternative approach to test opcode execution within a blob transaction.

Here are a couple of paths you might explore:

1. To align the test with your original intention: Deploy a contract with the desired bytecode in the pre-state allocation. Then, send a transaction to this contract's address to trigger code execution. You can refer to this example where a contract is deployed and interacted with: https://github.com/ethereum/execution-spec-tests/blob/85393456e75f5f3b476cb5e61c35d48172f9d543/tests/istanbul/eip1344_chainid/test_chainid.py#L28-L37

2. Alternatively, modify the test to focus on a specific behavior that is permissible within the constraints of EIP-4844.

If time permits, to verify what the EVM is executing during the test, you can use Geth's `evm` tool.

Here's how you can use it:

1. Install/download geth (note geth is no longer a prerequisite for filling - this is an **old version of the docs**): https://ethereum.github.io/execution-spec-tests/v3.0.0/getting_started/quick_start/

2. Fill the test using Geth's evm t8n command instead of EELS to get execution traces:
    ```
    fill tests/cancun/eip4844_blobs/eip4844_stopblob.py --evm-bin=~/bin/go-ethereum/bin/evm --traces --evm-dump-dir=/tmp/debug
    ```

3. Inspect the trace file to see if the EVM executed any code. If the EVM didn't use any gas, it indicates no execution took place (`/tmp/debug/cancun__eip4844_blobs__eip4844_stopblob__test_stop_opcode_with_transaction/fork_Cancun_blockchain_test/0/trace-0-0xce0288fdff9e0b7b3ec1bb8663b645d651da5a54df6941403dc1cc604267db2d.jsonl`).

Alternatively, you can use `consume` to execute the test against Geth's evm blocktest command:
```
consume direct --input=fixtures/blockchain_tests/cancun/eip4844_blobs/eip4844_stopblob/  --evm-bin=~/bin/go-ethereum/bin/evm -k blockchain_test -v
```

jcqln2:
Thanks for the feedback, I'll  try to modify and commit changes today by my midday (EST)

danceratopz:
> Thanks for the feedback, I'll try to modify and commit changes today by my midday (EST)

As I mentioned in DMs, feel free to take as much time as you like/can afford :+1:

PR #995, opened 2024-12-05
fix(plugins/filler): reuse cache in test_filler.py

marioevz:
## 🗒️ Description
Attempt to fix flakiness in `src/pytest_plugins/filler/tests/test_filler.py` by reusing the resolver cache across tests.

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/filler/tests/test_filler.py
+++ b/src/pytest_plugins/filler/tests/test_filler.py
@@ -5,14 +5,26 @@ Test the forks plugin.
 import configparser
 import json
 import os
+import tempfile
 import textwrap
 from datetime import datetime
 from pathlib import Path
 
+import platformdirs
 import pytest
 
 from ethereum_clis import ExecutionSpecsTransitionTool, TransitionTool
 from pytest_plugins.filler.filler import default_output_directory
+
+
+@pytest.fixture(scope="session")
+def cache_temp_dir():
+    """
+    Return a temporary directory for caching files.
+    """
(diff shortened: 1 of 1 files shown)


danceratopz (commented):
Good work on finding the cause of the CI fails. 

So, as far as I understand, the resolver downloads the EELS branch for the fork if:
- It's currently not cached, or if,
- Some resolver-specific configured time delta has elapsed.

In which case, even if the tests are "new" because we re-write the modules, the resolver should still take the cached fork from the `<user_profile>/.cache/ethereum-spec-evm-resolver` directory? Did I misunderstand something? If not, this looks like it should be fixed in the resolver?

Either way, messing with the temporary directories that `pytester` uses to execute `pytest` from is a real battle from prior experience, so I proposed an alternative solution in #997. 

Not a fan of having a different way of executing things between CI and developer, but #997 can be improved if you like it, or we see if it can be fixed in EELS.

danceratopz:
Closing in favor of #997.

danceratopz:
> In which case, even if the tests are "new" because we re-write the modules, the resolver should still take the cached fork from the `<user_profile>/.cache/ethereum-spec-evm-resolver` directory? Did I misunderstand something? If not, this looks like it should be fixed in the resolver?

The answer was: `pytester` doesn't use the user-specified value of `platformdirs.user_cache_dir()`, it uses a directory relative to the `temp_dir` used by the `pytester` pytest session. Perhaps it monkeypatches it.

PR #1027, opened 2024-12-17
Export int as hex for t8n env

winsvega:
## 🗒️ Description
Export values in it's hex representation when interacting with t8n

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/992

## ✅ Checklist


winsvega:
But this affects all exports. Do we have issues with that?

marioevz:
I think the solution to this is not changing the `Number` to be `HexNumber` but to use `HexNumber` where we want to see hexadecimals.

winsvega:
XNumber type is to be exported as hex not Number

marioevz:
I just checked the failing unit test and it seems to be an error in EELS, so the change is breaking the interface.

I think that we need to implement some changes in EELS before this change goes in, otherwise we might break test generation.

Another thing is that I compared the generated fixtures for Cancun before and after this PR and there are some differences, so the change is somehow ending up in the fixtures too.

winsvega:
the fixtures were being generated with error in evmone because it treats all input as hex. 
so we provide 131072  dec input meaning value of (0x020000)  but evmone reads it as 0x131072  hex and that will affect the fixtures

gurukamath:
> I think that we need to implement some changes in EELS before this change goes in, otherwise we might break test generation.

Could you point me to a particular json fixture where I can replicate the issue?

winsvega:
> > I think that we need to implement some changes in EELS before this change goes in, otherwise we might break test generation.
> 
> Could you point me to a particular json fixture where I can replicate the issue?

@gurukamath  try this command described in here 
https://github.com/ethereum/execution-specs/issues/1078

PR #1028, opened 2024-12-18
 fix(eof): eofwrap.py to ignore invalid blocks 

pdobacz:
## 🗒️ Description

There is no support for invalid FixtureTransactions to be read from JSON, so we cannot include invalid blocks easily. However as this is not central to the behavior eofwrap tests are intended to test, we can just skip these invalid blocks with the same end effect on the test.

~Unfortunately, the fix isn't limited to that. It seems like the fix to requests serialization/hashing (#990) missed a detail when the `requests_lists` is used to construct requests (like from t8n output). Please double check carefully if this fix is fine.~

~Note also that for devnet-5 this would be partially hidden by the fact, that we don't include empty requests in the commitment anymore, meaning that all empty requests will not trigger the problem~

EDIT: request handling remains unchanged, see thread below

## 🔗 Related Issues

NA

## ✅ Checklist

### DIFF
--- a/src/cli/eofwrap.py
+++ b/src/cli/eofwrap.py
@@ -91,6 +91,8 @@ class EofWrapper:
     FIXTURES_CANT_WRAP = "fixtures_cant_wrap"
     # Test fixtures with EOF code but test doesn't pass and generation fails
     FIXTURES_CANT_GENERATE = "fixtures_cant_generate"
+    # Invalid blocks in fixtures skipped
+    INVALID_BLOCKS_SKIPPED = "invalid_blocks_skipped"
     # State accounts with code wrapped into valid EOF
     ACCOUNTS_WRAPPED = "accounts_wrapped"
     # State accounts with code wrapped into valid unique EOF
@@ -111,6 +113,7 @@ class EofWrapper:
             self.FIXTURES_GENERATED: 0,
             self.FIXTURES_CANT_WRAP: 0,
             self.FIXTURES_CANT_GENERATE: 0,
+            self.INVALID_BLOCKS_SKIPPED: 0,
             self.ACCOUNTS_WRAPPED: 0,
             self.UNIQUE_ACCOUNTS_WRAPPED: 0,
             self.ACCOUNTS_INVALID_EOF: 0,
@@ -283,15 +286,15 @@ class EofWrapper:
                         **fixture_tx_dump,
                     )
                     block.txs.append(tx)
+
+                test.blocks.append(block)
             elif isinstance(fixture_block, InvalidFixtureBlock):
-                block = Block(
-                    rlp=fixture_block.rlp,
-                    exception=fixture_block.expect_exception,
-                )
+                # Skip - invalid blocks are not supported. Reason: FixtureTransaction doesn't
+                # support expected exception. But we can continue and test the remaining
+                # blocks.
+                self.metrics[self.INVALID_BLOCKS_SKIPPED] += 1
             else:
                 raise TypeError("not a FixtureBlock")
-
-            test.blocks.append(block)
 
         result = test.generate(
             request=None,  # type: ignore


pdobacz:
It seems the request hash problem affects filling `state_tests` with `evmone` too, presumably with other t8n's too?

chfast:
This fixes filling of EOF tests for me.

src/ethereum_test_types/types.py
```diff
@@ -1248,12 +1248,17 @@ def __init__(
...
+        ascending, starting from 0.
         """
         if requests_lists is not None:
             assert len(requests) == 0, "requests must be empty if list is provided"
             self.requests_list = []
-            for requests_list in requests_lists:
-                self.requests_list.append(requests_list_to_bytes(requests_list))
+            for type, requests_list in enumerate(requests_lists):
+                self.requests_list.append(
+                    Bytes(bytes([type]) + requests_list_to_bytes(requests_list))
```
marioevz, line 1260: This breaks EIP-7685 tests, because of this change in the spec: https://github.com/ethereum/EIPs/commit/5ac0297b9ac29255e785ff305feea940db5ac6b4

I think the requests lists from the blocks that are being parsed and passed to this function might be outdated, so the source might need to be fixed first.

marioevz (commented):
One comment that is a breaking issue, there must be some other approach to this.

src/ethereum_test_types/types.py
pdobacz: Ha, I thought this change isn't yet implemented. My change is only about parity between the two ifs in this function here, but still in prague-devnet-4 world.

pdobacz: ok, hold on, I think there indeed is another solution - never pass `b''` elements in the `requests_lists`, and then it works as prague-devnet-5. I think I was wrong about that particular piece, some debugging gone wrong on my part.

I can propose an alternative fix in a second

pdobacz: OK, take a look now pls. I still ended up adjusting this function instead of source (I tried that too, but then it is problematic to work out the type of the request lists entry).

I might still be deeply confused here, I was certain I knew what was going on, but not anymore :sweat_smile:.

src/ethereum_test_types/types.py
```diff
@@ -1248,12 +1248,18 @@ def __init__(
...
         """
         if requests_lists is not None:
             assert len(requests) == 0, "requests must be empty if list is provided"
             self.requests_list = []
-            for requests_list in requests_lists:
-                self.requests_list.append(requests_list_to_bytes(requests_list))
+            for type, requests_list in enumerate(requests_lists):
+                if requests_list:
+                    self.requests_list.append(
+                        Bytes(bytes([type]) + requests_list_to_bytes(requests_list))
```
marioevz, line 1261: This still breaks our test generation because this is the code path we use to process the list returned by the t8n tool, and the list returned by t8n (correctly) already contains the type byte at the beginning of each request.

I think the interface could definitely be improved, and it could be that we need a different parameter for this use case in this PR, and a different one for the values returned by t8n.

It would also help me to better understand this use case with an example test that causes an issue when wrapped with an EOF container.

pdobacz: Okay, I'm back to this after holidays, sorry about the wait @marioevz .

evmone's t8n doesn't include the type in the result, and it does include empty requests. (mind that in the commitment it follows the spec, i.e. empty requests are excluded from it per [the recent change](https://github.com/ethereum/EIPs/pull/8989)). And from what I understand, the eofwrap use case also works on the t8n output.

Where is the t8n output format formally specified / agreed upon?

(I'll try to produce an example)

pdobacz: Meanwhile, I think I've found the answer to my own question - t8n follows the execution-apis as spec https://github.com/ethereum/execution-apis/pull/599/files ?

As such I guess I should fix on `evmone`'s end, let me try that.

pdobacz: OK, that worked, I force-pushed those commits away from this PR and fixed in `evmone`. Thank you for the thorough review and patience, I really appreciate this.

src/ethereum_test_types/types.py
pdobacz: See other thread for continuation

marioevz (approved):
Thanks! We can revisit the requests issue in a follow up PR if that still is necessary, I think it should be solved at a different level to be honest, but we can discuss about it further.

PR #1036, opened 2024-12-19
new(tests): Add generic precompile-absence test

marioevz:
## 🗒️ Description
Adds a simple test that, starting from Byzantium, calls all addresses less than or equal to `0x101` with:
- zero calldata
- 31 bytes of calldata
- 32 bytes of calldata

and expects success from every call.

Normally if a pre-compile is present in the address, it should return failure if the size is not what it expects, so in this test we use three different sizes to try to ensure that at least one of them is an invalid size to a would-be accidentally-active pre-compile in one of these addresses.

We also check the `RETURNDATASIZE` to be zero.

## 🔗 Related Issues
https://github.com/ethereum/EIPs/pull/8945

## ✅ Checklist


tests/frontier/precompiles/test_precompile_absence.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    active_precompiles = fork.precompiles()
+    storage = Storage()
+    call_code = Bytecode()
+    for address in range(1, UPPER_BOUND + 1):
+        if Address(address) in active_precompiles:
+            continue
+        call_code += Op.SSTORE(
+            address,
+            Op.CALL(address=address, args_size=calldata_size),
```
winsvega, line 43: So the call is 1 in either case. Is it?

Also, need to check that precompiles before are also precompiles

winsvega (commented):
But the call return 1 in either case if you call  defined or undefined precompile?

winsvega:
I sent random bytes data to precompiles with the call. And if it's a precompile it consumes more gas than a call to empty account

tests/frontier/precompiles/test_precompile_absence.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+        storage[address] = 1
+        if Op.RETURNDATASIZE in fork.valid_opcodes():
+            call_code += Op.SSTORE(
+                address + RETURNDATASIZE_OFFSET,
+                Op.RETURNDATASIZE,
```
marioevz, line 49:
```suggestion
                Op.RETURNDATASIZE(),
```
To commit the suggestion above.

tests/frontier/precompiles/test_precompile_absence.py
marioevz: The idea is that the call should fail if there's a precompile in the address. If there's no precompile, the call will return without executing anything successfully.

I think I can make this more reliable by also passing zero gas in the call, then there is absolutely no way the call can succeed if there's a precompile.

> Also, need to check that precompiles before are also precompiles

That should be tested in a separate dedicated test that aims to test each specific EIP that adds the precompile. It's outside the scope of this PR.

marioevz:
@winsvega I changed the call to have zero gas, so it will undoubtedly fail in case there's a precompile where it should not. It worked great.

I checked by removing the BLS precompiles in the EEST's forks definitions but not in EELS, and this test failed as expected.

winsvega:
Ah yes, can restrict the gas, I shall go an correct my test )

tests/frontier/precompiles/test_precompile_absence.py
```diff
@@ -0,0 +1,72 @@
+"""
+abstract: Test Calling Precompile Range (close to zero)
+
```
marioevz, line 3:
```suggestion
```

winsvega (approved):

PR #1037, opened 2024-12-19
feat(clis,specs): Tests marked as `slow` have greater t8n server `timeout`

marioevz:
## 🗒️ Description

Allow tests marked as slow (with `@pytest.mark.slow`) to have a higher timeout when making a request to the t8n server.

I was writing a test to verify the maximum discount BLS multi-scalar-multiplication and it takes the t8n server 30 seconds to respond, which exceeds the 20 second timeout that is hard-coded by default.

## 🔗 Related Issues
None

## ✅ Checklist


winsvega:
That's a good flag, but not to confuse with time consuming. Some tests we have just run so many loops or have so many transactions, I marked them as time-consuming to be able to disable from test runs if needed

winsvega (approved):
Also timeout is same for all t8ns?

marioevz:
> not to confuse with time consuming

Indeed, I think  it should be fine by now to not be so granular, but if required we could prepare separate markers for:
1) A single t8n call is the cause for execution delay
2) Test requires multiple block fills and hence the slowness

> Also timeout is same for all t8ns?
This might be EELS only, but need to check with other fillers

PR #1040, opened 2024-12-21
feat(forks,specs,fixtures): add eip-7840 blob schedule config

marioevz:
## 🗒️ Description
Adds blob schedule as described in this PR: https://github.com/ethereum/EIPs/pull/9129 and https://github.com/ethereum/EIPs/pull/9240

### Transition Tool Interface Changes

- Adds `blobSchedule` to the `state` object passed when using t8n in server mode.
- The blob schedule is not passed to any tools that only work via command line, as they need to update support first.

E.g.
```json
"state": {
    "fork": "Cancun",
    "chainid": 1,
    "reward": 0,
    "blobSchedule": {
            "Cancun": {
                "target": "0x3",
                "max": "0x6",
                "baseFeeUpdateFraction": "0x32f0ed"
            }
        }
  }
```

### State Test Fixture Format Changes

Adds a `config` field to the root of the every fixture contain in the state tests JSON fixture files. The object inside the `config` field contains a `blobSchedule` field that contains the blob schedule.

```json
{
    "tests/cancun/eip1153_tstore/test_basic_tload.py::test_basic_tload_after_store[fork_Cancun-state_test]": {
        "env": {
            ...
        },
        "pre": {
            ...
        },
        "transaction": {
            ...
        },
        "post": {
            ...
        },
        "config": {
            "blobSchedule": {
                "Cancun": {
                    "max": "0x06",
                    "target": "0x03",
                    "baseFeeUpdateFraction": "0x32f0ed"
                }
            },
        },
        "_info": {
            ...
        }
    }
}
```

### Blockchain Test Fixture Format Changes

Adds a `config` field to the root of the every fixture contain in the blockchain and blockchain engine tests JSON fixture files.

The object inside the `config` field contains a `blobSchedule` field that contains the blob schedule, and also a `network` field, which is a copy of the `network` field in the root of the fixture. The `network` field at the root of the fixture is set to be deprecated.

```json
{
    "tests/cancun/eip1153_tstore/test_basic_tload.py::test_basic_tload_after_store[fork_Cancun-blockchain_test]": {
        "network": "Cancun",
        "genesisBlockHeader": {
            ...
        },
        "pre": {
            ...
        },
        "postState": {
            ...
        },
        "lastblockhash": "...",
        "config": {
            "network": "Cancun",
            "blobSchedule": {
                "Cancun": {
                    "max": "0x06",
                    "target": "0x03",
                    "baseFeeUpdateFraction": "0x32f0ed"
                }
            },
        },
        "genesisRLP": "...",
        "blocks": [
            ...
        ],
        "sealEngine": "NoProof",
        "_info": {
            ...
        }
    }
}
```

**Note:** The format of the `blobSchedule` elements is not exactly as represented in the EIP:
- The EIP uses integers to represent the values of the numbers, but zero-padded hexadecimal strings are used in the fixtures to maintain the format of the rest of the file.
- The fork name has the first letter capitalized instead of all-lowercase.

### Remove Target Blobs from Block Header

Target blobs fields are completely removed from the header fields throughout the repository, which was only used for EIP-7742.

### `parent` method in `Fork`

All forks now contain the `parent` method which returns the parent fork or `None` for Frontier exclusively.

## ✅ Checklist


marioevz:
PR is ready for review but all changes need to be documented, and a changelog entry needs to be added.

src/ethereum_test_specs/tests/fixtures/chainid_cancun_blockchain_test_engine.json
```diff
@@ -0,0 +1,121 @@
...
+                "storage": {}
+            }
+        },
+        "blobSchedule": {
+            "cancun": {
+                "max": "0x06",
+                "target": "0x03",
+                "baseFeeUpdateFraction": "0x32f0ed"
+            }
+        }
```
danceratopz, line 119: I don't think it's strictly necessary to add the blob schedule to `blockchain_test_engine` fixtures. Although it doesn't hurt, I don't think they will be used. These parameters will be additionally defined in the hive "ruleset" (resp `mapper.jq`s) and communicated to the client via generation of the `genesis.json`. Is that correct?

danceratopz (commented):
One quick question below.

src/ethereum_test_specs/tests/fixtures/chainid_cancun_blockchain_test_engine.json
danceratopz: Ah (at least for the eest hive simulators) we could dynamically set the values of the necessary environment variables with these fixtures values for the client to use. Was that your thinking?

marioevz: Indeed, right now everything is hard-coded, but I think at some point we can generate tests where these values are dynamic.

src/ethereum_clis/types.py
```diff
@@ -93,3 +101,19 @@ class TransitionToolOutput(CamelModel):
     alloc: Alloc
     result: Result
     body: Bytes | None = None
+
+
+class TransitionToolState(CamelModel):
+    """Transition tool state."""
```
winsvega, line 107: the naming here looks misleading. because the first thing I think of is ethereum state

src/ethereum_clis/types.py
```diff
@@ -93,3 +101,19 @@ class TransitionToolOutput(CamelModel):
...
+
+    fork: str
+    chain_id: int = Field(..., alias="chainid")
+    reward: int
+    blob_schedule: BlobSchedule | None
+
+
+class TransitionToolPost(CamelModel):
```
winsvega, line 115: it has input, but again reading Post I would expect it to be post Alloc

src/ethereum_test_base_types/composite_types.py
```diff
@@ -462,3 +462,29 @@ class AccessList(CamelModel):
     def to_list(self) -> List[Address | List[Hash]]:
         """Return access list as a list of serializable elements."""
         return [self.address, self.storage_keys]
+
+
+class ForkBlobSchedule(CamelModel):
+    """Representation of the blob schedule of a given fork."""
```
winsvega, line 468: is this applied retrospectively to cancun tests?

src/ethereum_test_specs/tests/fixtures/chainid_cancun_state_test.json
```diff
@@ -0,0 +1,91 @@
...
+            "currentGasLimit": "0x02540be400",
+            "currentNumber": "0x01",
+            "currentTimestamp": "0x03e8",
+            "currentRandom": "0x0000000000000000000000000000000000000000000000000000000000000000",
+            "currentDifficulty": "0x00",
+            "currentBaseFee": "0x07",
+            "currentExcessBlobGas": "0x00"
+        },
+        "blobSchedule": {
+            "cancun": {
```
winsvega, line 18: ok since we changin the state format anyway. I have rather radical proposal. 
if it is 1 test per file now. then it could have network field as blockchain test.
an no need to create subdict of cancun here and in other places.
and post can be postState directly without list of forks. 
and transaction could have no arrays.

src/ethereum_test_base_types/composite_types.py
danceratopz: Yes.

src/ethereum_clis/types.py
danceratopz: agreed

danceratopz, line 107: @winsvega how about?
```suggestion
class TransitionToolConfig(CamelModel):
    """Transition tool configuration."""
```

danceratopz, line 107: Or?
```suggestion
class TransitionToolContext(CamelModel):
    """Transition tool context."""
```

src/ethereum_test_base_types/composite_types.py
danceratopz: Here's an example included in the test fixtures of this PR:
https://github.com/ethereum/execution-spec-tests/blob/c73f68596c8a47a2370ddc2f9e6c74e56d8327ce/src/ethereum_test_specs/tests/fixtures/chainid_cancun_blockchain_test.json#L131-L136

src/ethereum_test_specs/tests/fixtures/chainid_cancun_state_test.json
danceratopz: I like this idea, but whereas adding one new top-level field is a relatively easy change for client teams, larger modifications of the fixture format wouldn't be welcomed by them at this already somewhat hectic stage of development. 

Perhaps we can discuss and schedule this for a calmer period (hopefully in a few weeks)? How about creating a follow-up issue.

danceratopz (commented):
I really like this approach, great to have this config in the fork classes.

I added documentation as a PR to this branch: #1099.

I'm not sure we whether we should start adding new, very specific config objects to the top level of our fixture structures. We could better future-proof this addition by adding a top-level `config` entry to the JSON fixtures. This would allow us to expand it later if other configs are required.

### Suggestion

#### State Test Fixture Format Changes

Adds `blobSchedule` to a new top-level `config` entry in state test JSON fixture files:

```json
{
    "tests/cancun/eip1153_tstore/test_basic_tload.py::test_basic_tload_after_store[fork_Cancun-state_test]": {
        "env": {
            ...
        },
        "pre": {
            ...
        },
        "transaction": {
            ...
        },
        "post": {
            ...
        },
        "config": {
          "blobSchedule": {
              "cancun": {
                  "max": "0x06",
                  "target": "0x03",
                  "baseFeeUpdateFraction": "0x32f0ed"
              },
            },
        },
        "_info": {
            ...
        }
    }
}
```

#### Blockchain Test Fixture Format Changes

- The `blobSchedule` structure and `network` are added to the new field top-level `config` entry in the blockchain fixture files.
- The top-level `network` field is left as-is, to be potentially deprecated in the future.

```json
{
    "tests/cancun/eip1153_tstore/test_basic_tload.py::test_basic_tload_after_store[fork_Cancun-blockchain_test]": {
        "network": "Cancun",
        "genesisBlockHeader": {
            ...
        },
        "pre": {
            ...
        },
        "postState": {
            ...
        },
        "lastblockhash": "...",
        "config": {
          "blobSchedule": {
              "cancun": {
                  "max": "0x06",
                  "target": "0x03",
                  "baseFeeUpdateFraction": "0x32f0ed"
              },
              "network": "Cancun",
            },
        },
        "genesisRLP": "...",
        "blocks": [
            ...
        ],
        "sealEngine": "NoProof",
        "_info": {
            ...
        }
    }
}
```

Just an idea, let me know what you think. cc @winsvega

src/ethereum_clis/types.py
winsvega: Context is better

chfast:
I'm not sure I'm able to grasp every information at once, but my generic comments from the consumer POV:

1. Reuse sub-objects between different formats (statetest, blockchaintest, t8n) if possible.
2. What works best for "structured JSON parsers" is that we have agreed default values and ability to overwrite them in tests. E.g. omitting a key should not mean something different than having the key with default value.
3. I'm quite confused why "cancun" is even here and repeated twice. For me the ability to overwrite by-the-spec values is wrong because I usually don't give such ability in API. E.g. if the test says `blobSchedule.cancun.max = 7` should I reject/skip/fail the test?

src/ethereum_clis/types.py
marioevz: Changed in latest commit 👍

marioevz: Changed to TransitionToolContext in latest commit.

marioevz: Changed to TransitionToolContext in latest commit 👍

src/ethereum_clis/types.py
marioevz: Ah, this was named after the post request (as in html) that is made to the t8n server when requesting a state transition.
Bad name decision, changing it to `TransitionToolRequest` 👍

src/ethereum_test_specs/tests/fixtures/chainid_cancun_state_test.json
marioevz: Created the new issue, we can discuss further there.

marioevz:
> 3\. I'm quite confused why "cancun" is even here and repeated twice. For me the ability to overwrite by-the-spec values is wrong because I usually don't give such ability in API. E.g. if the test says `blobSchedule.cancun.max = 7` should I reject/skip/fail the test?

I think it would be nice to have input from other clients with regards what we want to see in this object.

Particularly, are we ok with this field containing different values than mainnet, and then producing a test that verifies these value deviations somehow during the test? Like setting a max blobs to 100 and then verifying that a transaction with these amount of blobs is correctly accepted.

As to why is Cancun duplicated, we have the value contained in network, which can be `Cancun`, `Prague` or `CancunToPragueAtTime15k`, and for example for `CancunToPragueAtTime15k` we require `cancun` and `prague` inside of the blob schedule because Cancun's `target` and `max` will be applied before the block with 15k timestamp, and Prague's values will be applied afterwards. I hope this made sense but I agree it's confusing, so I can also add this description to the docs.

src/ethereum_test_specs/tests/fixtures/chainid_cancun_state_test.json
danceratopz: https://github.com/ethereum/execution-spec-tests/issues/1102

src/ethereum_clis/types.py
```diff
@@ -52,3 +52,19 @@ class TransitionToolOutput(CamelModel):
...
+    """Transition tool context."""
+
+    fork: str
+    chain_id: int = Field(..., alias="chainid")
+    reward: int
+    blob_schedule: BlobSchedule | None
+
+
+class TransitionToolRequest(CamelModel):
+    """Transition tool post."""
```
danceratopz, line 67:
```suggestion
    """Transition tool server request data."""
```

src/ethereum_test_forks/forks/forks.py
```diff
@@ -948,6 +944,29 @@ def max_blobs_per_block(cls, block_number: int = 0, timestamp: int = 0) -> int:
...
+            blob_schedule = BlobSchedule()
+        else:
+            last_blob_schedule = blob_schedule.last()
+        current_blob_schedule = ForkBlobSchedule(
+            target_blobs_per_block=cls.target_blobs_per_block(block_number, timestamp),
+            max_blobs_per_block=cls.max_blobs_per_block(block_number, timestamp),
+            base_fee_update_fraction=cls.blob_base_fee_update_fraction(block_number, timestamp),
+        )
+        if last_blob_schedule is None or last_blob_schedule != current_blob_schedule:
+            blob_schedule.append(fork=cls.__name__.lower(), schedule=current_blob_schedule)
```
danceratopz, line 967: The EIP uses lowercase for this object, but the inconsistency between the network field (capitalized) and blob schedule key (lowercase) really stands out now, for example:
https://github.com/ethereum/execution-spec-tests/blob/d4b5894ad767b03d89750583cac74d3cac03a0b7/src/ethereum_test_specs/tests/fixtures/chainid_cancun_blockchain_test.json#L131-L136

Can we take the capitalized fork name for consistency?
```suggestion
            blob_schedule.append(fork=cls.__name__, schedule=current_blob_schedule)
```

danceratopz (approved):
LGTM! Thanks for adding the `config` field. One minor nit about consistency (I'd prefer to use the fork name capitalized consistently in the JSON) but I can live with it as-is, if you prefer.

src/ethereum_test_forks/forks/forks.py
marioevz: Agreed, let's change it for consistency.

chfast:
If you plan to go with it, I think the nesting is backward. I'd imagine it to be organized as:

```json
"config": {
            "cancun": {
                "blobSchedule": {
                    "max": "0x06",
                    "target": "0x03",
                    "baseFeeUpdateFraction": "0x32f0ed"
                },
                "network": "Cancun",
            },
        },
```

And then you can skip unnecessary grouping. Grouping may look nice, but it rather makes consumers more complex.


```json
"config": {
            "cancun": {
                "maxBlobs": "0x06",
                "targetBlobs": "0x03",
                "blobBaseFeeUpdateFraction": "0x32f0ed",
                "network": "Cancun",
            },
        },
```

danceratopz:
> And then you can skip unnecessary grouping. Grouping may look nice, but it rather makes consumers more complex.
> 
> ```json
> "config": {
>             "cancun": {
>                 "maxBlobs": "0x06",
>                 "targetBlobs": "0x03",
>                 "blobBaseFeeUpdateFraction": "0x32f0ed",
>                 "network": "Cancun",
>             },
>         },
> ```

Appreciate the input @chfast.

I appreciate aspiring to avoiding unnecessary nesting, but I don't think we can nest as you propose as the following example demonstrates.

Here's the current proposal for a `config` section for a transition fork test from Cancun to Prague (it would need #1082 to generate this transition). This is closer to the [shape proposed in the EIP](https://github.com/ethereum/EIPs/blob/1b4be1945de37e9e379c193188994e691df73ef9/EIPS/eip-7840.md#specification).

```json
        "config": {
            "network": "CanunToPragueAtTime15k",
            "blobSchedule": {
                "Cancun": {
                    "target": "0x03",
                    "max": "0x06",
                    "baseFeeUpdateFraction": "0x32f0ed"
                },
                "Prague": {
                    "target": "0x06",
                    "max": "0x09",
                    "baseFeeUpdateFraction": "0x4c6964"
                }
            }
        }
```

To reduce nesting, how about this? 
```json
        "config": {
            "network": "CanunToPragueAtTime15k",
            "Cancun": {
                "targetBlobs": "0x03",
                "maxBlobs": "0x06",
                "blobBaseFeeUpdateFraction": "0x32f0ed"
            },
            "Prague": {
                "targetBlobs": "0x06",
                "maxBlobs": "0x09",
                "blobBaseFeeUpdateFraction": "0x4c6964"
            }
        }
```

cc @jochem-brouwer

winsvega:
for normal forks 
```
        "config": {
            "network": "fork",
            "blobSchedule": {
                    "target": "0x03",
                    "max": "0x06",
                    "baseFeeUpdateFraction": "0x32f0ed"
            }
        }
```

chfast:
I take it back. I just now noticed that `"blobSchedule"` is coming from EIP-7840 so if the EIP is accepted we should use it directly, by my rule number 1 :)

> 1. Reuse sub-objects between different formats (statetest, blockchaintest, t8n) if possible.

I just don't understand what is `"baseFeeUpdateFraction"` for.

danceratopz:
> I take it back. I just now noticed that `"blobSchedule"` is coming from EIP-7840 so if the EIP is accepted we should use it directly, by my rule number 1 :)

Ok! :slightly_smiling_face: 

> > 1. Reuse sub-objects between different formats (statetest, blockchaintest, t8n) if possible.

:+1: let me check what the difference is to the format used in t8n.

> I just don't understand what is `"baseFeeUpdateFraction"` for.

JICYMI, this is still open, but seems widely agreed upon: https://github.com/ethereum/EIPs/pull/9240

From the [(unmerged) rationale](https://github.com/ethereum/EIPs/blob/1b4be1945de37e9e379c193188994e691df73ef9/EIPS/eip-7840.md#rationale):
> Additionally, the baseFeeUpdateFraction parameter was added to adjust the responsiveness of blob gas pricing per fork.

winsvega:
or how about this one
```
       "network" : "forkAtoforkb",
        "config": [
           {
               "network": "forkA",
               "blobSchedule": {
                       "target": "0x03",
                       "max": "0x06",
                       "baseFeeUpdateFraction": "0x32f0ed"
               }
          },
          {
               "network": "forkB",
               "blobSchedule": {
                       "target": "0x03",
                       "max": "0x06",
                       "baseFeeUpdateFraction": "0x32f0ed"
               }
          }
        ]
```

jochem-brouwer:
For the config, I would take it directly from EIP-7840, otherwise this config would override the EIP. (Or change  the EIP)

danceratopz:
@marioevz @winsvega: As it stands, listening to @chfast:
> I take it back. I just now noticed that "blobSchedule" is coming from EIP-7840 so if the EIP is accepted we should use it directly, by my rule number 1 :)

and @jochem-brouwer (thanks for the input!):
> For the config, I would take it directly from EIP-7840, otherwise this config would override the EIP. (Or change the EIP)

... the format below is the current agreement on the fixture change. This is the current state of this PR (as of now, [6d05af4](https://github.com/ethereum/execution-spec-tests/pull/1040/commits/6d05af467527e6d86195bddf7385c4105cc93ae3)).

From the [comment above](https://github.com/ethereum/execution-spec-tests/pull/1040#issuecomment-2606832792):
> Here's the current proposal for a `config` section for a transition fork test from Cancun to Prague (it would need #1082 to generate this transition). This is closer to the [shape proposed in the EIP](https://github.com/ethereum/EIPs/blob/1b4be1945de37e9e379c193188994e691df73ef9/EIPS/eip-7840.md#specification).
> 
> ```json
>         "config": {
>             "network": "CanunToPragueAtTime15k",
>             "blobSchedule": {
>                 "Cancun": {
>                     "target": "0x03",
>                     "max": "0x06",
>                     "baseFeeUpdateFraction": "0x32f0ed"
>                 },
>                 "Prague": {
>                     "target": "0x06",
>                     "max": "0x09",
>                     "baseFeeUpdateFraction": "0x4c6964"
>                 }
>             }
>         }
> ```

If there are no objections, I think we can merge this PR as-is.

winsvega:
only one concern. if you get more then one section. 
if there would be blobSchedule2, then you overcopy forks for each section in the config.

marioevz:
> only one concern. if you get more then one section. if there would be blobSchedule2, then you overcopy forks for each section in the config.

I don't think I can see a `blobSchedule2` being needed, do you mean:
1) "Another fork is added": It would be just appended to the current object, e.g. "Osaka" would simply be a new field in the dictionary.
2) or "A new field is added to the blob schedule": Unless the field is not backwards compatible, which I don't think will happen, we can just append the new field to the current `blobSchedule` and should not be an issue.

Either way I don't see why we shouldn't be including fields we need now over potential future fields that would incompatible IMO.

winsvega:
I meant what if there would be another config besides blobschedule

marioevz:
> I meant what if there would be additional another config besides blobschedule in the future
> 
> Something like gastargetschedule

Got it, yes I think something like this would fit nicely within the new `config` object, just adding every new config property as a new field.

lightclient:
FWIW I feel like we should be moving away from hexadecimal representation where it makes sense and we're confident the number can be represented natively in js -- so much easier to read and reason about decimal representation.

PR #1044, opened 2024-12-26
feat(consume): Add `consume cache`, and `--input` flag accepts versioned release names

### LINKED ISSUE #787: feat(consume): refine the `--input` flag for fixture releases
Implement the comments and suggests from #745.

From https://github.com/ethereum/execution-spec-tests/pull/745#discussion_r1723225970, add the following:
> I think we should use the Github API to determine which release is the latest and then define input_source to the URL that points to a specific release (i.e., a URL containing v[0-9]+\.[0-9]+\.[0-9]+). This has a few benefits:
> 
> 1. The pytest header will display the correct release version (this is a must imho, so that you can quickly see which release was actually being executed).
> 2. We can display the exact release in the consume commands displayed in hiveview to help reproduce tests.
> 3. The extracted fixtures will be extracted to the correct cache directory.
> 
> Another approach would be to grab the version from the .meta/fixtures.ini file, but I thinkI prefer resolving latest. If we're being very careful, we could check that the resolved version matches that of .meta/fixtures.ini, but it's a bit paranoid.

From https://github.com/ethereum/execution-spec-tests/pull/745#issuecomment-2288383673, consider adding the following:
> 1) I was hoping a while ago to be able to add the release tag and fixture "tag" as input options.
> 
> So we could have --input=<release_tag|latest>-<fixture_tag>-release, for example:
> - `--input=latest-stable-release` equivalent to `--input=v3.0.0-stable-release`.
> - `--input=eip7692@v1.0.7-eip7692-release` for the EOF v1.0.7 fixtures.
> At this point we are may be going overkill! But at least if we include the former it covers all cases regarding releases.
> 
> I do still like URLs especially if pointing to releases outside of the main repo but then...
> 2) we could add git user names to the input!

marioevz:
### `consume cache` command

Takes same inputs as other `consume` commands, but it's only used to download and cache fixtures.

Useful to create docker containers and download the fixtures during the image creation process instead of each time the container runs.

The `--cache-folder` flag can also be specified to use a specific folder as cache, and when omitted it defaults to `ethereum-execution-spec-tests/cached_downloads` placed inside of the folder returned by [`platformdirs.user_cache_dir`](https://platformdirs.readthedocs.io/en/latest/api.html#platformdirs.user_data_dir).

### Release name parsing for `--input` flag

The `--input` flag now accepts release tags in the form `<RELEASE_NAME>@<RELEASE_VERSION>`, where `<RELEASE_VERSION>` can be in format `v1.2.3` or the keyword `latest` to automatically fetch the latest version.

It uses the GitHub API to fetch the list of releases and automatically download the specified version for the given release name.

The list of releases is cached and only re-downloaded after 10 minutes have passed since the last query, except when running inside of a docker container or in CI, in order to cache only during image creation.

Keywords `latest-stable-release` and `latest-develop-release` are deprecated in favor of `stable@latest` or `develop@latest` respectively.

### Hive instance information

The `/hive` endpoint of the hive server is now used to fetch information about the hive instance being used to execute the tests.

Information such as the client file used to configure the clients is now fetched and placed inside of the test description to aid on reproducibility.

## 🔗 Related Issues
Closes #787

## ✅ Checklist


src/cli/pytest_commands/consume.py
```diff
@@ -125,3 +121,15 @@ def engine() -> None:
 def hive() -> None:
     """Client consumes via all available hive methods (rlp, engine)."""
     pass
+
+
+@consume.command(
+    context_settings={"ignore_unknown_options": True},
+)
+@common_click_options
```
spencer-tb, line 129: Why commond click options?

src/pytest_plugins/consume/releases.py
```diff
@@ -0,0 +1,201 @@
...
+def download_release_information(destination_file: Path | None) -> List[ReleaseInformation]:
+    """Download the latest stable and develop releases and optionally save them to a file."""
+    response = requests.get(RELEASE_INFORMATION_URL)
+    response.raise_for_status()
+    release_information = response.json()
+    if destination_file is not None:
+        destination_file.parent.mkdir(parents=True, exist_ok=True)
+        with open(destination_file, "w") as file:
+            json.dump(release_information, file)
+    return parse_release_information(release_information)
```
spencer-tb, line 157: I had to add pageination to get this work as githubs api only returns the most recent 30 releases. Suggestion below worked for me so feel free to try it :D
```suggestion
def download_release_information(destination_file: Path | None) -> List[ReleaseInformation]:
    """
    Download all releases from the GitHub API, handling pagination properly.

    GitHub's API returns releases in pages of 30 by default. This function
    follows the pagination links to ensure we get every release, which is
    crucial for finding older version or latest releases.
    """
    all_releases = []
    current_url = RELEASE_INFORMATION_URL

    while current_url:
        response = requests.get(current_url)
        response.raise_for_status()
        all_releases.extend(response.json())
        current_url = None
        if "Link" in response.headers:
            links = requests.utils.parse_header_links(response.headers["Link"])
            for link in links:
                if link["rel"] == "next":
                    current_url = link["url"]
                    break

    if destination_file is not None:
        destination_file.parent.mkdir(parents=True, exist_ok=True)
        with open(destination_file, "w") as file:
            json.dump(all_releases, file)

    return parse_release_information(all_releases)
```

spencer-tb (commented):
Awesome PR, thanks for adding this - its so much better! Some comments below :)

**`consume cache`**

I love the hacky `test_cache.py` solution, I did try to find an alternative nonetheless. I think we can remove the `--cache-only` flag. Please follow comments for suggestions. With some of these the `--help` flag for both cache and non-cache commands look clean.

**`--input` and releases**

I validated that we could download any pre-release or url on my end, as well as `stable@latest` and `develop@latest`, including `eip7692@latest` which is awesome. I had to add pagination for the github api to get it work - more below. It could be nice to have an non-url method to download older stable releases but I don't think its necessary.

**`hive`**
I still need to validate the hiveview updates are propagated but looks good from my end.

src/pytest_plugins/consume/releases.py
marioevz: I'm adding also a `max_pages` to only fetch the first two pages.

I think 60 releases should be enough, but we can increase it in case of failures.

marioevz:
@spencer-tb Applied all the comments, thanks for the review!

src/pytest_plugins/consume/consume.py
```diff
@@ -33,7 +38,7 @@ def default_html_report_file_path() -> str:
     Filepath (default) to store the generated HTML test report. Defined as a
     function to allow for easier testing.
     """
-    return ".meta/report_consume.html"
+    return "./report_consume.html"
```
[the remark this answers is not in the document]
marioevz: Tbh I don't quite remember, I'll roll it back.
```suggestion
    return ".meta/report_consume.html"
```

spencer-tb (approved):
LGTM! :)

PR #1056, opened 2025-01-07
convert calldatacopy test

pacrob:
## 🗒️ Description

Converts `calldatacopy` tests from [here](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/VMTests/vmTests/calldatacopyFiller.yml) 

PR to delete from `ethereum/tests` is [here](https://github.com/ethereum/tests/pull/1425)

Tests pass run against `evm statetest` and `evm blocktest` `geth version 1.14.12-stable-293a300d`

Questions:

 - I'm not sure how to address the coverage CI failure. I tried running it locally but didn't get any more clarifying info from it.

## ✅ Checklist


tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,255 @@
...
+        "sec",
+    ],
+)
+def test_calldatacopy(
+    state_test: StateTestFiller,
+    code: Bytecode,
+    tx_data: bytes,
+    pre: Alloc,
+    code_address_storage: Account,
+    to_address_storage: Account,
```
marioevz, line 203:
```suggestion
    to_address_storage: Account | None,
```
From some of the parametrized values.

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,255 @@
...
+            + Op.PUSH1[0x20]
+            + Op.PUSH1[0x10]
+            + Op.PUSH1[0xF]
+            + Op.PUSH1[0x0]
+            + Op.PUSH1[0x4]
+            + Op.CALLDATALOAD
+            + Op.PUSH2[0x1000]
+            + Op.ADD
+            + Op.PUSH3[0xFFFFFF]
+            + Op.CALL
```
marioevz, line 226: I think this and most if not all lines above can be rewritten using parentheses like `Op.CALL(0xFFFFFF, Op.ADD( ...`.

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,255 @@
...
+            + Op.PUSH1[0x0]
+            + Op.SSTORE
+            + Op.PUSH1[0x40]
+            + Op.MLOAD
+            + Op.PUSH1[0x1]
+            + Op.SSTORE
+            + Op.STOP
+        ),
+        nonce=0,
+        balance=0x0BA1A9CE0BA1A9CE,
```
marioevz, line 239: I think this amount of balance might be unnecessary.

A bit of background is that this value was common in ethereum/tests, but now we can use the `execute` command to send these transactions to a live network, and if we use this amount the pre-deploy account has to be funded with this. So it's better to keep it to the minimum necessary.

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,255 @@
...
+    )
+
+    tx = Transaction(
+        data=tx_data,
+        gas_limit=0x04C4B400,
+        gas_price=0x0A,
+        nonce=0x00,
```
marioevz, line 246:
```suggestion
```
Can be removed since the sender that is returned by `pre.fund_eoa()` has a nonce counter that is automatically increased and used for each transaction that is generated with it.

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,255 @@
...
+            + Op.SSTORE
+            + Op.STOP
+        ),
+        nonce=0,
+        balance=0x0BA1A9CE0BA1A9CE,
+    )
+
+    tx = Transaction(
+        data=tx_data,
+        gas_limit=0x04C4B400,
```
marioevz, line 244: I think this also can be reduced due to similar reason as the balance of the contract.
```suggestion
        gas_limit=100_000,
```

marioevz (commented):
Thanks for this! I've left some comments to make the code more readable, I can fully review once these are applied 👍

marioevz:
> Questions:
> 
> * I'm not sure why tests before Byzantium are failing.

This could potentially be due to the `protected` field in the transaction, since I think the transactions can only be protected on Byzantium and afterwards. You can try `protected=fork>=Byzantium` in the transaction building call (requires adding `fork: Fork` as parameter to the test function, and importing `Fork` from `ethereum_test_forks`).

> * If the file is to be deleted from `ethereum/tests`, how should I reference the source in the docstring?

We could use the permalink to the file: https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/VMTests/vmTests/calldatacopyFiller.yml

> * I'm not sure how to address the coverage CI failure. I tried running it locally but didn't get any more clarifying info from it.

This might be because of the way the code is compiled for the ethereum/tests version of the test, so there's a difference between the opcodes used, but we could analyze the result and if deemed unimportant we can merge as is, no problem.

pyproject.toml
```diff
@@ -44,12 +44,13 @@ dependencies = [
...
     "pytest-json-report>=1.5.0,<2",
-    # TODO: bump questionary to a newer release, when it becomes available. 
+    # TODO: bump questionary to a newer release, when it becomes available.
     # The current release questionary 2.0.1 requires `prompt_toolkit = ">=2.0,<=3.0.36"`.
     # This conflicts with ipython; while not an EEST dependency, ipython a very useful tool:
     # https://ethereum.github.io/execution-spec-tests/main/dev/interactive_usage/
     "questionary @ git+https://github.com/tmbo/questionary@ff22aeae1cd9c1c734f14329934e349bec7873bc",
-    "prompt_toolkit>=3.0.48,<4",  # ensure we have a new enough version for ipython
+    "prompt_toolkit>=3.0.48,<4", # ensure we have a new enough version for ipython
+    "ruff>=0.8.2",
```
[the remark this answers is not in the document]
pacrob: This is me learning how `uv` works. I just installed it to try it out, didn't know `uv` would add it to the dep list. Removing.

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,255 @@
...
+@pytest.mark.valid_from("Byzantium")
+@pytest.mark.parametrize(
+    "code,tx_data,code_address_storage,to_address_storage",
+    [
+        (
+            (
+                Op.PUSH1[0x2]
+                + Op.PUSH1[0x1]
+                + Op.PUSH1[0x0]
+                + Op.CALLDATACOPY
```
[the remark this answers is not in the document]
winsvega: +1

tests/frontier/opcodes/test_calldatacopy.py
winsvega: +1

winsvega:
or you can reference legacy test file. because ethereum/tests:: develop tests are essentially a copy of tests in legacytests/cancun. 
legacytests are frozen from modifications
ethereum/tests  are to be converted to .py and deleted. so no development or new modifications

pacrob:
OK! I've addressed all the comments. Last questions from the checklist:

 - I can't set tags. What should they be?
 - How do you want the tests to be run? I've successfully fed them to `evm statetest` and `evm blocktest`, but not sure how to use a filled test file with the geth `t8n` tool.
 - `mkdocs serve` runs fine, but there doesn't seem to be any documentation generated in this case. Is that correct?

marioevz:
Hey! thanks, giving it another look just now :)

> * I can't set tags. What should they be?
Do you mean the ones in the PR? I think it should be ok to set the labels to `scope: tests`.

> * How do you want the tests to be run? I've successfully fed them to `evm statetest` and `evm blocktest` (geth version 1.14.12-stable-293a300d), but not sure how to use a filled test file with the geth `t8n` tool.

This is perfect, `evm statetest` and `evm blocktest` should be more than enough.

> * `mkdocs serve` runs fine, but there doesn't seem to be any documentation generated in this case. Is that correct?

I tried it out and it generated ok, checkout "Test Case Reference":
image
 
Then in the sidebar you should see the test case you wrote here:
image


> Also, I got the last failing tests to run by raising the gas limit in the tx from 100_000 to 100_000_000, but didn't try anything in between. Is it important here to try to be specific in the gas amount, or is just adding a bunch OK?

If there's no way around it then it's ok to add a bunch of gas, but if we can keep the gas limit to less than 30,000,000 it means that the test can run in live networks.
Otherwise we can just add the
```python
@pytest.mark.execute(pytest.mark.skip("too much gas"))
```
marker (the string is not standardized) and it should be gracefully skipped when running in live networks. 👍

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,176 @@
+"""test `CALLDATACOPY` opcode."""
+
+import pytest
+
+from ethereum_test_forks import Fork
+from ethereum_test_forks.forks.forks import Byzantium
```
marioevz, line 6:
```suggestion
from ethereum_test_forks import Byzantium, Fork
```
Can be reformatted.

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Based on https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/VMTests/vmTests/calldatacopyFiller.ym
+    """
+    code_address = pre.deploy_contract(code)
+    to = pre.deploy_contract(
+        code=(
+            Op.MSTORE(offset=0x0, value=0x1234567890ABCDEF01234567890ABCDEF0)
+            + Op.CALL(
+                0xFFFFFF, Op.ADD(0x1000, Op.CALLDATALOAD(offset=0x4)), 0x0, 0xF, 0x10, 0x20, 0x40
+            )
```
marioevz, line 154: The `Op.ADD(0x1000, Op.CALLDATALOAD(offset=0x4))` is unnecessary here.

The reason it was used in `ethereum/tests` it's because the parametrization happened using the transaction data, which `CALLDATALOAD` fetched and then it redirected the test flow to the correct contract.

But in EEST the return value of `pre.deploy_contract` is not entirely deterministic. It was working during fill because the default is indeed `0x1000`, and since the transaction had no data, `0x1000 + 0x00`, so we're good because we ended up in the correct contract. But in other contexts, such as live networks, the contract is completely different.

This can be changed to:
```suggestion
            + Op.CALL(
                gas=0xFFFFFF,
                address=code_address,
                value=0x0,
                args_offset=0xF,
                args_size=0x10,
                ret_offset=0x20,
                ret_size=0x40,
            )
```

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,176 @@
...
+            + Op.SSTORE(key=0x0, value=Op.MLOAD(offset=0x20))
+            + Op.SSTORE(key=0x1, value=Op.MLOAD(offset=0x40))
+            + Op.STOP
+        ),
+        nonce=0,
+    )
+
+    tx = Transaction(
+        data=tx_data,
+        gas_limit=0x100_000_000,
```
marioevz, line 165:
```suggestion
        gas_limit=100_000,
```
Yes should definitely be reduced, but I'm having a hard time figuring out why is it running out of gas.

It only happens in Frontier and Homestead, and it seems to be call on the root contract.

If you run with traces using `uv run fill --fork=Frontier "tests/frontier/opcodes/test_calldatacopy.py::test_calldatacopy[fork_Frontier-blockchain_test-cdc 0 1 2]" --evm-bin evm --traces`, the last call is trying to use 16 million gas for some reason:

```
Step 10:
{ 'depth': 1,
  'error': 'out of gas',
  'gas': '0x2bb13',
  'gasCost': '0x100002d',
  'memSize': 32,
  'op': 241,
  'opName': 'CALL',
  'pc': 38,
  'refund': 0,
  'stack': ['0x40', '0x20', '0x10', '0xf', '0x0', '0x1000', '0xffffff']}

Step 11:
{'error': 'out of gas', 'gasUsed': '0x2bb34', 'output': ''}
```

tests/frontier/opcodes/test_calldatacopy.py
```diff
@@ -0,0 +1,176 @@
...
+            Op.MSTORE(offset=0x0, value=0x1234567890ABCDEF01234567890ABCDEF0)
+            + Op.CALL(
+                0xFFFFFF, Op.ADD(0x1000, Op.CALLDATALOAD(offset=0x4)), 0x0, 0xF, 0x10, 0x20, 0x40
+            )
+            + Op.POP
+            + Op.SSTORE(key=0x0, value=Op.MLOAD(offset=0x20))
+            + Op.SSTORE(key=0x1, value=Op.MLOAD(offset=0x40))
+            + Op.STOP
+        ),
+        nonce=0,
```
marioevz, line 160:
```suggestion
```

Can be removed.

marioevz (commented):
Hey looks good so far, I have a couple more comments.

The gas issue I haven't been able to figure out but one thing that we could to is just limit the test to be filled only on Byzantium and after.

Let me know what you think.

marioevz:
Figured out the gas issue: It seems like passing something above the current available gas results in the `CALL` opcode charging a really high amount of gas in Frontier and Homestead, which I think was fixed in Byzantium with [EIP-150](https://eips.ethereum.org/EIPS/eip-150).

The solution seems to be something along the lines of:
```python
Op.CALL(
                gas=Op.SUB(Op.GAS(), 0x100),
                address=code_address,
                value=0x0,
                args_offset=0xF,
                args_size=0x10,
                ret_offset=0x20,
                ret_size=0x40,
            )
```
so the call does not try to send a gas amount higher than it currently gas.

With this change, sending the transaction with a gas limit of `100_000` works for all tests 🎉

pacrob:
OK! All requested updates pushed. I found them in docs, just didn't know where to look :smile:

winsvega:
the only missing coverage is `calldataload` and `add` opcodes. can be ignored.

tests/frontier/opcodes/test_calldatacopy.py
marioevz: Hey @pacrob I think we could still commit this suggestion, it should be possible now with the fix on line 152 👍

docs/CHANGELOG.md
```diff
@@ -28,7 +28,8 @@ Release tarball changes:
...
 - 🔀 Update EIP-7002 according to [spec updates](https://github.com/ethereum/EIPs/pull/9119) ([#1024](https://github.com/ethereum/execution-spec-tests/pull/1024)).
 - 🔀 Update EIP-2935 according to [spec updates](https://github.com/ethereum/EIPs/pull/9144) ([#1046](https://github.com/ethereum/execution-spec-tests/pull/1046))
-- ✨ [EIP-7691](https://eips.ethereum.org/EIPS/eip-7691) Blob throughput increase tests by parametrization of existing EIP-4844 tests ([#1023](https://github.com/ethereum/execution-spec-tests/pull/1023), [#1082](https://github.com/ethereum/execution-spec-tests/pull/1082))
+- ✨ [EIP-7691](https://eips.ethereum.org/EIPS/eip-7691) Blob throughput increase tests by parametrization of existing EIP-4844 tests ([#1023](https://github.com/ethereum/execution-spec-tests/pull/1023))
+- ✨ Port [calldatacopy test](https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/VMTests/vmTests/calldatacopyFiller.yml) ([#1056](https://github.com/ethereum/execution-spec-tests/pull/1056)).
```
marioevz, line 32:
```suggestion
- ✨ [EIP-7691](https://eips.ethereum.org/EIPS/eip-7691) Blob throughput increase tests by parametrization of existing EIP-4844 tests ([#1023](https://github.com/ethereum/execution-spec-tests/pull/1023), [#1082](https://github.com/ethereum/execution-spec-tests/pull/1082))
- ✨ Port [calldatacopy test](https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/VMTests/vmTests/calldatacopyFiller.yml) ([#1056](https://github.com/ethereum/execution-spec-tests/pull/1056)).
```
Potential issue during merge to main branch.

marioevz:
I've verified the coverage loss and it's down to a couple of things:
- The `ADD` and `CALLDATALOAD` opcodes are no longer used, which is ok
- There are no empty accounts in the test's pre-state, also ok
- MPT state differences (due to different accounts used), also ok

We should be able to merge without issues 👍

PR #1066, opened 2025-01-09
Refactor `left_pad_zeros_up_to_size` to a parameter in the `FixedSizeBytes` constructor

marioevz:
Changes `left_pad_zeros_up_to_size` to be instead an optional parameter in all classes that inherit `FixedSizeBytes`.

So for example `Hash(left_pad_zeros_up_to_size("0x00", 32))` is equivalent to `Hash("0x00", left_padding=True)`.

If neither `left_padding` or `right_padding` are specified, and the input is shorter than the expected size, an exception is raised.

### DIFF
--- a/tests/cancun/eip1153_tstore/test_tstorage_execution_contexts.py
+++ b/tests/cancun/eip1153_tstore/test_tstorage_execution_contexts.py
@@ -8,13 +8,13 @@ from typing import Dict, Mapping
 
 import pytest
 
-from ethereum_test_base_types.conversions import left_pad_zeros_up_to_size
 from ethereum_test_tools import (
     Account,
     Address,
     Alloc,
     Bytecode,
     Environment,
+    Hash,
     StateTestFiller,
     Transaction,
 )
@@ -308,7 +308,7 @@ def tx(pre: Alloc, caller_address: Address, callee_address: Address) -> Transact
     return Transaction(
         sender=pre.fund_eoa(),
         to=caller_address,
-        data=left_pad_zeros_up_to_size(callee_address, 32),
+        data=Hash(callee_address, left_padding=True),
         gas_limit=1_000_000,
     )
 
--- a/tests/shanghai/eip4895_withdrawals/test_withdrawals.py
+++ b/tests/shanghai/eip4895_withdrawals/test_withdrawals.py
@@ -10,7 +10,6 @@ from typing import Dict, List, Mapping
 import pytest
 
 from ethereum_clis import TransitionTool
-from ethereum_test_base_types.conversions import left_pad_zeros_up_to_size
 from ethereum_test_forks import Cancun, Fork
 from ethereum_test_tools import (
     EOA,
@@ -19,6 +18,7 @@ from ethereum_test_tools import (
     Alloc,
     Block,
     BlockchainTestFiller,
+    Hash,
     Transaction,
     TransactionException,
     Withdrawal,
@@ -201,7 +201,7 @@ def test_balance_within_block(blockchain_test: BlockchainTestFiller, pre: Alloc)
                     sender=sender,
                     gas_limit=100000,
                     to=contract_address,
-                    data=left_pad_zeros_up_to_size(recipient, 32),
+                    data=Hash(recipient, left_padding=True),
                 )
             ],
             withdrawals=[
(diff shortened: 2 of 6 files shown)


winsvega:
That undo the very same idea we wanted to ibtroduce is to restrict assigment of fixed bytes of different sizes

marioevz:
> The point was to make it explicit, so we think about bytes size when writing the tests

But isn't the parameter explicit enough? if no parameter is passed and the byte length is shorter, an error will be raised, so the tester has to explicitly:
- Before this PR: Use the `left_pad_zeros_up_to_size`.
- After this PR: Pass either the `left_padding` or the `right_padding` parameter as `True`.

I guess my point is that both are explicit ways to ensure that the tester knows what they are doing, just that the second one is a bit more convenient.

tests/cancun/eip1153_tstore/test_tstorage_execution_contexts.py
```diff
@@ -308,7 +308,7 @@ def tx(pre: Alloc, caller_address: Address, callee_address: Address) -> Transact
     return Transaction(
         sender=pre.fund_eoa(),
         to=caller_address,
-        data=left_pad_zeros_up_to_size(callee_address, 32),
+        data=Hash(callee_address, left_padding=True),
```
winsvega, line 311: again the number 32 is implicit

winsvega:
I see, so to conclude: 
1) any fixedBytes can be initialized from the string "0x11" of a smaller size if we provide padding flag. 
2) fixedBytes can be init with int, which will mean it is left padded by 00 as hash

it is a bit confusing if we provide no padding flag, we can write code that will throw an exception without realizing it. 
I usually try to put there enum as a required parameter in such cases, but here there is a scenario when str does not need padding. 

I agree that it is handy compared to importing a converter function. 

Shall the unit tests be rewritten back with the flag? because I put explicit long strings there

tests/cancun/eip1153_tstore/test_tstorage_execution_contexts.py
marioevz: Yes, but I think at this point the tester has already selected `Hash` so they know they needed a 32-byte constructor.

Another entirely different discussion could be that the hashes become longer in the future, but in that case I think we need a completely new class.

marioevz:
> Ah, can we overload constructor so if it takes only str, it will throw if string byte representation has incorrect size But we can use another constructor where it takes str and enum. so
> 
> Hash("0x01", ZEROPAD::LEFT) = Hash(0x01) will make Hash `0x00..00001` Hash("0x01") throws an exception 0x01 has incorrect size
> 
> ah I see python has no overload constuctors. so the default value of padding will be ZEROPAD.NONE

Yes exactly, the magic happens here (this is in the `FixedSizeBytes` class so applies to all subclasses):

```python
    def __new__(
        cls,
        input_bytes: FixedSizeBytesConvertible | T,
        *,
        left_padding: bool = False,
        right_padding: bool = False,
    ):
```

The asterisk on the fourth line means that no other positional arguments to the function are accepted, so if you do `Hash(0x1)`, the `0x1` is taken as the single and only positional argument `input_bytes`.

The arguments after the asterisk are keyword arguments, which means they have to be explicitly named when the function called, so `Hash(0x1, True)` will not work, it has to be either `Hash(0x1, left_padding=True)` or `Hash(0x1, right_padding=True)`.

Also since they have a default of `False` it means that the function can be called without them.

I thought about doing perhaps `Hash(0x1, padding="left")` or `Hash(0x1, padding="right")` but I don't really like letting the user have to write values to a string input. With an enum like `Hash(0x1, padding=FixedSizeBytesPadding.Left)` the downside is that `FixedSizeBytesPadding` would have to be imported.

tests/shanghai/eip4895_withdrawals/test_withdrawals.py
```diff
@@ -373,7 +373,7 @@ def test_self_destructing_account(
         sender=sender,
         gas_limit=100000,
         to=self_destruct_contract_address,
-        data=left_pad_zeros_up_to_size(recipient, 32),
+        data=Hash(recipient, left_padding=True),
```
winsvega, line 376: lets name Hash Hash32 because in tests we have Hash8 (block nonce) and Hash20 (address)

PR #1067, opened 2025-01-09
feat(tests): EIP-152 precompiled contracts with Blake2

reedsa:
## 🗒️ Description
Add tests for [EIP-152: BLAKE2 compression function `F` precompile](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-152.md).

Test cases pulled from https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stPreCompiledContracts/blake2BFiller.yml

## 🔗 Related Issues

ethereum/tests#1431

## ✅ Checklist

### DIFF
--- a/tests/istanbul/eip152_blake2/test_blake2.py
+++ b/tests/istanbul/eip152_blake2/test_blake2.py
@@ -1,0 +1,581 @@
+"""
+abstract: Tests [EIP-152: BLAKE2b compression precompile](https://eips.ethereum.org/EIPS/eip-152)
+    Test cases for [EIP-152: BLAKE2b compression precompile](https://eips.ethereum.org/EIPS/eip-152).
+"""
+
+from dataclasses import dataclass
+
+import pytest
+
+from ethereum_test_tools import (
+    Account,
+    Alloc,
+    Environment,
+    StateTestFiller,
+    TestParameterGroup,
+    Transaction,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "EIPS/eip-152.md"
+REFERENCE_SPEC_VERSION = "5510973b40973b6aa774f04c9caba823c8ff8460"
+
+
+@dataclass(kw_only=True, frozen=True, repr=False)
+class Blake2bInput(TestParameterGroup):
+    """
+    Helper class that defines the BLAKE2b precompile inputs and creates the
+    call data from them. Returns all inputs encoded as bytes.
+
+    Attributes:
+        rounds_length (int): An optional integer representing the bytes length
+            for the number of rounds. Defaults to the expected length of 4.
+        rounds (int | str): A hex string or integer value representing the number of rounds.
+        h (str): A hex string that represents the state vector.
+        m (str): A hex string that represents the message block vector.
+        t_0 (int | str): A hex string or integer value that represents the first offset counter.
+        t_1 (int | str): A hex string or integer value that represents the second offset counter.
+        f (bool): An optional boolean that represents the final block indicator flag.
+            Defaults to True.
+
+    """
+
+    rounds_length: int = 4
+    rounds: int | str
+    h: str
+    m: str
+    t_0: int | str
+    t_1: int | str
+    f: bool | int = True
+
+    def create_blake2b_tx_data(self):
+        """Generate input for the BLAKE2b precompile."""
+        _rounds = self.rounds.to_bytes(length=self.rounds_length, byteorder="big")
+        _h = bytes.fromhex(self.h)
+        _m = bytes.fromhex(self.m)
+        _t_0 = (
+            bytes.fromhex(self.t_0)
+            if isinstance(self.t_0, str)
+            else self.t_0.to_bytes(length=8, byteorder="little")
+        )
+        _t_1 = (
+            bytes.fromhex(self.t_1)
+            if isinstance(self.t_1, str)
+            else self.t_1.to_bytes(length=8, byteorder="little")
+        )
+        _f = int(self.f).to_bytes(length=1, byteorder="big")
+
+        return _rounds + _h + _m + _t_0 + _t_1 + _f
+
+
+@dataclass(kw_only=True, frozen=True, repr=False)
+class ExpectedOutput(TestParameterGroup):
+    """
+    Expected test result.
+
+    Attributes:
+        call_succeeds (str | bool): A hex string or boolean to indicate whether the call was
+            successful or not.
+        data_1 (str): String value of the first updated state vector.
+        data_2 (str): String value of the second updated state vector.
+
+    """
+
+    call_succeeds: str | bool
+    data_1: str
+    data_2: str
+
+
+@pytest.mark.valid_from("Istanbul")
+@pytest.mark.parametrize(
+    ["data", "output"],
+    [
+        pytest.param(
+            Blake2bInput(
+                rounds=0,
+                rounds_length=0,
+                h="",
+                m="",
+                t_0="",
+                t_1="",
+            ),
+            ExpectedOutput(
+                call_succeeds=False,
+                data_1="0x00",
+                data_2="0x00",
+            ),
+            id="EIP-152-case0",
+        ),
+        pytest.param(
+            Blake2bInput(
+                rounds=12,
+                rounds_length=3,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=3,
+                t_1=0,
+            ),
+            ExpectedOutput(
+                call_succeeds=False,
+                data_1="0x00",
+                data_2="0x00",
+            ),
+            id="EIP-152-case1",
+        ),
+        pytest.param(
+            Blake2bInput(
+                rounds=12,
+                rounds_length=5,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=3,
+                t_1=0,
+            ),
+            ExpectedOutput(
+                call_succeeds=False,
+                data_1="0x00",
+                data_2="0x00",
+            ),
+            id="EIP-152-case2",
+        ),
+        pytest.param(
+            Blake2bInput(
+                rounds=12,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=3,
+                t_1=0,
+                f=2,
+            ),
+            ExpectedOutput(
+                call_succeeds=False,
+                data_1="0x00",
+                data_2="0x00",
+            ),
+            id="EIP-152-case3",
+        ),
+        pytest.param(
+            Blake2bInput(
+                rounds=0,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=3,
+                t_1=0,
+            ),
+            ExpectedOutput(
+                call_succeeds=True,
+                data_1="0x08c9bcf367e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5",
+                data_2="0xd282e6ad7f520e511f6c3e2b8c68059b9442be0454267ce079217e1319cde05b",
+            ),
+            id="EIP-152-case4",
+        ),
+        pytest.param(
+            Blake2bInput(
+                rounds=12,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=3,
+                t_1=0,
+            ),
+            ExpectedOutput(
+                call_succeeds=True,
+                data_1="0xba80a53f981c4d0d6a2797b69f12f6e94c212f14685ac4b74b12bb6fdbffa2d1",
+                data_2="0x7d87c5392aab792dc252d5de4533cc9518d38aa8dbf1925ab92386edd4009923",
+            ),
+            id="EIP-152-case5",
+        ),
+        pytest.param(
+            Blake2bInput(
+                rounds=12,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=3,
+                t_1=0,
+                f=False,
+            ),
+            ExpectedOutput(
+                call_succeeds=True,
+                data_1="0x75ab69d3190a562c51aef8d88f1c2775876944407270c42c9844252c26d28752",
+                data_2="0x98743e7f6d5ea2f2d3e8d226039cd31b4e426ac4f2d3d666a610c2116fde4735",
+            ),
+            id="EIP-152-case6",
+        ),
+        pytest.param(
+            Blake2bInput(
+                rounds=1,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=3,
+                t_1=0,
+            ),
+            ExpectedOutput(
+                call_succeeds=True,
+                data_1="0xb63a380cb2897d521994a85234ee2c181b5f844d2c624c002677e9703449d2fb",
+                data_2="0xa551b3a8333bcdf5f2f7e08993d53923de3d64fcc68c034e717b9293fed7a421",
+            ),
+            id="EIP-152-case7",
+        ),
+        # Excessive number of rounds expects to run out of gas
+        pytest.param(
+            Blake2bInput(
+                rounds=4294967295,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=3,
+                t_1=0,
+            ),
+            ExpectedOutput(
+                call_succeeds=False,
+                data_1="0x0",
+                data_2="0x0",
+            ),
+            id="EIP-152-case8",
+        ),
+        # Case from https://github.com/ethereum/tests/pull/948#issuecomment-925964632
+        pytest.param(
+            Blake2bInput(
+                rounds=12,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162636465000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                t_0=5,
+                t_1=0,
+            ),
+            ExpectedOutput(
+                call_succeeds=True,
+                data_1="0xf3e89a60ec4b0b1854744984e421d22b82f181bd4601fb9b1726b2662da61c29",
+                data_2="0xdff09e75814acb2639fd79e56616e55fc135f8476f0302b3dc8d44e082eb83a8",
+            ),
+            id="EIP-152-case9",
+        ),
+        pytest.param(
+            Blake2bInput(
+                rounds=16,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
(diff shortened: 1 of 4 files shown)


reedsa:
I wasn't familiar with the opcodes needed to successfully write the pre state contract, so I ended up copying over the yul from the original test. I may not have it all correct since the output state indicates there is no data returned in each case. I have the test cases written out so it'd be great if I could get some eyes on the potential issues with my implementation. Thanks!

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,501 @@
...
+        # Slow case with large number of rounds
+        # pytest.param(
+        #     Blake2Input(rounds=b"\x08"),
+        #     ExpectedOutput(
+        #         call_return_code="0x01",
+        #         data_1="0xfc59093aafa9ab43daae0e914c57635c5402d8e3d2130eb9b3cc181de7f0ecf9",
+        #         data_2="0xb22bf99a7815ce16419e200e01846e6b5df8cc7703041bbceb571de6631d2615",
+        #     ),
+        #     id="EIP-192-case8",
+        # ),
```
danceratopz, line 149: It is possible to apply the `slow` marker to indicate these tests as such:
```python
        pytest.param(
            Blake2Input(rounds=b"\x08"),
            ExpectedOutput(
                call_return_code="0x01",
                data_1="0xfc59093aafa9ab43daae0e914c57635c5402d8e3d2130eb9b3cc181de7f0ecf9",
                data_2="0xb22bf99a7815ce16419e200e01846e6b5df8cc7703041bbceb571de6631d2615",
            ),
            id="EIP-192-case8",
            marks=pytest.mark.slow,
        ),
```
and then execute them, respectively exclude them, via:
```
uv run pytest ./tests/istanbul/eip152_blake2/test_blake2.py -m slow
uv run pytest ./tests/istanbul/eip152_blake2/test_blake2.py -m "not slow"
```
But tbh honest it's probably not slow compared to the block history tests:
```
tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py::test_block_hashes_history[fork_Prague-blockchain_test-full_history_plus_one_check_blockhash_first]
tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py::test_block_hashes_history[fork_Prague-blockchain_test_engine-full_history_plus_one_check_blockhash_first]
```

So I would prefer to skip the mark and just test locally with:
```
uv run pytest ./tests/istanbul/eip152_blake2/test_blake2.py -k "not case8" --collect-only -q
```

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,501 @@
...
+    data_1: str
+    data_2: str
+
+
+@pytest.mark.valid_from("Istanbul")
+@pytest.mark.parametrize(
+    ["data", "output"],
+    [
+        pytest.param(
+            Blake2Input(rounds=b"\x00"),
```
winsvega, line 69: thats a bit ugly. do we have a class that makes Bytes from string ("0x112233") so that bytes are 11 22 33?
@marioevz

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,501 @@
...
+    yul: YulCompiler,
+    data: Blake2Input,
+    output: ExpectedOutput,
+    pre: Alloc,
+):
+    """Test BLAKE2 precompile."""
+    env = Environment()
+
+    account = pre.deploy_contract(
+        code=yul(
```
winsvega, line 325: the end goal is to get rid of yul as it is external dependency and the compiled code might be altered by solc.
rewrite the code using Op.  (see in existing tests)

winsvega (changes requested):
rewrite the yul contract using Op 

Also since you start blake test, the exising blake tests can be converted 
https://github.com/search?q=repo%3Aethereum%2Ftests%20blake&type=code

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,501 @@
...
+                // The output is bytes 0x200-0x23F
+                let retVal := call(gas(), 9, 0, 0, bufLength, 0x200, 0x40)
+
+                // Return the results
+                sstore(0, retVal)   // Success? Failure?
+                // The output
+                sstore(1, mload(0x200))
+                sstore(2, mload(0x220))
+            }
+            """
```
danceratopz, line 475: This will allow lines > 100 in the Yul block for convenience :)
```suggestion
            """  # noqa: E501 (line-too-long)
```

tests/istanbul/eip152_blake2/test_blake2.py
danceratopz, line 69: This could be more convenient, particularly for longer bytes objects. I think you could also potentially define this data as a string here in the test definitions for convenience and then convert to bytes in `create_blake2_tx_data`?
```suggestion
            Blake2Input(rounds=bytes.fromhex("00")),
```

danceratopz (commented):
Nice! Looking good so far, leaving the pre-state contract as Yul is a good first approach to getting the tests working up front. 

After all the tests fill successfully with this Yul contract, we could consider splitting it up and deploying test case-specific contracts matching the pytest parameterization.

Below are some cosmetic suggestions. Happy to help debug the fails, just let me know. But a starting point could be to fill using geth and enable traces (these are currently unavailable with EELS, but coming soon). Something like:
```
uv run pytest ./tests/istanbul/eip152_blake2/test_blake2.py -k "not case8" --evm-bin=~/bin/evm --traces -x
```

tests/istanbul/eip152_blake2/test_blake2.py
reedsa: That's a great point, I could actually just plug in the numbers as integers and convert those to bytes in the class itself. I think explicit numbers would make it really clear and easy to understand, would that be preferable to using hex values?

reedsa:
> rewrite the yul contract using Op
> 
> Also since you start blake test, here are the existing blake tests that need to be converted https://github.com/search?q=repo%3Aethereum%2Ftests%20blake&type=code

Should I keep all of these tests together in the same file but as separate tests? I could also keep them as separate files entirely, but maybe it's better to keep them all in one spot?

tests/istanbul/eip152_blake2/test_blake2.py
reedsa: Awesome callout, good to know that's there! I'll just enable this test.

reedsa:
@danceratopz @winsvega Just pushed up the latest revisions. As you can see there are test failures but I am not seeing anything from the trace output that would indicate an issue. It seems the inputs are not producing the same outputs as in the original [ethereum/tests](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stPreCompiledContracts/blake2BFiller.yml) cases.

From what I can tell, the `t_0` value should be set to the offset of the message input. In the cases where there are 16 bytes of data, this should be `0x10`. The cases which set the message to 16 bytes in length (https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/stPreCompiledContracts/blake2BFiller.yml#L145) do not seem to update this value. I'm not certain if the original test is sending in the wrong offset value but based on the spec and what was done for [case 9](https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/stPreCompiledContracts/blake2BFiller.yml#L134), it seems the value should be changed to `0x10`. Case 9 was created based on this thread: https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/stPreCompiledContracts/blake2BFiller.yml#L134

It's possible the output values should be changed based on this logic. I'm not sure what that means for the original test cases or the clients which rely on those.

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,557 @@
...
+                data_1="0x00",
+                data_2="0x00",
+            ),
+            id="EIP-192-case0",
+        ),
+        pytest.param(
+            Blake2Input(
+                rounds=12,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="616263000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
```
winsvega, line 113: here it is missing one byte of 00

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,557 @@
...
+                data_2="0x2858fd9c60fae53f366d7e2820040a8662b336de6d859764f20747acbb8999fe",
+            ),
+            id="EIP-192-0032-120",
+        ),
+        pytest.param(
+            Blake2Input(
+                rounds=64,
+                h="48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b",
+                m="6162636465666768696a6b6c6d6e6f707172737475767778797a7b7c7d7e7f808182838485868788898a8b8c8d8e8f909192939495969798999a9b9c9d9e9fa0a1a2a3a4a5a6a7a8a9aaabacadaeafb0b1b2b3b4b5b6b7b8b9babbbcbdbebfc0c1c2c3c4c5c6c7c8c9cacbcccdcecfd0d1d2d3d4d5d6d7d8d900000000000000",
+                t_0="7800000000000000",
```
winsvega, line 494: here the original test seems to be incorrect as it actually passes length byte of 30 not 78
you can debug original call with geth and retesteth
```
rt blake2BFiller.yml  -d test-0064-120 --vmtraceraw out
```
search for line with "CALL"  a line above will have memory and the call parameters are there. 
so far I see that original test is mistaken because it was using complex logic 
```
// The message starts at byte 0x44
           if gt(msgLen, 0) {
               for { } gt(msgLen, 0) { } {
                 // Write a byte
                 mstore8(add(0x44, msgLen), add(0x60, msgLen))
                 msgLen := sub(msgLen, 1)
               }
           }
```

winsvega: if you use unix, retesteth nightly bin available here http://retesteth.ethdevops.io/retesteth

reedsa: Thanks for the pointers, I think you are saying the `t_0` value here should be `78`? Should the expected outputs be changed to the values given in the pytest errors I'm getting?

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,547 @@
...
+
+    """
+
+    rounds_length: int = 4
+    rounds: int
+    h: str
+    m: str
+    t_0: str
+    t_1: str
+    f: bool = True
```
marioevz, line 48:
```suggestion
    rounds: int | str
    h: Tuple[int, ...] | str
    m: Tuple[int, ...] | str
    t_0: int | str
    t_1: int | str
    f: bool | str = True
```
We could change the types here to either int or string:
- int: the value gets converted to bytes using the endianness specified in the spec automatically.
- str: the string is parsed as hex bytes and copied verbatim into the formatted precompile call data that is generated in `create_blake2_tx_data`.

And the reasoning is that it's easier to differentiate between the cases that call the precompile with the correct data size from those that don't IMO. 

Maybe for `h` and `m` it makes less sense because the precompile expects a tuple of ints, but for `rounds`, `t_0`, `t_1` and `f` I think it should be ok and it should be a bit easier to read.

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,547 @@
...
+        )
+        + Op.STOP(),
+    )
+    sender = pre.fund_eoa()
+
+    tx = Transaction(
+        ty=0x0,
+        to=account,
+        data=data.create_blake2_tx_data(),
+        gas_limit="0xF000000000",
```
marioevz, line 530: This is certainly an extreme amount of gas 😅 

```suggestion
        gas_limit=1_000_000,
```

I tested locally and all cases pass with this reduction, except `case8`.

It seems like case 8 round count and 0xF000000000 as gas combined result in the EELS server hanging indefinitely and all subsequent tests fail because of this.

I think it should be ok to expect that the case gracefully runs out of gas if this many rounds are requested.

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,547 @@
...
+        gas_price=10,
```
marioevz, line 531:
```suggestion
```
We are trying to remove `gas_price` from having fixed values, since the cases hang indefinitely when running on a live network because the minimum gas there is 1 Gwei, so omitting this value results in the `execute` command handling this automatically for us 👍

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,547 @@
...
+    env = Environment()
+
+    account = pre.deploy_contract(
+        # Store all CALLDATA into memory (offset 0)
+        Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+        # Store the returned CALL status (success = 1, fail = 0) into slot 0:
+        + Op.SSTORE(
+            0,
+            # Setup stack to CALL into Blake2 with the CALLDATA and CALL into it (+ pop value)
+            Op.CALL(Op.GAS(), 9, 0, 0, Op.CALLDATASIZE(), 0x200, 0x40),
```
marioevz, line 512:
```suggestion
            Op.CALL(
                address=9,
                args_offset=0,
                args_size=Op.CALLDATASIZE(),
                ret_offset=0x200,
                ret_size=0x40,
            ),
```
This is a trick to let the `Op.CALL` handle the positioning of the stack, and also make the code a bit easier to follow (`gas` defaults to `Op.GAS()` so the result is exactly the same).

It works with most opcodes, and in the docstrings in the `vm/opcodes.py` you can find the keyword argument names for every opcode.

marioevz (commented):
A few comments but it looks pretty good overall, thanks for implementing this!

tests/istanbul/eip152_blake2/test_blake2.py
reedsa: I like that idea so the test parameters are all consistent integers. To be certain, the parameterized tests will end up with h, m, t_* integer values instead? For example:
```
            Blake2bInput(
                rounds=0,
                h=[
                    32922972323324069763321526652467788264663358122255454854356121218204912930725,
                    94764667551462817271282565646542853509102241772507082255866415388681393004635,
                ],
                m=[6382179, 0, 0, 0],
                t_0=3,
                t_1=0,
            ),
```

marioevz: > I like that idea so the test parameters are all consistent integers. To be certain, the parameterized tests will end up with h, m, t_* integer values instead? I'm looking at how to encode the Tuple of integers together as one bytes value, is there a standard way of doing that?
> 
> For example:
> 
> ```
>             Blake2bInput(
>                 rounds=0,
>                 h=[
>                     32922972323324069763321526652467788264663358122255454854356121218204912930725,
>                     94764667551462817271282565646542853509102241772507082255866415388681393004635,
>                 ],
>                 m=[6382179, 0, 0, 0],
>                 t_0=3,
>                 t_1=0,
>             ),
> ```

Yeah, I think it makes more sense for `round` and `t_*` because they are smaller numbers, and to convert them is a single `to_bytes`. Following the [EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-152.md#specification) it should be something like:
```python
self.rounds.to_bytes(length=4, byteorder="big")
self.t_0.to_bytes(length=8, byteorder="little")
self.t_1.to_bytes(length=8, byteorder="little")
```
> I'm looking at how to encode the Tuple of integers together as one bytes value, is there a standard way of doing that?

I was thinking about using `join` and `to_bytes` for each integer in the Tuple, and use the length and byte order in the EIP.

So e.g. for `h`, "8 unsigned 64-bit little-endian words" in the EIP, the integer representation could be `(0, 0, 0, 0, 0, 0, 0, 0)`, so eight numbers where the max is `2**64-1`, and then I think maybe using
```python
b"".join(i.to_bytes(length=8, byteorder="little") for i in self.h)
```

*But*... I think the main annoyance right now is taking the strings we have and converting that to integer tuples, so I'm not sure if it's worth it since `m` and `h` is not something that can be easily readable as `rounds` is for example.

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,580 @@
...
+        protected=True,
+        sender=sender,
+        value=100000,
+    )
+
+    post = {}
+    post = {
+        account: Account(
+            storage={
+                0: output.call_succeeds,
```
winsvega, line 574: Put some value in pre so if vector fails for oog, it will se the call result is deadbeef instead of 00

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,580 @@
...
+        ty=0x0,
+        to=account,
+        data=data.create_blake2b_tx_data(),
+        gas_limit=1_000_000,
+        protected=True,
+        sender=sender,
+        value=100000,
+    )
+
+    post = {}
```
winsvega, line 570: Typo

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,580 @@
...
+            ),
+            ExpectedOutput(
+                call_succeeds="0x01",
+                data_1="0x5bb981381beb687d5fdbe5e7c096fbd1ce193b780948c1d74ebbb7c58db364c7",
+                data_2="0xb7695d32f918444dbdcbdcff476fc70a926e228c4cbb7d05473711d3b56e5b33",
+            ),
+            id="EIP-152-0064-120",
+        ),
+        pytest.param(
+            Blake2bInput(
```
winsvega, line 506: Does it fill the vectors from original tests now no problem?
Which original vectors were incorrect?

reedsa:
@marioevz @winsvega I've updated the code based on your feedback, let me know if there's anything else. All cases are consistent with the cases that are in the original precompiled tests.

It's hard to tell if any edge cases may be missing. A [code search](https://github.com/search?q=repo:ethereum/tests%20blake&type=code) shows other json files but I think most of those cases are already covered by the new tests. I believe the json fixture  https://github.com/ethereum/tests/blob/c30599fb23d0470301f68bc69188ef444dcbc861/src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json#L2 is already covered in this PR.

There are several files such as https://github.com/ethereum/tests/blob/c30599fb23d0470301f68bc69188ef444dcbc861/GeneralStateTests/stPreCompiledContracts2/CALLCODEBlake2f.json#L2 which are checking `hash`, `logs` and `txbytes`. How are these generated? Should those json fixtures be captured as a separate test module since they use a different opcode `CALLCODE`?

Thanks for checking these and helping me understand if I missed anything!

tests/istanbul/eip152_blake2/test_blake2.py
reedsa: I added this case from this specific test: https://github.com/ethereum/tests/blob/c30599fb23d0470301f68bc69188ef444dcbc861/src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json#L197

reedsa: > Does it fill the vectors from original tests now no problem? Which original vectors were incorrect?

The expected output for each of the cases changed for any `t_0` value that was updated. So those with `16` bytes of data and `120` bytes of data.

EIP-152-16-16
...
EIP-152-1024-16

EIP-152-0016-120
EIP-152-0032-120
EIP-152-0064-120

tests/istanbul/eip152_blake2/test_blake2.py
reedsa: Added `Op.MSTORE(0x200, 0xDEADBEEF)` and updated the expectation of case8. Is that similar to what you have in mind? It looks like the CALL still updates the `0x200` location with 0 even if it fails.

tests/istanbul/eip152_blake2/test_blake2.py
winsvega: yes. because the vector name says 120 (0x78) but in actual test it passes 0x30. I assume ori mistaken in that test logic. therefor python version is correct

winsvega:
> @marioevz @winsvega I've updated the code based on your feedback, let me know if there's anything else. All cases are consistent with the cases that are in the original precompiled tests.
> 
> It's hard to tell if any edge cases may be missing. A [code search](https://github.com/search?q=repo:ethereum/tests%20blake&type=code) shows other json files but I think most of those cases are already covered by the new tests. I believe the json fixture https://github.com/ethereum/tests/blob/c30599fb23d0470301f68bc69188ef444dcbc861/src/GeneralStateTestsFiller/stPreCompiledContracts2/CALLBlake2fFiller.json#L2 is already covered in this PR.
> 
> There are several files such as https://github.com/ethereum/tests/blob/c30599fb23d0470301f68bc69188ef444dcbc861/GeneralStateTests/stPreCompiledContracts2/CALLCODEBlake2f.json#L2 which are checking `hash`, `logs` and `txbytes`. How are these generated? Should those json fixtures be captured as a separate test module since they use a different opcode `CALLCODE`?
> 
> Thanks for checking these and helping me understand if I missed anything!

hash logs and txbytes generated by the framework, you don't need to worry about it. 
the copy of test you provided is just the parametrization of same test but using a different callcode to the blake precompile

tests/istanbul/eip152_blake2/test_blake2.py
winsvega: where? if the call fails it should not modify the return memory.
isn't 200 too close to calldatasize?

tests/istanbul/eip152_blake2/test_blake2.py
reedsa: Sounds good, just wanted to be sure. Thanks!

tests/istanbul/eip152_blake2/test_blake2.py
reedsa: I was thinking the contract would overwrite the `deadbeef` value if the call succeeds, however I see you mentioned putting the value in pre but I am not aware of a solution for this.

I see a test which seems to have expectations for what is in storage, like https://github.com/ethereum/execution-spec-tests/blob/be1d1c9f483f69d5d2c644c9cb1ee3b3acd1e855/tests/osaka/eip7692_eof_v1/eip7069_extcall/test_calls.py#L535

> Put some value in pre so if vector fails for oog, it will se the call result is deadbeef instead of 00

For the oog case, `case8`, the call returns 0 but the transaction succeeds. Is there an example of what you mean for checking the `deadbeef` value is present?

reedsa: There's now a value in memory and saved in the result storage after `CALL`. The issue I'm seeing is that if the call fails, the value is `0x0` rather than `0xdeadbeef`. If the call succeeds, `0xdeadbeef` is untouched. Shouldn't it be the other way around?

Does this make sense to you? I don't understand why a failing call would alter that value to zero.

marioevz: So the place to add a pre-storage value would be during the `pre.deploy_contract` call:
```python
    account = pre.deploy_contract(..., storage={0: 0xDEADBEEF})
```
What happens here is that, at the time of deployment, the code is deployed but also the storage is pre-set so that the 0th key's value is equal to `0xDEADBEEF`.

So now, even before the transaction for your test is executed, there's a value already there.

This is only important in cases where a zero value in the storage is also a passing condition for the test.

For example, in some of the tests where we are expecting a zero as the return code of the call to the precompile. If for some reason the transaction failed, the test would still pass because zero is the default storage value of a any key, but since we are now setting `0xDEADBEEF` as the initial value of the key, the test will only pass if the transaction succeeded because only then the zero return code will overwrite `0xDEADBEEF`.

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,586 @@
...
+    output: ExpectedOutput,
+    pre: Alloc,
+):
+    """Test BLAKE2b precompile."""
+    env = Environment()
+
+    account = pre.deploy_contract(
+        # Store all CALLDATA into memory (offset 0)
+        Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
+        + Op.MSTORE(0x1000, 0xDEADBEEF)
```
marioevz, line 536:
```suggestion
```
This can be removed of the storage initial value alternative.

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,586 @@
...
+            Op.MLOAD(0x200),
+        )
+        + Op.SSTORE(
+            2,
+            Op.MLOAD(0x220),
+        )
+        + Op.SSTORE(
+            3,
+            Op.MLOAD(0x1000),
+        )
```
marioevz, line 561:
```suggestion
```
Also can be removed.

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,586 @@
...
+        + Op.STOP(),
+    )
```
marioevz, line 563:
```suggestion
        + Op.STOP(),
        storage={0: 0xDEADBEEF},
    )
```

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -0,0 +1,586 @@
...
+        value=100000,
+    )
+
+    post = {
+        account: Account(
+            storage={
+                0: 0x1 if output.call_succeeds else 0x0,
+                1: output.data_1,
+                2: output.data_2,
+                3: 0xDEADBEEF if output.call_succeeds else 0x0,
```
marioevz, line 582:
```suggestion
```
Can also be removed.

marioevz (commented):
Looks great, just the comments about the storage initial state that Dimitry suggested.

marioevz (approved):
LGTM, thanks for the changes!

I went over the coverage report and it seems related to the different approach we use in the python tests where we generate a different contract per test case instead of one big contract and handle the cases via jumps, and the lost coverage comes from the jump instructions not being covered anymore, which is completely reasonable to me since we are not looking to test this area with these cases.

PR #1073, opened 2025-01-14
new(tests): add EOF CALLF stack overflow validation tests

chfast:
## 🗒️ Description

Add new EOF / EIP-4750 CALLF tests that exercise the validator around stack height / overflow.

## ✅ Checklist


tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
```diff
@@ -343,3 +347,68 @@ def test_unreachable_code_sections(
...
+
+
+@pytest.mark.parametrize("callee_outputs", [1, 2, MAX_CODE_OUTPUTS])
+def test_callf_stack_height_limit_exceeded(eof_test, callee_outputs):
+    """Test for invalid EOF code containing CALLF instruction exceeding the stack height limit."""
+    callf_stack_height = 1024 - callee_outputs
+    container = Container(
+        sections=[
+            Section.Code(
+                Op.PUSH0 * callf_stack_height + Op.CALLF[1] + Op.POP + Op.STOP,
```
pdobacz, line 359: Op.POP actually needed or leftover?

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
```diff
@@ -343,3 +347,68 @@ def test_unreachable_code_sections(
...
+        ],
+    )
+    eof_test(data=container, expect_exception=EOFException.STACK_OVERFLOW)
+
+
+@pytest.mark.parametrize("callee_stack_height", [2, 3, MAX_OPERAND_STACK_HEIGHT])
+def test_callf_stack_overflow_by_height(eof_test, callee_stack_height):
+    """
+    Test for invalid EOF code containing CALLF instruction exceeding the runtime stack height limit
+    by calling a function with non-zero max stack height.
```
pdobacz, line 399: "non-zero" is inaccurate, because 1 would work. so maybe elaborate why 1 would work, ideally by including it in the params.

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
```diff
@@ -343,3 +347,68 @@ def test_unreachable_code_sections(
...
+    eof_test(data=container, expect_exception=EOFException.MAX_STACK_HEIGHT_ABOVE_LIMIT)
+
+
+@pytest.mark.parametrize("callee_outputs", [1, 2, MAX_CODE_OUTPUTS])
+def test_callf_stack_overflow_by_outputs(eof_test, callee_outputs):
+    """
+    Test for invalid EOF code containing CALLF instruction exceeding the runtime stack height limit
+    by calling a function with non-zero outputs.
+    """
+    callf_stack_height = 1025 - callee_outputs
```
pdobacz, line 378: It would help to expand that 1025 is runtime max stack limit of 1024 (not to confuse with `MAX_OPERAND_STACK_HEIGHT`, which is 1 less) plus one.

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
```diff
@@ -343,3 +347,68 @@ def test_unreachable_code_sections(
...
+    container = Container(
+        sections=[
+            Section.Code(
+                Op.PUSH0 * callf_stack_height + Op.CALLF[1] + Op.STOP,
+                max_stack_height=MAX_OPERAND_STACK_HEIGHT,  # this value is invalid
```
pdobacz, line 383: re the comment - it is invalid but that is not what invalidates the code here, correct? maybe expand on this in the comment, this confused me for a second here. Like so "this value doesn't match the code, but should be irrelevant to the behavior being tested".

Expanding on this, maybe it's actually worth to parametrize this value with some values like 0, 1, 2, MAX-1, MAX, expecting them all to fail the same way?

pdobacz (commented):
LGTM, left some ideas to discuss

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
chfast: Sadly there is no constant for the runtime stack height limit (1024).

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
chfast: correct. I don't fully understand how `max_stack_height` in pytest work. I can put value `1024` and the test will work but I cannot put `1025`...

marioevz: I think that, with `max_stack_height=1025`, the error that is returned by a correct client could be either `EOFException.STACK_OVERFLOW` or `EOFException.MAX_STACK_HEIGHT_ABOVE_LIMIT` depending on which check is performed first on this first section.

I agree that it might be worth it to at least also try `max_stack_height=callf_stack_height`, which will be 1025, with an expected exception of `EOFException.STACK_OVERFLOW | EOFException.MAX_STACK_HEIGHT_ABOVE_LIMIT`, just for the off chance that a faulty client's fault is being obscured by this other check (I.e. rejecting this EOF bytecode because it really did not match the max stack height expected even though it exceeds the limit).

chfast: I think in this case we should also put the invalid but within the range value: `max_stack_height=1025`. We want the error come from the code section validation, not from just inspecting the header.

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
chfast: I removed the POP.

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
chfast: Added constant.

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
chfast: Parametrized `max_stack_height` in the type section for some values. All are invalid if you compare with the value computed (1025).

tests/osaka/eip7692_eof_v1/eip4750_functions/test_code_validation.py
chfast: Explained.

PR #1081, opened 2025-01-16
feat(forks,plugins): Fork transition test marker

marioevz:
## 🗒️ Description
Adds two keyword parameters to the `valid_at_transition_to` marker:

- `subsequent_forks: bool`: If set to True, the test will be generated for the transition fork to the specified fork name, but also for all subsequent transition forks. Default is False.
- `until: str | None`: Specify an end to the fork transitions to be filed when using `subsequent_forks=True`. Implies `subsequent_forks=True`. Default is `None`.

E.g.
```python
@pytest.mark.valid_at_transition_to("Cancun", subsequent_forks=True)
```

produces tests on `ShanghaiToCancunAtTime15k` and `CancunToPragueAtTime15k`, and any transition fork after that.

Also added unit tests in `src/pytest_plugins/forks/tests/test_markers.py`.

Implementing tests for EIP-7691 in a separate PR, as required by lack of coverage pointed out in this thread: https://discord.com/channels/595666850260713488/1329093054698881087

### Validity Markers Refactor

All validity markers are now defined as classes that subclass a base validity marker class, [`ValidityMarker`](https://github.com/ethereum/execution-spec-tests/blob/73ab2c94b931ffbf66c6f51122a7cf9d24e8fbe4/src/pytest_plugins/forks/forks.py#L569).

This allows simpler validity marker definition, and easier handling of keyword arguments if it is the case that a validity marker supports them. See [`ValidAtTransitionTo`](https://github.com/ethereum/execution-spec-tests/blob/73ab2c94b931ffbf66c6f51122a7cf9d24e8fbe4/src/pytest_plugins/forks/forks.py#L711) for an example of how such a validity marker is defined.

## 🔗 Related Issues
None

## ✅ Checklist


marioevz:
> How about:
> 
> ```
> @pytest.mark.valid_at_transition_to("Cancun", subsequent_transitions=True, until="Osaka")
> ```

Addresed on the latest commit, which also includes a nice refactor of all validity markers into separate classes:

https://github.com/ethereum/execution-spec-tests/blob/73ab2c94b931ffbf66c6f51122a7cf9d24e8fbe4/src/pytest_plugins/forks/forks.py#L711-L747

src/ethereum_test_forks/tests/test_forks.py
```diff
@@ -153,6 +161,50 @@ def test_forks():
...
+    we expect the following logic:
+
+    Marker    Comparison   A->B Included
+    --------- ------------ ---------------
+    From A    fork >= A    True
+    Until A   fork <= A    False
+    From B    fork >= B    True
+    Until B   fork <= B    True
+    """
+    assert BerlinToLondonAt5 >= Berlin
```
winsvega, line 180: Finally! does it also work for fork >= Cancun  and fork < Osaka ?

src/pytest_plugins/forks/forks.py
```diff
@@ -552,151 +560,354 @@ def fork(request):
...
+
+    produces tests on `ShanghaiToCancunAtTime15k` and `CancunToPragueAtTime15k`, and any transition
+    fork after that.
+
+    And:
+    ```python
+    @pytest.mark.valid_at_transition_to("Cancun", subsequent_transitions=True, until="Prague")
+    ```
+
+    produces tests on `ShanghaiToCancunAtTime15k` and `CancunToPragueAtTime15k`, but no forks after
```
winsvega, line 802: what would it mean the test is produced at transition?  what are the input arguments to such test function? 
transition test comes as a blockchain test with a few block before transition, a transition block and a few blocks after. in each block we put some test logic.  so it is   x-1, x, x + 1 at least. how would it look like in python?

src/ethereum_test_forks/tests/test_forks.py
```diff
@@ -153,6 +161,50 @@ def test_forks():
...
+    assert not BerlinToLondonAt5 <= Berlin
```
winsvega, line 181: if berlin to london >= Berlin and   berlintolondon  not <= Berlin then berlintolondon != Berlin ? its tricky because the answer depends on what blocknumber it is. I would rather make transition forks uncomparable to standard forks.

winsvega (commented):
have a couple of questions

src/ethereum_test_forks/tests/test_forks.py
marioevz: Yes exactly, I agree that the comparison doesn't quite make sense between normal forks and transition forks.

These comparisons were used in my previous implementation of this PR, where I had a combination of markers like `valid_from("Cancun")` and `fork_transition_test`, so we needed to compare `Cancun` against all transition forks to see if the test needed to be filled for a given transition fork, but this is not the case anymore.

I can give it a try to modify the behavior to disallow such comparisons.

marioevz: > I can give it a try to modify the behavior to disallow such comparisons.

Yes that is correct, they are treated as different kinds of tests.

src/pytest_plugins/forks/forks.py
marioevz: Same as a normal blockchain test, with the difference that you need to add enough blocks to reach the `15_000` timestamp, which can happen even at block 1.

```python
@pytest.mark.valid_at_transition_to("Prague")
    def test_something_that_happens_during_the_fork_transition_to_prague(
        pre: Alloc
    ):
        blocks = [
           Block(timestamp=1_000, tx=[...]),  # Cancun block
           Block(timestamp=2_000, tx=[...]),  # Cancun block
           ...
           Block(timestamp=15_000, tx=[...]),  # First Prague block
           Block(timestamp=16_000, tx=[...]),  # Second Prague block
```

src/ethereum_test_forks/tests/test_forks.py
marioevz: I think `CancunToPrague >= Cancun` is a clear `True` IMO,  and also `CancunToPrague < Osaka`, but `PragueToOsaka < Osaka` is a bit harder to pin down.

marioevz:
@danceratopz merged your suggestion from the other PR, this one should be ready 😄

danceratopz (approved):
The refactoring of the validity markers is really nice! It adds much more structure to this plugin (I think this is one of the most complex parts of our pytest code). And really nice additional tests for these, btw! Few minor comments below.

src/pytest_plugins/forks/forks.py
```diff
@@ -552,151 +560,354 @@ def fork(request):
...
+        # Limit to non-transition forks if no validity markers were applied
+        test_fork_set: Set[Fork] = metafunc.config.all_forks  # type: ignore
     else:
-        if not valid_from:
-            valid_from = get_forks_with_no_parents(forks)
+        # Start with all forks and transitions if any validity markers were applied
+        test_fork_set: Set[Fork] = metafunc.config.all_forks_with_transitions  # type: ignore
+        for validity_marker in validity_markers:
+            # Apply the validity markers to the test function if applicable
+            test_fork_set = test_fork_set & validity_marker.process()
```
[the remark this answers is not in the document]
marioevz: Glad you liked it! I love this approach of subclassing to create stuff 😄

src/pytest_plugins/forks/forks.py
```diff
@@ -552,151 +560,354 @@ def fork(request):
...
-    forks: Set[Fork] = metafunc.config.forks  # type: ignore
-
-    intersection_set: Set[Fork] = set()
-    if valid_at_transition_to:
-        for fork in valid_at_transition_to:
-            if fork in fork_set:
-                intersection_set = intersection_set | set(transition_fork_to(fork))
+    validity_markers: List[ValidityMarker] = ValidityMarker.get_all_validity_markers(metafunc)
 
+    if not validity_markers:
```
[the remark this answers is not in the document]
marioevz: Yeah, this edge case is why this `if` exists, we could raise an error here instead very easily and fix that frontier test, in other PR perhaps 😄

src/ethereum_test_forks/tests/test_forks.py
marioevz: I'm going to keep an eye on new tests that make this kind of comparisons to see if it's worth it to handle comparing a transition fork to a normal fork, and we can make it stricter if necessary, but at this moment I think this should be ok.

PR #1095, opened 2025-01-20
feat(plugins/consume): Check port 8551 on Engine API tests

marioevz:
## 🗒️ Description
Use the [`"HIVE_CHECK_LIVE_PORT"` client environment value](https://github.com/ethereum/hive/blob/7910caeacba62a7172fd98dbe73322724c146297/internal/libhive/api.go#L265-L275) when starting a client for an Engine API test in the `consume` command.

This seems to be a recurrent issue when instantiating parallel client containers for hive runs, as it can be seen in https://hive.pectra-devnet-5.ethpandaops.io/suite.html?suiteid=1737336360-cdc9f3b55f2e828c17ab8b6c364934e8.json&suitename=eest%2Fconsume-engine#

The solution suggested by this PR is to signal hive to wait for the Engine API port (8551) instead of the RPC port (8545) before returning the client to the simulator (`consume`).

## 🔗 Related Issues
None

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [ ] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: A PR with removal of converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been opened.
- [x] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [x] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.

PR #1098, opened 2025-01-21
Add check live port as dedicated fixture

danceratopz:
This is a suggestion for #1095. 

It adds the check live as a dedicated fixture alongside the existing `environment` fixture in order to keep the hive client  config in one place.

I verified that the variable is correctly propagated locally for #1095 and for this PR.


marioevz (approved):
Thanks!

PR #1099, opened 2025-01-21
Add blob schedule docs

danceratopz:
Adds the blob schedule types to the consume section of the docs that describe our fixture formats.


marioevz (approved):
Thanks!

PR #1114, opened 2025-01-23
new(tests): CALLF with truncated immediate bytes

chfast:
## ✅ Checklist


pdobacz (approved):
you can also repeat the `imm_from_next_section` case with a byte in the data section - even better if it's the data section of the parent container, not unlike the bug which Danno's fuzzing found recently.

But maybe that would be too much

chfast:
Adding raw bytes to bytecode strikes back. I need to check what is the best solution so far.

chfast:
> you can also repeat the `imm_from_next_section` case with a byte in the data section - even better if it's the data section of the parent container, not unlike the bug which Danno's fuzzing found recently.
> 
> But maybe that would be too much

This test is quite sophisticated: if you overflow reading of one section to the next one the validation will pass. This only works for the `CALLF + "00"` case. Therefore, I'm going to pass on parametrizing it.

Issue #159, opened 2023-06-22
Fixtures missing from JSON when using pytest-xdist


danceratopz:
The `pytest-xdist` allows concurrent execution of test cases. If we enable xdist when running pytest for test cases, certain fixtures are not written to file even though the same amount of tests are reported to be executed from python.

To reproduce:
```console
rm -rf fixtures-without-xdist fixtures-with-xdist
fill -v --output=fixtures-without-xdist
fill -v -n auto --output=fixtures-with-xdist
ack -c "0\d\d-fork=" fixtures-without-xdist/ | awk -F: '{ total += $2 } END { print total }'
# -> 134
ack -c "0\d\d-fork=" fixtures-with-xdist/ | awk -F: '{ total += $2 } END { print total }'
# -> 35
```
Additionally, we should make the fork+test ordering and naming deterministic within fixtures. The top-level fork+test entries in the JSON fixtures are not deterministically ordered by fork+test. The goal would be to generate fixtures that are identical to those generated without using the xdist plugin.

Originally spotted by @spencer-tb!

danceratopz:
When python-xdist is enabled, each worker performs collection and sends the results back to the controller. Then the controller sends indices corresponding to a collected test id to the worker to execute the test. Fixtures, even session-scoped fixtures are executed in the pytest execution on each worker and the data is not joined automatically on the controller by default. In our case, the FixtureCollector is doing its job only for the subset of tests that get executed on each worker.

A nice partial and quick solution is to execute pytest-xdist using `--dist loadscipe` which seems sufficient for our current needs. I.e.,
```console
fill -n auto --dist loadscope
```
where the `--dist` flag:
```console
--dist=distmode       set mode for distributing tests to exec environments.
<-- snip -->
                        loadscope: load balance by sending pending groups of tests in the same scope to any available environment.
```
This ensures that all parametrizations of a test are executed on the same xdist worker, instead of being distributed equally across workers. In particular, this means that all the fixtures for one test case are executed on one work and the FixtureCollector will collect all the parametrized cases for this test and write them correctly to file, as for the `-n 0` case.

This has the consequence that running single test cases with xdist won't receive any speed-up, as they'll only use one worker. In this case, however, run time is probably not very critical. 

Benchmark on 4c402abcac4d13efde9f2a67f9a70363088d70a5 with 4844 tests on a laptop with 16 cores:
- `fill --until Cancun -n 0` - 112 seconds.
[...]

PR #1118, opened 2025-01-23
feat(fill): improve performance when xdist is enabled

marioevz:
## 🗒️ Description
Improves xdist workload distribution method by reverting `--dist` flag to its default value, `load`, and also make the fixture collector multi-process safe by creating a lock file per fixture file generated, and locking->reading->writing the file instead of overwriting it.

## 🔗 Related Issues
Description of xdist scoping here https://github.com/ethereum/execution-spec-tests/issues/159#issuecomment-1604250000 

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_fixtures/file.py
+++ b/src/ethereum_test_fixtures/file.py
@@ -4,6 +4,7 @@ import json
 from pathlib import Path
 from typing import Annotated, Any, Dict, Optional
 
+from filelock import FileLock
 from pydantic import Discriminator, Tag
 
 from ethereum_test_base_types import EthereumTestRootModel
@@ -67,8 +68,15 @@ class BaseFixturesRootModel(EthereumTestRootModel):
         json_fixtures: Dict[str, Dict[str, Any]] = {}
         for name, fixture in self.items():
             json_fixtures[name] = fixture.json_dict_with_info()
-        with open(file_path, "w") as f:
-            json.dump(json_fixtures, f, indent=4)
+        lock_file_path = file_path.with_suffix(".lock")
+        with FileLock(lock_file_path):
+            initial_data = {}
+            if file_path.exists():
+                with open(file_path, "r") as f:
+                    initial_data = json.load(f)
+            initial_data.update(json_fixtures)
+            with open(file_path, "w") as f:
+                json.dump(initial_data, f, indent=4)
 
     @classmethod
     def from_file(
--- a/src/ethereum_test_fixtures/file.py
+++ b/src/ethereum_test_fixtures/file.py
@@ -66,17 +66,16 @@ class BaseFixturesRootModel(EthereumTestRootModel):
         add the hash to the info field on per-fixture basis.
         """
         json_fixtures: Dict[str, Dict[str, Any]] = {}
-        for name, fixture in self.items():
-            json_fixtures[name] = fixture.json_dict_with_info()
         lock_file_path = file_path.with_suffix(".lock")
         with FileLock(lock_file_path):
-            initial_data = {}
             if file_path.exists():
                 with open(file_path, "r") as f:
-                    initial_data = json.load(f)
-            initial_data.update(json_fixtures)
+                    json_fixtures = json.load(f)
+            for name, fixture in self.items():
+                json_fixtures[name] = fixture.json_dict_with_info()
+
             with open(file_path, "w") as f:
-                json.dump(initial_data, f, indent=4)
+                json.dump(dict(sorted(json_fixtures.items())), f, indent=4)
 
     @classmethod
     def from_file(
--- a/src/pytest_plugins/filler/filler.py
+++ b/src/pytest_plugins/filler/filler.py
@@ -167,8 +167,16 @@ def pytest_addoption(parser: pytest.Parser):
         "--index",
         action="store_true",
         dest="generate_index",
-        default=False,
-        help="Generate an index file for all produced fixtures.",
+        default=True,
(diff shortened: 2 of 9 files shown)


src/ethereum_test_fixtures/file.py
```diff
@@ -65,10 +66,16 @@ def collect_into_file(self, file_path: Path):
...
+        lock_file_path = file_path.with_suffix(".lock")
+        with FileLock(lock_file_path):
+            if file_path.exists():
+                with open(file_path, "r") as f:
+                    json_fixtures = json.load(f)
+            for name, fixture in self.items():
+                json_fixtures[name] = fixture.json_dict_with_info()
+
+            with open(file_path, "w") as f:
+                json.dump(dict(sorted(json_fixtures.items())), f, indent=4)
```
danceratopz, line 78: I would prefer to have the fixtures ordered as pytest executes them with xdist disabled. But I'm guessing that you added `sorted` as it's no longer possible and otherwise the ordering with xdist is now non-deterministic?

One way to maintain this would be to apply sorted on test id per test function in `pytest_collection_modifyitems`.

src/pytest_plugins/filler/filler.py
```diff
@@ -167,8 +167,16 @@ def pytest_addoption(parser: pytest.Parser):
         "--index",
         action="store_true",
         dest="generate_index",
-        default=False,
-        help="Generate an index file for all produced fixtures.",
+        default=True,
```
danceratopz, line 170: This option now has no-effect, removed in #1130. Keeping it would maintain backward compatibility, but I don't think we need to worry about that yet?

src/pytest_plugins/filler/tests/test_filler.py
```diff
@@ -102,9 +102,57 @@ def test_shanghai_two(state_test, x):
                 Path("fixtures/state_tests/shanghai/module_shanghai/shanghai_two.json"),
             ],
             [2, 2, 2, 2, 2, 2, 2, 2, 2, 6, 6, 6],
-            False,
+            True,
             id="default-args",
         ),
+        pytest.param(
+            ["--index"],
```
danceratopz, line 109: I think we can clean these out now, done in #1130.

src/pytest_plugins/filler/filler.py
```diff
@@ -809,3 +760,44 @@ def pytest_collection_modifyitems(config: pytest.Config, items: List[pytest.Item
                     item.add_marker(mark)
         if "yul" in item.fixturenames:  # type: ignore
             item.add_marker(pytest.mark.yul_test)
+
+
+def pytest_sessionfinish(session: pytest.Session, exitstatus: int):
```
danceratopz, line 765: This seems a much better location for all these actions! :+1:

src/pytest_plugins/filler/filler.py
```diff
@@ -809,3 +760,44 @@ def pytest_collection_modifyitems(config: pytest.Config, items: List[pytest.Item
...
+    - Remove any lock files that may have been created.
+    - Generate index file for all produced fixtures.
+    - Create tarball of the output directory if the output is a tarball.
+    """
+    if xdist.is_xdist_worker(session):
+        return
+
+    output: Path = session.config.getoption("output")
+    is_output_tarball = output.suffix == ".gz" and output.with_suffix("").suffix == ".tar"
+    output_dir = strip_output_tarball_suffix(output)
```
danceratopz, line 778: V minor nit, I'd prefer to define these closer to their use for readability.

src/pytest_plugins/filler/filler.py
```diff
@@ -809,3 +760,44 @@ def pytest_collection_modifyitems(config: pytest.Config, items: List[pytest.Item
...
+
+    if is_output_stdout(output):
+        # Don't perform any further actions if the output is stdout.
+        return
```
danceratopz, line 782: ...especially so this "exit early" case is more prominently at the top of the function, see #1130.

danceratopz (commented):
Awesome, thanks! Few comments and suggestions in #1130.

src/ethereum_test_fixtures/file.py
marioevz: > One way to maintain this would be to apply sorted on test id per test function in `pytest_collection_modifyitems`.

Even then I think it's not possible to know which worker is going to finish filling first, even if the workload is sent out in order.

The reason why I thought of adding this is because it makes it a bit easier to visually compare the output fixtures and the whole fixture file is already in memory so it should not be that much extra overhead.

src/pytest_plugins/filler/filler.py
marioevz: PR merged so this is done now.

src/pytest_plugins/filler/tests/test_filler.py
marioevz: PR merged so this is done now.

src/pytest_plugins/filler/filler.py
marioevz: PR merged so this is done now.

src/ethereum_test_fixtures/file.py
danceratopz: I think the main thing is to ensure that the behavior of "xdist" vs "no xdist" is consistent, which it now will be as `sorted` is applied in either case. I don't mind changing the behavior, thanks for the follow-up!

danceratopz (approved):
LGTMMMMM!

PR #1119, opened 2025-01-23
feat(fw): allow adding verbatim bytes to Bytecode

chfast:
## 🗒️ Description

This provides a basic requested feature to allow concatenating `Bytecode` with verbatim `bytes`.
This implementation overload the `__add__` and make the basic use case work.
However, this is not very consistent. E.g. 
- `code + b"\x00"` means something different than `code + Bytecode(b"\x00")`
- `Bytecode() + b"\x00"` works in places where `Bytecode()` is expected but `b"\x00"` don't.

The reason this is not a complete solution is that `Bytecode()` already has a constructor from `bytes` but its purpose is different and unknown to me. We should consider changing the semantic of conversion from bytes to `Bytecode()` by just copying the bytes.

There is one more option for handling truncated opcode immediate bytes: take verbatim bytes from `[]`, e.g. `Op.PUSH3[b"\x00"]`. This currently compiles, but `Bytecode` also tries to be smart here and extends the provided bytes to the desired length. I'm not 100% sure what it does but I couldn't built a code with truncated immediates this way. However, there is one example that seems to be working: https://github.com/ethereum/execution-spec-tests/blob/ca912404cf7d8e95c7b06c3a5adcac4d37d7fbce/tests/osaka/eip7692_eof_v1/eip4200_relative_jumps/test_rjumpv.py#L273

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_vm/tests/test_vm.py
+++ b/src/ethereum_test_vm/tests/test_vm.py
@@ -413,15 +413,17 @@ def test_opcode_comparison():
     assert Op.ADD != Op.STOP
     assert Op.ADD > Op.STOP
 
-def test_bytecode_concatenation_with_bytes():
-    """Test that the bytecode can be concatenated with bytes.
-    Bytes work as verbatim code and don't affect the bytecode properties."""
 
-    base = Op.PUSH1[0xff] + Op.NOT
+def test_bytecode_concatenation_with_bytes():
+    """
+    Test that the bytecode can be concatenated with bytes.
+    Bytes work as verbatim code and don't affect the bytecode properties.
+    """
+    base = Op.PUSH1[0xFF] + Op.NOT
     assert str(base) == ""
 
     code = base + b"\x01\x02"
-    assert code == bytes([0x60, 0xff, 0x19, 0x01, 0x02])
+    assert code == bytes([0x60, 0xFF, 0x19, 0x01, 0x02])
 
     assert str(code) == ""
     assert code.popped_stack_items == code.popped_stack_items
--- a/src/ethereum_test_vm/tests/test_vm.py
+++ b/src/ethereum_test_vm/tests/test_vm.py
@@ -412,3 +412,20 @@ def test_opcode_comparison():
     assert Op.ADD == Op.ADD
     assert Op.ADD != Op.STOP
     assert Op.ADD > Op.STOP
+
+def test_bytecode_concatenation_with_bytes():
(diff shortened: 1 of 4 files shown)


marioevz (approved):
LGTM. I pushed some minor fixes and will merge once the CI passes.

marioevz:
Regarding the inconsistencies, this feature should be handled with care anyway, and I think we should only use it when absolutely necessary, so I will keep an eye on usages during reviews to make sure it doesn't get out of control :P

src/ethereum_test_vm/tests/test_vm.py
```diff
@@ -412,3 +412,22 @@ def test_opcode_comparison():
...
+    Bytes work as verbatim code and don't affect the bytecode properties.
+    """
+    base = Op.PUSH1[0xFF] + Op.NOT
+    assert str(base) == ""
+
+    code = base + b"\x01\x02"
+    assert code == bytes([0x60, 0xFF, 0x19, 0x01, 0x02])
+
+    assert str(code) == ""
+    assert code.popped_stack_items == code.popped_stack_items
```
pdobacz, line 429: sorry for noticing late - these assertions are void - should be ` == base...` instead?

tests/osaka/eip7692_eof_v1/eip4200_relative_jumps/test_rjump.py
```diff
@@ -116,7 +116,7 @@ def test_rjump_truncated_rjump_2(
 ):
     """EOF1I4200_0002 (Invalid) EOF code containing truncated RJUMP."""
     eof_test(
-        data=Container.Code(Op.RJUMP + Op.STOP),
+        data=Container.Code(Op.RJUMP + b"\x00"),
```
pdobacz, line 119: to address @marioevz 's concerns, maybe this should be `Op.RJUMP + verbatim(b"\x00")`, and `verbatim` produces some type which can be added to `Bytecode` but isn't `bytes` explicitly, possibly `Bytecode` would be fine? This would make spotting abuse much easier.

Basically `verbatim` could be what @chfast envisioned for the constructor `Bytecode`.

pdobacz (commented):
I had some late comments

src/ethereum_test_vm/tests/test_vm.py
danceratopz: Fixed by @chfast in #1137

PR #1120, opened 2025-01-23
feat(tests) Precompile Checks

reedsa:
## 🗒️ Description

Add tests to check each precompile at specific addresses.

Tests from: https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stPreCompiledContracts/idPrecompsFiller.yml

Removed in:
https://github.com/ethereum/tests/pull/1444

## ✅ Checklist

### DIFF
--- a/tests/frontier/precompiles/test_precompiles.py
+++ b/tests/frontier/precompiles/test_precompiles.py
@@ -1,0 +1,111 @@
+"""Tests supported precompiled contracts."""
+
+from typing import Iterator, Tuple
+
+import pytest
+
+from ethereum_test_forks import Fork
+from ethereum_test_tools import (
+    Account,
+    Alloc,
+    Environment,
+    StateTestFiller,
+    Transaction,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+UPPER_BOUND = 0xFF
+NUM_UNSUPPORTED_PRECOMPILES = 1
+
+
+def precompile_addresses(fork: Fork) -> Iterator[Tuple[str, bool]]:
+    """
+    Yield the addresses of precompiled contracts and their support status for a given fork.
+
+    Args:
+        fork (Fork): The fork instance containing precompiled contract information.
+
+    Yields:
+        Iterator[Tuple[str, bool]]: A tuple containing the address in hexadecimal format and a
+            boolean indicating whether the address is a supported precompile.
+
+    """
+    supported_precompiles = fork.precompiles()
+
+    num_unsupported = NUM_UNSUPPORTED_PRECOMPILES
+    for address in range(1, UPPER_BOUND + 1):
+        if address in supported_precompiles:
+            yield (hex(address), True)
+        elif num_unsupported > 0:
+            # Check unsupported precompiles up to NUM_UNSUPPORTED_PRECOMPILES
+            yield (hex(address), False)
+            num_unsupported -= 1
+
+
+@pytest.mark.valid_from("Berlin")
+@pytest.mark.parametrize_by_fork("address,precompile_exists", precompile_addresses)
+def test_precompiles(
+    state_test: StateTestFiller, address: str, precompile_exists: bool, pre: Alloc
+):
+    """
+    Tests the behavior of precompiled contracts in the Ethereum state test.
+
+    Args:
+        state_test (StateTestFiller): The state test filler object used to run the test.
+        address (str): The address of the precompiled contract to test.
+        precompile_exists (bool): A flag indicating whether the precompiled contract exists at the
+            given address.
+        pre (Alloc): The allocation object used to deploy the contract and set up the initial
+            state.
+
+    This test deploys a contract that performs two CALL operations to the specified address and a
+    fixed address (0x10000), measuring the gas used for each call. It then stores the difference
+    in gas usage in storage slot 0. The test verifies the expected storage value based on
+    whether the precompiled contract exists at the given address.
+
+    """
+    env = Environment()
+
+    account = pre.deploy_contract(
+        Op.MSTORE(0, 0)  # Pre-expand the memory so the gas costs are exactly the same
+        + Op.GAS
+        + Op.CALL(
+            address=address,
+            value=0,
+            args_offset=0,
+            args_size=32,
+            output_offset=32,
+            output_size=32,
+        )
+        + Op.POP
+        + Op.SUB(Op.SWAP1, Op.GAS)
+        + Op.GAS
+        + Op.CALL(
+            address=pre.fund_eoa(amount=0),
+            value=0,
+            args_offset=0,
+            args_size=32,
+            output_offset=32,
+            output_size=32,
+        )
+        + Op.POP
+        + Op.SUB(Op.SWAP1, Op.GAS)
+        + Op.SWAP1
+        + Op.SUB
+        + Op.SSTORE(0, Op.ISZERO)
+        + Op.STOP,
+        storage={0: 0xDEADBEEF},
+    )
+
+    tx = Transaction(
+        to=account,
+        sender=pre.fund_eoa(),
+        gas_limit=1_000_000,
+        protected=True,
+    )
+
+    # A high gas cost will result from calling a precompile
+    # Expect 0x00 when a precompile exists at the address, 0x01 otherwise
+    post = {account: Account(storage={0: "0x00" if precompile_exists else "0x01"})}
+
+    state_test(env=env, pre=pre, post=post, tx=tx)
--- a/converted-ethereum-tests.txt
+++ b/converted-ethereum-tests.txt
@@ -82,3 +82,6 @@ GeneralStateTests/VMTests/vmTests/push.json
 
 ([#1067](https://github.com/ethereum/execution-spec-tests/pull/1067))
 GeneralStateTests/stPreCompiledContracts/blake2B.json
+
+([#1067](https://github.com/ethereum/execution-spec-tests/pull/1120))
+GeneralStateTests/stPreCompiledContracts/idPrecomps.json


tests/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,341 @@
...
+        ("02", True),
+        ("03", True),
+        ("04", True),
+        ("05", True),
+        ("06", True),
+        ("07", True),
+        ("08", True),
+        ("09", True),
+        ("0A", True),
+        ("0B", False),
```
marioevz, line 26: This list would only be valid for Cancun, but we could use this marker [`@pytest.mark.parametrize_by_fork`](https://ethereum.github.io/execution-spec-tests/main/writing_tests/test_markers/#pytestmarkparametrize_by_fork) to dynamically generate this list by fork.

I would also reduce the amount of cases to the first `exists=False` fork.

marioevz (commented):
Gave this one a quick look and I wrote one comment.

We also have this similar test over here: https://github.com/ethereum/execution-spec-tests/blob/main/tests/frontier/precompiles/test_precompile_absence.py

reedsa:
> Gave this one a quick look and I wrote one comment.
> 
> We also have this similar test over here: https://github.com/ethereum/execution-spec-tests/blob/main/tests/frontier/precompiles/test_precompile_absence.py

Those tests are very similar. Do you think it's still worth migrating these tests?

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,107 @@
...
+    account = pre.deploy_contract(
+        Op.MSTORE(args_size, 0x20)
+        + Op.MSTORE(input_buf, 0xFF)
+        + Op.MSTORE(output_buf, 0xFF)
+        + Op.MSTORE8(input_buf, 0xFF)
+        + Op.MSTORE8(output_buf, 0xFF)
+        + Op.MSTORE(gas_test, Op.GAS())
+        # Setup stack to CALL into precompile with the CALLDATA and CALL into it (+ pop value)
+        + Op.CALL(
+            address=Op.CALLDATALOAD(0),
```
reedsa, line 70: Trying to pass in the precompile address in the `data` of the transaction for this call. It doesn't seem like I can pass the correct value from the bytes value passed, even if I try to convert the value from the `CALLDATALOAD` to an integer. Is there a recommendation for how to approach this?

reedsa:
@winsvega @marioevz now I have the data (precompile address) being sent into the transaction. It looks like all tests are returning 0 however. Is there something I may have missed with the logic in the contract? In the original test, the gas cost difference is found based on a condition that the precompile gas is greater than the gas for the call at the address `0x10000`.

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Args:
+        fork (str): The fork name.
+
+    Returns:
+        list[tuple[str, bool]]: The precompile addresses and whether they exist.
+
+    """
+    return [
+        ("01", True),
```
winsvega, line 26: you can access fork class from the test if you add  fixture in the parameter list.
fork: Fork 

and then it has a function that returns all precompiles currently active on that fork as integer list.

tests/frontier/precompiles/test_precompiles.py
winsvega: here the data must be padded as EVM word of 32 bytes. so not enough just fromhex(1 byte) you need to padd it with zeros to the left. like Hash32(data, left_padding=true)

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,99 @@
...
+        + Op.MSTORE(gas_10000, Op.SUB(Op.GAS(), Op.MLOAD(gas_10000)))
+        + Op.SSTORE(
+            0,
+            Conditional(
+                condition=Op.GT(Op.MLOAD(gas_test), Op.MLOAD(gas_10000)),
+                if_true=Op.SUB(Op.MLOAD(gas_test), Op.MLOAD(gas_10000)),
+                if_false=Op.SUB(Op.MLOAD(gas_10000), Op.MLOAD(gas_test)),
+            ),
+        )
+        + Op.SSTORE(0, Op.LT(Op.SLOAD(0), 0x10))
```
winsvega, line 81: here also a pattern we use. there is class Storage.
so you can call `Op.SSTORE(storage.store_next(expected_value, "debug hint"),  logic)`

winsvega (commented):
added comments

reedsa:
@marioevz @winsvega Latest revisions are creating parameterized tests for each fork with valid precompiles as well as a single invalid precompile. I noticed the gas cost is slightly higher than what the original tests were using. Instead of the gas being compared with `0x10` [in the original test](https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/stPreCompiledContracts/idPrecompsFiller.yml#L68), it now checks that the cost is less than `0x1A4`.

This however fails with Byzantium and Constantinople for precompile 5 and I'm not sure why.

winsvega:
I think the tes was not designed for that forks. Check in ethereum/legacytests if it has a revision for other forks.

But yes, need to understand what's going on

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,114 @@
...
+    args_offset = 0x1000
+    args_size = 0x20
+    output_offset = 0x2000
+    output_size = 0x20
+
+    gas_test = 0x00
+    gas_10000 = 0x20
+
+    account = pre.deploy_contract(
+        Op.MSTORE(gas_test, Op.GAS)
```
winsvega, line 70: @marioevz it works `Op.GAS` as well as `Op.GAS()` ?

.vscode/launch.json
```diff
@@ -0,0 +1,133 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  // If the VS Code "Run and Debug" button, respecively launch selector are not visible, see this answer:
+  // https://stackoverflow.com/a/74245823
+  // 
+  "version": "0.2.0",
```
winsvega, line 8: why the settings got exported?

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,114 @@
...
+
+    This test deploys a contract that performs two CALL operations to the specified address and a
+    fixed address (0x10000), measuring the gas used for each call. It then stores the difference
+    in gas usage in storage slot 0. The test verifies the expected storage value based on
+    whether the precompiled contract exists at the given address.
+
+    """
+    env = Environment()
+
+    args_offset = 0x1000
```
winsvega, line 61: I see you didn't want to copy, but then you can actually use default values of this args leaving it empty. 
and it is always good when reading the code If it is visible what is the offset without scrolling up to the defenition (IMHO)

winsvega (changes requested):
when implementing test for the first time, we, actually take our understanding of vm for granted, but newcomers must use vmtrace. and it always good to use vmtrace to see what is actually happening.  (yes by now we have little compiler in our head) 
I remember the first year I was using vmtrace all the time, it is essential to get an idea of what is going on and how your code is executed. 

saying that we might want to use/develop better tools for vmtrace visualizing and debugging. and manuals.

@marioevz @danceratopz

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,114 @@
...
+        fork (Fork): The fork instance containing precompiled contract information.
+
+    Yields:
+        Iterator[Tuple[str, bool]]: A tuple containing the address in hexadecimal format and a
+            boolean indicating whether the address is a supported precompile.
+
+    """
+    supported_precompiles = fork.precompiles()
+
+    for address in range(1, len(supported_precompiles) + 2):
```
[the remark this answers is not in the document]
reedsa: I can extract this into a variable to describe it a little better. The range is set to include the first address where a precompile does not exist. That will cover what @marioevz mentioned in https://github.com/ethereum/execution-spec-tests/pull/1120#discussion_r1928970559

To support disjointed ranges, perhaps this should be changed to some maximum and ignore most addresses that do not have a supported precompile address. This would effectively be the inverse of the "precompile absence" tests: https://github.com/ethereum/execution-spec-tests/blob/093d119e804ae4b0e99701214b252579bd39b8d8/tests/frontier/precompiles/test_precompile_absence.py#L40

tests/frontier/precompiles/test_precompiles.py
reedsa: I've seen this in a few places like https://github.com/ethereum/execution-spec-tests/blob/093d119e804ae4b0e99701214b252579bd39b8d8/tests/osaka/eip7692_eof_v1/gas_test.py#L81

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,114 @@
...
+        + Op.MSTORE(gas_test, Op.SUB(Op.GAS, Op.MLOAD(gas_test)))
+        + Op.MSTORE(gas_10000, Op.GAS)
+        + Op.CALL(
+            address=0x10000,
+            args_offset=args_offset,
+            args_size=args_size,
+            output_offset=output_offset,
+            output_size=output_size,
+        )
+        + Op.MSTORE(gas_10000, Op.SUB(Op.GAS, Op.MLOAD(gas_10000)))
```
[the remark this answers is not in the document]
reedsa: Good catch, I think I got these backward.

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,103 @@
...
+    This test deploys a contract that performs two CALL operations to the specified address and a
+    fixed address (0x10000), measuring the gas used for each call. It then stores the difference
+    in gas usage in storage slot 0. The test verifies the expected storage value based on
+    whether the precompiled contract exists at the given address.
+
+    """
+    env = Environment()
+
+    account = pre.deploy_contract(
+        Op.MSTORE(0x00, Op.GAS)
```
winsvega, line 71: yes now 0x00 can become a variable like  

slot_precompile_call_gas = 0x00
slot_empty_address_call_gas = 0x20

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,103 @@
...
+        + Op.CALL(address=address)
+        + Op.MSTORE(0x00, Op.SUB(Op.MLOAD(0x00), Op.GAS))
+        + Op.MSTORE(0x20, Op.GAS)
+        + Op.CALL(address=0x10000)
```
winsvega, line 75: usually we use input 0, length 32  output 32 length 32 or so. thats why ori marked this args as 0xff to not confuse the memory. because call can touch our memory if we are not careful.

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,103 @@
...
+    tx = Transaction(
+        to=account,
+        sender=pre.fund_eoa(),
+        gas_limit=1_000_000,
+        protected=True,
+    )
+
+    # A high gas cost will result from calling a precompile
+    # Expect 0x00 when a precompile exists at the address, 0x01 otherwise
+    post = {account: Account(storage={0: "0x00" if precompile_exists else "0x01"})}
```
winsvega, line 101: here I don't clearly understand why abs(empty_call_gas - precompile_call_gas) < 0x1a4 means this logic. 
if you can rewrite it more clean like 
`if precompile_exists then precompile_call_gas  is x` else y  if it is possible.

tests/precompiles/test_precompiles.py
winsvega: yes. but what if we will have gaps? like new precompiles defined 0x20, 0x21, 0x22 but not at 0x1f, 0x1e

winsvega (approved):
I think now it covers the original test. a little optimised  with undefined precompiles check. 

just a few notes if we can now make it even better

tests/frontier/precompiles/test_precompiles.py
reedsa: I'm not clear on this, since the actual gas value varies by precompile call. I'm not sure what logic could be set outside of the contract code. Could you explain a bit more on this?

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -0,0 +1,103 @@
...
+    supported_precompiles = fork.precompiles()
+
+    num_unsupported = NUM_UNSUPPORTED_PRECOMPILES
+    for address in range(1, UPPER_BOUND + 1):
+        if address in supported_precompiles:
+            yield (hex(address), True)
+        elif num_unsupported > 0:
+            # Check unsupported precompiles up to NUM_UNSUPPORTED_PRECOMPILES
+            yield (hex(address), False)
+            num_unsupported -= 1
```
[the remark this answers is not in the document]
reedsa: Since there are other tests for no precompile in `test_precompile_absence`, having those checks here as well would be redundant. I think that's why @marioevz said we should just check the first address that a precompile does not exist. I could change the upper bound to 0xff if need be, but I think the `fork.precompiles` for the forks being checked in this test are mostly just 0x1 - 0x9, and for cancun it includes 0xa as well.

> I think now we have separate range of special addresses starting from 0x100 too.

Should the fork that supports this be included here? I don't see those specific ranges in the fork classes.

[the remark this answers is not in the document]
reedsa: I've changed this to loop through the upper bound of `0xFF` and only add a single unsupported address otherwise.

marioevz (commented):
Looks great! Great test scenario, we have a similar one where we use zero gas for the calls, but I think this expands the coverage since there could be some cases where zero gas behaves differently in some implementations.

A few comments on how we can skip the conditional in the bytecode entirely, and also how to guarantee we can use an account with no code in it.

marioevz:
Checked locally and the CI error is because this precompile is disabled in evmone by default, adding flag `-DEVMONE_PRECOMPILES_SILKPRE=1` fixes the state test, and if it becomes a problem we can add it to the CI.

Merging now 👍

PR #1131, opened 2025-01-27
export fork and chain id fixture config

winsvega:
## 🗒️ Description
export chain id and network to the test config field so it won't be empty in state tests

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_fixtures/blockchain.py
+++ b/src/ethereum_test_fixtures/blockchain.py
@@ -401,6 +401,7 @@ class FixtureConfig(CamelModel):
     """Chain configuration for a fixture."""
 
     fork: str = Field(..., alias="network")
+    chain_id: int = Field(1, alias="chainid")
     blob_schedule: FixtureBlobSchedule | None = None
 
 
--- a/src/ethereum_test_fixtures/state.py
+++ b/src/ethereum_test_fixtures/state.py
@@ -98,6 +98,8 @@ class FixtureConfig(CamelModel):
     """Chain configuration for a fixture."""
 
     blob_schedule: FixtureBlobSchedule | None = None
+    chain_id: int = Field(1, alias="chainid")
+    fork: str = Field(..., alias="network")
 
 
 class StateFixture(BaseFixture):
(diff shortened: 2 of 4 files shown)


src/ethereum_test_fixtures/blockchain.py
```diff
@@ -401,6 +401,7 @@ class FixtureConfig(CamelModel):
     """Chain configuration for a fixture."""
 
     fork: str = Field(..., alias="network")
+    chain_id: int = Field(1, alias="chainid")
```
marioevz, line 404: I think this would be the sole `int` type in the entire fixture format.

Should we simply use `ZeroPaddedHexNumber` or `HexNumber` here to make it easier for parsers?

src/ethereum_test_fixtures/state.py
```diff
@@ -98,6 +98,8 @@ class FixtureConfig(CamelModel):
     """Chain configuration for a fixture."""
 
     blob_schedule: FixtureBlobSchedule | None = None
+    chain_id: int = Field(1, alias="chainid")
```
marioevz, line 101: Same comment as in the Blockchain test format, although here the `FixtureForkPostIndexes` are actually `int`, so should not be as problematic as in the other one.

marioevz (commented):
Just two small comments, and we need to add these fields to the documentation, but I can help out with that.

winsvega:
need to add into hive too

src/ethereum_test_fixtures/blockchain.py
danceratopz: I think that `ZeroPaddedHexNumber` would be the most consistent choice.

src/pytest_plugins/consume/hive_simulators/conftest.py
```diff
@@ -143,7 +143,7 @@ def environment(
         blockchain_fixture.fork in ruleset
     ), f"fork '{blockchain_fixture.fork}' missing in hive ruleset"
     return {
-        "HIVE_CHAIN_ID": "1",
+        "HIVE_CHAIN_ID": blockchain_fixture.config.chain_id,
```
marioevz, line 146:
```suggestion
        "HIVE_CHAIN_ID": str(Number(blockchain_fixture.config.chain_id)),
```

Need to make this cast here to make it work because all client mappers expect a decimal number as a string.

marioevz (approved):
LGTM, thanks!

PR #1139, opened 2025-01-28
fix(consume/engine): check if client fails to raise an expected Engine API error

danceratopz:
## 🗒️ Description
This PR adds a missing check for the case of Engine API exceptions.

If the fixture specifies a payload `errorCode`, then check that it is indeed raised by the client. If not, fail the test.

## 🔗 Related Issues
Small follow-up to #1133.

## ✅ Checklist


marioevz (approved):
Thanks so much for debugging and fixing this!

PR #1142, opened 2025-01-28
🧹 chore(docs): wider, optimised banner

raxhvl:
Wider banner. Size is ~94% smaller from 1.2mb to 65kb; should speed up loading the landing page.


danceratopz (approved):
Nice, thanks for this!

PR #1151, opened 2025-01-29
feat(tests): add worst-case benchmark for `MOD`

chfast:
## 🗒️ Description

Add a worst-case test running a block with as many MOD instructions with arguments of the parametrized range.

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/1571

## ✅ Checklist

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -32,6 +32,21 @@ KECCAK_RATE = 136
 ECRECOVER_GAS_COST = 3_000
 
 
+def neg(x: int) -> int:
+    """Negate the given integer in the two's complement 256-bit range."""
+    assert 0 <= x < 2**256
+    return 2**256 - x
+
+
+def make_dup(index: int) -> Opcode:
+    """
+    Create a DUP instruction which duplicates the index-th (counting from 0) element
+    from the top of the stack. E.g. make_dup(0) → DUP1.
+    """
+    assert 0 <= index < 16
+    return Opcode(0x80 + index, pushed_stack_items=1, min_stack_height=index + 1)
+
+
 @pytest.mark.valid_from("Cancun")
 def test_worst_keccak(
     blockchain_test: BlockchainTestFiller,
@@ -596,7 +611,7 @@ def test_worst_shifts(
     rng = random.Random(1)  # Use random with a fixed seed.
     initial_value = 2**256 - 1  # The initial value to be shifted; should be negative for SAR.
 
-    # Create the list of shift amounts if length 15 (max reachable by DUPs instructions).
+    # Create the list of shift amounts with 15 elements (max reachable by DUPs instructions).
     # For the worst case keep the values small and omit values divisible by 8.
     shift_amounts = [x + (x >= 8) + (x >= 15) for x in range(1, 16)]
 
@@ -613,18 +628,13 @@ def test_worst_shifts(
             if new_v != 0:
                 return new_v, index
 
-    def make_dup(i):
-        """Create a DUP instruction to get the i-th shift amount constant from the stack."""
-        # TODO: Create a global helper for this.
-        return Opcode(0x80 + (len(shift_amounts) - i))
-
     code_body = Bytecode()
     v = initial_value
     while len(code_body) <= code_body_len - 4:
         v, i = select_shift_amount(shl, v)
-        code_body += make_dup(i) + Op.SHL
+        code_body += make_dup(len(shift_amounts) - i) + Op.SHL
         v, i = select_shift_amount(shift_right_fn, v)
-        code_body += make_dup(i) + shift_right
+        code_body += make_dup(len(shift_amounts) - i) + shift_right
 
     code = code_prefix + code_body + code_suffix
     assert len(code) == MAX_CODE_SIZE - 2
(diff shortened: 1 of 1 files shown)


tests/benchmarks/test_mod.py
```diff
@@ -0,0 +1,128 @@
...
+            code += dup + Op.MOD
+        code += Op.POP
+
+    account = pre.deploy_contract(code)
+
+    tx = Transaction(
+        to=account,
+        gas_limit=30_000_000,
+        gas_price=10,
+        sender=sender,
```
marioevz, line 76:
```suggestion
        sender=sender,
        protected=False,
```
Just so the transaction is successful in Frontier too.

tests/benchmarks/test_mod.py
```diff
@@ -0,0 +1,128 @@
...
+            code += Op.MOD
+        code += Op.POP
+
+    account = pre.deploy_contract(code)
+
+    tx = Transaction(
+        to=account,
+        gas_limit=30_000_000,
+        gas_price=10,
+        sender=sender,
```
marioevz, line 125:
```suggestion
        sender=sender,
        protected=False,
```

marioevz (commented):
I think we should rather create a marker, e.g.:

```python
@pytest.mark.benchmark
```

And then you can place these tests in `tests/frontier/opcodes/` but still select them with `uv run fill -m benchmark`.

How do you see the clients consuming these tests for benchmarks? Is there a command in evmone to run benchmark state tests?

chfast:
Sorry, I didn't put a proper description of this yet. But here are my thoughts about benchmarks in EEST.

1. Benchmarks are formally regular tests and the existing types (state tests, blockchain tests, even EOF tests and transaction tests) are already very useful. They should be produced with `fill` and be consumed in a normal way.
2. We still should separate them from the main tests, maybe by using a marker like `@pytest.mark.benchmark` and/or in a separate directory.
3. They should not contribute to the test coverage. Therefore, not be part of the `fixtures` release or physically separated.
4. We want to generate them for a single fork and likely a recent one. It is an open question if there should be a default fork for all benchmarks and how to upgrade the fork. But in generally, we are mostly interested about EVM performance near the tip of the chain, not in the historical execution.
5. That's also the reason why sorting them by the fork names is not good idea. Potential categories are something like: real use cases (e.g. SHA1), micro-benchmarks, worst-case, etc. But I don't think we need to decide categories right now.
6. I imagine used to be like this: if this is the regular test execution:
         ```evm test benchmark.json```
   then benchmarking is:
       ```evm test --bench benchmark.json```
   The difference is just with `--bench` the test is executed in the loop and the execution time is reported.
7. Because of this loop execution we may need to put some additional requirements on the benchmarks. E.g. it should not modify a state, if it modifies a stare the state diff should be the same in every iteration, the test itself should execute for at least 1 ms.

marioevz:
@chfast I like the idea, and basically everything you mention is somewhat easily doable in EEST.

One question, for (6), is there a pass/fail condition? Or is this something that is left to the client test consumer discretion?

chfast:
> One question, for (6), is there a pass/fail condition? Or is this something that is left to the client test consumer discretion?

In the example `evm test benchmark.json` is a regular test execution: no change in the fail/pass conditions.

The additional `--bench` likely will execute a benchmark in a loop and report the timing statistics. What I try to do in my current tooling is to execute it once and check assertions. Only then switch to the loop execution. In general, you want to make sure the benchmarks are executed correctly otherwise there is no point to benchmark incorrect implementation (or the test itself). In practice the exact method my depend on the time for single execution. It is indeed up to the client teams to decided. But also up to us to audit these decisions.

tests/zkevm/test_worst_compute.py
```diff
@@ -644,3 +647,104 @@ def make_dup(i):
...
+    The "numerators" is a pool of 15 constants pushed to the EVM stack at the program start.
+    The order of accessing the numerators is selected in a way the mod value remains in the range
+    as long as possible.
+    """
+    num_numerators = 15
+    numerator_bits = 256
+    numerator_max = 2**numerator_bits - 1
+    numerator_min = 2 ** (numerator_bits - 1)
+
+    # Pick the modulus min value so that it will not drop to the lower word count.
```
jsign, line 674: Worth adding an assertion that `mod_bits % 64 == 63`? Since that's a requirement of `mod_bit` to satisfy your intention explained in the comment.

tests/zkevm/test_worst_compute.py
```diff
@@ -644,3 +647,104 @@ def make_dup(i):
...
+        assert len(indexes) > numerators_min_len  # Disable if you want to find longer MOD chains.
+        if len(indexes) > numerators_min_len:
+            break
+        seed += 1
+        print(f"{seed=}")
+
+    # TODO: Don't use fixed PUSH32. Let Bytecode helpers to select optimal push opcode.
+    code_constant_pool = sum(Op.PUSH32[n] for n in numerators)
+    code_prefix = code_constant_pool + Op.JUMPDEST
+    code_suffix = Op.JUMP(len(bytes(code_constant_pool)))  # TODO: len(Bytecode) doesn't work.
```
jsign, line 723: I tested this TODO since I already used `len(Bytecode)` in other place, so wanted to make sure this wasn't broken elsewhere.

Looks like it should work fine as defined [here](https://github.com/ethereum/execution-spec-tests/blob/main/src/ethereum_test_vm/bytecode.py#L96-L98).

To double check, I added a `print(f"{len(bytes(code_constant_pool))=} {len(code_constant_pool)=}")` and got when filling
```
len(bytes(code_constant_pool))=495 len(code_constant_pool)=495
len(bytes(code_constant_pool))=495 len(code_constant_pool)=495
len(bytes(code_constant_pool))=495 len(code_constant_pool)=495
len(bytes(code_constant_pool))=495 len(code_constant_pool)=495
```
so looks fine?

tests/zkevm/test_worst_compute.py
```diff
@@ -644,3 +647,104 @@ def make_dup(i):
...
+    code_body_len = MAX_CODE_SIZE - len(code_prefix) - len(code_suffix)
+    code_segment = (
+        Op.CALLDATALOAD(0) + sum(make_dup(len(numerators) - i) + Op.MOD for i in indexes) + Op.POP
```
jsign, line 726: Nice, very cool idea. Long chain repeated as many times as possible.

jsign (approved):
Left some comments for your consideration, but overall LGTM.

Cycles:
```
tests/zkevm/test_worst_compute.py::test_worst_mod[fork_Cancun-blockchain_test-mod_bits_63]-1    3918422193
tests/zkevm/test_worst_compute.py::test_worst_mod[fork_Cancun-blockchain_test-mod_bits_255]-1   6321807633
tests/zkevm/test_worst_compute.py::test_worst_mod[fork_Cancun-blockchain_test-mod_bits_127]-1   6559030064
tests/zkevm/test_worst_compute.py::test_worst_mod[fork_Cancun-blockchain_test-mod_bits_191]-1   8734621043
```

tests/zkevm/test_worst_compute.py
chfast: This works also for multiplies of 64 (e.g. 256). I skip these values because they are not works cases. But I can add an assert for this, but I think it is better to leave it as is. In the end, it works correctly if someone wants to try different values for `mod_bits`. I will update the comment though not to indicate this is required.

tests/zkevm/test_worst_compute.py
chfast: True. This is mypy error.

chfast:
Added also `SMOD` variant.

marioevz (approved):
LGTM, thanks!

PR #1159, opened 2025-01-30
Add docs to #1131

marioevz:
## 🗒️ Description
Adds documentation for #1131.

If we decide to change the format of the `chainid` field to `HexNumber` or `ZeroPaddedHexNumber`, these changes should also be updated.

### DIFF
--- a/docs/consuming_tests/state_test.md
+++ b/docs/consuming_tests/state_test.md
@@ -66,7 +66,13 @@ Chain configuration object.
 
 ### `FixtureConfig`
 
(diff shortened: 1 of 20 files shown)


docs/consuming_tests/state_test.md
```diff
@@ -66,7 +66,13 @@ Chain configuration object.
 
 ### `FixtureConfig`
 
-At the moment this object can contain only the `blobSchedule` that is necessary to apply the correct cap to the maximum amount of blobs that a transaction can have. Otherwise, in forks prior to Cancun for example, it will be an empty JSON object.
+#### - `network`: [`Fork`](./common_types.md#fork)
```
winsvega, line 69: It can be array but it is not array in config class

winsvega (approved):

PR #1166, opened 2025-02-02
✨ feat(gentest): Support for transaction types

### LINKED ISSUE #861: ✨(gentest): add support for other transaction types
Currently, `gentest` only supports type 0 transaction types, which is quite limiting in a post-burn world :fire: 

We should add support for other types, prioritising type 2 transactions:
- 0x2 - EIP-1559
- 0x1 - EIP-2930 (access list)
- 0x4 - EIP-7702

raxhvl:
## 🗒️ Description

Support for type 1 and 2 transaction.

I have not created a test case to test type 1 transaction. Let me know if that is okay.

## 🔗 Related Issues
closes #861 

## ✅ Checklist

### DIFF
--- a/src/cli/gentest/request_manager.py
+++ b/src/cli/gentest/request_manager.py
@@ -12,31 +12,15 @@ Classes:
 
 from typing import Dict
 
-from pydantic import BaseModel
-
 from config import EnvConfig
-from ethereum_test_base_types import Hash, HexNumber
+from ethereum_test_base_types import Hash
 from ethereum_test_rpc import BlockNumberType, DebugRPC, EthRPC
-from ethereum_test_types import Transaction
+from ethereum_test_rpc.types import TransactionByHashResponse
+from ethereum_test_types import Environment
 
 
 class RPCRequest:
     """Interface for the RPC interaction with remote node."""
-
-    class RemoteTransaction(Transaction):
-        """Model that represents a transaction."""
-
-        block_number: HexNumber
-        tx_hash: Hash
-
-    class RemoteBlock(BaseModel):
-        """Model that represents a block."""
-
-        coinbase: str
-        difficulty: str
-        gas_limit: str
-        number: str
-        timestamp: str
 
     node_url: str
     headers: dict[str, str]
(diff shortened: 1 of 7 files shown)


raxhvl:
While working on this issue I realized that we have two models to represent a Transaction:

1. Transaction used by tests

https://github.com/ethereum/execution-spec-tests/blob/fef9a263f2a4c9cb7d6568b5dde1e0eea8fe1da1/src/ethereum_test_types/types.py#L601-L624

2. Transaction received from an RPC

https://github.com/ethereum/execution-spec-tests/blob/fef9a263f2a4c9cb7d6568b5dde1e0eea8fe1da1/src/ethereum_test_rpc/types.py#L29-L49


We should combine them.

marioevz:
> We should combine them.

There are slight differences but I agree we could minimize code duplication by at least inheriting `Transaction` from `TransactionByHashResponse`.

src/cli/gentest/request_manager.py
```diff
@@ -55,11 +55,11 @@ def eth_get_transaction_by_hash(self, transaction_hash: Hash) -> RemoteTransacti
         block_number = res.block_number
         assert block_number is not None, "Transaction does not seem to be included in any block"
 
-        assert (
-            res.ty == 0
-        ), f"Transaction has type {res.ty}: Currently only type 0 transactions are supported."
+        assert res.ty < 4, f"Unsupported transaction type :{res.ty}"
```
marioevz, line 58: I think this implies we support type 3 transactions too, and the only missing thing would be the versioned hashes.

marioevz (commented):
Thanks! Some small comments.

raxhvl:
I explored ways to make Gentest more robust. In our codebase, the [`Transaction`](https://github.com/ethereum/execution-spec-tests/blob/83942b0011e32625ec8e74598673b1cce7b3db16/src/ethereum_test_types/types.py#L601) model is responsible for defining new transaction types. The [`TransactionResponseByHash`](https://github.com/ethereum/execution-spec-tests/blob/83942b0011e32625ec8e74598673b1cce7b3db16/src/ethereum_test_rpc/types.py#L29) is the RPC equivalent of a Transaction, with added post-execution metadata like the transaction hash, block number etc.

Previously, Gentest had its own `Transaction` (3rd!) model that performed separate validation. This redundant code essentially dealt with a single entity, leading to unnecessary complexity.

I have refactored the code to make the Transaction model from `ethereum_test_types` the source of truth. This model has comprehensive [validation logic](https://github.com/ethereum/execution-spec-tests/blob/83942b0011e32625ec8e74598673b1cce7b3db16/src/ethereum_test_types/types.py#L715-L777) that handles most use cases. The RPC transaction now extends this model, ensuring that RPC transactions are automatically validated. It is possible we face some short term issues due to this change in our code using RPC data. I have tested and have not encountered any problems yet.

The primary takeaway of this exercise is that Gentest will now _automatically_ support both current and future transactions as long as the root Transaction model is kept up to date.

```mermaid
classDiagram
    Transaction <|-- TransactionByHashResponse : extends
    TransactionByHashResponse <|-- Gentest : consumes
    class Transaction{
      +..tx fields..
      +validate()
    }

    class TransactionByHashResponse{
        + transaction_hash
        + block_number
        ....
    }
```

marioevz:
So I just tried out `uv run execute` with a type 4 test and it's currently failing when parsing this:
```json
{
  "type": "0x4",
  "chainId": "0x1",
  "nonce": "0x0",
  "gas": "0x7a120",
  "maxFeePerGas": "0x3b9aca00",
  "maxPriorityFeePerGas": "0x3b9aca00",
  "to": "0xd69dcd96dce63dddc8921cc4efc8b8027223084e",
  "value": "0x0",
  "accessList": [],
  "authorizationList": [
    {
      "chainId": "0x0",
      "address": "0x2da5fa1a13279ff30b734d02370058ce60e3c478",
      "nonce": "0x0",
      "yParity": "0x0",
      "r": "0x18fc889a2b9cf91b2c57bfa5359f1d1150aefbaf474daa418a3e720dff1ff52b",
      "s": "0xa570fceb4c95d81d8d308075111274588402b0931b680d58ca295c466a9d5cb"
    }
  ],
  "input": "0x",
  "r": "0x1292b87437eb92c88584e91dfa2dd7f1e0ee46823b0afec04d1b2e4e727122d9",
  "s": "0x772385b2de94a80c5e0ddd0d991c2999532e7f75ef1047ffcd66e04ee5f39fd1",
  "yParity": "0x0",
  "v": "0x0",
  "hash": "0x2dd9536121dead91dc6e738518aef5fe4f328d3e909a02a5886586ef5c166899",
  "blockHash": "0xfac2310773979fc4760c1564f01a78dfca6c499e7d74a6f553effca00ee768c4",
  "blockNumber": "0x4",
  "transactionIndex": "0x0",
  "from": "0xe5d78afb66042679f59c579d2212bbf5ea0f19ae",
  "gasPrice": "0x3b9aca00"
}
```
I'm not sure why would it give me `gasPrice` alongside `maxFeePerGas` and `maxPriorityFeePerGas`, but I still need to check other clients to see if this is common behavior.

We can add a simple preprocessor to the pydantic model to remove this redundant field if it's present.

marioevz:
I've tried all clients onhttps://github.com/raxhvl/execution-spec-tests/pull/2 and only Besu is currently not passing because of how they are formatting the authorization lists, so I think we should be ok with merging that PR and then this one.

marioevz (approved):
LGTM, thanks!

raxhvl:
Hey @marioevz  I tested this further. There is some confusion still in the model field names for sender and recipient addresses of a transaction:

## The RPC model uses `from_address` and `to_address`

https://github.com/ethereum/execution-spec-tests/blob/08a3b3b0aa02a121e3730f9f5a3528c9c36046cd/src/ethereum_test_rpc/types.py#L37-L39

## The Transaction model has `sender` and `to`
https://github.com/ethereum/execution-spec-tests/blob/08a3b3b0aa02a121e3730f9f5a3528c9c36046cd/src/ethereum_test_types/types.py#L625

https://github.com/ethereum/execution-spec-tests/blob/08a3b3b0aa02a121e3730f9f5a3528c9c36046cd/src/ethereum_test_types/types.py#L679

## Assertion failure because of this confusion
This means that when tx is constructed from RPC the `sender` is not set. Which causes this assertion failure:

https://github.com/ethereum/execution-spec-tests/blob/08a3b3b0aa02a121e3730f9f5a3528c9c36046cd/src/cli/gentest/request_manager.py#L58

I could rewrite this assertion to suit what the RPC model has, but I think we should fix this confusion.

cc: @danceratopz

marioevz:
@raxhvl Maybe just not calling the `Transaction.model_post_init` in the RPC class could be a solution, because the point the transaction one is to validate tester writing a good tx, whereas on RPC we know that most likely these attributes are correct, wdyt?

raxhvl:
@marioevz suppressing the error might not work here. 

Full context: The code is trying to fetch an RPC transaction and to create a test. The test requires a `Transaction` object. When I cast `TransactionByHashResponse` to `Transaction` in the test, it will not have a `sender` or `to` field since those are `from_address` and `to_address` address in the other model.

The JSON response from RPC does infact has the correct field names, I wonder why `TransactionByHashResponse` sticks out?

Issue #1200, opened 2025-02-10
running `uv run fill --collect-only` fails when the fixtures folder does not already exist


felix314159:
Someone who just cloned the repo and follow some guide might run:
* `uv venv --python 3.12`
* `uv sync --all-extras`
* `uv run fill --collect-only`

However, the last command fails with `pluggyTeardownRaisedWarning: FileNotFoundError: The directory fixtures does not exist.` if the folder fixtures does not already exist in the root of the repo. Since it by default doesn't exist, a fix would be to create the folder, if it doesn't exist, when fill() from fill.py is run.

LouisTsai-Csie:
Hi @felix314159 ,

I also encounter the problem today, but I solve the issue by manually creating the `fixtures/` folder. I then run the command documented in the README

`uv run fill -k "push0 and delegatecall" tests/shanghai/ -m state_test -v`

And then I got the following error:

```bash
__________________ TestPush0CallContext.test_push0_contract_during_call_contexts[fork_Shanghai-state_test-delegatecall] ___________________
../../../.pyenv/versions/3.10.9/lib/python3.10/encodings/idna.py:163: in encode
    raise UnicodeError("label empty or too long")
E   UnicodeError: label empty or too long

The above exception was the direct cause of the following exception:
tests/shanghai/eip3855_push0/test_push0.py:145: in test_push0_contract_during_call_contexts
    state_test(env=env, pre=pre, post=post, tx=tx)
src/pytest_plugins/filler/filler.py:675: in __init__
    fixture = self.generate(
src/ethereum_test_specs/state.py:199: in generate
    return self.make_state_test_fixture(t8n, fork, eips, slow=is_slow_test(request))
src/ethereum_test_specs/state.py:132: in make_state_test_fixture
    transition_tool_output = t8n.evaluate(
src/ethereum_clis/transition_tool.py:531: in evaluate
    return self._evaluate_server(
src/ethereum_clis/transition_tool.py:348: in _evaluate_server
    response = self._server_post(
src/ethereum_clis/transition_tool.py:290: in _server_post
    response = Session().post(
.venv/lib/python3.10/site-packages/requests/sessions.py:637: in post
    return self.request("POST", url, data=data, json=json, **kwargs)
.venv/lib/python3.10/site-packages/requests/sessions.py:579: in request
    settings = self.merge_environment_settings(
.venv/lib/python3.10/site-packages/requests/sessions.py:760: in merge_environment_settings
[...]

PR #1201, opened 2025-02-10
create fixtures folder when it does not exist when fill is run (#1200)

felix314159:
## 🗒️ Description
create fixtures folder when it does not exist when fill is run

## ✅ Checklist

### DIFF
--- a/src/cli/pytest_commands/fill.py
+++ b/src/cli/pytest_commands/fill.py
@@ -1,6 +1,8 @@
 """CLI entry point for the `fill` pytest-based command."""
 
 import sys
+from os import getcwd, makedirs
+from os.path import join
 from typing import List
 
 import click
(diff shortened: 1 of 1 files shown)


marioevz (changes requested):
I think this is happening because during `pytest_sessionfinish` we are not exiting early if we are only collecting tests, we only exit if the output is to the `stdout`:
https://github.com/ethereum/execution-spec-tests/blob/8397062390a2c27466e35cd32d7216b4144755e5/src/pytest_plugins/filler/filler.py#L771-L776

We could simply add another condition for early exit like we do here: 
https://github.com/ethereum/execution-spec-tests/blob/8397062390a2c27466e35cd32d7216b4144755e5/src/pytest_plugins/filler/filler.py#L212-L213

It's a bit cleaner than creating the directory since the directory isn't really used for collect only.

danceratopz:
Closed in favor of 
- #1254

PR #1202, opened 2025-02-11
feat(docs): added fast build mechanic with usage examples

felix314159:
## 🗒️ Description
When using `tests` as arg the build time can be more than 40 sec, when using `tests/frontier` it is around 7 sec on my machine. While working on the docs it is nice to have the options to have access to a faster build mode that doesn't spend most of its time rebuilding docs you are not modifying anyways. It is advised to create a bashrc alias for the usage example commands. Before creating a PR for any docs changes it is advised to run `uvx --with=tox-uv tox -e docs` once to ensure that there are no issues when building using the `--strict` flag.

## ✅ Checklist

### DIFF
--- a/docs/gen_test_case_reference.py
+++ b/docs/gen_test_case_reference.py
@@ -30,7 +30,7 @@ logger = logging.getLogger("mkdocs")
 # USAGE 1 (use fast mode):
 #       export FAST_DOCS=true && uv run mkdocs serve
 # USAGE 2 (use fast mode + hide side-effect warnings):
-#       export FAST_DOCS=true && uv run mkdocs serve 2>&1 | sed '/is not found among documentation files/d'
+#       export FAST_DOCS=true && uv run mkdocs serve 2>&1 | sed '/is not found among documentation files/d' # noqa: E501
 test_arg = "tests"
 fast_mode = getenv("FAST_DOCS")
 if fast_mode is not None:
--- a/docs/gen_test_case_reference.py
+++ b/docs/gen_test_case_reference.py
@@ -51,9 +51,6 @@ args = [
     "(not blockchain_test_engine) and (not eip_version_check)",
     "-s",
     test_arg,
-    # "tests/shanghai",
-    # "tests/osaka/eip7692_eof_v1",  # noqa: SC100
-    # "tests/prague/eip2537_bls_12_381_precompiles",  # noqa: SC100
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -42,7 +42,6 @@ commands = python -c "import src.cli.tox_helpers; src.cli.tox_helpers.markdownli
 description = Build documentation in strict mode (mkdocs)
 extras = docs
 setenv =
-    SPEC_TESTS_AUTO_GENERATE_FILES = true
```
danceratopz, line 45: This variable was the old version of `FAST_DOCS` that got lost :smile: Seems good to clean it out in the scope of this PR.

It's actually not necessary to set `FAST_DOCS` to `""` here as tox shields the `mkdocs` command from the user's environment and only passes the list of env vars defined here:
https://tox.wiki/en/latest/config.html#pass_env

danceratopz (approved):
Thanks! Long term I'd prefer to add an env var or add user config profiles to specify the following inputs to `fill`:
- `--from`
- `--until`
- tests path: i.e. `tests/` or `tests/prague`
The reason for the latter is to have a faster way of generating the docs for a specific fork, because that's what you might want to view with `mkdocs serve`.

docs/gen_test_case_reference.py
```diff
@@ -24,6 +25,19 @@
 
 logger = logging.getLogger("mkdocs")
 
+
+# if docs are generated while FAST_DOCS is true, then use "tests/frontier" otherwise use "tests"
+# USAGE 1 (use fast mode):
+#       export FAST_DOCS=true && uv run mkdocs serve
+# USAGE 2 (use fast mode + hide side-effect warnings):
+#       export FAST_DOCS=true && uv run mkdocs serve 2>&1 | sed '/is not found among documentation files/d' # noqa: E501
```
danceratopz, line 33: @felix314159 I appended `# noqa: E501` to this comment to disable the line-too=long check; there's always an exception to the rule! In this case, it makes no sense to split the shell command over two lines and make it difficult for the user to copy-paste.

tox.ini
felix314159: Yep, I also tried what happens running `uvx --with=tox-uv tox -e docs` after you had manually set `export FAST_DOCS=true`. Seems to work fine / as intended.

PR #1204, opened 2025-02-11
refactor(plugins/execute): Refactor account refunds logic

marioevz:
## 🗒️ Description

Makes the following changes to help the refund logic complete in case of a failure during fixture setup:

### Move Refund Logic to `pre` fixture

Previously the sender refund logic was placed in the `base_test_parametrizer` function, which is responsible for setting up the `state_test`, `blockchain_test`, etc, fixtures that are used in tests to actually generate the test.

The problem was that this fixture is not guaranteed to be the first fixture prepared during the pytest routine, so it could be that we had a fixture order preparation of:

`pre` -> `sender` -> `blockchain_test`

If an error occurred during the `sender` fixture setup, which potentially used `pre` fixture to fund a sender account, the refund logic would not be reached because the setup failed before reaching `blockchain_test`

The fix in this case is to put all the refund logic right in the `pre` fixture, and that way we can make sure that there is no potential fixture setup between the `pre` and the `blockchain_test` that could make the refund logic not take place.

### Use A Replacement Transaction For Refunds

In the case of refunds, we now use a replacement transaction because it was difficult to estimate the real remaining balance of the sender account if there was a pending transaction in the mempool, so now we simply reuse the nonce to replace any transaction still present in the mempool.

### Handle `eth_getTransactionByHash` Returning `null`

Handles the case where the transaction could not be found by the client and a `null` (`None`) response is returned.

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/execute/sender.py
+++ b/src/pytest_plugins/execute/sender.py
@@ -30,6 +30,15 @@ def pytest_addoption(parser):
     )
 
     sender_group.addoption(
+        "--sender-funding-txs-gas-price",
+        action="store",
+        dest="sender_funding_transactions_gas_price",
+        type=Wei,
+        default=None,
+        help=("Gas price set for the funding transactions of each worker's sender key."),
+    )
+
+    sender_group.addoption(
         "--sender-fund-refund-gas-limit",
         action="store",
         dest="sender_fund_refund_gas_limit",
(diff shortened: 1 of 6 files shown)


src/pytest_plugins/execute/sender.py
```diff
@@ -104,7 +89,7 @@ def sender_key_initial_balance(
             assert seed_sender_balance_per_worker > 100, "Seed sender balance too low"
             # Subtract the cost of the transaction that is going to be sent to the seed sender
             sender_key_initial_balance = seed_sender_balance_per_worker - (
-                sender_fund_refund_gas_limit * sender_funding_transactions_gas_price
+                sender_fund_refund_gas_limit * default_gas_price
```
huyngopt1994, line 92: sender_funding_transactions_gas_price I think that we should keep previous logic for supporting passing the price funding, the only change only need to pump refunding gas price from funding gas price ?

marioevz: Returned the parameter, but it now defaults to the same value as `--default-gas-price`, and we still bump the gas price for the refunds.

winsvega (commented):
looks good, does it work?

marioevz:
> looks good, does it work?

Might need a rebase, but I can retry locally and see that is working properly before merging.

marioevz:
@winsvega Retested locally and seems to work, I also fixed the commits.

winsvega (approved):

PR #1209, opened 2025-02-12
chore(tooling,deps): update mypy to latest release

### LINKED ISSUE #1207: Update mypy to newest version (1.15)
As shown [here](https://mypy-lang.blogspot.com/2025/02/mypy-115-released.html) it should be fine for us to update to the newest version, they drop Python 3.8 and we already are using a newer Python version

danceratopz:
## 🗒️ Description
Updates `mypy` from `0.991` to `>=1.15.0,<1.16`.

An update is long overdue, but @felix314159 ran into a fatal error with `mypy==0.991`, which was fixed by upgrading. 

Note: According to the [pydantic docs](https://docs.pydantic.dev/latest/integrations/mypy/#supported-mypy-versions):
> Pydantic supports the mypy versions released less than 6 months ago. Older versions may still work with the plugin but won't be tested.

I didn't encounter any issues with the pydantic mypy pluging, although we don't take advantage of any of the stricter checks offered by the plugin, cf #1218.

**Speed-up**
Upgrading gives a nice speed-up compared to 0.991, here's the timings on my machine to run:
```
rm -rf .mypy_cache/; uv run mypy src tests docs .github/scripts
```
- `main`: 36s
- `tooling/update-mypy`: 16s
(be sure to run `uv sync --all-extras` between switching branches).

## 🔗 Related Issues
Fixes #1207.

## ✅ Checklist


spencer-tb (approved):
LGTM :)

marioevz (approved):
LGTM, thanks!

PR #783, opened 2024-09-03
feat(specs): EOF: `eof_test` generates state tests, and `execute` in live devnets

marioevz:
## 🗒️ Description

~~REQUIRES #1220~~ Merged

### - `eof_test` produces state tests (and blockchain tests) 

Create a state test out of an `eof_test` by:

1) in the case of a run-time container, wrapping the container under test within an init container and sending a contract-creating transaction
2) in the case of an init container, the container is wrapped in a factory container (which would execute the initcode to create a contract) and the factory container is wrapped within an init container, which is finally sent as a contract-creating transaction

### - State test generated from EOF tests have extra markers

All state tests generated automatically from the `eof_test` spec now have a different test ID, along with an extra marker applied.

The test ID will show `state_test_from_eof_test` instead of simply `state_test`, e.g.:
`test_swapn_on_max_stack[fork_Osaka-state_test_from_eof_test-swapn_operand_0]`
intead of:
`test_swapn_on_max_stack[fork_Osaka-state_test-swapn_operand_0]`

Also, all tests will have the extra markers `state_test_from_eof_test` or `blockchain_test_from_eof_test` or `blockchain_test_engine_from_eof_test` in order to be able to fill only these specific tests.

E.g. using `-m state_test` will select all tests, including `state_test_from_eof_test` tests:

```
collected 28 items / 22 deselected / 6 selected

<Package eip663_dupn_swapn_exchange>
  <Module test_swapn.py>
    <Function test_swapn_all_valid_immediates[fork_Osaka-state_test]>
    <Function test_swapn_on_max_stack[fork_Osaka-state_test_from_eof_test-swapn_operand_0]>
    <Function test_swapn_on_max_stack[fork_Osaka-state_test_from_eof_test-swapn_operand_255]>
    <Function test_swapn_stack_underflow[fork_Osaka-state_test_from_eof_test-stack_height_0]>
    <Function test_swapn_stack_underflow[fork_Osaka-state_test_from_eof_test-stack_height_1]>
    <Function test_swapn_stack_underflow[fork_Osaka-state_test_from_eof_test-stack_height_21]>
    <Function test_swapn_stack_underflow[fork_Osaka-state_test_from_eof_test-stack_height_255]>
```

But using `-m state_test_from_eof_test` will only select state tests that have been automatically generated from an EOF test:

```
collected 28 items / 23 deselected / 5 selected

<Package eip663_dupn_swapn_exchange>
  <Module test_swapn.py>
[...]

PR #1220, opened 2025-02-14
feat(plugins,fixtures,specs): Labeled fixture and execute formats

marioevz:
### Labeled Fixture Formats in Spec Types

Adds labeled fixture and execute formats that can be defined by the spec tests for the fixtures they produce.

For example, `state_test` generated Blockchain and Blockchain engine fixtures now have the following labels:

- `LabeledFixtureFormat(BlockchainFixture, "blockchain_test_from_state_test")`
- `LabeledFixtureFormat(BlockchainEngineFixture, "blockchain_test_engine_from_state_test")`

This translates in the test IDs now showing "blockchain_test_from_state_test" and "blockchain_test_engine_from_state_test" instead of "blockchain_test" and "blockchain_test_engine" respectively.

These test cases now also have an extra marker that is the label.

- `blockchain_test_from_state_test` and `blockchain_test` markers for Blockchain test.
- `blockchain_test_engine_from_state_test` and `blockchain_test_engine` markers for Blockchain test.

This enables a bit more granularity when selecting the cases to fill (or execute in the case of execute), as we don't lose the `-m blockchain_test`, instead we now also can use `-m blockchain_test_from_state_test` to only fill blockchain tests that were generated from a State test.

I.e. `-m blockchain_test` still selects Blockchain tests generated from state tests, but we now also have `-m blockchain_test_from_state_test` which selects only Blockchain tests that were generated from a state test.

### Remove `StateTestOnly`and `BlockchainTestEngine`

Removes `StateTestOnly` and `BlockchainTestEngine` (and `EOFTestOnly` after #783) classes in favor of using markers to signal that a test should only be filled for a given fixture format:

- `pytest.mark.state_test_only`
- `pytest.mark.blockchain_test_only`
- `pytest.mark.blockchain_test_engine_only`
- `pytest.mark.eof_test_only` in #783

This makes it easier to add the marker to a single parameter during parametrization, instead of having to write a new test function to single out tests that have this requirement.

### Notes

Required for #783.

## ✅ Checklist


src/ethereum_test_specs/blockchain.py
```diff
@@ -287,14 +293,33 @@ class BlockchainTest(BaseTest):
...
+        fixture_format: FixtureFormat,
+        fork: Fork,
+        markers: List[pytest.Mark],
+    ) -> bool:
+        """Discard a fixture format from filling if the appropriate marker is used."""
+        if "blockchain_test_only" in [m.name for m in markers]:
+            return fixture_format != BlockchainFixture
+        if "blockchain_test_engine_only" in [m.name for m in markers]:
+            return fixture_format != BlockchainEngineFixture
+        return False
```
spencer-tb, line 321: Just wondering why we don't do the same for `discard_execute...` here too? I assume because we only care about the transaction object within the test or execute only applies to state test transaction objects?

src/pytest_plugins/execute/execute.py
```diff
@@ -340,21 +340,56 @@ def pytest_generate_tests(metafunc: pytest.Metafunc):
...
+                            format_with_or_without_id,
+                            id=format_with_or_without_id.execute_format_name.lower(),
+                            marks=[
+                                getattr(
+                                    pytest.mark,
+                                    format_with_or_without_id.execute_format_name.lower(),
+                                )
+                            ],
+                        )
+                    )
```
spencer-tb, line 374: We could move this `if`/`else` component of the loop into a more generic helper function that both `filler.py` and `execute.py` import, checking for `hasattr(format, "label")` in the `if` component.

No strong feeling on this, happy as is, just a small thought

src/ethereum_test_specs/base.py
```diff
@@ -51,8 +51,30 @@ class BaseTest(BaseModel):
...
+
+    @classmethod
+    def discard_execute_format_by_marks(
+        cls,
+        execute_format: ExecuteFormat,
+        fork: Fork,
+        markers: List[pytest.Mark],
+    ) -> bool:
+        """Discard an execute format from executing if the appropriate marker is used."""
+        return False
```
spencer-tb, line 77: It could be nice to make these abstract to force us to explicitly state this per test type. For example I was unsure why we don't state the `discard_execute...` function for the blockchain test type. No strong feeling on this though

src/pytest_plugins/execute/execute.py
```diff
@@ -340,21 +340,56 @@ def pytest_generate_tests(metafunc: pytest.Metafunc):
...
 
 
+def get_spec_execute_format_for_item(
+    params: Dict[str, Any],
+) -> Tuple[Type[BaseTest], ExecuteFormat]:
+    """Return the spec type and execute format for the given test item."""
+    for spec_type in SPEC_TYPES:
+        if spec_type.pytest_parameter_name() in params:
+            return spec_type, params[spec_type.pytest_parameter_name()]
+    raise ValueError("No spec type found in the test item.")
```
spencer-tb, line 390: Similar here where we could create a generic helper that returns something like:
```python
SpecFormat = Union[FixtureFormat, ExecuteFormat]
...
-> Tuple[Type[BaseTest], SpecFormat]:
```

spencer-tb (approved):
Nice PR. Easy to see what's going on. Only a few thoughts but nothing major!

Approved and LGTM (as I would be happy to merge as is)!

src/ethereum_test_specs/blockchain.py
marioevz: My thinking was that, since we already have the `pytest.mark.execute` marker which can be used to skip a test in execute mode with a given reason, we didn't need more granularity since there's only one execute format.

I think this can be left as is and we can revisit it once we have more than one execute format.

src/pytest_plugins/execute/execute.py
marioevz: Done, thanks!

src/pytest_plugins/execute/execute.py
marioevz: Done 👍

src/ethereum_test_specs/base.py
marioevz: It was done to set a default and not having to repeat a lot of code, I think it can be left as is for now and we can revisit it later if it poses an issue.

PR #1221, opened 2025-02-14
feat,test(consume): use regex for `--sim.limit`; don't use xdist if `--sim.parallelism=1`

danceratopz:
## 🗒️ Description
This PR improves how the `consume` hive simulators handle the `--sim.parallelism` (^= `HIVE_PARALLELISM`) and `--sim.limit` (`HIVE_TEST_PATTERN`) args.

### `HIVE_PARALLELISM`

Don't activate `pytest-xdist` if `--sim.parallelism` is 1 (the default). This sets `-n=1` in the pytest simulator which superflously activates the pytest-xdist plugin. This has no benefits and obfuscates pytest output; xdist is now only used if `HIVE_PARALLELISM > 1`.

### `HIVE_TEST_PATTERN`

The previous behavior was to pass `HIVE_TEST_PATTERN` to pytest's `-k` argument, which is very pytest specific. This PR adds a dependency on `pytest-regex` and allows users to provide a regex to `--sim.limit` which is more consistent with other simulators.

To ensure transparent behavior between running between the two modes: 

1. Directly `./hive --sim eest/consume-engine,eest/consume-rlp`, and,
2. Sending tests from a `consume` session to `./hive --dev`

a `--sim.limit` cli option has been added to all the consume commands (including `consume direct`). This flag has does have some special non-breaking extra behavior:
 
1. If the value is exactly `collectonly`, enable collection mode without
    filtering.
2. If the value starts with `collectonly:`, enable collection mode and use the
    rest as a regex pattern.
3. If the (now possibly transformed) value starts with `id:`, treat the rest as a
    literal test ID and escape special regex characters for the user.

### Test Cases

Adds the first `pytester` based framework tests for `consume` for the new `--sim.limit` arg in `--collect-only` mode (no dependences required). These tests serve as a start and could be improved. They currently do an end-to-end test from `fill` to `consume` (in collect-only mode). Instead of running `fill`, this could be simplified to load an `index.json` file instead, but as-is, serves as a starting point to generate fixtures to actually be executed by `consume` in framework tests (if an `evm` bin is available).

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -17,6 +17,7 @@ Test fixtures for use by clients are available for each release on the [Github r
 - ✨ Allow `consume direct --collect-only` without specifying a fixture consumer binary on the command-line ([#1237](https://github.com/ethereum/execution-spec-tests/pull/1237)).
 - ✨ Report the (resolved) fixture tarball URL and local fixture cache directory when `consume`'s `--input` flag is a release spec or URL  [#1239](https://github.com/ethereum/execution-spec-tests/pull/1239).
 - ✨ EOF Container validation tests (`eof_test`) now generate container deployment state tests, by wrapping the EOF container in an init-container and sending a deploy transaction ([#783](https://github.com/ethereum/execution-spec-tests/pull/783), [#1233](https://github.com/ethereum/execution-spec-tests/pull/1233)).
+- ✨ Use regexes for Hive's `--sim.limit` argument and don't use xdist if `--sim.parallelism==1` in the `eest/consume-rlp` and `eest/consume-rlp` simulators ([#1220](https://github.com/ethereum/execution-spec-tests/pull/1220)).
 
 ### 📋 Misc
 
@@ -294,7 +295,6 @@ Test fixtures for use by clients are available for each release on the [Github r
 - 🔀 `ethereum_test_tools` library has been split into multiple libraries ([#645](https://github.com/ethereum/execution-spec-tests/pull/645)).
 - ✨ Add the consume engine simulator and refactor the consume simulator suite ([#691](https://github.com/ethereum/execution-spec-tests/pull/691)).
 - 🐞 Prevents forcing consume to use stdin as an input when running from hive ([#701](https://github.com/ethereum/execution-spec-tests/pull/701)).
-- ✨ Use regexes for Hive's `--sim.limit` argument and don't use xdist if `--sim.parallelism==1` in the `eest/consume-rlp` and `eest/consume-rlp` simulators ([#1220](https://github.com/ethereum/execution-spec-tests/pull/1220)).
 
 ### 📋 Misc
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -292,8 +292,9 @@ Test fixtures for use by clients are available for each release on the [Github r
 - 🔀 Refactor `gentest` to use `ethereum_test_tools.rpc.rpc` by adding to `get_transaction_by_hash`, `debug_trace_call` to `EthRPC` ([#568](https://github.com/ethereum/execution-spec-tests/pull/568)).
 - ✨ Write a properties file to the output directory and enable direct generation of a fixture tarball from `fill` via `--output=fixtures.tgz`([#627](https://github.com/ethereum/execution-spec-tests/pull/627)).
 - 🔀 `ethereum_test_tools` library has been split into multiple libraries ([#645](https://github.com/ethereum/execution-spec-tests/pull/645)).
-- ✨ Add the consume engine simulator and refactor the consume simulator suite. ([#691](https://github.com/ethereum/execution-spec-tests/pull/691)).
-- 🐞 Prevents forcing consume to use stdin as an input when running from hive. ([#701](https://github.com/ethereum/execution-spec-tests/pull/701)).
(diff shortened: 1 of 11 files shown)


docs/CHANGELOG.md
```diff
@@ -177,8 +177,9 @@ Release tarball changes:
...
 - ✨ Write a properties file to the output directory and enable direct generation of a fixture tarball from `fill` via `--output=fixtures.tgz`([#627](https://github.com/ethereum/execution-spec-tests/pull/627)).
 - 🔀 `ethereum_test_tools` library has been split into multiple libraries ([#645](https://github.com/ethereum/execution-spec-tests/pull/645)).
-- ✨ Add the consume engine simulator and refactor the consume simulator suite. ([#691](https://github.com/ethereum/execution-spec-tests/pull/691)).
-- 🐞 Prevents forcing consume to use stdin as an input when running from hive. ([#701](https://github.com/ethereum/execution-spec-tests/pull/701)).
+- ✨ Add the consume engine simulator and refactor the consume simulator suite ([#691](https://github.com/ethereum/execution-spec-tests/pull/691)).
+- 🐞 Prevents forcing consume to use stdin as an input when running from hive ([#701](https://github.com/ethereum/execution-spec-tests/pull/701)).
+- ✨ Use regexes for Hive's `--sim.limit` argument and don't use xdist if `--sim.parallelism==1` in the `eest/consume-rlp` and `eest/consume-rlp` simulators ([#1220](https://github.com/ethereum/execution-spec-tests/pull/1220)).
```
marioevz, line 182: Might need to move these to the next release upon rebasing this branch to main.

src/pytest_plugins/consume/consume.py
```diff
@@ -68,6 +70,51 @@ def download_and_extract(url: str, base_directory: Path) -> Path:
...
+def process_sim_limit(pattern: str) -> SimLimitBehavior:
+    """
+    Process the `--sim.limit` argument to determine test filtering behavior.
+
+    The function handles three cases:
+    1. If the value is exactly "collectonly", enable collection mode without
+        filtering.
+    2. If the value starts with "collectonly:", enable collection mode and use the
+        rest as a regex pattern.
+    3. If the (possibly transformed) value starts with "id:", treat the rest as a
```
marioevz, line 90: I think one thing that is not completely clear is whether something like `collectonly:id:some_test` is valid.

src/pytest_plugins/consume/consume.py
```diff
@@ -114,6 +161,21 @@ def pytest_addoption(parser):  # noqa: D103
             "The --html flag can be used to specify a different path."
         ),
     )
+    consume_group.addoption(
+        "--sim.limit",
+        action="store",
+        dest="sim_limit",
+        default=None,
+        help=(
```
marioevz, line 169: `type=SimLimitBehavior` if we define `process_sim_limit` as the init function for the class.

src/pytest_plugins/consume/consume.py
```diff
@@ -68,6 +70,51 @@ def download_and_extract(url: str, base_directory: Path) -> Path:
...
+    Returns:
+        A SimLimitBehavior object with appropriate settings.
+
+    """
+    if pattern == "collectonly":
+        return SimLimitBehavior(collectonly=True)
+
+    sim_limit = SimLimitBehavior(pattern=pattern)
+
+    collect_only_prefix = "collectonly:"
```
marioevz, line 105: This would be a lot easier if there was a way to pass custom environment variables to the simulator from the hive executable.

marioevz (commented):
Looks great overall!

Just a couple of comments.

src/pytest_plugins/consume/consume.py
danceratopz: I'm a bit wary of adding generic per-simulator args/config (even thought that's what we're doing here :laughing:). A  simple, consistent UX is quite nice for people that don't use it regularly. Having said that, there will definitely be other use cases for this.

I do think we should discuss/explore whether we can add other standard flags to hive (like --collect-only, for all simulators).

marioevz:
Tried the changes locally and unfortunately `consume cache` is breaking because it cannot find `option.sim_limit` for some reason during setup. This is the command I used:
```
uv run consume cache --input stable@latest
```

danceratopz:
> Tried the changes locally and unfortunately `consume cache` is breaking because it cannot find `option.sim_limit` for some reason during setup. This is the command I used:
> 
> ```
> uv run consume cache --input stable@latest
> ```

Thanks for testing this, I'll add a framework test to this PR to catch this. We definitely need more testing for consume! 

This happens because `consume cache` returns from `pytest_addoption()` early before the `--sim.limit` option is defined, but `pytest_configure()` still tries to use this value. This will be fixed by #1239 which refactors `consume cache` to exit `pytest_configure()` early after the fixture download instead of running a dummy test session. Let's merge #1239 before this PR and rebase on those changes.

marioevz (approved):
LGTM, just tested the rebased branch and works perfectly. I'll open a PR in the hive repo to update the docker files to support the cache.

danceratopz:
> LGTM, just tested the rebased branch and works perfectly. I'll open a PR in the hive repo to update the docker files to support the cache.

Amazing, thanks so much for the rebase, @marioevz! Was just about to do it myself.

PR #1230, opened 2025-02-18
block header validation in context of blockchain test

winsvega:
## ✅ Checklist


winsvega:
But in any code if we don't provide fork value to FixtureHeader  before pydantic validation, it will not validate the corectness of the fields. 

there is one approach is the we derive fork field when validating FixtureHeader by looking at what fields are there. 
and then on blockchain fixture level we will verify the header fork value. 
but that requires rethinking FixtureHeader validation style.

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -455,6 +478,33 @@ class BlockchainFixture(BlockchainFixtureCommon):
     blocks: List[FixtureBlock | InvalidFixtureBlock]
     seal_engine: Literal["NoProof"] = Field("NoProof")
 
+    @field_validator("blocks", mode="before")
```
marioevz, line 481: I think this could be done more elegantly and automatic by changing the type of the [fork field](https://github.com/ethereum/execution-spec-tests/blob/5c7583f98c6eb7b291a2e38c4bf2e3a93ea9d91c/src/ethereum_test_fixtures/blockchain.py#L442) to `Fork`, and then implementing the logic to parse a `Fork` type from a string that contains the name, like it's being done in #1191:
https://github.com/ethereum/execution-spec-tests/blob/af577cd929805121383d4a7a5a8dbec701a4e16c/src/ethereum_test_forks/base_fork.py#L112-L126

@spencer-tb do you think we could split out this functionality into another PR that does this without the full EIP refactoring?

Ideally, pydantic should be able to pick up the string and convert it automatically into the appropriate `type[BaseFork]`.

I think we would also require to implement `__new__` in [`class BaseType`](https://github.com/ethereum/execution-spec-tests/blob/af577cd929805121383d4a7a5a8dbec701a4e16c/src/ethereum_test_forks/base_fork.py#L208) for it to return the class instead of an instance of the fork.

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -32,7 +40,16 @@
...
+from ethereum_test_forks import (
+    Cancun,
+    Constantinople,
+    Fork,
+    Frontier,
+    Homestead,
+    Istanbul,
+    Paris,
+    Prague,
+)
```
marioevz, line 52: We could use `get_forks` to make the fork resolution by name more dynamic.
```suggestion
from ethereum_test_forks import Fork, Paris, get_forks
```

And then simply define something like:
```python
all_forks_by_name = {fork.name(): fork for fork in get_forks()}
```

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -455,6 +478,33 @@ class BlockchainFixture(BlockchainFixtureCommon):
...
+                    elif network == "Constantinople":
+                        header["fork"] = Constantinople
+                    elif network == "Istanbul":
+                        header["fork"] = Istanbul
+                    elif network == "Paris":
+                        header["fork"] = Paris
+                    elif network == "Cancun":
+                        header["fork"] = Cancun
+                    elif network == "Prague":
+                        header["fork"] = Prague
```
marioevz, line 505:
```suggestion
                    if network in all_forks_by_name:
                        header["fork"] = all_forks_by_name[network]
```
Follow up to the previous comment.

marioevz (commented):
Looks good overall, just a minor suggestion to make this more future-proof. Thanks!

PR #1236, opened 2025-02-19
feat(tests): converting `calldataload` and `calldatasize` tests

pacrob:
## ✅ Checklist
- https://github.com/ethereum/tests/pull/1477
- used `geth version 1.15.2-stable-c8c62daf`


tests/frontier/opcodes/test_calldatasize.py
```diff
@@ -0,0 +1,156 @@
...
+        (
+            b"\x1a\x84Q\xe6\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01",
+            Account(storage={0x00: 0x02}),
+        ),
+        (
+            b"\x1a\x84Q\xe6\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x11\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01",
+            Account(storage={0x00: 0x11}),
+        ),
+        (
+            b"\x1a\x84Q\xe6\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00!\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01",
```
winsvega, line 22: is there better way to declare bytes ? more readable

winsvega:
evmone build script failing

tests/frontier/opcodes/test_calldatasize.py
pacrob: shortened and labeled!

pacrob:
OK, I think it's gtg. Still getting coverage loss though. How can I see what's lost , and is there a solution? I think my previous PR had the same issue.

```
===============
COVERAGE REPORT
/tests/DIFF/index.html
New covered Lines: 180
New covered function Paths: 77

ERROR: coverage is lost!
Lost coverage Lines: 26
Lost coverage function Paths: 5
```

tests/frontier/opcodes/test_calldataload.py
```diff
@@ -0,0 +1,116 @@
...
+    tx_data: bytes,
+    pre: Alloc,
+    address_a_storage: Account,
+):
+    """
+    Test `CALLDATALOAD` opcode.
+
+    Based on https://github.com/ethereum/tests/blob/ae4791077e8fcf716136e70fe8392f1a1f1495fb/src/GeneralStateTestsFiller/VMTests/vmTests/calldatacopyFiller.yml
+    """
+    address_a = pre.deploy_contract(code_for_address_a)
```
marioevz, line 75: It might be unnecessary to get the entire code as a parameter since it's very similar in all cases, so we could change it only having a `calldataload_offset` parameter and change this to:
```suggestion
    address_a = pre.deploy_contract(Op.SSTORE(0, Op.CALLDATALOAD(calldataload_offset)) + Op.STOP)
```

tests/frontier/opcodes/test_calldataload.py
```diff
@@ -0,0 +1,116 @@
...
+from ethereum_test_tools import Account, Alloc, Bytecode, StateTestFiller, Transaction
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+
+@pytest.mark.parametrize(
+    "mstore,args_size,code_for_address_a,tx_data,address_a_storage",
+    [
+        (
+            (Op.MSTORE8(offset=0x0, value=0x25) + Op.MSTORE8(offset=0x1, value=0x60)),
+            0x2,
```
marioevz, line 15: It might be better to change these two parameters to a single one called `calldata`, which could contain a `bytes` value as follows:
- two_bytes: b'\x25\x60'
- word_n_byte: b'\xff' * 32 + b'\x23'
- 34_bytes: bytes.fromhex('123456789ABCDEF00000000000000000000000000000000000000000000000000024')

or something similar, then in the code we could use the `Om.MSTORE(calldata, 0x00)` (Macros can be imported by `from ethereum_test_tools import Macros as Om`) to put the call data in memory and `args_size = len(calldata)`, then finally `Op.CALL` to the `address_a`.

Advantages to this method are:
- IMO it makes it easier to understand what the test is trying to verify by showing exactly what the calldata is when calldataload is being called.
- We can parametrize for generate a test where `address_a` is the entry point of the transaction and `calldata` is the data of the transaction.

tests/frontier/opcodes/test_calldataload.py
```diff
@@ -0,0 +1,116 @@
...
+                args_size=0x0,
+                ret_offset=0x0,
+                ret_size=0x0,
+            )
+            + Op.STOP
+        ),
+    )
+
+    tx = Transaction(
+        data=tx_data,
```
marioevz, line 107: I think `tx_data` is not really used anywhere and could be safely removed.

But ideally it should be replaced by `calldata` as described in the previous comment 👍

tests/frontier/opcodes/test_calldataload.py
```diff
@@ -0,0 +1,116 @@
...
+    "mstore,args_size,code_for_address_a,tx_data,address_a_storage",
+    [
+        (
+            (Op.MSTORE8(offset=0x0, value=0x25) + Op.MSTORE8(offset=0x1, value=0x60)),
+            0x2,
+            (Op.PUSH1[0x0] + Op.CALLDATALOAD + Op.PUSH1[0x0] + Op.SSTORE),
+            b"\x00",
+            Account(
+                storage={0x00: 0x2560000000000000000000000000000000000000000000000000000000000000}
+            ),
```
marioevz, line 20: Could be simplified to:
```suggestion
            0x2560000000000000000000000000000000000000000000000000000000000000,
```
And then the function defines the `Account(storage=...` part.

tests/frontier/opcodes/test_calldatasize.py
```diff
@@ -0,0 +1,98 @@
...
+
+from ethereum_test_forks import Byzantium, Fork
+from ethereum_test_tools import Account, Alloc, StateTestFiller, Transaction
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+
+@pytest.mark.parametrize(
+    "args_size,sha3_len,storage",
+    [
+        ("0002", "01", Account(storage={0x00: 0x02})),
```
marioevz, line 13:
```suggestion
        ("0002", "01", 0x02),
```
Similar comment to previous file, `Account(storage=...` can be defined inside the function to make the parametrization more readable.

tests/frontier/opcodes/test_calldataload.py
```diff
@@ -0,0 +1,116 @@
...
+                ret_size=0x0,
+            )
+            + Op.STOP
+        ),
+    )
+
+    tx = Transaction(
+        data=tx_data,
+        gas_limit=100_000,
+        gas_price=0x0A,
```
marioevz, line 109:
```suggestion
```
Can be safely removed, so we can execute this same test in other environments.

tests/frontier/opcodes/test_calldataload.py
```diff
@@ -0,0 +1,116 @@
...
+        protected=fork >= Byzantium,
+        sender=pre.fund_eoa(),
+        to=to,
+        value=0x01,
```
marioevz, line 113:
```suggestion
```
I think this can also be safely removed since it's not used anywhere.

tests/frontier/opcodes/test_calldatasize.py
```diff
@@ -0,0 +1,98 @@
...
+                address=address,
+                value=0x0,
+                args_offset=0x0,
+                args_size=Op.CALLDATALOAD(offset=0x4),
+                ret_offset=0x0,
+                ret_size=0x0,
+            )
+        )
+    )
+    tx_data = bytes.fromhex("1a8451e6" + "00" * 30 + args_size + "00" * 31 + sha3_len)
```
marioevz, line 86: No need to encode this as call data to be decoded by the contract, `args_size` and `sha3_len` could be integers and they could be placed inside of the contract directly:
```python
    to = pre.deploy_contract(
        code=(
            Op.MSTORE(offset=0x0, value=Op.SHA3(offset=0x0, size=sha3_len))
            + Op.CALL(
                gas=Op.SUB(Op.GAS(), 0x100),
                address=address,
                value=0x0,
                args_offset=0x0,
                args_size=args_size,
                ret_offset=0x0,
                ret_size=0x0,
            )
        )
    )
```
Although I'm having trouble understanding why the original test is using `SHA3` in the first place since the memory contents are irrelevant, and my best guess is to just have a non-zero filled memory.

I think we should replace all of this with `calldata` just like the comment in `test_calldataload.py`, and also parametrize this function to test calldatasize at the top-level call (returns `len(tx.data)`) or a sub-level call (returns arg_size).

marioevz (commented):
Hey, some comments which I think should make the test a bit more EEST-like, but we might lose irrelevant coverage (Like the `SHA3` function which is IMO is out-of-scope for these tests) but that's ok.

winsvega:
lets add  push32, mstore8, add opcodes to default opcodes exempt from coverage check requirements.
it is in the file tests/homestead/coverage/test_coverage.py

pacrob:
> lets add `push32, mstore8, add` opcodes to default opcodes exempt from coverage check requirements. it is in the file tests/homestead/coverage/test_coverage.py

Not all fixed, but better. Thanks!

before:
```
ERROR: coverage is lost!
Lost coverage Lines: 34
Lost coverage function Paths: 11
```
after:
```
ERROR: coverage is lost!
Lost coverage Lines: 33
Lost coverage function Paths: 3
```

tests/frontier/opcodes/test_calldatasize.py
pacrob: @marioevz I simplified it a bit, but I'm pretty sure it's still checking correctly. Anything I'm missing?

pacrob:
Latest coverage fail:
```
Lost coverage Lines: 56
Lost coverage function Paths: 9
```

Looking into it:

expected:
 - Lost 22 lines and 6 functions from `lib/evmone` due to no longer using `keccak` or `calldataload` in the `calldatasize` test
 - Lost 7 lines `test/state/host.cpp`, 3 lines and one function in `test/state/host.cpp` due to not including a value in the transaction in the `calldatasize` test.

unknown:
 - Lost 2 lines in `evmc/include/evmc`, showing that the test is no longer checking Prague. Why might that be happening?
 - Leaving 22 lines and 2 functions across `mpt.cpp`, `state.cpp`, and `system_contracts.cpp`. Not sure of source.

I'll keep poking, but if you know whether the last 2 can be fixed or should be ignored, let me know!

pacrob:
Hey, @marioevz, I've addressed your recent feedback. Let me know if you have more!

marioevz (approved):
LGTM, thanks!

marioevz:
Checking the lost coverage and it's due of the following reasons:
- Removal of SHA3 opcode.
- Account deletion code branches.
- Value transferring code branches.
The lost coverage is not relevant to the test, so we can safely proceed to merge.

Issue #1232, opened 2025-02-19
consume: Add support to `consume direct` for EOF test fixtures


marioevz:
Command `consume direct` should support EOF fixtures, since tools like `evmone-eofparse` allow consumption of EOF containers.

The command technically already supports EOF fixtures with the merging of #935, but currently there is no fixture-consumer class defined to process EOF fixtures.

The only thing required is to create a `FixtureConsumerTool` subclass that contains the logic to process the fixtures using the `evmone-eofparse` external tool, like the one defined for Geth's `evm` tool here:

https://github.com/ethereum/execution-spec-tests/blob/2a1e031ca8489e925da760bea572af2524a7f941/src/ethereum_clis/clis/geth.py#L246-L251

During `consume direct` execution, this new class would receive the EOF fixture path in this endpoint:

https://github.com/ethereum/execution-spec-tests/blob/2a1e031ca8489e925da760bea572af2524a7f941/src/ethereum_clis/clis/geth.py#L368-L374

The main difference between Geth's class and evmone's class is that `evmone-eofparse` is a different binary than the one used in the transition tool (`evmnone-t8n`).

This comment describes partially what this entails:

https://github.com/ethereum/execution-spec-tests/pull/935/files#r1907541619

danceratopz:
EOF container verification binaries:

client | branch | binary/command |
--- | --- | --- |
geth | [osaka-mega-eof](https://github.com/shemnon/go-ethereum/tree/osaka-mega-eof) | `build/bin/evm eofparse` |

felix314159:
Can be re-opened when EOF becomes relevant again

PR #1237, opened 2025-02-20
feat(consume): allow `consume direct --collect-only` without `--bin`

danceratopz:
## 🗒️ Description
Changes to `consume direct` behavior:
1. Rename the `--fixture-consume-bin` option to `--bin`.
2. Removes the default value for `--fixture-consume-bin`, previously this was `["evm"]` geth. Now this value must be explicitly specified (some preparation for #1232), unless ran with `--collect-only`.
3. Enables `--collect-only` without `--bin`.
4. Fixes `--bin`'s type: `list` -> `Path`.
 
The main motivation of this PR was 3. in order to enable consume unit testing without requiring a fixture consumer binary (required for #1221).

## 🔗 Related Issues
- #1232
- #935
- #1221 

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/consume/direct/conftest.py
+++ b/src/pytest_plugins/consume/direct/conftest.py
@@ -7,17 +7,36 @@ For example, via go-ethereum's `evm blocktest` or `evm statetest` commands.
 
 import json
 import tempfile
+import warnings
 from pathlib import Path
 from typing import List
 
 import pytest
 
+from ethereum_clis.ethereum_cli import EthereumCLI
 from ethereum_clis.fixture_consumer_tool import FixtureConsumerTool
 from ethereum_test_base_types import to_json
+from ethereum_test_fixtures import BaseFixture
 from ethereum_test_fixtures.consume import TestCaseIndexFile, TestCaseStream
 from ethereum_test_fixtures.file import Fixtures
 
 from ..consume import FixturesSource
+
+
+class CollectOnlyCLI(EthereumCLI):
+    """A dummy CLi for use with `--collect-only`."""
+
+    def __init__(self):  # noqa: D107
+        pass
+
+
+class CollectOnlyFixtureConsumer(
+    FixtureConsumerTool, CollectOnlyCLI, fixture_formats=list(BaseFixture.formats.values())
+):
+    """A dummy fixture consumer for use with `--collect-only`."""
+
+    def consume_fixture(self):  # noqa: D102
+        pass
 
 
 def pytest_addoption(parser):  # noqa: D103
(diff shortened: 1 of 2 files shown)


src/pytest_plugins/consume/direct/conftest.py
```diff
@@ -7,34 +7,52 @@
...
 from ethereum_test_base_types import to_json
+from ethereum_test_fixtures import BaseFixture
 from ethereum_test_fixtures.consume import TestCaseIndexFile, TestCaseStream
 from ethereum_test_fixtures.file import Fixtures
 
 from ..consume import FixturesSource
 
 
+class CollectOnlyCLI(EthereumCLI):
+    """A dummy CLi for use with `--collect-only`."""
```
danceratopz, line 27:
```suggestion
    """A dummy CLI for use with `--collect-only`."""
```

src/pytest_plugins/consume/direct/conftest.py
```diff
@@ -7,34 +7,52 @@
...
     )
 
     consume_group.addoption(
-        "--fixture-consumer-bin",
+        "--bin",
         action="append",
         dest="fixture_consumer_bin",
-        type=list,
-        default=[Path("evm")],
+        type=str,
```
marioevz, line 51:
```suggestion
        type=Path,
```
Works too, I just checked.

marioevz (commented):
My only issue with this approach is the ids that it prints on the collect-only output:
```
CollectOnlyFixtureConsumer-tests/cancun/eip6780_selfdestruct/test_reentrancy_selfdestruct_revert.py::test_reentrancy_selfdestruct_revert[fork_Shanghai-blockchain_test_from_state_test-second_suicide_DELEGATECALL-first_suicide_CALL]
CollectOnlyFixtureConsumer-tests/cancun/eip6780_selfdestruct/test_reentrancy_selfdestruct_revert.py::test_reentrancy_selfdestruct_revert[fork_Shanghai-blockchain_test_from_state_test-second_suicide_DELEGATECALL-first_suicide_DELEGATECALL]
```
If you plug these into `-k`, and then add the `--bin` it will run no tests for you because the suffix of the test id will change to the actual geth fixture consumer name.

I like that `evm` is no longer the default because it forces the user to think about which consumer will run the tests.

src/pytest_plugins/consume/direct/conftest.py
danceratopz: Oh yeah, that's perfect, thanks.

marioevz (approved):
LGTM, thanks!

PR #1246, opened 2025-02-24
feat(consume): allow geth to validate eof test vectors via direct

felix314159:
## 🗒️ Description
Geth supports eofparse and can be used to validate a given eof test (code valid yes/no without running the code). The command I used for fill was `uv run fill -vv -m "not slow" -m eof_test --fork=osaka --evm-bin=$HOME/Documents/evmone/build/bin/evmone-t8n` and the consume command I tested the code with was `uv run consume direct --input=./fixtures/eof_tests --bin=$HOME/Documents/go-ethereum/build/bin/evm -s` (I added `-s` to see the output of some print statements I put in `test_via_direct.py` to see more detailed info about what is going on). It shows all 3895 tests passed in less than a minute, lmk what can be improved. Next I would try to do the same for consume direct via evmone.

## ✅ Checklist


src/ethereum_clis/clis/geth.py
```diff
@@ -290,26 +368,71 @@ def consume_blockchain_test(
...
 
-    @cache  # noqa
-    def consume_state_test_file(
+    # ------------------------------------ eof test -----------------------------------------------
+    def extract_int_from_subprocess_string(
+        self, s: str, substring_of_interest_prefix: str
+    ) -> int | None:
+        """Hacky method to extract relevant substring from string returned by subprocess execution."""  # noqa: E501
+        # get everything after this substring occurrence, then remove everything not in [0,9]
+        relevant_substring = re.sub(r"\D", "", s.split('total executed"=', 1)[1])
```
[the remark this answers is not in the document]
danceratopz: I think this function could be made more explicit by re-writing it to match 2 regex groups, one for the "passed" and one for the "total" count and return them both in a single call.

danceratopz (commented):
Thanks for adding this @felix314159! 

This needs to be tested against some test fixtures that are expected to fail, as 
```
INFO [02-25|10:26:56.473] Executed tests                           passed=3 "total executed"=4
```
still leads to all fixtures in the file passing the test.

Two (obvious) comments:
1. We're forced into hacky methods as the `eofparse` tool doesn't provide any structured output. Perhaps we should add this to geth's [`eofparse`](https://github.com/ethereum/go-ethereum/blob/7d8aca95d28c4e8560a657fd1ff7852ad4eee72c/cmd/evm/eofparse.go#L71) instead of writing brittle code in EEST. I looked at Besu's tool yesterday and they do return structured output. We could adopt this:
```json
{
    "fileName": "/home/dtopz/.cache/ethereum-execution-spec-tests/cached_downloads/v4.0.0/fixtures_eip7692/fixtures/eof_tests/osaka/eip7692_eof_v1/eip3540_eof_v1/eof_example/eof_example_custom_fields.json",
    "group": "tests/osaka/eip7692_eof_v1/eip3540_eof_v1/test_eof_example.py::test_eof_example_custom_fields[fork_Osaka-eof_test]",
    "name": "0",
    "fork": "Osaka",
    "pass": true,
    "expectedError": null,
    "actualError": null
}
```
2. Regardless of whether we improve `eofparse` or not, better testing of `consume direct` is a must. We can use `pytester`-level tests (which can test the behavior of entire pytest test sessions) to catch fails like the one above. I can start adding some in a separate PR.

Let's discuss 1. in our meeting today.

src/ethereum_clis/clis/geth.py
```diff
@@ -246,10 +246,88 @@ def is_fork_supported(self, fork: Fork) -> bool:
 class GethFixtureConsumer(
     GethEvm,
     FixtureConsumerTool,
-    fixture_formats=[StateFixture, BlockchainFixture],
+    fixture_formats=[StateFixture, BlockchainFixture, EOFFixture],
 ):
     """Geth's implementation of the fixture consumer."""
 
+    # ------------------------------------ state test ---------------------------------------------
+    @cache  # noqa: B019
```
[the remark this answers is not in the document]
marioevz: I think the memory leaks should not be a problem for `consume` since we have a single instance per run as you mention, and also `consume` is not a long-running process anyway, so it should be ok for the sake of code simplicity.

src/ethereum_clis/clis/geth.py
marioevz, line 377: I think it should be:
```suggestion
        relevant_substring = re.sub(r"\D", "", s.split(substring_of_interest_prefix, 1)[1])
```

src/ethereum_clis/clis/geth.py
```diff
@@ -290,26 +368,71 @@ def consume_blockchain_test(
                 f"Unexpected exit code:\n{' '.join(command)}\n\n Error:\n{result.stderr}"
             )
 
-    @cache  # noqa
-    def consume_state_test_file(
+    # ------------------------------------ eof test -----------------------------------------------
+    def extract_int_from_subprocess_string(
+        self, s: str, substring_of_interest_prefix: str
+    ) -> int | None:
```
marioevz, line 374:
```suggestion
    @staticmethod
    def extract_int_from_subprocess_string(
        s: str, substring_of_interest_prefix: str
    ) -> int | None:
```
This should be static since it doesn't access any instance properties, then you can call it with `GethFixtureConsumer.extract_int_from_subprocess_string`.

src/ethereum_clis/clis/geth.py
marioevz: This one is failing after the change for me:
For `s` equal to `'INFO [03-17|17:33:22.741] Executed tests passed=4 "total executed"=4'`, `relevant_substring` becomes `44` when `substring_of_interest_prefix=='tests passed='`.

I think this function can be replaced with:
```python
m = re.search(r'tests passed=(\d+) "total executed"=(\d+)', s)
```
and then tests passed is `int(m.group(1))` and tests executed is `int(m.group(2))`.

felix314159:
> I think we can safely close this since EOF is out of scope for Fusaka, wdyt @felix314159 ?

Yes and thanks for going through old PRs, it's a good time to close a lot of them

PR #1257, opened 2025-02-25
fix(test): remove test violating EIP-7523

gumb0:
This case has an account with empty code, balance, nonce=0 and non-empty stor.age in prestate, which is forbidden since The Merge according to [EIP-7523](https://eips.ethereum.org/EIPS/eip-7523).

Background: not having to satisfy this test simplifies the implementation of refund logic in EIP-7702, where we can assume that if account is empty, then it does not exist in the state (no need to additionally check if it has non-empty storage), and then no refund is assigned when processing authorizations.

### DIFF
--- a/tests/prague/eip7702_set_code_tx/test_set_code_txs.py
+++ b/tests/prague/eip7702_set_code_tx/test_set_code_txs.py
@@ -212,11 +212,6 @@ def test_set_code_to_sstore(
 @pytest.mark.parametrize(
     "auth_signer_nonce",
     [
-        pytest.param(
-            0,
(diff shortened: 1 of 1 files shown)


chfast (approved):
I'm very much in favor of this change because checking if an account has any storage is actually a PITA in EVM.

marioevz:
@petertdavies I believe this could be the test you were referring to the other day that contains an empty account? (Not entirely empty but empty with storage)

marioevz (approved):
LGTM

chfast:
:tada:

PR #1244, opened 2025-02-21
feat(tests) Additional blake2 precompile tests

reedsa:
## 🗒️ Description

Convert remaining Blake2 precompile test cases for CALL and DELEGATECALL

## 🔗 Related Issues

#972 (stPreCompiledContracts / stPreCompiledContracts2)

## 🗑️ ethereum/tests removals:
* https://github.com/ethereum/tests/pull/1479

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: A PR with removal of converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been opened.
- [ ] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [ ] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.

PR #1268, opened 2025-02-26
fix(types): ensure `Environment` fields have even byte length

danceratopz:
## 🗒️ Description
These changes ensure that the fields from `Environment` passed to `t8n` have even byte length by setting their type to `ZeroPaddedHexNumber`. This is required by `evmone-t8n`.

This fix is taken verbatim from @winsvega's fix from #1263 for a quick merge to unblock #1244.

## 🔗 Related Issues
This became an issue after:
- #1027

Should fix the issues with the "Evmone Coverage Script" workflow in:
- #1244 

## ✅ Checklist


chfast:
This is problem probably for difficulty because evmone don't treat the difficulty as a number. And almost all tests use value `0x20000`. If it used value `0x2000` it would had been fine.

Anyway, we can correct it in evmone-t8n.

However, if you plan to depend on t8n tools from multiple projects long term you should create a t8n conformance test suite.

danceratopz:
> This is problem probably for difficulty because evmone don't treat the difficulty as a number. And almost all tests use value `0x20000`. If it used value `0x2000` it would had been fine.

Yup, realised that :)

> Anyway, we can correct it in evmone-t8n.

Thanks @chfast, if we can address it here, there's no need to change it here in evmone-t8n.

> However, if you plan to depend on t8n tools from multiple projects long term you should create a t8n conformance test suite.

Yup, we're going to add this.

marioevz (approved):
LGTM, thanks!

PR #1290, opened 2025-03-08
fix(types): implement `__eq__` for Removable

SamWilsn:
## 🗒️ Description

I need this for my unit tests for my structured data library.

### DIFF
--- a/src/ethereum_test_types/types.py
+++ b/src/ethereum_test_types/types.py
@@ -75,7 +75,7 @@ class Removable:
     (`None` normally means "do not modify").
     """
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: Any) -> bool:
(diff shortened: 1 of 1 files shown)


src/ethereum_test_types/types.py
```diff
@@ -75,7 +75,11 @@ class Removable:
     (`None` normally means "do not modify").
     """
 
-    pass
+    def __eq__(self, other) -> bool:
```
danceratopz, line 78:
```suggestion
    def __eq__(self, other: Any) -> bool:
```

danceratopz (approved):
LGTM @SamWilsn, would you mind sharing the library you're referring to? Or is it still in the works? :slightly_smiling_face:

SamWilsn:
Still in the works!

PR #1293, opened 2025-03-10
new(tests): EOF - ensure empty/maxcode deployments work in legacy

pdobacz:
## 🗒️ Description

Otherwise this seems to only be (unintentionally) tested by `eofwrap` tests, which was raised by the Nethermind team after the recent tests release. Not specifically EOF, but I'm adding a dedicated EOF -related test, as the failure might be triggered by paths validating EOF version and/or deciding whether we have an EOF code being handled.

I validated by triggering such broken paths in `evmone`, the new tests and eofwrap tests failed then, existing EOF tests seemed to be passing.

## 🔗 Related Issues

NA

## ✅ Checklist


tests/osaka/eip7692_eof_v1/eip7620_eof_create/test_legacy_eof_creates.py
```diff
@@ -132,3 +141,55 @@ def test_legacy_initcode_eof_contract_fails(
     )
 
     state_test(env=env, pre=pre, post=post, tx=tx)
+
+
+@pytest.mark.parametrize(
+    "legacy_create_opcode",
+    [
+        pytest.param(Op.CREATE, id="CREATE"),
```
chfast, line 149: I don't think `id` is needed, `Op`s are converted to meaningful names. In this case they are going to be exactly `CREATE` and `CREATE2`.

tests/osaka/eip7692_eof_v1/eip7620_eof_create/test_legacy_eof_creates.py
```diff
@@ -132,3 +141,55 @@ def test_legacy_initcode_eof_contract_fails(
...
+@pytest.mark.parametrize(
+    "init_code",
+    [
+        pytest.param(Bytecode(), id="empty_initcode"),
+        pytest.param(LegacyInitcode(initcode_length=MAX_INITCODE_SIZE), id="max_initcode"),
+        pytest.param(LegacyInitcode(deploy_code=Bytecode()), id="empty_code"),
+        pytest.param(LegacyInitcode(deploy_code=Op.STOP * MAX_BYTECODE_SIZE), id="max_code"),
+    ],
+)
+def test_legacy_create_edge_code_size_works(
```
chfast, line 162:
```suggestion
def test_legacy_create_edge_code_size(
```

tests/osaka/eip7692_eof_v1/eip7620_eof_create/test_legacy_eof_creates.py
```diff
@@ -132,3 +141,55 @@ def test_legacy_initcode_eof_contract_fails(
...
+    state_test: StateTestFiller,
+    pre: Alloc,
+    legacy_create_opcode: Opcodes,
+    init_code: Bytecode,
+):
+    """Verifies that legacy initcode/deploycode having 0 or max size works with EOF enabled."""
```
chfast, line 168:
```suggestion
    """Verifies that legacy initcode/deploycode having 0 or max size works in the fork where EOF is enabled."""
```

chfast (approved):

chfast (commented):
After review comment: maybe the test should go to [EIP-3860](https://eips.ethereum.org/EIPS/eip-3860)?

pdobacz:
> After review comment: maybe the test should go to [EIP-3860](https://eips.ethereum.org/EIPS/eip-3860)?

I presume you've meant 3670? But that has no CREATE/CREATE2 whatsoever. 7620 introduces rules which might interfere with CREATE/CREATE2, like figuring out if we're in eof or not on creation

pdobacz:
> Shanghai fork would makes sense here from my side. As the test is testing EIP-3860, it should go in the fork folder that the EIP under test is introduced

Done, PTAL if this is what you've meant @spencer-tb @chfast

chfast (commented):
Yes, as long as the placement is considered.

But I don't understand what EOF has to do with this test. It should work the same before EOF. I.e. you can enable it since Shanghai?

pdobacz:
Well, I don't know anymore tbh. The reason why it is EOF related is that implementing EOF in clients might break this (and has). Also, this is not really 3860-specific, as one of the (previously untested?) cases were zero-sized legacy codes.

So giving this an EOF label provides explanation why this test is specifically important. Having reconsidered this, I'd move this back into eof tests.

And the topic seems to be wider - I feel like it's not the first time when we are struggling to find a good spot for tests that come from bugs found in the open. One way to think about this is to collocate them with the cause, i.e. put under the EOF tests in Osaka. Maybe mark them as regression tests?

pdobacz:
That being said, I marked the test as valid from Shanghai, which makes sense. LMK what you think about placement @spencer-tb :point_up: .

spencer-tb:
Placed in Shanghai marked from Shanghai works for me. Maybe we add to the docstring the EOF reasoning?

pdobacz:
> Placed in Shanghai marked from Shanghai works for me. Maybe we add to the docstring the EOF reasoning?

Done, pushed, will merge soon. Thx!

chfast:
Nice!

PR #1294, opened 2025-03-11
feat(test-convert) : Fillerconvert WIP parse .json test vectors

winsvega:
## 🗒️ Description
WIP prototype of .json/yml test fillers 

I added pydantic models (WIP) of state test fillers. 
and a function that will build pyspec transition test out of it. 

lost of logic to port and reimplement in python. here is just an example of how it will look like. 

General Idea:
-take .json test filler  ->   get list of individual  test vectors (id, env, pre, tx, post)  ->  run state_test()  normally

the complexity raises because original state test filler hash vector of [tx] and [post]  that need to be parsed back to single test again. which I think is in the end better idea. due to a complexity of having many test vectors in one file, we will have 1 vector 1 file. even though env,pre is copied

## ✅ Checklist


src/cli/fillerconvert/structures/common/common.py
```diff
@@ -0,0 +1,63 @@
...
+def parse_hash32(i: str) -> str:
+    """Check if the given string is a valid hash32."""
+    if i.startswith("0x"):
+        i = i[2:]
+    b = bytes.fromhex(i)
+    assert len(b) == 32
+    return "0x" + b.hex()
+
+
+AddressInFiller = Annotated[str, BeforeValidator(parse_address)]
```
marioevz, line 60: Seems to me that `Address` should work here, but is there any reason why it doesn't?

src/cli/fillerconvert/structures/common/common.py
```diff
@@ -0,0 +1,63 @@
...
+ValueInFiller = Annotated[str, BeforeValidator(parse_hex_number)]
```
marioevz, line 61: Similar to previous comment, would `HexNumber` work here?

src/cli/fillerconvert/structures/common/common.py
```diff
@@ -0,0 +1,63 @@
...
+CodeInFiller = Annotated[str, BeforeValidator(parse_code)]
```
marioevz, line 62: I agree that the `BeforeValidator` is definitely needed here, but I think it would be better to have:
```suggestion
CodeInFiller = Annotated[Bytes, BeforeValidator(parse_code)]
```
Just to not have to work with strings.

src/cli/fillerconvert/structures/state_test_filler.py
```diff
@@ -0,0 +1,155 @@
...
+        vector.tx = tr
+
+        post = Alloc()
+        for address, account in expect_result.items():
+            storage = Storage()
+            if account.storage is not None:
+                for key, value in account.storage.items():
+                    storage[key] = value
+
+            # TODO looks like pyspec post state verification will not work for default values?
```
marioevz, line 143: We can use `model_fields_set` of the `account` to see if a value was explicitly set.

For example, if `"balance" in account.model_fields_set`, it means the value was explicitly set to zero instead the value being zero because if fell back to the default.

src/cli/fillerconvert/structures/common/common.py
```diff
@@ -0,0 +1,63 @@
...
+Hash32InFiller = Annotated[str, BeforeValidator(parse_hash32)]
```
marioevz, line 63: It seems like `Hash` should work here without issues.

marioevz (commented):
Looks good overall, just a couple of comments.

src/cli/fillerconvert/structures/common/common.py
winsvega: It won't serialize if its Bytes

src/cli/fillerconvert/structures/common/common.py
winsvega: I thought it was better to separate pydantic for the testFiller scope. Otherwise, schemas get mixed here and there.

winsvega:
Moved to:

https://github.com/ethereum/execution-spec-tests/pull/1362

PR #1295, opened 2025-03-11
chore(fill,spec): write `fixture-format` not `fixture_format` to fixtures

danceratopz:
## 🗒️ Description

Renames the `info["fixture_format"]` field to `info["fixture-format"]`.

Both `fixture-format` and `fixture_format` can be used when loading fixtures from now on -> old fixtures can be used with this new change. But this will break loading (new) fixtures using this format from older versions of EEST.

@winsvega noticed that this was inconsistent with the existing field names in `_info`, so we decided to fix this now, as it's only been recently introduced.

## 🔗 Related Issues
None

## ✅ Checklist
.


spencer-tb (approved):
LGTM!

winsvega:
https://github.com/ethereum/tests/pull/1468

PR #1305, opened 2025-03-12
feat(tests) EIP-2929: Gas cost increase tests

reedsa:
## ✅ Checklist

### DIFF
--- a/tests/berlin/eip2929_gas_cost_increases/test_gas_cost.py
+++ b/tests/berlin/eip2929_gas_cost_increases/test_gas_cost.py
@@ -1,0 +1,138 @@
+"""Tests supported precompiled contracts."""
+
+from typing import Callable, Dict
+
+import pytest
+
+from ethereum_test_tools import (
+    Account,
+    Alloc,
+    Environment,
+    StateTestFiller,
+    Transaction,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "EIPS/eip-2929.md"
+REFERENCE_SPEC_VERSION = "0e11417265a623adb680c527b15d0cb6701b870b"
+
+UPPER_BOUND = 0xFF
+NUM_UNSUPPORTED_PRECOMPILES = 8
+
+
+def contract_call_bytecode(address, call_params):
+    """Return the bytecode for a contract that measures the gas cost of a CALL operation."""
+    return (
+        Op.MSTORE(0, 0)  # Pre-expand the memory so the gas costs are exactly the same
+        + Op.MSTORE(0x100, 0xDEADBEEF)  # Initialize memory for gas measurement
+        + Op.CALL(
+            address=0x101157,
+            value=1,
+            args_offset=0,
+            args_size=0,
+            output_offset=0,
+            output_size=0,
+        )  # Call to a fixed address with wei to warm up
+        + Op.POP
+        + Op.GAS
+        + Op.CALL(
+            address=address,
+            value=call_params["value"],
+            args_offset=call_params["args_offset"],
+            args_size=call_params["args_size"],
+            output_offset=call_params["output_offset"],
+            output_size=call_params["output_size"],
+        )
+        + Op.POP
+        + Op.SUB(Op.SWAP1, Op.GAS)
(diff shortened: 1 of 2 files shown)


tests/berlin/eip2929_gas_cost_increases/test_gas_cost.py
```diff
@@ -0,0 +1,1071 @@
...
+
+def initial_setup_bytecode(sender: Account) -> Bytecode:
+    """Generate bytecode for the initial setup to be tested."""
+    return (
+        Op.MSTORE(0, 0)
+        + Op.MSTORE(0x100, 0xDEADBEEF)
+        + Op.CALL(
+            address=sender,
+            value=1,
+            args_offset=0,
```
winsvega, line 62: i think this are the default values anyway

tests/berlin/eip2929_gas_cost_increases/test_gas_cost.py
```diff
@@ -0,0 +1,1071 @@
...
+            id="call_address_1_params_0",
+        ),
+        pytest.param(
+            {
+                "address": 0x02,
+                "value": 0,
+                "args_offset": 0,
+                "args_size": 0,
+                "output_offset": 0,
+                "output_size": 0,
```
winsvega, line 202: if it is all the time 0, 0, 0, 0, 0, can just omit it to default values

tests/berlin/eip2929_gas_cost_increases/test_gas_cost.py
```diff
@@ -0,0 +1,1071 @@
...
+                "args_size": 0,
+                "output_offset": 0,
+                "output_size": 1,
+            },
+            id="call_address_7_params_value_1_output_size_1",
+        ),
+        pytest.param(
+            {
+                "address": 0x08,
+                "value": 1,
```
winsvega, line 1038: so looks like the parametrization in all this tests really is 

adress, value, output size. 

it can be one test with that many parametrizations 

for each addesss
  for each value
    for each output size  

and so on. if the test body does not change.

reedsa: Had issues with parametrizing with expected values that depend on specific inputs. Ended up changing these to parametrize the inputs and created separate tests for the permutations of the expected output.

I noticed the expected gas cost for many cases are different from the original tests. I'm looking at those closely to figure out the actual values to use, there may be additional changes to the contract based on certain scenarios. I'll check dr to figure out the differences there.

Does the current approach seem reasonable? If so, I plan to add the remaining test cases that use `CALLCODE`, `DELEGATECALL`, etc.

tests/berlin/eip2929_gas_cost_increases/test_gas_cost.py
reedsa: I'll omit the zero value arguments to use defaults 👍

reedsa:
@marioevz @winsvega I'm seeing some differences for the final gas cost in the failing tests. The original values seem to be used to account for the cost to create a trie. It's possible these tests have a different cost due to initial allocations. Should I update the expectations to reflect the actual costs that I'm seeing here?

tests/berlin/eip2929_gas_cost_increases/test_call_gas_cost_change.py
```diff
@@ -0,0 +1,349 @@
...
+        to=account,
+        sender=pre.fund_eoa(),
+        gas_limit=1_000_000,
+        protected=True,
+    )
+
+    expected_gas_cost_change = hex(0x0)
+    if not precompile_exists:
+        expected_gas_cost_change = hex(0x6B6C)
+    elif fork >= Cancun and address in ["0x1", "0x2", "0x3", "0x4", "0x5", "0x6", "0x7", "0x8"]:
```
winsvega, line 162: The fork class iteslf has a list of precompiles defined in it. Use that function

reedsa: The logic in each of these tests is set for only certain precompiles, for example the precompiles at `0x9` and `0xa` for Cancun should expect to have no change in the cost of gas where the others will expect to have a cost difference of `0x61A8`.

PR #1310, opened 2025-03-13
chore(deps): Bump EELS dependency

petertdavies:
## 🗒️ Description

The `py_ecc` dependency has been changed from a git dependency to a standard PyPI dependency. Since EEST indirectly depends on `py_ecc` through EELS it needs its EELS dependency bumped.

## ✅ Checklist
- [x] Add CHANGELOG entry.

### DIFF
--- a/.github/workflows/tox_verify.yaml
+++ b/.github/workflows/tox_verify.yaml
@@ -102,7 +102,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: ethereum/execution-specs
-          ref: 78fb726158c69d8fa164e28f195fabf6ab59b915
+          ref: fa847a0e48309debee8edc510ceddb2fd5db2f2e
           path: execution-specs
           sparse-checkout: |
             src/ethereum
@@ -139,7 +139,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: ethereum/execution-specs
-          ref: 78fb726158c69d8fa164e28f195fabf6ab59b915
+          ref: fa847a0e48309debee8edc510ceddb2fd5db2f2e
           path: execution-specs
           sparse-checkout: |
             src/ethereum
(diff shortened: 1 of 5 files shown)


danceratopz (changes requested):
@petertdavies Do you see any reason to not use ethereum-execution? The plan was to bump to this version?
https://pypi.org/project/ethereum-execution/1.17.0rc6.dev1/

We'll need to update `uv.lock`, too. I can handle it.

danceratopz:
> @petertdavies Do you see any reason to not use ethereum-execution? The plan was to bump to this version: https://pypi.org/project/ethereum-execution/1.17.0rc6.dev1/
> 
> We'll need to update `uv.lock`, too. I can handle it.

This shouldn't be urgent as both `ethereum` and `ethereum-spec-evm-resolver` are pinned.

Afk, but if someone picks this up in the meantime: If we bump `ethereum` / `ethereum-execution` we'll need to bump the resolver to the latest version at the same time, too.

petertdavies:
> This shouldn't be urgent as both `ethereum` and `ethereum-spec-evm-resolver` are pinned.

`ethereum-spec-evm-resolver` is not pinned. The entry in `pyproject.toml` is `ethereum-spec-evm-resolver @ git+https://github.com/petertdavies/ethereum-spec-evm-resolver`.

I've been developing some BLS tests as a separate python project which depends on `execution-spec-tests`. This mostly works, I will make an issue with the minor issues I encountered doing this. When I updated the `py-ecc` dependency in `ethereum-spec-evm-resolver` it broke that project.

danceratopz:
> > This shouldn't be urgent as both `ethereum` and `ethereum-spec-evm-resolver` are pinned.
> 
> `ethereum-spec-evm-resolver` is not pinned. The entry in `pyproject.toml` is `ethereum-spec-evm-resolver @ git+https://github.com/petertdavies/ethereum-spec-evm-resolver`.

Thanks for reporting - it should be. Repinned.

danceratopz (approved):
LGTM. We just need to do fixture diff using `hasher` to `main` for pre-Prague forks before merging.

spencer-tb:
Checking now!

spencer-tb (approved):
LGTM! Thanks!

PR #1304, opened 2025-03-12
feat(consume): add `test_case.fork` as a marker for each test

danceratopz:
## 🗒️ Description
Allow filtering of test cases by fork by marker (e.g., via `-m "Cancun or Prague"`) using the `fork` field in `index.json`: For each test case generated in `pytest_generate_tests()`, apply its `fork` as a marker to the test case.

#### Expected behavior

```
consume direct --markers
/home/dtopz/code/github/danceratopz/eest/src/pytest_plugins/consume/direct/conftest.py:85: UserWarning: No fixture consumer binaries provided; using a dummy consumer for collect-only; all possible fixture formats will be collected. Specify fixture consumer(s) via `--bin` to see actual collection results.
  warnings.warn(
@pytest.mark.ConstantinopleFix: Mark test for ConstantinopleFix fork

@pytest.mark.London: Mark test for London fork

@pytest.mark.Paris: Mark test for Paris fork

@pytest.mark.Byzantium: Mark test for Byzantium fork

@pytest.mark.ShanghaiToCancunAtTime15k: Mark test for ShanghaiToCancunAtTime15k fork
...
```

## 🔗 Related Issues
- https://github.com/ethereum/execution-spec-tests/issues/1260

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [ ] All: Considered squashing commits to improve commit history.
- [ ] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).

PR #1318, opened 2025-03-17
fix(consume): allow consumption of fixtures with unknown forks

danceratopz:
## 🗒️ Description
This is a follow-up to #1304 that addresses two issues:
1. Running `consume` would raise an error if an unknown fork was encountered, i.e., if the index file contained a fork that was not supported by EEST. This is fixed by adding a `strict_mode` flag to `get_relative_fork_markers` and setting it to `False` when called from `consume`. This is required because ethereum/{tests,legacyTests} contains forks that EEST doesn't know about (yet).
2. With 1. added, `consume` would then issue warnings for unregistered marks for the marks that EEST doesn't know about. This can be fixed by registering markers for all the forks in the test cases. As a convenience, these are now added as top-level metadata to the index file.

#### Example Index File

```json
{
  "root_hash": "0x94ae27ca07dcf0714a6b8d7344049988067bde2611c3a17c21c47f5572b08b9c",
  "created_at": "2025-03-17T11:45:00.405238",
  "test_count": 39109,
  "forks": [
    "ParisToShanghaiAtTime15k",
    "HomesteadToEIP150At5",
    "Prague",
    "Paris",
    "BerlinToLondonAt5",
    "HomesteadToDaoAt5",
    "ByzantiumToConstantinopleFixAt5",
    "Istanbul",
    "Cancun",
    "Shanghai",
    "EIP158ToByzantiumAt5",
    "FrontierToHomesteadAt5",
    "Berlin",
    "ArrowGlacierToParisAtDiffC0000",
    "London"
  ],
  "fixture_formats": [
    "blockchain_test"
  ],
  "test_cases": [
    {
      "id": "AmIOnEIP150_Cancun",
      "fixture_hash": "0x291ef4e307e3c3fa38a0d01b0e2b6fd299c0cfb372f5d6ed803158f3f2b1b42d",
      "fork": "Cancun",
      "format": "blockchain_test",
      "json_path": "ValidBlocks/bcForkStressTest/AmIOnEIP150.json"
    },
    {
    ...
```

## 🔗 Related Issues
- #1304 

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/consume/consume.py
+++ b/src/pytest_plugins/consume/consume.py
@@ -16,7 +16,7 @@ import rich
 
 from cli.gen_index import generate_fixtures_index
 from ethereum_test_fixtures import BaseFixture
-from ethereum_test_fixtures.consume import TestCases
+from ethereum_test_fixtures.consume import IndexFile, TestCases
 from ethereum_test_forks import get_forks, get_relative_fork_markers, get_transition_forks
 from ethereum_test_tools.utility.versioning import get_current_commit_hash_or_tag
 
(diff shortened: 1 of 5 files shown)


src/pytest_plugins/consume/consume.py
```diff
@@ -291,16 +291,23 @@ def pytest_configure(config):  # noqa: D103
...
+    index = IndexFile.model_validate_json(index_file.read_text())
+    config.test_cases = index.test_cases
+
     for fixture_format in BaseFixture.formats.values():
         config.addinivalue_line(
             "markers",
             f"{fixture_format.format_name}: Tests in `{fixture_format.format_name}` format ",
         )
+
+    all_forks_with_transitions = {  # type: ignore
```
spencer-tb, line 304: I think the name we've had has been misleading. To me the current name implies only transition forks.
```suggestion
    # All forked defined within EEST
    all_forks = {  # type: ignore
```

src/pytest_plugins/consume/consume.py
```diff
@@ -291,16 +291,23 @@ def pytest_configure(config):  # noqa: D103
...
+        fork for fork in set(get_forks()) | get_transition_forks() if not fork.ignore()
+    }
+    # add markers for each fork in the index file
+    if hasattr(index, "forks"):  # backwards compatibility
+        for fork in index.forks:
+            all_forks_with_transitions.add(fork)
```
spencer-tb, line 310:
```suggestion
    # Append all forks within the index file (compatibility with `ethereum/tests`)
    all_forks.update(getattr(index, "forks", []))
```

spencer-tb (approved):
LGTM! Couple small comments. Nicely laid out PR :D

src/pytest_plugins/consume/consume.py
danceratopz: mastery

src/pytest_plugins/consume/consume.py
danceratopz: Is better

PR #1321, opened 2025-03-17
feat(consume): improve test case ids for ethereum/tests fixtures

danceratopz:
## 🗒️ Description
Improve index generation of `ethereum/tests` fixtures: Create better test case IDs by prepending the relative JSON file path to the ID.

EEST fixture IDs remain the same for `consume`; they stay the same as generated by `fill`.

**Before**
```
chainId_d0g0v0_Cancun
chainIdGasCost_d0g0v0_Cancun
```
**After**
```
GeneralStateTests/stChainId/chainId.json::chainId_d0g0v0_Cancun
GeneralStateTests/stChainId/chainIdGasCost.json::chainIdGasCost_d0g0v0_Cancun
```

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/cli/gen_index.py
+++ b/src/cli/gen_index.py
@@ -86,6 +86,14 @@ def generate_fixtures_index_cli(
     )
 
 
+def get_test_case_id(fixture_name, relative_file_path) -> str:
+    """Create an appropriate test case id based on its source."""
+    # TODO: Update this function for multi-fork/multi-tx-index state test fixtures
+    if ".py" in fixture_name:  # EEST fixture
+        return fixture_name
+    return f"{relative_file_path}::{fixture_name}"  # ethereum/tests fixture
+
+
 def generate_fixtures_index(
     input_path: Path,
     quiet_mode: bool = False,
@@ -155,9 +163,10 @@ def generate_fixtures_index(
 
             relative_file_path = Path(file).absolute().relative_to(Path(input_path).absolute())
             for fixture_name, fixture in fixtures.items():
+                test_case_id = get_test_case_id(fixture_name, relative_file_path)
                 test_cases.append(
                     TestCaseIndexFile(
-                        id=fixture_name,
+                        id=test_case_id,
                         json_path=relative_file_path,
                         # eest uses hash; ethereum/tests uses generatedTestHash
                         fixture_hash=fixture.info.get("hash")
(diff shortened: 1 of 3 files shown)


src/cli/gen_index.py
```diff
@@ -86,6 +86,14 @@ def generate_fixtures_index_cli(
     )
 
 
+def get_test_case_id(fixture_name, relative_file_path) -> str:
+    """Create an appropriate test case id based on its source."""
+    # TODO: Update this function for multi-fork/multi-tx-index state test fixtures
+    if ".py" in fixture_name:  # EEST fixture
+        return fixture_name
+    return f"{relative_file_path}::{fixture_name}"  # ethereum/tests fixture
```
spencer-tb, line 94: Could we append at the start `ethereum/tests/`?

src/cli/gen_index.py
```diff
@@ -86,6 +86,14 @@ def generate_fixtures_index_cli(
     )
 
 
+def get_test_case_id(fixture_name, relative_file_path) -> str:
+    """Create an appropriate test case id based on its source."""
+    # TODO: Update this function for multi-fork/multi-tx-index state test fixtures
+    if ".py" in fixture_name:  # EEST fixture
```
spencer-tb, line 92: Maybe a future issue: adding `repo:` to the `_info:` section of the fixture. Then using this here instead of `.py`.

spencer-tb (commented):
Looks good to me so far! Couple thoughts - not blockers by all means :)

src/cli/gen_index.py
danceratopz: @winsvega the idea of adding the repo field came up here, btw.

danceratopz:
@winsvega revisited multi-fork, mulit-transaction index state test fixtures and discussed how we could best represent these in our `index.json`. @winsvega offered to simply re-fill these fixtures and write every parametrized variant (via fork and transaction) as a separate test fixture within the JSON instead of a single fixture as is now. In this case, ethereum/tests and EEST fixtures will be "compatible" and both have a 1-to-1 mapping ("test fixture entry"-to-"test case") and `consume` will be able to offer per-test-case results without any changes to `genindex` or `consume` logic.

This doesn't fix the problem for `legacyTests`, but probably `consume` can operate in a simpler way for these fixtures (and simply not offer per-test case execution resolution, rather per test fixture).

PR #1322, opened 2025-03-17
Fillci

marioevz:
Changes #1263 to dynamically detect installed t8n tools instead of forcing the unit tests to fail for all users if they don't install Geth and Evm-one.

Even with this change, CI still requires an update to install Evm-one and Geth in `pytest_framework` in https://github.com/ethereum/execution-spec-tests/blob/main/.github/workflows/tox_verify.yaml.

### DIFF
--- a/src/ethereum_clis/clis/geth.py
+++ b/src/ethereum_clis/clis/geth.py
@@ -164,14 +164,12 @@ class GethEvm(EthereumCLI):
 
     def __init__(
         self,
-        binary: Path,
+        binary: Optional[Path] = None,
         trace: bool = False,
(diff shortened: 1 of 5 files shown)


winsvega:
But we have to see it fail if somth get broken for the t8ns

marioevz:
@winsvega added extra unit test to force CI runs to have evmone and geth installed. It currently fails and will continue to fail until we update the CI to pre-install both t8n tools:
https://github.com/ethereum/execution-spec-tests/actions/runs/13910398219/job/38923118046?pr=1322#step:5:338

PR #1319, opened 2025-03-17
fix(ci,docs): fix docs deployment

danceratopz:
## 🗒️ Description
This the error introduced in #1317

## 🔗 Related Issues
- #1317

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.

PR #1323, opened 2025-03-18
chore(ci): fix github pages docs push

spencer-tb:
## 🗒️ Description
Gives the ephemeral github token write access to github pages.  

## 🔗 Related Issues
Fixes the CI failure here: https://github.com/ethereum/execution-spec-tests/actions/runs/13910619884/job/38923833604
(hopefully :P)
- This didn't work #1319
- The original solution that did work #179
 
## ✅ Checklist
- [x] ~~All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~
- [x] ~~All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.~~

### DIFF
--- a/.github/workflows/docs_deploy.yaml
+++ b/.github/workflows/docs_deploy.yaml
@@ -1,5 +1,8 @@
 name: Deploy Docs
 
+permissions:
+  pages: write # Required for deploying to GitHub Pages with `secrets.GITHUB_TOKEN`
+
 on:
   push:
     tags:


danceratopz (approved):
I don't think this will work as `gh-pages` is protected. I don't think there's anyway to push to a protected branch from a github action using github's token-based authentification. Let's try and see. But I don't think there's any additional config we change that would enable it. Let's try. If not we revert back to:
- #179

PR #1327, opened 2025-03-18
docs: Add EIP checklist templates

marioevz:
## 🗒️ Description

This PR aims to introduce EIP checklist templates that must be followed when implementing tests for a given EIP.

## ✅ Checklist


docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,286 @@
...
+### Test Vectors
+
+- [ ] Memory expansion
+    - [ ] Verify that the opcode execution results in the correct memory expansion, being by offset or size or interaction of both parameters (Size of zero should never result in memory expansion, regardless of offset value). Test at least the following memory expansion sizes
+        - [ ] Zero bytes expansion
+        - [ ] Single byte expansion
+        - [ ] 31 bytes expansion
+        - [ ] 32 bytes expansion
+        - [ ] 33 bytes expansion
+        - [ ] 64 or more bytes expansion
```
winsvega, line 26: + boundary test   int_max +/- 1  bytes expansion.   int64_max+/-1 bytes expansion. 
such tests must OOG right away without trying to allocate the memory.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,286 @@
...
+- [ ] Stack over/underflows
+    - [ ] If the opcode pushes one or more items to the stack, and the opcode pushes more elements than it pops, verify that the opcode execution results in exeptional abort when pushing elements to the stack would result in the stack having more than 1024 elements.
+    - [ ] If the opcode pops one or more items to the stack, or it has a minimum stack height of one or more, verify that the opcode execution results in exeptional abort then stack has 1 less item than the minimum stack height expected.
+- [ ] Execution context
+    - [ ] Normal call
```
winsvega, line 31: - in create/create2  constructor.
- in transaction constructor.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,286 @@
...
+    - [ ] If an opcode has data portion that affects its behavior, verify checklist items with multiple interesting values (E.g. if data portion size is 1 byte, use at least 0x00, 0x01, 0x7F, 0xFF).
+- [ ] Contract creation
+    - [ ] Verify contract is created at the expected address given multiple inputs to the opcode parameters.
+    - [ ] Verify that contract is not created in case of
+        - [ ] Out-of-gas when available gas is less than minimum contract creation stipend.
+        - [ ] Contract creation would result in an address collision with an existing contract or eoa-delegated address.
+- [ ] Fork transition
+    - [ ] Verify that the opcode results in exeptional abort if executed before its activation fork.
+    - [ ] Verify that the opcode results in invalid EOF container if attempted to deploy before its activation fork.
+
```
winsvega, line 83: I think we can automate some of this checks with something like test_scenarios. although I see it will become too complicated. So I suggest to make a vector of automatic checks.

lets say we have an opcode(x,y,z)  we give it to a class that auto generates the tests for values of x,y,z  and verifies the exceptions. not a fuzzing but like a verifiaction against static template of test inputs.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,286 @@
...
+### Test Vectors
+
+- [ ] Genesis value
+    - [ ] Verify, if possible, that the value can be set at genesis if the network starting fork is the activation fork, and that clients can consume such genesis.
+- [ ] Value behavior
+    - [ ] Verify, given multiple initial values, that the value is correctly modified for the current and subsequent blocks as expected, depending on the circumstances that affect the value as defined in the EIP.
+- [ ] Fork transition
+    - [ ] Verify initial value of the field at the first block of the activation fork.
+    - [ ] Verify that a block containing the new header field before the activation of the fork is invalid.
+    - [ ] Verify that a block lacking the new header field at the activation of the fork is invalid.
```
winsvega, line 237: if field is a list, verify the validity status of empty list, and max_allowed_size +/-1

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,286 @@
...
+    - [ ] Verify if the transaction has out-of-bounds conditions in its fields and verify:
+        - [ ] Max value for each field
+        - [ ] Max value + 1 for each field
+- [ ] Contract creation
+    - [ ] Verify that the transaction can create new contracts, or that it fails to do so if it's not allowed
+- [ ] Sender account modifications
+    - [ ] Verify that the sender account of the new transaction type transaction has its nonce incremented by one after the transaction is included in a block
+    - [ ] Verify that the sender account of the new transaction type transaction has its balance reduced by the correct amount (gas consumed and value) after the transaction is included in a block
+- [ ] Fork transition
+    - [ ] Verify that a block prior to fork activation where the new transaction type is introduced and containing the new transaction type is invalid.
```
winsvega, line 215: same here if a new transaction type has a field which is a list, verify if it is allowed to have an empty_list, and if there is a limit on number of elements verify that it is rejected if limit is overflown

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,286 @@
...
+        - [ ] Verify exact gas consumption minus one results in out-of-gas error.
+    - [ ] Precompile has dynamic gas usage
+        - [ ] Verify exact gas consumption, given different valid inputs.
+        - [ ] Verify exact gas consumption minus one results in out-of-gas error, given different valid inputs.
+- [ ] Excessive Gas Cases
+    - [ ] Verify spending all block gas in calls to the precompile (100 million gas or more).
+- [ ] Fork transition
+    - [ ] Verify that calling the precompile before its activation fork results in a valid call even for inputs that are expected to be invalid for the precompile.
+    - [ ] Verify that calling the precompile before its activation fork with zero gas results in a valid call.
+
```
winsvega, line 134: aha here then we need to verify that a new system contract can be executed by 7702 delegation. which I think shall be done automatically by our test framework

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: Thanks! This should be covered by lines 39-42.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+        - [ ] 32 bytes expansion
+        - [ ] 33 bytes expansion
+        - [ ] 64 or more bytes expansion
+- [ ] Stack over/underflows
+    - [ ] If the opcode pushes one or more items to the stack, and the opcode pushes more elements than it pops, verify that the opcode execution results in exeptional abort when pushing elements to the stack would result in the stack having more than 1024 elements.
+    - [ ] If the opcode pops one or more items to the stack, or it has a minimum stack height of one or more, verify that the opcode execution results in exeptional abort then stack has 1 less item than the minimum stack height expected.
+- [ ] Execution context
+    - [ ] Normal call
+    - [ ] Static call
+        - [ ] Verify exeptional abort if the opcode attempts to modify the code, storage or balance of an account
```
jochem-brouwer, line 33: In some cases opcodes are also banned if they do not do not modify code (for instance I think that PAY opcode is banned even if the value sent is 0). Another example is SSTORE where the current value is not modified (although this is more obvious that it should not happen in a STATIC context I think)

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+    - [ ] Delegate call
```
jochem-brouwer, line 34: Also verify here that the STATICCALL context stacks. So STATIC -> DELEGATE should still run in static mode. (also applies to CALL/CALLCODE)

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+        - [ ] If the opcode modifies the storage of the account currently executing it, verify that only the account that is delegating execution is the one that has its storage modified.
+        - [ ] If the opcode modifies the balance of the account currently executing it, verify that only the account that is delegating execution is the one that has its balance modified.
+        - [ ] If the opcode modifies the code of the account currently executing it, verify that only the account that is delegating execution is the one that has its code modified.
+    - [ ] Code call
```
jochem-brouwer, line 38: What is a code call?

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+    - [ ] Initcode
+        - [ ] Verify opcode behaves as expected when running during the initcode phase of contract creation
+            - [ ] Initcode of a contract creating transaction.
+            - [ ] Initcode of a contract creating opcode (including itself if opcode creates a contract).
```
jochem-brouwer, line 42: So CREATE and CREATE2.

Also be aware of weird quirks like contract A doing CREATE2, then in this create context calling back into A, which then attempts to deploy exactly the same CREATE2 (so tries to deploy at same address. This is prevented because the nonce of the to-be-created address gets bumped to 1 prior to executing, and prior to trying to create a new address the nonce of the target address should be 0, so the second attempt will immediately fail).

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+    - [ ] Transaction context
+        - [ ] If opcode changes behavior depending on particular transaction properties, test using multiple values for each property.
+    - [ ] Block context
+        - [ ] If opcode changes behavior depending on particular block properties, test using multiple values for each property.
+    - [ ] EOF Container Context
+        - [ ] If opcode changes behavior depending on particular EOF container properties, test using multiple values for each property.
+- [ ] Return data
+    - [ ] Verify proper return data buffer modification if the opcode is meant to interact with it, otherwise verify that the return data buffer is unnaffected
```
jochem-brouwer, line 50: Nit: return data cannot be modified, it can only be overwritten

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+        - [ ] At current call context.
+        - [ ] At parent call context.
+- [ ] Gas usage
```
jochem-brouwer, line 53: Most bugs in implementations arise due to "order-of-operations" gas calculations. For instance when should the 63/64 gas cap be calculated? Before or after gas deduction?

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+    - [ ] If an opcode is terminating, meaning it results in the current call context to end execution, test the following scenarios
+        - [ ] Top-level call termination
+        - [ ] Sub-level call termination
+        - [ ] Initcode termination
+- [ ] Out-of-bounds checks
+    - [ ] Verify if the opcode has out-of-bounds conditions in its parameters and verify:
+        - [ ] Max value for each parameter
+        - [ ] Max value + 1 for each parameter
+- [ ] Exeptional Abort
+    - [ ] Verify behavior that is supposed to cause an exeptional abort, other than stack over or underflow, or out-of-gas errors.
```
jochem-brouwer, line 72: This is often not clear when I read EIPs what an "exceptional abort" means. Usually this means: consume all gas. But in other contexts (REVERT) the call frame is aborted/reverted, but any gas left in the current frame is returned to the caller (the "nice" option :)). Often have to go back to EIP author to figure out what is meant here.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+- [ ] Data portion
+    - [ ] If an opcode has data portion that affects its behavior, verify checklist items with multiple interesting values (E.g. if data portion size is 1 byte, use at least 0x00, 0x01, 0x7F, 0xFF).
+- [ ] Contract creation
+    - [ ] Verify contract is created at the expected address given multiple inputs to the opcode parameters.
+    - [ ] Verify that contract is not created in case of
+        - [ ] Out-of-gas when available gas is less than minimum contract creation stipend.
+        - [ ] Contract creation would result in an address collision with an existing contract or eoa-delegated address.
+- [ ] Fork transition
+    - [ ] Verify that the opcode results in exeptional abort if executed before its activation fork.
```
jochem-brouwer, line 81: Also verify that opcode is availble at fork block, and in some cases also that it yields the correct value (for instance in EIP-1559 the base fee depends on gas used/target gas of previous block, but previous block technically has no "target gas" so a "magic value"/"initial value" is used here)

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+
+- [ ] Call contexts
+    - [ ] Normal call to precompile from contract
+    - [ ] Delegate call to precompile from contract
+    - [ ] Static call to precompile from contract
+    - [ ] Code call to precompile from contract
+    - [ ] Precompile as transaction entry-point
+    - [ ] Call from initcode
+        - [ ] Contract creating transaction
+        - [ ] Contract creating opcode
```
jochem-brouwer, line 101: Also: EIP-7702 delegated call to precompile: should treat the precompile as having no-code, so if delegated to this address should not execute any precompile logic.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+        - [ ] Calldata too long, given different inputs, where the value represents a correct input to the precompile with padded zeros.
+- [ ] Gas usage
+    - [ ] Precompile has constant gas usage
+        - [ ] Verify exact gas consumption
+        - [ ] Verify exact gas consumption minus one results in out-of-gas error.
+    - [ ] Precompile has dynamic gas usage
+        - [ ] Verify exact gas consumption, given different valid inputs.
+        - [ ] Verify exact gas consumption minus one results in out-of-gas error, given different valid inputs.
+- [ ] Excessive Gas Cases
+    - [ ] Verify spending all block gas in calls to the precompile (100 million gas or more).
```
jochem-brouwer, line 130: Sort of a benchmark case too?

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+- [ ] Fork transition
+    - [ ] Verify that the opcode results in exeptional abort if executed before its activation fork.
+    - [ ] Verify that the opcode results in invalid EOF container if attempted to deploy before its activation fork.
+
+### Framework Changes
+
+- [ ] Add opcode to `src/ethereum_test_vm/opcode.py`
+- [ ] Add opcode to relevant methods in the fork where the EIP is introduced in `src/ethereum_test_forks/forks/forks.py`
+
+## New Precompile
```
jochem-brouwer, line 89: General for precompiles: we have this nasty rule that precompiles are added to EIP-2929 "warm addresses" at the fork. So, also checks that the precompile is cold before fork, and warm after fork. (This can be added to some general precompile testing framework)

jochem-brouwer: Especially important to check that before fork the address is cold. Can accidentally mark them as warm.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+        - [ ] Calldata too short, given different inputs, where the value represents a correct but truncated input to the contract.
+        - [ ] Calldata too long, given different inputs, where the value represents a correct input to the contract with padded zeros.
+- [ ] Excessive Gas Cases
+    - [ ] If possible, simulate a scenario where the execution of the contract at the end of the block execution by the system address would result in excessive gas usage (100 million gas or more).
+    - [ ] Verify spending all block gas in calls to system contract (100 million gas or more).
+- [ ] System Contract Deployment
+    - [ ] Verify block execution behavior after fork activation if the system contract has not been deployed.
+    - [ ] Verify deployment transaction results in the system contract being deployed to the expected address.
+- [ ] Contract Variations
+    - [ ] Verify execution of the different variations of the contract for different networks (if any) results in the expected behavior
```
jochem-brouwer, line 179: This is an interesting point, if we would assume that the bytecode on different networks is the same, the only opcode which comes to mind which might diverge chains is the `CHAINID` opcode at this point (but there might be more chain-specific opcodes, besides contextual opcodes like `TIMESTAMP`)

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+- [ ] Contract Substitution
+    - [ ] Substitute the contract to modify its behavior when called by the system address (at the end of the block execution):
+        - [ ] Modified return value lengths
+        - [ ] Modify emitted logs
+- [ ] Fork transition
+    - [ ] Verify calling the system contract before its activation fork results in correct behavior (depends on the system contract implementation).
```
jochem-brouwer, line 185: Also good test, this should test for current EIPs that the system call is not done before the fork (but when the contract exists)

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Framework Changes
+
+- [ ] Add system contract address to relevant methods in the fork where the EIP is introduced in `src/ethereum_test_forks/forks/forks.py`
+- [ ] Add system contract bytecode to the returned value of `pre_allocation_blockchain` in the fork where the EIP is introduced in `src/ethereum_test_forks/forks/forks.py`
```
jochem-brouwer, line 190: Special case (if it is being treated as system contract) would be deposit contract since code/address differs on mainnet/Sepolia.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+## New Transaction Type
+
+### Test Vectors
+
+- [ ] Intrinsic Gas Costs
+    - [ ] Transaction validity: For each new field that affects the intrinsic gas cost of the transaction:
+        - [ ] Verify the transaction (and the block it is included in) is valid by providing the exact intrinsic gas as `gas_limit` value to the transaction with all multiple combinations of values to the field.
+        - [ ] Verify the transaction (and the block it is included in) is invalid by providing the exact intrinsic gas minus one as `gas_limit` value to the transaction with all multiple combinations of values to the field.
+- [ ] Encoding Tests (RLP, SSZ)
+    - [ ] Verify correct transaction rejection due to incorrect field sizes
```
jochem-brouwer, line 201: Maybe to add: for a field, add leading 0s, this should invalidate the field. (In case of Address field this should either be 0/20 bytes (to: field of tx), so adding 0s will also invalidate it but now for a different reason, namely: length is wrong instead of leading zeros (or both, depends upon which order the client checks the errors))

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+    - [ ] Verify correct transaction rejection if the fields particular to the new transaction types are missing
+    - [ ] Verify correct transaction rejection if the transaction type contains extra fields
+    - [ ] If the transaction contains fields with new serializable types, perform all previous tests on the new type/field
+- [ ] RPC Tests
+    - [ ] * Verify `eth_estimateGas` behavior for different valid combinations of the new transaction type
```
jochem-brouwer, line 206: How to test this? This depends upon the parent block it is being called on, and it will also depend upon what params are used in the block we are adding this tx (for instance: timestamp, could have a skipped block so this block is not T+12 but T+24)

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+    - [ ] Verify `eth_sendRawTransaction` using `execute`
+- [ ] Out-of-bounds checks
+    - [ ] Verify if the transaction has out-of-bounds conditions in its fields and verify:
+        - [ ] Max value for each field
+        - [ ] Max value + 1 for each field
+- [ ] Contract creation
+    - [ ] Verify that the transaction can create new contracts, or that it fails to do so if it's not allowed
+- [ ] Sender account modifications
+    - [ ] Verify that the sender account of the new transaction type transaction has its nonce incremented by one after the transaction is included in a block
```
jochem-brouwer, line 215: Nit: does not apply to 7702 if self-delegated since this bumps nonce by more than 1.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+    - [ ] Verify that the sender account of the new transaction type transaction has its balance reduced by the correct amount (gas consumed and value) after the transaction is included in a block
```
jochem-brouwer, line 216: Also verify that the sender has enough balance to send the tx (intrinsic gas + value)

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,289 @@
...
+        - [ ] Modify emitted logs
+- [ ] Fork transition
+    - [ ] Verify calling the system contract before its activation fork results in correct behavior (depends on the system contract implementation).
+
+### Framework Changes
+
+- [ ] Add system contract address to relevant methods in the fork where the EIP is introduced in `src/ethereum_test_forks/forks/forks.py`
+- [ ] Add system contract bytecode to the returned value of `pre_allocation_blockchain` in the fork where the EIP is introduced in `src/ethereum_test_forks/forks/forks.py`
+
+## New Transaction Type
```
jochem-brouwer, line 192: Also tests regarding txs within the same block. For instance, the "warm addresses" and "warm slots" are cleared after a transaction (so does not stack to next transaction). Also things like Transient Storage (EIP 1153) should be cleared after each tx.

jochem-brouwer (commented):
I left some comments, this is after a first pass of this checklist :) Will check again later and when I think of more "exotic" cases, will add them. I might have commented on some cases which are already added, but which I missed (sorry about that).

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: Was trying to convey "call made using CALLCODE opcode", but I think it's not clear enough.
I've changed it to CALLCODE, and also the rest of the entries to use the opcode name.

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: Added:
> Verify recursive contract creation using the opcode: Factory contract uses the opcode, and initcode calls back to factory contract

To the `Contract creation section below.

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: This is a nice observation. I think REVERT is more of a case of being a "terminating opcode", because it doesn't consume all gas, and is an expected abort rather than exceptional.

Nonetheless I added another bullet-point to the "Terminating opcode" section to account for terminating opcodes that rollback the call frame.

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: Absolutely, great point! Added.

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: Added an item to the "Fork transition", thanks!

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: Added to "Call contexts" section, thanks!

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: Good point, I think this is something that needs to be brought up during audits of system contracts bytecode, CHAINID and similar codes should never be part of system contracts IMO.

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: Yes, we cannot ban calls to the system contract, so basically a test that is heavily dependent on the specific contract behavior needs to be written.

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: We actually don't support network specific changes in our framework, which could definitely change in the future!

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,329 @@
...
+    - [ ] Verify that the transaction can create new contracts if the transaction type supports it.
+- [ ] Sender account modifications
+    - [ ] Verify that the sender account of the new transaction type transaction has its nonce incremented by one after the transaction is included in a block
+    - [ ] Verify that the sender account of the new transaction type transaction has its balance reduced by the correct amount (gas consumed and value) after the transaction is included in a block
+- [ ] Fork transition
+    - [ ] Verify that a block prior to fork activation where the new transaction type is introduced and containing the new transaction type is invalid.
+- [ ] RPC Tests
+    - [ ] * Verify `eth_estimateGas` behavior for different valid combinations of the new transaction type
+    - [ ] Verify `eth_sendRawTransaction` using `execute`
+
```
spencer-tb, line 260: What about the following within a single block. The first 2 assume we can hardcode the block gas limit in the environment.
```
- [ ] Tx Block Interactions
     - [ ] Verify that a transaction of the new type is rejected if its gas limit exceeds the block gas limit.
     - [ ] Verify that a transaction of the new type is accepted if the gas limit equals the block gas limit.
     - [ ] Verify that a block with all transactions types including the new type is executed correctly.
```

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,329 @@
...
+    - [ ] Verify, given multiple initial values, that the value is correctly modified for the current and subsequent blocks as expected, depending on the circumstances that affect the value as defined in the EIP.
+- [ ] Fork transition
+    - [ ] Verify initial value of the field at the first block of the activation fork.
+    - [ ] Verify that a block containing the new header field before the activation of the fork is invalid.
+    - [ ] Verify that a block lacking the new header field at the activation of the fork is invalid.
+
+
+### Framework Changes
+
+**TBD**
```
spencer-tb, line 285: Adding changes to execution payload if required.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,329 @@
...
+
+## New Block Body Field
+
+### Test Vectors
+
+**TBD**
+
+### Framework Changes
+
+**TBD**
```
spencer-tb, line 295: Adding changes to new payload if required.

spencer-tb (commented):
Awesome so far! I will come back to this after you have pushed your changes. Some thoughts you _might_ want to include.

docs/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,373 @@
+# EIP Execution Layer Testing Checklist Template
+
+Depending on the changes introduced by an EIP, the following template is the minimum baseline to guarantee test coverage of the Execution Layer features.
+
+## General
+
+- [ ] Code coverage
+    - [ ] Run produced tests against [EELS](https://github.com/ethereum/execution-specs) and verify that line code coverage of new added lines for the EIP is 100%, with only exeptions being unreachable code lines.
```
danceratopz, line 8:
```suggestion
    - [ ] Run produced tests against [EELS](https://github.com/ethereum/execution-specs) and verify that line code coverage of new added lines for the EIP is 100%, with only exeptions being unreachable code lines.
    - [ ] Run coverage on the test code itself (as a basic logic sanity check), i.e., `uv run fill --cov tests`.
```

.markdownlint.yaml
```diff
@@ -1,7 +1,9 @@
...
-MD024: false  # no-duplicate-heading - We use duplicate headings in the changelog.
-MD033: false  # no-inline-html - Too strict.
+default: true # Default state for all rules
+MD013: false # line-length: We don't fill paragaraphs/limit line length
+MD034: false # no-bare-urls - We use pymdownx.magiclink which allows bare urls
+MD046: false # code-block-style - This doesn't play well with material's admonitions)
+MD024: false # no-duplicate-heading - We use duplicate headings in the changelog.
+MD033: false # no-inline-html - Too strict.
+MD007:
+  indent: 4
```
danceratopz, line 9: @marioevz I committed this here, but then cherry-picked it and made a separate PR that has already made it to main (#1460), if you rebase to main, this will change will disappear and all the indentation warnings here will (still) be fixed.

spencer-tb:
I think we should consider merging this and build upon it during fusaka. Added it to the final pectra release issue:
https://github.com/ethereum/execution-spec-tests/issues/1508

@marioevz - let me know what you think, once you mark the PR for as ready for review I'll re-review it all!

marioevz:
> I think we should consider merging this and build upon it during fusaka. Added it to the final pectra release issue: #1508
> 
> @marioevz - let me know what you think, once you mark the PR for as ready for review I'll re-review it all!

This is a great idea, I'm going to push to have this finished today and merged.

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: I think https://github.com/ethereum/execution-apis has the correct setup to make this more deterministic.

In this repository we could try to use `execute` to make these tests deterministic.

docs/checklist_templates/eip_testing_checklist_template.md
marioevz: This is a great suggestion, it reminded me of an important case missing from this checklist, will be adding 👍

docs/writing_tests/checklist_templates/eip_testing_checklist_template.md
```diff
@@ -0,0 +1,1157 @@
...
+| ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------ | ----- |
+| `general/code_coverage/eels`          | Run produced tests against [EELS](https://github.com/ethereum/execution-specs) and verify that line code coverage of new added lines for the EIP is 100%, with only exceptions being unreachable code lines. |        |       |
+| `general/code_coverage/test_coverage` | Run coverage on the test code itself (as a basic logic sanity check), i.e., `uv run fill --cov tests`.                                                                                                       |        |       |
+| `general/code_coverage/second_client` | Optional - Run against a second client and verify sufficient code coverage over new code added for the EIP.                                                                                                  |        |       |
+
+#### Fuzzing
```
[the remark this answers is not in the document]
marioevz: Let me put a small snippet in there, will rebase and then merge.

PR #1330, opened 2025-03-19
fix(consume): fix loading of multi-index, mixed tx type fixtures

danceratopz:
## 🗒️ Description

Test fixtures from `ethereum/{tests,legacyTests}` can be parametrized by a data, gas and value index. In this case, the tx type can be either be an access list type transaction (`0x1`), or not (`0x0`). For the latter, the entry in the transaction `accessList` field is set to `null`. This change handles this case. An empty list can also be specified as entry in this field, but then this is a `0x1` type tx.

## 🔗 Related Issues
None

## ✅ Checklist


marioevz (approved):
Thanks!

I think if we ever do a v2 for the state tests fixture formats, access lists is something that needs to be corrected.

PR #1336, opened 2025-03-20
feat(plugins): Static test filler

marioevz:
## 🗒️ Description

Introduces the refiller plugin which is capable of loading static test specs from static files.

The plugin uses static specs which inherit from base in `src/ethereum_test_specs/base_static.py`.

The inheriting classes must implement the `fill_function` method which returns the function that will be used by refiller to fill the static tests when the static file is loaded through the class.

This PR contains no implementation of classes that inherit from the base static test class, and these should be implemented by follow-up PRs separately.

An (non-working) example can be seen in the following file: https://github.com/ethereum/execution-spec-tests/blob/filler-reader/src/ethereum_test_specs/state_json.py

Refiller mimics the normal python test filler id convention and therefore the files it loads must always contain sub-tests.

I.e. the loaded static test files must be a JSON/YAML with an object at root level, of which every key is a sub-test.

Example of test IDs from a static test file:
```
tests/homestead/contractCreationOOGdontLeaveEmptyContractFiller.json::contractCreationOOGdontLeaveEmptyContract[fork_Berlin-state_test] FILLED
tests/homestead/contractCreationOOGdontLeaveEmptyContractFiller.json::contractCreationOOGdontLeaveEmptyContract[fork_Berlin-blockchain_test] FILLED
tests/homestead/contractCreationOOGdontLeaveEmptyContractFiller.json::contractCreationOOGdontLeaveEmptyContract[fork_Berlin-blockchain_test_engine] FILLED
```

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_specs/base_static.py
+++ b/src/ethereum_test_specs/base_static.py
@@ -1,0 +1,133 @@
+"""Base class to parse test cases written in static formats."""
+
+from abc import abstractmethod
+from typing import Any, Callable, ClassVar, Dict, List, Type, Union
+
+from pydantic import BaseModel, TypeAdapter, ValidatorFunctionWrapHandler, model_validator
+
+
+class BaseStaticTest(BaseModel):
+    """Represents a base class that reads cases from static files."""
+
+    formats: ClassVar[List[Type["BaseStaticTest"]]] = []
+    formats_type_adapter: ClassVar[TypeAdapter]
+
+    format_name: ClassVar[str] = ""
+
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs):
+        """
+        Register all subclasses of BaseStaticTest with a static test format name set
+        as possible static test format.
+        """
+        if cls.format_name:
+            # Register the new fixture format
+            BaseStaticTest.formats.append(cls)
+            if len(BaseStaticTest.formats) > 1:
+                BaseStaticTest.formats_type_adapter = TypeAdapter(
+                    Union[tuple(BaseStaticTest.formats)],
+                )
+            else:
+                BaseStaticTest.formats_type_adapter = TypeAdapter(cls)
+
+    @model_validator(mode="wrap")
+    @classmethod
+    def _parse_into_subclass(
+        cls, v: Any, handler: ValidatorFunctionWrapHandler
+    ) -> "BaseStaticTest":
+        """Parse the static test into the correct subclass."""
+        if cls is BaseStaticTest:
+            return BaseStaticTest.formats_type_adapter.validate_python(v)
+        return handler(v)
+
+    @abstractmethod
+    def fill_function(self) -> Callable:
+        """
+        Return the test function that can be used to fill the test.
+
+        This method should be implemented by the subclasses.
+
+        The function returned can be optionally decorated with the `@pytest.mark.parametrize`
+        decorator to parametrize the test with the number of sub test cases.
+
+        Example:
+        ```
+        @pytest.mark.parametrize("n", [1])
+        @pytest.mark.parametrize("m", [1, 2])
+        @pytest.mark.valid_from("Homestead")
+        def test_state_filler(
+            state_test: StateTestFiller,
+            fork: Fork,
+            pre: Alloc,
+            n: int,
+            m: int,
+        ):
+            \"\"\"Generate a test from a static state filler.\"\"\"
+            assert n == 1
+            assert m in [1, 2]
+            env = Environment(**self.env.model_dump())
+            sender = pre.fund_eoa()
+            tx = Transaction(
+                ty=0x0,
+                nonce=0,
+                to=Address(0x1000),
+                gas_limit=500000,
+                protected=False if fork in [Frontier, Homestead] else True,
+                data="",
+                sender=sender,
+            )
+            state_test(env=env, pre=pre, post={}, tx=tx)
+
+        return test_state_filler
+        ```
+
+        To aid the generation of the test, the function can be defined and then the decorator be
+        applied after defining the function:
+
+        ```
+        def test_state_filler(
+            state_test: StateTestFiller,
+            fork: Fork,
+            pre: Alloc,
+            n: int,
+            m: int,
+        ):
+            ...
+        test_state_filler = pytest.mark.parametrize("n", [1])(test_state_filler)
+        test_state_filler = pytest.mark.parametrize("m", [1, 2])(test_state_filler)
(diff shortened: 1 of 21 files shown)


src/ethereum_test_specs/base_static.py
```diff
@@ -0,0 +1,133 @@
...
+    format_name: ClassVar[str] = ""
+
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs):
+        """
+        Register all subclasses of BaseStaticTest with a static test format name set
+        as possible static test format.
+        """
+        if cls.format_name:
+            # Register the new fixture format
```
marioevz, line 24:
```suggestion
            # Register the new static test format
```

src/ethereum_test_specs/base_static.py
```diff
@@ -0,0 +1,133 @@
...
+                nonce=0,
+                to=Address(0x1000),
+                gas_limit=500000,
+                protected=False if fork in [Frontier, Homestead] else True,
+                data="",
+                sender=sender,
+            )
+            state_test(env=env, pre=pre, post={}, tx=tx)
+
+        return test_state_filler
```
winsvega, line 81: the function return itself?

winsvega (commented):
can you show an example how would it work from my test filler parser returning TestVector class. (the one that has env,pre,post,tx,id) ?

src/ethereum_test_specs/base_static.py
```diff
@@ -0,0 +1,133 @@
...
+        def test_state_filler(
+            state_test: StateTestFiller,
+            fork: Fork,
+            pre: Alloc,
+            n: int,
+            m: int,
+        ):
+            \"\"\"Generate a test from a static state filler.\"\"\"
+            assert n == 1
+            assert m in [1, 2]
```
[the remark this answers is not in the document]
marioevz: Yes that is correct, you can parametrize it on the fly before returning the function:
```python
@pytest.mark.parametrize("m", [1, 2, 3])
@pytest.mark.parametrize("n", [-1])
def test_fn(...
```
And the parametrization can be dynamic and depend on the values that you find in the state test filler for example.

src/ethereum_test_specs/base_static.py
marioevz: Yes, the returned function is dynamically generated based on the properties of the specific static test that was read from the file.

src/pytest_plugins/refiller/refiller.py
```diff
@@ -0,0 +1,295 @@
...
+        """Resolve and apply fixtures before test execution."""
+        self._fixtureinfo = self.session._fixturemanager.getfixtureinfo(
+            self,
+            None,
+            None,
+            funcargs=False,
+        )
+        request = FixtureRequest(self, _ispytest=True)
+        for fixture_name in self.fixturenames:
+            self.params[fixture_name] = request.getfixturevalue(fixture_name)
```
marioevz, line 287: One caveat I found when using the static state test loader is that this line fails if the yml or json file is found in a folder that does not have a `__init__.py` file with the following error:
```
Exception has occurred: AssertionError
Could not obtain a node for scope "Scope.Module" for function <FillerTestItem NoSrcAccount1559[fork_Prague-state_test-v=0-g=0-d=0]>
AssertionError: Could not obtain a node for scope "Scope.Module" for function <FillerTestItem NoSrcAccount1559[fork_Prague-state_test-v=0-g=0-d=0]>

During handling of the above exception, another exception occurred:

  File "/home/marioevz/Development/Eth/execution-spec-tests/src/pytest_plugins/refiller/refiller.py", line 287, in setup
    self.params[fixture_name] = request.getfixturevalue(fixture_name)
                                ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
AssertionError: Could not obtain a node for scope "Scope.Module" for function <FillerTestItem NoSrcAccount1559[fork_Prague-state_test-v=0-g=0-d=0]>
```

Note: `NoSrcAccount1559` is the name of the test.

src/ethereum_test_specs/base_static.py
```diff
@@ -0,0 +1,229 @@
...
+            if isinstance(v, dict):
+                v = BaseStaticTest.remove_comments(v)
+            elif isinstance(v, list):
+                v = [BaseStaticTest.remove_comments(i) if isinstance(i, dict) else i for i in v]
+            result[k] = v
+        return result
+
+    @model_validator(mode="before")
+    @classmethod
+    def remove_comments_from_model(cls, data: Any) -> Any:
```
danceratopz, line 139: Instead of discarding these, they should be added to `_info`'s `description` field. Could be a bit messy, as they could have come from any location in the YAML? We can create a follow-up issue?

src/ethereum_test_specs/base_static.py
```diff
@@ -0,0 +1,229 @@
...
+ALL_FORKS = get_forks()
+
+
+def fork_by_name(fork_name: str) -> Fork:
+    """Get a fork by name."""
+    for fork in ALL_FORKS:
+        if fork.name() == fork_name:
+            return fork
+
+    raise Exception(f'Fork "{fork_name}" could not be identified.')
```
danceratopz, line 155: I'd suggest refactoring and moving to `src/ethereum_test_forks/helpers.py`.

src/ethereum_test_specs/base_static.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+class ForkRangeDescriptor(BaseModel):
```
danceratopz, line 158: I'd actually suggest moving this to `src/ethereum_test_forks/helpers.py` too :slightly_smiling_face: 

Although I'm struggling to find other use cases tbh. We don't need it in the cli as we have `--from` and `--until`.

But at least we'd be able to find it more easily if we did need it again in the future.

danceratopz (approved):
Amazing! Couple of small comments. The `refiller` plugin itself is pretty crazy :)

`refill` is catchy but I'm not sure it correctly conveys its purpose, I think that `static_filler` would be better. A `refiller` sounds like it refills the fixtures, not the fillers?

src/ethereum_test_specs/base_static.py
marioevz: I'll open an issue because I don't have a good idea on how to do it now 👍

src/ethereum_test_specs/base_static.py
marioevz: Also done in latest commit.

PR #1337, opened 2025-03-21
fix(test): improve EIP-7702 test for type 4 transaction with empty to

gumb0:
The transaction in this test is invalid for 2 reasons: empty `to` and also empty `authorization_list`. Some test runners that don't check for exact error can pass it for the wrong reason.


winsvega:
is there a test also with valid to and empty auth list?

winsvega (approved):

winsvega:
We now check the exception string returned by the tooling. But it is rather experimental feature

PR #1280, opened 2025-03-05
new(tests): Add EIP-6110 Sepolia Variant Contract test

marioevz:
## 🗒️ Description
Creates a test that replaces the mainnet contract with the Sepolia variant.

## ✅ Checklist
- [ ] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [ ] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [ ] Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been added to [converted-ethereum-tests.txt](/ethereum/execution-spec-tests/blob/main/converted-ethereum-tests.txt).
- [x] Tests: A PR with removal of converted JSON/YML tests from [ethereum/tests](/ethereum/tests) have been opened.
- [x] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65.
- [x] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://ethereum.github.io/execution-spec-tests/main/tests/) are correctly formatted.

PR #1340, opened 2025-03-21
new(tests): add more test cases for request types

felix314159:
## 🗒️ Description
WIP

## ✅ Checklist


tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
```diff
@@ -343,29 +474,29 @@ def invalid_requests_block_combinations(fork: Fork) -> List[Any]:
...
     # - Empty requests list with invalid hash
-    combinations = [
+    combinations: List[ParameterSet] = [
         pytest.param(
             [],
             [
                 bytes([i]) for i in range(fork.max_request_type() + 1)
             ],  # Using empty requests, calculate the hash using an invalid calculation method:
-            # sha256(sha256(b"\0") ++ sha256(b"\1") ++ sha256(b"\2") ++ ...)
+            # sha256(sha256(b'0x00') ++ sha256(b'0x01') ++ sha256(b'0x02') ++ ...)
```
marioevz, line 497: This is actually slightly different to what the test is actually doing:
- `b'\0'.hex()` is equal to `00`
- `b'0x00'.hex()` is equal to `30783030`, the ascii representation of string `"0x00"`

tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
```diff
@@ -491,6 +622,15 @@ def invalid_requests_block_combinations(fork: Fork) -> List[Any]:
...
     )
+    # too many withdrawals per block (16 is max - EIP-7002)
+    # TODO: how to make sth like below work?
+    # combinations.append(
+    #     pytest.param(
+    #         *[single_withdrawal_from_contract(i) for i in range(0, 17)],
+    #         BlockException.INVALID_REQUESTS,
+    #         id="more_than_16_withdrawals_per_block",
+    #     ),
+    # )
```
marioevz, line 633: Copy-pasting from discord:
The limit is included in the system contract, not anywhere else, I don't think it's even included in the consensus layer clients.

So since it's not enforced by the execution clients other than by smart contract logic, it's theoretically possible to modify the system contract in a testnet to manually overload the limit, and see what happens.

Given the issue we had already with Sepolia, where the smart contract was different in the testnet, I think it would be interesting to inject a modified contract to force 17 withdrawals and at least verify that the ELs behave the same.

Either:
- all EL clients mark the block as invalid
- all EL clients successfully process the block

I think the latter is the one that's going to happen.

See this PR to see how to inject a different contract in the place of the normal system contract: https://github.com/ethereum/execution-spec-tests/pull/1280

tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
```diff
@@ -5,10 +5,11 @@
 """  # noqa: E501
 
 from itertools import permutations
-from typing import Any, Dict, Generator, List, Tuple
+from typing import Dict, Generator, List, Tuple
```
marioevz, line 8: I think we should take the opportunity to rename this file to something more generic like `test_multi_type_requests.py`, so it can be updated in the future if a new request type is added.

tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
```diff
@@ -99,9 +102,15 @@ def single_consolidation_from_contract(i: int) -> ConsolidationRequestContract:
     return ConsolidationRequestContract(requests=[single_consolidation(i)])
 
 
-def get_permutations(n: int = 3) -> Generator[Any, None, None]:
+def get_permutations(
+    n: int = 3,
+) -> Generator[
+    ParameterSet,
```
marioevz, line 108: This is more of an union:
```python
DepositRequest | WithdrawalRequest | ConsolidationRequest
```
`ParameterSet` only applies if the `pytest.param` is called.

tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
```diff
@@ -214,10 +215,140 @@ def get_contract_permutations(n: int = 3) -> Generator[Any, None, None]:
...
+            id="consolidation_from_eoa+withdrawal_from_contract+deposit_from_contract",
+        ),
+        # testing upper limits of each request type per slot if it exists
+        pytest.param(
+            [
+                single_consolidation_from_contract(0),
+                single_consolidation_from_contract(1),
+                # single_consolidation_from_contract(2),  # i=2 is not allowed cuz
+                # only 2 MAX CONSOLIDATIONS PER BLOCK (EIP-7251), but why does the error message
+                # not inform about this? only reports hash mismatch
```
marioevz, line 312: That is because the limit not actually enforced by restricting the number of calls that can be made to the consolidation contract, rather, the consolidations executed after the limit is reached will queue up and only will be dequeued on the next block.

In general, the functions in this file that generate calls to the system contracts cannot know if their requests are going to be included in the block or not.

That's the reason behind the requests hash mismatch: the tx that generates the call is added to the block, it's executed, we add it to the list of expected requests (by our tests) in the header, but it will not match because of the queuing mechanism in the contract.

Hope I explained myself ok, but if not let me know 😄

marioevz (commented):
Some comments on this preliminary review 🙌 

Let me know if any of them needs more info.

tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
felix314159: You are right, I think I wanted to do sth like `bytes([0x00])` but we can just leave it as it was

tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
felix314159: Thanks, I tried that actually but then was shown a mypy type mismatch warning. I have silenced this warning now with `# type: ignore` and the filling still works without issues when using `DepositRequest | WithdrawalRequest | ConsolidationRequest` instead of `ParameterSet`

marioevz: Oh really? It might be then that there's another type that is actually being returned, because mypy is usually correct.

tests/prague/eip7685_general_purpose_el_requests/test_deposits_withdrawals_consolidations.py
marioevz: Could also do `b"\x00"` because it's more common and `b"\0"` is a bit obscure.

tests/prague/eip7685_general_purpose_el_requests/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,97 @@
...
+    pre[Spec_EIP7002.WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS] = Account(
+        code=modified_code,
+        nonce=1,
+        balance=0,
+    )
+
+    sender: EOA = pre.fund_eoa()
+    tx: Transaction = Transaction(
```
marioevz, line 69: See my previous comment, transactions to trigger the withdrawals can be obtained using the `transactions` method mentioned in there.

tests/prague/eip7685_general_purpose_el_requests/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,97 @@
...
+        gas_limit=100_000,
+    )
+
+    blockchain_test(
+        pre=pre,
+        blocks=[
+            Block(
+                txs=[tx],
+                # header_verify=Header(
+                # requests_hash=Requests(
```
marioevz, line 81: Taking the previous comments into account, I think the modifications should be:
- Create test parameter `requests_list`, that you can parametrize with values from this class: https://github.com/ethereum/execution-spec-tests/blob/87131722a1bd16fc6209f0f63d4db838c2210d5d/tests/prague/eip7002_el_triggerable_withdrawals/helpers.py#L15
- The `requests_list` is passed to either [`WithdrawalRequestTransaction`](https://github.com/ethereum/execution-spec-tests/blob/87131722a1bd16fc6209f0f63d4db838c2210d5d/tests/prague/eip7002_el_triggerable_withdrawals/helpers.py#L90) or [`WithdrawalRequestContract`](https://github.com/ethereum/execution-spec-tests/blob/87131722a1bd16fc6209f0f63d4db838c2210d5d/tests/prague/eip7002_el_triggerable_withdrawals/helpers.py#L123), and then from the instance created you can call `update_pre` to automatically prepare the senders of the withdrawals, and then `transactions` to obtain the transactions to put in the block to trigger the withdrawals.
- Finally `requests_list` is used in this line too to obtain the expected requests hash:
```suggestion
                requests_hash=Requests(requests_list),
```

Ideally, given the scenario here, `requests_list` should contain the 18 elements you want to introduce for this test case.

marioevz (commented):
I mainly reviewed [test_modified_withdrawal_contract.py](https://github.com/ethereum/execution-spec-tests/pull/1340/files#diff-47edb59ceb9083d50e4b12729690b8250595fff7126e65b7f4b1fae61e0c034e) here, but some thoughts:
- I would use `--traces` when running `uv run fill` to make sure that the transactions are indeed correctly calling the system contract and are not failing or running out of gas.
- My main concern with the modified contract is that it will be a bit difficult to debug if the system call is not working correctly because currently the t8n does not provide traces for the system call (although that would be a nice feature!).

If you have the modified sys-asm contract to take a look at we could also help out reviewing that :+1:

tests/prague/eip7685_general_purpose_el_requests/test_modified_withdrawal_contract.py
felix314159: In my latest commit I tried to do these things however there are three issues: 
* The first issue is that the parameterized version (which I hoped would do the same as the code actually being used in my commit) does not work. If you uncomment it out and fill with `uv run fill -k "modified_withdrawal_contract" tests/prague/ -m blockchain_test --fork=Prague -vv` you can see what I mean
* The second issue is that with my current code I can not pass `Requests(requests_list)`
* The third issue is that I can not add `--traces` to the fill command because I get odd errors like 
```
with open(src, 'rb') as fsrc:
E   FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpuxnl0ouu/trace-0-0x14cfc1c6268f1e8f70866357f10ad719591f03bdc1b62cfb03330195a70d5b54.jsonl'

```

The code of my latest commit fills without errors (and is consumed without errors) but does not do what I want.

marioevz (commented):
Hey, I addressed your questions in each comment, it works for me locally although I didn't check for 18 withdrawals only two.

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,90 @@
...
+#             amount=0,
+#             fee=Spec_EIP7002.get_fee(0),
+#         ),
+#         WithdrawalRequest(
+#             validator_pubkey=0x01,
+#             amount=0,
+#             fee=Spec_EIP7002.get_fee(0),
+#         ),
+#     ],
+# )
```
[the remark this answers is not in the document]
felix314159: Thanks, I forgot about `pytest.param`

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,261 @@
...
+        ),
+        pytest.param(
+            [
+                *[
+                    single_withdrawal_with_custom_fee(i)
+                    for i in range(
+                        0,
+                        15,
+                    )
+                ],
```
marioevz, line 62: Since we are already using a function to make this easier here, we can put the for loop inside the function:
```suggestion
                *withdrawal_list_with_custom_fee(15),
```

And then:
```python
def withdrawal_list_with_custom_fee(n: int) -> List[WithdrawalRequest]:  # noqa: D103
    return [WithdrawalRequest(
        validator_pubkey=i + 1,
        amount=0,
        fee=Spec_EIP7002.get_fee(0),
    ) for i in range(n)]
```
Or something similar.

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,261 @@
...
+):
+    """Test how clients were to behave when more than 16 withdrawals would be allowed per block."""
+    modified_code: bytes = b"3373fffffffffffffffffffffffffffffffffffffffe1460cb5760115f54807fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff146101f457600182026001905f5b5f82111560685781019083028483029004916001019190604d565b909390049250505036603814608857366101f457346101f4575f5260205ff35b34106101f457600154600101600155600354806003026004013381556001015f35815560010160203590553360601b5f5260385f601437604c5fa0600101600355005b6003546002548082038060121160df575060125b5f5b8181146101835782810160030260040181604c02815460601b8152601401816001015481526020019060020154807fffffffffffffffffffffffffffffffff00000000000000000000000000000000168252906010019060401c908160381c81600701538160301c81600601538160281c81600501538160201c81600401538160181c81600301538160101c81600201538160081c81600101535360010160e1565b910180921461019557906002556101a0565b90505f6002555f6003555b5f54807fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff14156101cd57505f5b6001546002828201116101e25750505f6101e8565b01600290035b5f555f600155604c025ff35b5f5ffd"  # noqa: E501
```
marioevz, line 110: It would be nice to have the source code to this linked somewhere.

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,261 @@
...
+        modified_code += Op.MSTORE(memory_offset, withdrawal_request_chunk_1_3_32bytes)
+        memory_offset += 32
+
+        modified_code += Op.MSTORE(memory_offset, withdrawal_request_chunk_2_3_32bytes)
+        memory_offset += 32
+
+        modified_code += Op.MSTORE(memory_offset, withdrawal_request_chunk_3_3_12bytes)
+        # memory_offset += 32  # MSTORE ALWAYS writes 32 bytes, no need to update memory offset tho
+
+        amount_of_requests += 1
```
marioevz, line 233: We have a macro that would have made your life much easier for this (sorry about that!).

You should be able to use `Om.MSTORE` (imported using `from ethereum_test_tools import Macros as Om`) to simplify this a bit:
```suggestion
        modified_code += Om.MSTORE(bytes(withdrawal_request))
```

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,261 @@
...
+
+    modified_code += Op.RETURN(
+        0, 76 * amount_of_requests
```
marioevz, line 236:
```suggestion
        0, len(requests_list[0]) * len(requests_list)
```

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
marioevz: Another option is to convert everything to bytes in one go:
```python
withdrawal_requests_bytes = sum(bytes(w) for w in requests_list)
modified_code += Om.MSTORE(withdrawal_requests_bytes)
```
and then in the return line you can do `len(withdrawal_requests_bytes)`.

However you prefer is fine.

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,261 @@
...
+        nonce=1,
+        balance=0,
+    )
+
+    # given a list of withdrawal requests construct a withdrawal request transaction
+    withdrawal_request_transaction = WithdrawalRequestTransaction(requests=requests_list)
+    # prepare withdrawal senders
+    withdrawal_request_transaction.update_pre(pre=pre)
+    # get transaction list
+    txs: List[Transaction] = withdrawal_request_transaction.transactions()
```
marioevz, line 250: Interesting thing is that, with this contract, we don't even have to add any transactions to the block, but I think it's ok for the sake of completeness.

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
felix314159: Thanks, that's a lot more readable than my version

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
felix314159: Ah so it can be assumed that each request has the same size, therefore we don't need to know how many bytes each request consists of. Makes sense, ty. But what if the "empty request" test is first? Then we can't access the element with index 0

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
felix314159: Good idea, improves readability

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -204,37 +165,24 @@ def test_extra_withdrawals_pseudo_contract(
...
+        # update memory_offset with the correct value
+        withdrawal_request_bytes_amount: int = len(bytes(withdrawal_request))
+        assert withdrawal_request_bytes_amount == 76, f"Expected withdrawal request to be of size 76 but got size {withdrawal_request_bytes_amount}"  # noqa: E501
+        memory_offset += withdrawal_request_bytes_amount
+
+        # TODO: in opcodes.py the argument order and names of arguments are different from those
+        # on evm.codes website. this is dangerous, especially since with the macro I am not allowed
+        # to provide the parameters as named parameters like
+        # Om.MSTORE(data=bytes(withdrawal_request), offset=memory_offset)
+        # due to 'unexpected keyword argument.. for __call__ of'
```
felix314159, line 179: Is there be a simple change to allow `Macro.__call__()` to support named parameters? LMK what you think

marioevz: Ah you are absolutely right, I think this is a mistake.

The only way I can think of allowing keyword arguments is subclassing the `Macro` class and override `__call__` in a new `class MSTORE`.

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,209 @@
...
+        modified_code += Om.MSTORE(bytes(withdrawal_request), memory_offset)
+        amount_of_requests += 1
+
+    modified_code += Op.RETURN(
+        0, 76 * amount_of_requests
+    )
```
marioevz, line 185:
```suggestion
    modified_code += Op.RETURN(0, Op.MSIZE())
```

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
marioevz: Check my suggested change in the next comment.

marioevz (commented):
Some comments, but I think it's close to be good to go 👍

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,209 @@
...
+    """Test how clients were to behave when more than 16 withdrawals would be allowed per block."""
+    modified_code: Bytecode = Bytecode()
+    memory_offset: int = 0
+    amount_of_requests: int = 0
+
+
+    for withdrawal_request in requests_list:
+        # update memory_offset with the correct value
+        withdrawal_request_bytes_amount: int = len(bytes(withdrawal_request))
+        assert withdrawal_request_bytes_amount == 76, f"Expected withdrawal request to be of size 76 but got size {withdrawal_request_bytes_amount}"  # noqa: E501
```
marioevz, line 172:
```suggestion
        assert withdrawal_request_bytes_amount == 76, (
            "Expected withdrawal request to be of size 76 but got size "
            f"{withdrawal_request_bytes_amount}"
        )
```

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,212 @@
...
+    # WithdrawalRequestContract,
+    WithdrawalRequestTransaction,
+)
+from .spec import Spec as Spec_EIP7002
+from .spec import ref_spec_7002
+
+REFERENCE_SPEC_GIT_PATH: str = ref_spec_7002.git_path
+REFERENCE_SPEC_VERSION: str = ref_spec_7002.version
+
+
```
marioevz, line 35:
```suggestion

pytestmark = pytest.mark.valid_from("Prague")


```
Need to limit these tests to only run on Prague and after.

tests/prague/eip7685_general_purpose_el_requests/test_multi_type_requests.py
```diff
@@ -99,9 +102,15 @@ def single_consolidation_from_contract(i: int) -> ConsolidationRequestContract:
...
 
 
-def get_permutations(n: int = 3) -> Generator[Any, None, None]:
+def get_permutations(
+    n: int = 3,
+) -> Generator[
+    DepositRequest | WithdrawalRequest | ConsolidationRequest,
+    None,
+    None,
+]:
```
marioevz, line 111:
```suggestion
def get_permutations(n: int = 3) -> Generator[ParameterSet, None, None]:
```
I think this was my fault from last review, but this should be the way to go.

tests/prague/eip7685_general_purpose_el_requests/test_multi_type_requests.py
```diff
@@ -116,12 +125,12 @@ def get_permutations(n: int = 3) -> Generator[Any, None, None]:
         ),
     ]
     for perm in permutations(requests, n):
-        yield pytest.param([p[1] for p in perm], id="+".join([p[0] for p in perm]))
+        yield pytest.param([p[1] for p in perm], id="+".join([p[0] for p in perm]))  # type: ignore
```
marioevz, line 128:
```suggestion
        yield pytest.param([p[1] for p in perm], id="+".join([p[0] for p in perm]))
```
Only if the previous comment is applied.

tests/prague/eip7002_el_triggerable_withdrawals/test_modified_withdrawal_contract.py
```diff
@@ -0,0 +1,207 @@
...
+        assert withdrawal_request_bytes_amount == 76, (
+            "Expected withdrawal request to be of size 76 but got size "
+            f"{withdrawal_request_bytes_amount}"
+        )
+        memory_offset += withdrawal_request_bytes_amount
+
+        modified_code += Om.MSTORE(bytes(withdrawal_request), memory_offset)
+        amount_of_requests += 1
+
+    # modified_code += Op.RETURN(0, 76 * amount_of_requests)
```
marioevz, line 182:
```suggestion
```

marioevz (approved):
One last request, please add a item to `docs/CHANGELOG.md` with a one-line description of the tests added by this PR.

Thanks for the changes!

PR #1344, opened 2025-03-24
feat(tests): Identity precompile test conversions

reedsa:
## ✅ Checklist


reedsa:
@marioevz @winsvega There are a couple test cases failing, seeing differences in the expectations from the original tests. I have a couple questions about these cases, as follows:

* The original test for `identity_1_nonzero`, expects the call to succeed and for the stored value to be 0. However this implementation results in the call failing for some reason.
* The original test for `identity_5` does not include the call result in the expected output. In this case, it looks like the call should revert but expects the stored value to be set as `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF`. When running the converted test case I do not see the value being stored. Should the expected result just be `0xDEADBEEF` based on this implementation, or is there a better way to capture the original test? Does this happen because the call reverts and the contract code does not complete?

winsvega:
Have you checked the vmtrace of original tests? 

You can run retesteth test.json --vmtrace

You can select vector --singlenet Cancun -d index

reedsa:
> Have you checked the vmtrace of original tests?
> 
> You can run retesteth test.json --vmtrace
> 
> You can select vector --singlenet Cancun -d index

Yeah the trace for  the CALL in `identity_5` shows `"error" out of gas`, but the specified memory is still allocated with that value. I'm wondering if since the call reverts from the error, if these new tests do not end up allocating that memory.

winsvega:
If a subcall fails with OOG, the upper level call continues, thus stroing the value.

Sorry I have not time to review, but I can give hints.

winsvega:
Ah look

https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stPreCompiledContracts2/CallIdentity_1_nonzeroValueFiller.json

In original test the pre storage is empty (zero)

Thus if the call fails we don't see it as the storage was not set to deadbeef
Thats why it is good to always set the storage to nonzero value if we expect to verify return code of 0.

winsvega:
About identity5, if you see the original test makes little sense, it can be errors.

See if the test logic make sense, try to enhance it with more scenarios using pytest. 

If you see that original test does not do anything useful, see what was the intent, maybe it can be fixed.

winsvega:
Fill tests shows much more mismatch in the expectations. Need to check the logic, something is off.

reedsa:
I was expecting the identity precompile tests to be valid from `Frontier` but I see failures when I try to fill these. Is there something extra needed to cover these?

reedsa:
## Manual Coverage Analysis

### /evmone/evmc/include/evmc - helpers.h ✅
```
     302 LBC           0 :     case EVMC_PRAGUE:
     303               0 :         return "Prague";
```
Lines lost due to coverage script not filling Prague for python tests, and since latest release of ethereum/tests contains prague, this is expected.

### /evmone/test/state - mpt.cpp ✅
```
      95 LBC           0 :     static MPTNode ext(const Path& path, std::unique_ptr<MPTNode> child) noexcept
      96                 :     {
      97               0 :         assert(child->m_kind == Kind::branch);
      98               0 :         MPTNode node{Kind::ext, path};
      99               0 :         node.m_children[0] = std::move(child);
     100               0 :         return node;
```
...
```
     244 LBC           0 :     case Kind::ext:
     245                 :     {
     246               0 :         encoded = rlp::encode(m_path.encode(m_kind)) + encode_child(*m_children[0]);
     247               0 :         break;
```
Lost due how test accounts (in pre-alloc) are structured in EEST vs ethereum/tests, which is irrelevant.

tests/frontier/identity_precompile/common.py
```diff
@@ -0,0 +1,93 @@
...
+        ret_offset=0x0,
+        ret_size=0x20,
+    ):
+        """Create a new instance with the provided values."""
+        self.gas = gas
+        self.value = value
+        self.args_offset = args_offset
+        self.args_size = args_size
+        self.ret_offset = ret_offset
+        self.ret_size = ret_size
```
winsvega, line 38: @marioevz shall we add it to default codebase in Opcodes this struct could be useful

reedsa:
@marioevz if you have a few mintues for a final look, I think this is ready to merge. Thanks!

tests/frontier/identity_precompile/common.py
marioevz, line 38:
```suggestion
class CallArgs(BaseModel):
    """Defines inputs to CALL for the Identity precompile."""

    gas: int = 0x1F4
    address: int = Constants.IDENTITY_PRECOMPILE_ADDRESS
    value: int = 0x0
    args_offset: int = 0x0
    args_size: int = 0x20
    ret_offset: int = 0x0
    ret_size: int = 0x20
```

Pydantic would be helpful here, and also can be used to pass in the parameters more nicely to the opcode (see next comment).

tests/frontier/identity_precompile/common.py
```diff
@@ -0,0 +1,93 @@
...
+            storage.store_next(call_succeeds),
+            call_type(
+                gas=call_args.gas,
+                address=Constants.IDENTITY_PRECOMPILE_ADDRESS,
+                value=call_args.value,
+                args_offset=call_args.args_offset,
+                args_size=call_args.args_size,
+                ret_offset=call_args.ret_offset,
+                ret_size=call_args.ret_size,
+            ),
```
marioevz, line 87:
```suggestion
            call_type(**call_args.model_dump()),
```
If we use pydantic for `CallArgs`.

tests/frontier/identity_precompile/test_identity.py
```diff
@@ -0,0 +1,172 @@
...
+            CallArgs(gas=0x64),
+            (0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,),
+            True,
+            id="identity_4",
+        ),
+        pytest.param(
+            CallArgs(gas=0x11),
+            (0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,),
+            False,
+            id="identity_4_gas17",
```
marioevz, line 64:
```suggestion
            id="identity_4_insufficient_gas",
```

tests/frontier/identity_precompile/test_identity.py
```diff
@@ -0,0 +1,172 @@
...
+        ),
+        pytest.param(
+            CallArgs(gas=0x12),
+            (0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,),
+            True,
+            id="identity_4_gas18",
```
marioevz, line 70:
```suggestion
            id="identity_4_exact_gas",
```

tests/frontier/identity_precompile/test_identity_returndatasize.py
```diff
@@ -0,0 +1,63 @@
...
+from ethereum_test_tools import Opcodes as Op
+from tests.frontier.identity_precompile.common import Constants
+
+
+@pytest.mark.valid_from("Byzantium")
+@pytest.mark.parametrize(
+    ["args_size", "output_size", "expected_returndatasize"],
+    [
+        pytest.param(16, 32, 16, id="output_16"),
+        pytest.param(32, 16, 32, id="output_32"),
```
marioevz, line 22:
```suggestion
        pytest.param(16, 32, 16, id="output_size_greater_than_input"),
        pytest.param(32, 16, 32, id="output_size_less_than_input"),
```

marioevz (approved):
Awesome! Thanks for implementing these :)

I think it only needs a rebase for us to be able to delete the fillers from `./tests/static/state_tests/...`.

marioevz:
Finally got the coverage to run, and the lost coverage lines are expected, proceeding to merge.

PR #1343, opened 2025-03-24
chore(ci): explicity define version range for `pycryptodome`.

spencer-tb:
## 🗒️ Description
Fixes the macOS failures in CI.

### Problem
The issue stems from `solc-select`, due to the `pycryptodome` dependency, take a look at their [`pyproject.toml`](https://github.com/crytic/solc-select/blob/dev/pyproject.toml#L9):
```toml
dependencies = [
  "pycryptodome>=3.4.6",
  "packaging",
]
```

Within our CI we get the following [error](https://github.com/ethereum/execution-spec-tests/actions/runs/14039166908/job/39304679439):
```
× Failed to build `pycryptodome==3.15.0`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit
      status: 1)
      ...
setuptools.errors.InvalidConfigError: Invalid dash-separated key
      'py-limited-api' in 'bdist_wheel' (setup.cfg), please use the underscore
      name 'py_limited_api' instead.

      hint: This usually indicates a problem with the package or the build
      environment.
  help: `pycryptodome` (v3.15.0) was included because
        `ethereum-execution-spec-tests` (v1.0.0) depends on `solc-select`
        (v1.0.4) which depends on `pycryptodome`
```

### Solution

Adding `pycryptodome` to our `.toml` and restricting the upper version to `3.10`.

## 🔗 Related Issues
N/A

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.

PR #1345, opened 2025-03-24
chore(ci): revert pycryptodome change

spencer-tb:
## 🗒️ Description
Reverts this PR: https://github.com/ethereum/execution-spec-tests/pull/1343

## 🔗 Related Issues
Key issue was from `uv`. It was fixed very fast: https://github.com/astral-sh/uv/issues/12434

## ✅ Checklist


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools==70.0", "wheel"]
+requires = ["setuptools>=61.2", "wheel"]
```
danceratopz, line 2:
```suggestion
requires = ["setuptools==78.0.2", "wheel"]
```

danceratopz (commented):
The bug that #1343 addressed, wasn't a `uv` problem, it was due to a bug in `setuptools` `v78.0.0` (see [this comment](https://github.com/astral-sh/uv/issues/12434#issuecomment-2748594872) in the linked uv issue). It was fixed in `setuptools` `v78.0.1`.

As we originally didn't pin `setuptools` in `[build-system]` before your initial PR, #1343, eest was exposed to this bug. #1343 did pin it though, which is a good idea, so I suggest instead of simply reverting #1343 in this PR, we pin it to the latest patch release, `v78.0.2` to make we don't run into this type of problem again.

Then the question is: Should we pin the `setuptools` included in our package `dependencies`?
Which then begs the question: Why is `setuptools` in our package dependencies? :laughing: 

AFAICT, it's only there for `trie < 3`, as `trie` seemingly doesn't have it as a dependency, but imported `pkg_resources` from it's `__init__.py`:
```
src/ethereum_test_types/types.py:25: in <module>
    from trie import HexaryTrie
.venv/lib/python3.12/site-packages/trie/__init__.py:1: in <module>
    import pkg_resources
E   ModuleNotFoundError: No module named 'pkg_resources'
============================================================================================================================================================== short test summary info ===============================================================================================================================================================
ERROR  - ModuleNotFoundError: No module named 'pkg_resources'
```
Trie < 3 also gave us a setuptools related warning:
```
.venv/lib/python3.12/site-packages/trie/__init__.py:1
  /home/dtopz/code/github/new/execution-spec-tests/.venv/lib/python3.12/site-packages/trie/__init__.py:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    import pkg_resources
```
Luckily, updating to trie==3.1.0 solves all these problems :partying_face: 

So to summarize:
1. We should pin `setuptools` to 78.0.2 in the `[build-system]` `requires` section.
2. We should update `trie`'s dep constraints to `trie>=3.1.0,<4` in eest package `dependencies`.
3. We should remove `setuptools` and `setuptools-types` from our package `dependencies`.

This is getting a bit out of hand, but it seems worth addressing these issues :smile: 
﻿

pyproject.toml
danceratopz: Done locally

danceratopz:
>     1. We should pin `setuptools` to 78.0.2 in the `[build-system]` `requires` section.
Done by hand, doesn't effect uv.lock.

>     2. We should update `trie`'s dep constraints to `trie>=3.1.0,<4` in eest package `dependencies`.
```
uv add "trie>=3.1.0,<4"
```

>     3. We should remove `setuptools` and `setuptools-types` from our package `dependencies`.
```
uv remove setuptools types-setuptools
```

spencer-tb (commented):
LGTM! Thanks @danceratopz :)

danceratopz (approved):
:+1:

PR #1348, opened 2025-03-25
fix(fill): skip fixture index generation if output dir is not clean

### LINKED ISSUE #1030: bug(fill): insist that the output directory for generated JSON fixtures is initially empty
Since the test case index file generation was added as a teardown step to `fill` in #716, multiple developers have stumbled over the following issue. Background: the index file (`./fixtures/.meta/index.json`) gets used by `consume` for fast test case parsing and filtering.

1. Execute `fill` with output directory `./fixtures`.
2. Change `execution-spec-tests` branch to a branch with (breaking) changes to our JSON fixture pydantic models (which describe the fields and types defined in fixtures).
3. Execute `fill` with output directory `./fixtures`.
4. If the tests generated in 1. are effected by the breaking changes in 2., step 3. will report errors that the fixtures can not be parsed by the pydantic models.

This is very confusing to the user as the errors may include tests (or forks) that weren't even filled in step 3.

Possible options:
1. We could consider deleting the directory before writing new fixtures to it. Perhaps exiting cowardly unless a `--overwrite` flag or similar is specified.
2. Automatically create a unique sub-directory in the output directory using a high resolution timestamp, for example.
3. Simply refuse to write to the directory and exit, the user should specify a different directory or delete the old one.

mattgidsy:
## 🗒️ Description
This PR implements a safeguard around fixture index generation to prevent stale fixture parsing errors when switching between branches with schema-breaking changes.

Specifically:
- At `pytest_sessionstart`, the fixture output directory is checked for cleanliness (empty or not).
- The state is stored on the `pytest.config` object.
- At `pytest_sessionfinish`, the `generate_fixtures_index()` step is skipped if the output directory was not clean.
- A clear warning is issued in both setup and teardown phases when skipping index generation.

## 🔗 Related Issues
Fixes [#1030](https://github.com/ethereum/execution-spec-tests/issues/1030)

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/filler/filler.py
+++ b/src/pytest_plugins/filler/filler.py
@@ -250,6 +250,26 @@ def pytest_configure(config):
     config.stash[metadata_key]["Command-line args"] = f"<code>{command_line_args}</code>"
 
 
+def pytest_sessionstart(session: pytest.Session):
+    """Check if the output directory was clean at session start."""
+    output: Path = session.config.getoption("output")
+    output_dir = strip_output_tarball_suffix(output)
+
+    if is_output_stdout(output) or session.config.option.collectonly:
+        session.config._output_dir_was_clean = None
+        return
+
+    if output_dir.exists() and any(output_dir.iterdir()):
+        session.config._output_dir_was_clean = False
+        warnings.warn(
+            f"[filler] Output directory '{output_dir}' is not empty at session start. "
+            f"Index generation will be skipped to avoid parsing stale fixtures.",
+            stacklevel=2,
+        )
+    else:
+        session.config._output_dir_was_clean = True
+
+
 @pytest.hookimpl(trylast=True)
 def pytest_report_header(config: pytest.Config):
     """Add lines to pytest's console output header."""
@@ -817,11 +837,19 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int):
     for file in output_dir.rglob("*.lock"):
         file.unlink()
 
-    # Generate index file for all produced fixtures.
+    # Skip index generation if output dir was not clean at session start to avoid stale fixture
     if session.config.getoption("generate_index"):
-        generate_fixtures_index(
-            output_dir, quiet_mode=True, force_flag=False, disable_infer_format=False
-        )
+        was_clean = getattr(session.config, "_output_dir_was_clean", True)
+        if was_clean is True:
+            generate_fixtures_index(
+                output_dir, quiet_mode=True, force_flag=False, disable_infer_format=False
+            )
+        elif was_clean is False:
+            warnings.warn(
+                f"[filler] Skipping index file generation because the output directory "
+                f"'{output_dir}' was not clean at session start.",
+                stacklevel=2,
+            )
 
     # Create tarball of the output directory if the output is a tarball.
     is_output_tarball = output.suffix == ".gz" and output.with_suffix("").suffix == ".tar"


src/pytest_plugins/filler/filler.py
```diff
@@ -250,6 +250,26 @@ def pytest_configure(config):
...
+def pytest_sessionstart(session: pytest.Session):
+    """Check if the output directory was clean at session start."""
+    output: Path = session.config.getoption("output")
+    output_dir = strip_output_tarball_suffix(output)
+
+    if is_output_stdout(output) or session.config.option.collectonly:
+        session.config._output_dir_was_clean = None
+        return
+
+    if output_dir.exists() and any(output_dir.iterdir()):
```
danceratopz, line 262: Matt, the following was detected by debugging the failing unit test (https://github.com/ethereum/execution-spec-tests/actions/runs/14051860205/job/39344901008#step:5:48).

Unfortunately, this doesn't quite work due to some funky, non-obvious interaction with the pytest-html plugin. We hijack the pytest-html command-line args in `pytest_configure()` to write a test report in html format by default:
https://github.com/ethereum/execution-spec-tests/blob/b4e09b1e2750d36d205ccac345a499e868fc3286/src/pytest_plugins/filler/filler.py#L216-L221

It seems to me that this plugin ensures that the target directory for the generated html file exists at the end of its `pytest_configure()` hook, which means that the output directory is never empty in `any(output_dir.iterdir()):`) in our `pytest_sessionstart()` hook (decorating our `pytest_sessionstart()` with `@pytest.hookimpl(tryfirst=true)` which tries to prioritize our plugin hook over other plugins doesn't work).

I think the only way to fix this is to unfortunately move this check up to `pytest_configure()` hook, although I quite liked it here in `pytest_sessionstart()`.

I confirmed this by running the `fill` command used in the pytest setup `fixture`
https://github.com/ethereum/execution-spec-tests/blob/b4e09b1e2750d36d205ccac345a499e868fc3286/src/pytest_plugins/consume/tests/test_consume_args.py#L55-L73
from the problematic unit test:
```
fill -c pytest.ini --skip-evm-dump -m "(not blockchain_test_engine) and (not eip_version_check)" --from=Paris --until=Cancun --output=/tmp/pytest-of-dtopz/pytest-13/fixtures0 tests/istanbul/eip1344_chainid/test_chainid.py 
```
-> dir exists
```
ls -da /tmp/pytest-of-dtopz/pytest-13/fixtures0/.meta/
```

Adding the `--no-html` flag cures the problem:
```
rm -rf /tmp/pytest-of-dtopz/pytest-13/fixtures0
fill -c pytest.ini --skip-evm-dump -m "(not blockchain_test_engine) and (not eip_version_check)" --from=Paris --until=Cancun --output=/tmp/pytest-of-dtopz/pytest-13/fixtures0 tests/istanbul/eip1344_chainid/test_chainid.py 
```
-> the index file is generated as expected.

This was a tricky one!

src/pytest_plugins/filler/filler.py
```diff
@@ -250,6 +250,26 @@ def pytest_configure(config):
...
+        session.config._output_dir_was_clean = False
+        warnings.warn(
```
danceratopz, line 264: I know we discussed a warning here, but after seeing the output, which is quite noisy for little information:
```
/home/dtopz/code/github/new/execution-spec-tests/.venv/lib/python3.12/site-packages/pluggy/_callers.py:103: UserWarning: [filler] Output directory '<function output_dir at 0x7d7572caaac0>' is not empty at session start. Index generation will be skipped to avoid parsing stale fixtures.
  res = hook_impl.function(*args)
```

I think the better approach would be to append a line to the pytest session header using this hook:
https://github.com/ethereum/execution-spec-tests/blob/b4e09b1e2750d36d205ccac345a499e868fc3286/src/pytest_plugins/filler/filler.py#L253

Similar to the warning in the header generated from the forks plugin highlighted here:
image

danceratopz (commented):
Hey @mattgidsy thanks a lot for this! It's not quite working with default command-line options yet, I think once you're set up with WSL, then you'll be able to run the unit tests locally and have more confidence in your implementation.

It does work if you add `--no-html` to fill though, so it's very close! More details below!

winsvega:
hm. shall we require to clean the output directory unless user provide a flag?

danceratopz:
> hm. shall we require to clean the output directory unless user provide a flag?

Hi @winsvega, please see the reasoning here, happy to discuss options in the meeting:
https://github.com/ethereum/execution-spec-tests/issues/1030#issuecomment-2591861585

danceratopz:
Hi @mattgidsy, thanks for your efforts on this! As we discussed, the team revisited this issue and decided to opt for a safer solution which insists that the specified output directory must either not exist or be completely empty. This ensures that an output directory can never be "polluted" with incompatible test fixtures from multiple runs.

This was nonetheless a good effort, do reach out if you'd like to contribute again in the future!

Closing in favor of
- #1471
- #1473

danceratopz:
> hm. shall we require to clean the output directory unless user provide a flag?

Yup, following up from our discussions, we've opted for this solution instead:
- #1473

PR #1349, opened 2025-03-25
fix(tests): EIP-2935: change ret_offset to avoid duplicate

sonhv0212:
## 🗒️ Description

The `check_contract` uses `MSTORE` to store `check_block_number` at memory offset 0. However, if `CALL` to `HISTORY_STORAGE_ADDRESS` is reverted, `MLOAD(0)` returns the value of `check_block_number` instead of `0x0`, as expected.

I ran this test in `execute` mode and found that it failed. My solution was to change `ret_offset` to avoid duplication with `check_block_number`. I ran it again and it worked as expected.


sonhv0212:
cc @marioevz

marioevz (approved):
LGTM, filled locally and everything seems to be working fine.

Thanks for spotting this!

PR #1289, opened 2025-03-07
chore(tests): update bls12 test vectors

spencer-tb:
## 🗒️ Description
Following: #1275, this PR adds the updated test vectors to consolidate coverage for the BLS12 consensus bug.

From 1098 test cases to now 2073. There are still more we can add with parameterization.

Tested against geth using consume direct.

## 🔗 Related Issues
https://github.com/ethereum/EIPs/pull/9425

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.

PR #1350, opened 2025-03-25
feat(tests): extra BLS12-381 coverage

spencer-tb:
## 🗒️ Description

Primarily adds additional coverage to g1/g2 add/mul/msm and pairing functions.
Additionally a case for the fp map to g1.

Includes a BLS point generator class that is used in some of the test vectors. This may be used for future fuzzing. More tests will be added in a follow up PR using this generator.

## 🔗 Related Issues
Previous vector addition is included here: #1289 (to explain why we are deleting `tests/prague/msm_G1_bls.json`.

## ✅ Checklist


tests/prague/msm_G1_bls.json
spencer-tb, line 1: Seems this was my mistake, and a duplicate, it exists in `vectors/`.

tests/prague/eip2537_bls_12_381_precompiles/helpers.py
```diff
@@ -61,3 +65,579 @@ def vectors_from_file(filename: str) -> List:
...
+                continue
+            for sign in [1, -1]:
+                zt2 = (-1 + sign * discriminant.sqrt()) / 2
+                t2 = zt2 / ISO_3_Z
+                if t2.is_square():
+                    t = t2.sqrt()
+                    x1 = -ISO_3_B / ISO_3_A * (1 + 1 / (ISO_3_Z^2 * t^4 + ISO_3_Z * t^2))
+                    assert x1 == x, f"Verification failed: x1={x1} != x={x}"
+                    print(t)
+        ```
```
marioevz, line 624: I went over the script a couple hours today but I could not find an issue with it.

I even changed the expected outcome from `INF_G2` to what py_ecc outputs (and eels) and ran the tests with all clients in an effort to try to catch a bug in py_ecc, but all clients pass the test, so it seems to me that the `t` here is incorrect, which is surprising given the assertion that's in the code.

tests/prague/eip2537_bls_12_381_precompiles/helpers.py
```diff
@@ -61,3 +65,579 @@ def vectors_from_file(filename: str) -> List:
...
+        # 2. A non-trivial point P (what we generated with the sage script).
+        # 3. The point -P (negation of P).
+        #
+        # Note no field elements in FP2 map to P or -P through the SWU map!!
+        return FP2(
+            (
+                1321391878976200074119645034444121025665939532471580297403784397444487204385014689930429494040875145883332046150361,
+                2445707658947438392540268931313688646028792845302972805205282500613101640661654934098821383626938971038899010733909,
+            ),
+        )
```
marioevz, line 643: I'm thinking that for the time being this should return an empty list and we should make a comment that the value returned by the algorithm for some reason do not result in the `INF_G2` response we expect.

tests/prague/eip2537_bls_12_381_precompiles/helpers.py
```diff
@@ -61,3 +65,579 @@ def vectors_from_file(filename: str) -> List:
...
+            discriminant = 1 - 4 / (ISO_11_A / ISO_11_B * x + 1)
+            if not discriminant.is_square():
+                continue
+            for sign in [1, -1]:
+                zt2 = (-1 + sign * discriminant.sqrt()) / 2
+                t2 = zt2 / ISO_11_Z
+                if t2.is_square():
+                    t = t2.sqrt()
+                    assert x == -ISO_11_B / ISO_11_A * (1 + 1 / (ISO_11_Z**2 * t**4 + ISO_11_Z * t**2))
+                    print(t)
```
marioevz, line 571: This is absolutely amazing!

tests/prague/eip2537_bls_12_381_precompiles/test_bls12_map_fp_to_g1.py
```diff
@@ -61,8 +63,51 @@ def test_valid(
     )
 
 
+@pytest.mark.parametrize(
+    "input_data,expected_output,vector_gas_value",
```
marioevz, line 67:
```suggestion
@pytest.mark.parametrize("expected_output", [Spec.INF_G1], ids=[""])
@pytest.mark.parametrize(
    "input_data",
```

tests/prague/eip2537_bls_12_381_precompiles/test_bls12_map_fp2_to_g2.py
```diff
@@ -73,8 +76,73 @@ def test_valid(
     )
 
 
+@pytest.mark.parametrize(
+    "input_data,expected_output,vector_gas_value",
```
marioevz, line 80:
```suggestion
@pytest.mark.parametrize("expected_output", [Spec.INF_G2], ids=[""])
@pytest.mark.parametrize(
    "input_data",
```
Same as previous comment, but this should actually not run any tests because we don't have such points.

tests/prague/eip2537_bls_12_381_precompiles/helpers.py
```diff
@@ -61,3 +65,579 @@ def vectors_from_file(filename: str) -> List:
...
+        """
+        return cls.find_g1_point_by_x(x_value, in_subgroup=True, on_curve=False)
+
+    # G1 random points - all methods updated with seed parameter
+    @classmethod
+    def generate_random_g1_point_in_subgroup(cls, seed=None) -> PointG1:
+        """Generate a random G1 point that is in the r-order subgroup."""
+        seed_value = cls.BASE_SEED if seed is None else cls.BASE_SEED + seed * 10
+        random.seed(seed_value)
+        x = random.randrange(Spec.P)  # Random field element in Fp
```
marioevz, line 442: I would lean towards making `seed` obligatory and using a hash function instead.

I can see that `Spec.P` is 380 bits long, so we could perhaps use sha384? (I think it should be in the hashlib library).

tests/prague/eip2537_bls_12_381_precompiles/helpers.py
```diff
@@ -61,3 +65,579 @@ def vectors_from_file(filename: str) -> List:
...
+                x = FQ(point[0])  # type: ignore
+                y = FQ(point[1])  # type: ignore
+                base_point = (x, y)
+                result = multiply(base_point, cofactor)
+                return (int(result[0]), int(result[1]))  # type: ignore
+        except Exception as e:
+            raise ValueError("Failed to multiply point by cofactor") from e
+
+    @classmethod
+    def find_g1_point_by_x(cls, x_value: int, in_subgroup: bool, on_curve: bool = True) -> PointG1:
```
marioevz, line 250: I think we should cache this and `find_g2_point_by_x`, and cache them to disk in order to make loading this library faster.

I haven't used any cache decorator myself but after a quick chatgpt query it suggested `joblib`:
```python
from joblib import Memory

memory = Memory(location=".cache", verbose=0)

@memory.cache
def expensive_function(x, y):
    print("Computing...")
    return x + y
```

marioevz (commented):
Looks amazing, thanks for putting so much effort into the algorithms to find the kernels!

I think we could merge without a proper Fp2 that resolves to `INF_G2` and follow up afterwards because the rest of the tests included in this PR are extremely important.

I left a few comments.

tests/prague/eip2537_bls_12_381_precompiles/helpers.py
spencer-tb: I copied it from Peter! :)

tests/prague/eip2537_bls_12_381_precompiles/helpers.py
spencer-tb: Thanks so much for looking into this and confirming :D

marioevz (approved):
Fantastic job on getting so many cases implemented! 🎉

Issue #533, opened 2024-04-29
bug(fill): fix misleading test collection reporting in terminal output


danceratopz:
After the pytest header and test collection (but before test execution) pytest incorrectly reports how many tests are collected. This is visible with and without `--collect-only`.

For example:
```console
fill tests/istanbul/eip1344_chainid/ --fork=Cancun --collect-only
```
outputs (incorrect output is marked as such):
```console
================================== test session starts ===================================
platform linux -- Python 3.10.12, pytest-7.4.4, pluggy-1.5.0
Executing tests for: Cancun 
Only executing tests with stable/deployed forks: Specify an upcoming fork via --until=fork to add forks under development.
rootdir: /home/dtopz/code/github/danceratopz/execution-spec-tests
configfile: pytest.ini
plugins: cov-4.1.0, xdist-3.6.1
collecting ... start
collected 3 items / 1 deselected / 2 selected  # <-- INCORRECT, should be "collected 3 items"

<Package eip1344_chainid>
  <Module test_chainid.py>
    <Function test_chainid[fork_Cancun-blockchain_test]>
    <Function test_chainid[fork_Cancun-blockchain_test_hive]>
    <Function test_chainid[fork_Cancun-state_test]>

====================== 2/3 tests collected (1 deselected) in 0.01s =======================   # <-- INCORRECT, should be "3 tests collected in 0.01 s"
```

For every test module collected, one additional "test" will be incorrectly reported as deselected (above we only collect one module).

The error is due to the fact that we automatically add EIP version tests in the `spec_version_checker` plugin (`src/pytest_plugins/spec_version_checker/spec_version_checker.py`) but then also disable them by default in `fill`'s `pytest.ini` (L9 and L11):
https://github.com/ethereum/execution-spec-tests/blob/4783e6e7fe8337d492ad1d449ddd285cc35363cf/pytest.ini#L6-L11

The most reasonable solution would be to simply not use this plugin when executing `fill` and have yet another pytest command dedicated to version checking that can be ran regularly in CI. The reason to not enable these tests by default is that each of these version checking tests makes a request to the Github API (to retrieve info from ethereum/EIPs); without authentication setup, `fill`  quickly hits the Github API rate limit.

Note, as of 2024-04-29, eip version checking is currently entirely unused in our CI. To run these tests, it's required to explicitly select tests marked as `eip_version_check` with:
```
fill tests/istanbul/eip1344_chainid/ --fork=Cancun -m eip_version_check -v
```

[...]

PR #1357, opened 2025-03-26
fix(plugins/spec_version_checker): Sort modules

marioevz:
## 🗒️ Description
For some reason the xdist started to fail and it was due to the items being added to the list of tests by the spec version checker were in a different order in each worker.

This PR sorts the modules before adding the tests to have a deterministic list.

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/spec_version_checker/spec_version_checker.py
+++ b/src/pytest_plugins/spec_version_checker/spec_version_checker.py
@@ -5,6 +5,7 @@ modules matches that of https://github.com/ethereum/EIPs.
 
 import re
 from types import ModuleType
+from typing import List, Set
 
 import pytest
 from _pytest.nodes import Item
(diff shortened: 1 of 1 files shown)


danceratopz (commented):
@marioevz this looks fine, but tbh, I would really like to completely remove the `spec_version_checker` logic from `fill`. It messes with our test collection results and leads to confusing reporting in the terminal, cf 
- #533

Also, I think these checks are of limited use (do you run them btw?) and are certainly not executed anywhere in CI.

As this has crop up now, my suggestion would be to simply to remove this feature from `fill` and create a follow-up issue to add a separate command that does this and can run nightly in CI, for example. Yet another plugin? :laughing:

danceratopz:
I ran the eip version check tests locally:
```
fill -m eip_version_check -v
```
8 tests passed, here's a failure mode summary:
image

We now have so many checks that we get throttled by the Github API within a single test session (not the case back in the day :wink:).

We can avoid the throttling of course.

spencer-tb (approved):
Approving so we can fill Prague!

I'm in favour of removing this from `fill` too. A separate command in CI would be nice or something that just runs for all the EL EIPs and pings for changes? I think we should create a seperate issue for the latter @danceratopz?

The solution around the API throttling is to use the Trees API endpoint.
This would allow us to fetch the SHA hashes for all spec files in a single API call rather than making individual requests per file, `/git/trees/{branch}?recursive=1` to get all files in one request.

danceratopz (approved):
LGTM

PR #1359, opened 2025-03-26
refactor(base_types,types,tests): Create RLP Serialization Classes

marioevz:
## 🗒️ Description
Creates `RLPSerializable` and `SignableRLPSerializable` that can be inherited by a class to make it RLP serializable. 

### RLPSerializable
 When inherited by a class, and the class defines the `rlp_fields` class variable as a list of strings, the method `rlp` will now return an RLP encoded bytes that is composed by the fields named in the ordered `rlp_fields`.

The method will automatically convert the fields into values accepted by `ethereum_rlp.encode`.

If the field contains an object that is also `RLPSerializable` (or `SignableRLPSerializable`), it will call a method called `to_list` which returns the object's fields a list that is already converted to values accepted by `ethereum_rlp.encode`, so the encoding will treat this sub-field as a list of elements instead of an array of bytes.

### SignableRLPSerializable
Similar to `RLPSerializable` but it can also produce RLP encoded bytes that are meant to be signed with a private key, and also requires implementation of a `sign` method.

E.g. `AuthorizationTuple` contains the following RLP related fields:
```python
rlp_fields: ClassVar[List[str]] = ["chain_id", "address", "nonce", "v", "r", "s"]
rlp_signing_fields: ClassVar[List[str]] = ["chain_id", "address", "nonce"]
```
Which means that the `sign` method will use an RLP encoded bytes composed of only `["chain_id", "address", "nonce"]`, whereas the full RLP encoded object is composed of those same fields plus the `["v", "r", "s"]` of the signature.

Note: we could improve the names of both of these classes, suggestions welcome.

### Testing Opportunities
While this PR only refactors and does not introduce many more functionality, it introduces the possibility of automatically generating tests for RLP encodeable objects, since the objects and their composition based on their fields are now better defined.

### Test parity
Test fixtures were generated using `uv run fill --from=Cancun --until=Prague -n auto -m "not slow and not eip_version_check"` against main generated functions (plus #1357 and #1358) and the resulting fixtures were identical.

## 🔗 Related Issues
None

## ✅ Checklist


src/ethereum_test_base_types/serialization.py
```diff
@@ -0,0 +1,144 @@
...
+    rlp_override: Bytes | None = None
+
+    signable: ClassVar[bool] = False
+    rlp_fields: ClassVar[List[str]]
+    rlp_signing_fields: ClassVar[List[str]]
+
+    def get_rlp_fields(self) -> List[str]:
+        """
+        Return a list containing the ordered names of the list that need be included
+        in the list of RLP fields to serialize for the RLP object.
```
spencer-tb, line 40:
```suggestion
        Return an ordered list of field names to be included in RLP serialization.
```

src/ethereum_test_base_types/serialization.py
```diff
@@ -0,0 +1,144 @@
...
+        By default, rlp_fields class variable is used.
+
+        The list can be nested list up to one extra level to represent nested fields.
+        """
+        return self.rlp_fields
+
+    def get_rlp_signing_fields(self) -> List[str]:
+        """
+        Return a list contained the ordered names of the list that need be included
+        in the list of RLP fields to serialize for the object signature.
```
spencer-tb, line 53:
```suggestion
        Return an ordered list of field names to be included in the RLP serialization of the object signature.
```

src/ethereum_test_types/types.py
```diff
@@ -1071,6 +1035,27 @@ def created_contract(self) -> Address:
         return Address(hash_bytes[-20:])
 
 
+class NetworkWrappedTransaction(CamelModel, RLPSerializable):
+    """Network wrapped transaction."""
```
spencer-tb, line 1039: Love the refactor!

Could we add a bit more here to explain what the networked wrapped tx?

src/ethereum_test_base_types/serialization.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        field_list: List[str]
+        if signing:
+            if not self.signable:
+                raise Exception(f'Object "{self.__class__.__name__}" does not support signing')
+            field_list = self.get_rlp_signing_fields()
+        else:
+            if self.signable:
+                self.sign()
+            field_list = self.get_rlp_fields()
```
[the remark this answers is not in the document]
marioevz: It's tricky because it will attempt to sign signable objects embedded inside of a signable/non-signable outer object.

This is just to ensure that an embedded object is not RLP encoded with zeros where its signature should be.

Take for example an authorization tuple inside of a transaction: If we try to serialize the transaction and embed the authorization tuple inside of it, but the authorization tuple has not been signed, it will serialize the transaction with an authorization tuple that has zeros as its signature, even though the authorization tuple has a signer/sender embedded and we have everything we need for it to be signed.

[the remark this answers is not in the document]
marioevz: I just tried to remove it, but it's indeed very convenient for the sub-elements to be signed automatically.

When I removed it, one of the tests that was using the `NetworkWrappedTransaction` started failing because since it contains a full transaction, which is signable, doing `NetworkWrappedTransaction.rlp()` returned an rlp where the transaction was not signed.

So I think it's better to keep this auto-signing to avoid filling tests where the RLP does not contain the signature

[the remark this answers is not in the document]
spencer-tb: Thanks for the explaining! This makes sense. What if we add a small comment by the else, something like:
```
else:
    # Automatically sign signable objects during full serialization
    # Ensures nested objects have valid signatures in the final RLP
```

docs/CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ consume cache --help
 
 - 🔀 Bump the version of `execution-specs` used by the framework to the package [`ethereum-execution==1.17.0rc6.dev1`](https://pypi.org/project/ethereum-execution/1.17.0rc6.dev1/); bump the version used for test fixture generation for forks < Prague to current `execution-specs` master, [fa847a0](https://github.com/ethereum/execution-specs/commit/fa847a0e48309debee8edc510ceddb2fd5db2f2e) ([#1310](https://github.com/ethereum/execution-spec-tests/pull/1310)).
 - 🐞 Init `TransitionTool` in `GethTransitionTool` ([#1276](https://github.com/ethereum/execution-spec-tests/pull/1276)).
+- 🔀 Refactored RLP encoding of test objects to allow automatic generation of tests. ([#1359](https://github.com/ethereum/execution-spec-tests/pull/1359))
```
spencer-tb, line 55:
```suggestion
- 🔀 Refactored RLP encoding of test objects to allow automatic generation of tests ([#1359](https://github.com/ethereum/execution-spec-tests/pull/1359)).
```

spencer-tb: Sorry I noticed it and could not unsee it :D

spencer-tb (approved):
LGTM! ;)

PR #1362, opened 2025-03-27
feat(tests): Parse and fill original Filler.json using static test loader

winsvega:
Introduce StatiStateTest class that loads Filler.json/yml and makes fixtures out of vectors from that tests
fillconvert command to verify test parsing status and verify generated tests post hash
Pydantic schemas to read Filler.json/yaml
Fixes to pyspecs logic around state tests -> blockchain test, correctly generating genesis and block1


src/ethereum_test_specs/state_static.py
```diff
@@ -0,0 +1,223 @@
...
+from .base_static import BaseStaticTest
+from .state import StateTestFiller
+
+
+class StateStaticTest(BaseStaticTest):
+    """General State Test static filler from ethereum/tests."""
+
+    tests: Dict[str, StateTestInFiller] | None = Field(None)
+    vectors: List[StateTestVector] | None = Field(None)
+    # format_name: ClassVar[str] | None = "state_test"
```
marioevz, line 27:
```suggestion
    format_name: ClassVar[str] = "state_test"
```
This might have been an issue.

src/ethereum_test_specs/state_static.py
```diff
@@ -0,0 +1,223 @@
...
+from ethereum_test_types import Account, Alloc, Environment, Transaction
+
+from .base_static import BaseStaticTest
+from .state import StateTestFiller
+
+
+class StateStaticTest(BaseStaticTest):
+    """General State Test static filler from ethereum/tests."""
+
+    tests: Dict[str, StateTestInFiller] | None = Field(None)
```
marioevz, line 25: This should be changed because what refiller passes to the subclasses of `BaseStaticTest` is basically what `StateTestInFiller` is.

winsvega:
so if Pydantic parse fails. the test is not detected. but I want to see an error message to avoid tests being skipped

the fill errors are because in state test when env has currentBaseFee = 1000 it means the block that executes this transaction miust have baseFee equal 1000
when we fill state test as blockchain  genesis must have baseFee pre calculated in such a way so that 
b1.basefee = basefee(genesis.basefee) = 1000 
genesis.basefee = basefee-1(b1.basefee) = 1142
same for difficulty

why it fills into 
`stExample/add11Filler/add11.json`
instead of 
`stExample/add11.json`

winsvega:
I remove all the .json fillers for code review, but not squashing the commits for now to preserve the history

marioevz (approved):
Amazing job getting this implemented!

Applied some changes directly before merging:
- Renamed `legacy_state_tests` folder to `legacy/state`, and I think the follow up PR where we move these tests needs to move them to this new folder.
- Moved files from `cli/fillerconvert/structures` to `ethereum_test_specs/static_state` and renamed the imports.
- Remove hard-coded directory references

file.txt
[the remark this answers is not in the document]
winsvega: No, it is lllc test

Issue #1360, opened 2025-03-27
feat(tests): add tests for the new stricter rules of EIP-6110 (deposit requests)


jochem-brouwer:
These tests have to do with this merged PR: https://github.com/ethereum/EIPs/pull/9460

Note that the EIP does not specify the specific bytecode, it only specifies that the deposit contract, once it logs a DepositEvent (this has log topic 0 `DEPOSIT_EVENT_SIGNATURE_HASH`), the layout of this topic is strictly validated. If this validation fails, the block is invalid. It is therefore possible to use any deposit contract bytecode, but the log layout of the deposit event should be correct. Note that Sepolia and Mainnet/Holesky differ in bytecode from each other, and it is likely that devnets or maybe L2s introduce new bytecodes (Sepolia has permissioned deposits via an integrated ERC20 token).

The tests should test the deposit contract in the following way:

- [ ] LOG with correct layout (VALID)
- [ ] LOG with log data > 576
- [ ] LOG with log data < 576
- [ ] LOG with offsets not at the right location
- [ ] LOG with sizes not at the right sizes
- [ ] LOG with different topic (VALID)

These tests should be wrapped in a test which also has txs which submit the log, but where the tx (or the call) gets reverted.

This is low-prio since this will only fail for non-mainnet clients which have a deposit contract which does not use the mainnet bytecode or Sepolia bytecode, and this "custom" bytecode does not follow the log layout as specified in the EIP. 

I will assign myself for this :)

PR #1371, opened 2025-03-31
feat(tests): EIP-6110 `DepositEvent` layout+topic tests

jochem-brouwer:
## 🗒️ Description
PR built on top of https://github.com/ethereum/execution-spec-tests/pull/1280 to add EIP-6110 `DepositEvent` log layout tests as described in https://github.com/ethereum/EIPs/pull/9460

It tests:

- [x] LOG with correct layout and an ERC20 `Transfer` event (Sepolia-like case) (VALID)
- [x] LOG with log data > 576 (INVALID block)
- [x] LOG with log data < 576 (INVALID block)

For each param (`pubkey`, `withdrawal_credentials`, `amount`, `signature`, `index`)
- [x] LOG with offsets not at the right location (INVALID block)
- [x] LOG with sizes not at the right sizes (INVALID block)

## 🔗 Related Issues
Closes https://github.com/ethereum/execution-spec-tests/issues/1360

## ✅ Checklist


tests/prague/eip6110_deposits/helpers.py
```diff
@@ -17,6 +19,61 @@ def sha256(*args: bytes) -> bytes:
...
+    index_size=8,
+    index_data=b"",
+    index_offset=512,
+) -> bytes:
+    result = bytearray(576)
+    offset = 0
+
+    def write_uint256(value):
+        nonlocal offset
+        result[offset : offset + 32] = value.to_bytes(32, byteorder="big")
```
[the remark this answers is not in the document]
jochem-brouwer: Great question, I had not realized this but this is a python quirk. Notice that I re-assign `offset` later in this method, but I "edit" `result` (this does not re-assign the `result` value so not `result = x`, I instead to `result[x]=y`)

image

This works:

image

This does not:

image

(which is the code, need to edit this `offset`)

Note: this is a python thing, I got confused, because from JS side this is absolutely not necessary, you can just re-assign and it will just pick the first "scope" it can find it has the variable declared, or create a new one in the current scope :smile:

tests/prague/eip6110_deposits/helpers.py
```diff
@@ -5,9 +5,11 @@
 from hashlib import sha256 as sha256_hashlib
 from typing import Callable, ClassVar, List
 
-from ethereum_test_tools import EOA, Address, Alloc, Bytecode, Hash, Transaction
+from ethereum_test_tools import EOA, Address, Alloc, Bytecode
 from ethereum_test_tools import DepositRequest as DepositRequestBase
+from ethereum_test_tools import Hash
 from ethereum_test_tools import Opcodes as Op
+from ethereum_test_tools import Transaction
```
marioevz, line 12:
```suggestion
from ethereum_test_tools import EOA, Address, Alloc, Bytecode, Hash, Transaction
from ethereum_test_tools import DepositRequest as DepositRequestBase
from ethereum_test_tools import Opcodes as Op
```
Just to appease the linter.

tests/prague/eip6110_deposits/helpers.py
```diff
@@ -17,6 +19,61 @@ def sha256(*args: bytes) -> bytes:
...
+    withdrawal_credentials_offset=256,
+    amount_size=8,
+    amount_data=b"",
+    amount_offset=320,
+    signature_size=96,
+    signature_data=b"",
+    signature_offset=384,
+    index_size=8,
+    index_data=b"",
+    index_offset=512,
```
marioevz, line 37:
```suggestion
    pubkey_size: int = 48,
    pubkey_data: bytes = b"",
    pubkey_offset: int = 160,
    withdrawal_credentials_size: int = 32,
    withdrawal_credentials_data: bytes = b"",
    withdrawal_credentials_offset: int = 256,
    amount_size: int = 8,
    amount_data: bytes = b"",
    amount_offset: int = 320,
    signature_size: int = 96,
    signature_data: bytes = b"",
    signature_offset: int = 384,
    index_size: int = 8,
    index_data: bytes = b"",
    index_offset: int = 512,
```
Also for the linter.

tests/prague/eip6110_deposits/helpers.py
```diff
@@ -17,6 +19,61 @@ def sha256(*args: bytes) -> bytes:
...
+) -> bytes:
+    result = bytearray(576)
```
marioevz, line 39:
```suggestion
    """Create the deposit log bytes."""
    result = bytearray(576)
```

tests/prague/eip6110_deposits/helpers.py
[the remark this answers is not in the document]
marioevz: Had not used `nonlocal` before! Great to know :)

tests/prague/eip6110_deposits/helpers.py
```diff
@@ -17,6 +19,61 @@ def sha256(*args: bytes) -> bytes:
...
+    offset = 0
+
+    def write_uint256(value):
```
marioevz, line 42:
```suggestion
    def write_uint256(value: int):
```

tests/prague/eip6110_deposits/helpers.py
```diff
@@ -17,6 +19,61 @@ def sha256(*args: bytes) -> bytes:
...
+        nonlocal offset
+        result[offset : offset + 32] = value.to_bytes(32, byteorder="big")
+        offset += 32
+
+    def write_bytes(data, size):
```
marioevz, line 47:
```suggestion
    def write_bytes(data: bytes, size: int):
```

tests/prague/eip6110_deposits/test_modified_contract.py
```diff
@@ -0,0 +1,201 @@
...
+            ),
+        ],
+        post={},
+    )
+
+
+@pytest.mark.parametrize(
+    "log_argument,value",
+    [(log_argument, value) for log_argument in EVENT_ARGUMENTS for value in EVENT_ARGUMENT_VALUES],
+)
```
marioevz, line 158:
```suggestion
)
@pytest.mark.exception_test
```
This is required on the latest master, because we now track the tests that result in invalid block/transactions using this marker.

tests/prague/eip6110_deposits/test_modified_contract.py
```diff
@@ -0,0 +1,201 @@
...
+def test_invalid_layout(
+    blockchain_test: BlockchainTestFiller, pre: Alloc, log_argument: str, value: str
+):
+    """Test deposit contract emitting logs with invalid layouts (sizes/offsets)"""
```
marioevz, line 162:
```suggestion
    """Test deposit contract emitting logs with invalid layouts (sizes/offsets)."""
```
Ruff complains about periods at the end of docstrings.

tests/prague/eip6110_deposits/test_modified_contract.py
```diff
@@ -0,0 +1,201 @@
...
+
+    log_params = {**DEFAULT_DEPOSIT_REQUEST_LOG_DATA_DICT}
+    log_params[log_argument] = 0 if value == "zero" else int("FF" * 32, 16)
+
+    deposit_request_log = create_deposit_log_bytes(**log_params)
```
marioevz, line 167:
```suggestion
    log_params[log_argument] = 0 if value == "zero" else 2**256 - 1  # type: ignore

    deposit_request_log = create_deposit_log_bytes(**log_params)  # type: ignore
```
I'd use type ignore in these two because of how the layout is, which is ok but it's tricky to comply with mypy when using kwargs.

Another tiny change is `2**256 - 1` instead of `int("FF" * 32, 16)` to avoid hex parsing on each test run.

tests/prague/eip6110_deposits/test_modified_contract.py
```diff
@@ -0,0 +1,201 @@
...
+        code=bytecode,
+        nonce=1,
+        balance=0,
+    )
+    sender = pre.fund_eoa()
+
+    tx = Transaction(
+        to=Spec.DEPOSIT_CONTRACT_ADDRESS,
+        sender=sender,
+        gas_limit=100_000,
```
marioevz, line 186:
```suggestion
        gas_limit=100_000,
        error=BlockException.INVALID_REQUESTS,
```
We have to add the exception here too due to how EELS applies the logic verification to the deposit contract (has to be the same as the exception at the block level so we could end up changing it).

tests/prague/eip6110_deposits/test_modified_contract.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+    blockchain_test(
+        pre=pre,
+        blocks=[
+            Block(
+                txs=[tx],
+                exception=BlockException.INVALID_REQUESTS,  # TODO: use correct exception
```
marioevz, line 194: I think the most appropriate thing would be to create a new transaction exception.

Perhaps `TransactionException.INVALID_DEPOSIT_LOG` or similar.

Looking at all the different error messages that EELS currently with the incorrect deposit log, I think we will probably have to use regex.

First step would be to add the new transaction to this Enum:
https://github.com/ethereum/execution-spec-tests/blob/13aeeedff0b4c48cf464a80d30eec4ca2fadab98/src/ethereum_test_exceptions/exceptions.py#L113

Then add it to EELS exception mapper:
https://github.com/ethereum/execution-spec-tests/blob/13aeeedff0b4c48cf464a80d30eec4ca2fadab98/src/ethereum_clis/clis/execution_specs.py#L202

And finally to ethjs:
https://github.com/ethereum/execution-spec-tests/blob/13aeeedff0b4c48cf464a80d30eec4ca2fadab98/src/ethereum_clis/clis/ethereumjs.py#L47

marioevz (commented):
Thanks for these!

I think it looks pretty good overall, and just some minor linting issues to get to calm down ruff and mypy :)

If you need help with anything please just ping me and I can help out, but basically once we fix the comments we should be able to go ahead and merge 👍

jochem-brouwer:
CI passes, ready for review!

NOTE: the current tests will check for the "failed transaction error" for the invalid deposit logs. This will fail on client implementations which check the deposit contract logs after ALL transactions (EthJS) and not after EACH transaction (EELS). Any client which has this bug likely implements it the "EthJS" way and is therefore not suitable to fill the tests. Note that EIP-6110 specs parsing the deposit logs after all transactions, not after each transaction.

tests/prague/eip6110_deposits/test_modified_contract.py
```diff
@@ -0,0 +1,223 @@
...
+    Requests,
+    Transaction,
+)
+from ethereum_test_tools import Macros as Om
+from ethereum_test_tools import Opcodes as Op
+
+from .helpers import DepositRequest, create_deposit_log_bytes
+from .spec import Spec, ref_spec_6110
+
+pytestmark = pytest.mark.valid_from("Prague")
```
marioevz, line 21:
```suggestion
pytestmark = [
    pytest.mark.valid_from("Prague"),
    pytest.mark.execute(pytest.mark.skip(reason="modifies pre-alloc")),
]
```
This is a nice addition to have to avoid trying to execute these tests in live networks, since it requires a modified contract which we cannot pre-set there.

marioevz (approved):
Looks great, thanks so much for these!

PR #1379, opened 2025-04-01
feat(tests): Additional EIP-2935 coverage

reedsa:
## 🗒️ Description

Additional coverage for checklist items in https://github.com/ethereum/execution-spec-tests/blob/eip-checklist/docs/checklist_templates/eip_testing_checklist_template.md#-new-system-contract
* Add cases for calling history storage address with all call opcode variants.

## ✅ Checklist

### DIFF
--- a/tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py
+++ b/tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py
@@ -315,6 +315,55 @@ def test_block_hashes_history(
     )
 
 
+@pytest.mark.valid_from("Prague")
+@pytest.mark.with_all_call_opcodes
+def test_block_hashes_call_opcodes(
+    blockchain_test: BlockchainTestFiller, pre: Alloc, call_opcode: Op
+):
+    """Test that the call opcodes can be used to call the history contract and get the block hashes."""  # noqa: E501
+    blocks = []
+    blocks.append(Block())
+
+    storage = Storage()
+    return_code_slot = storage.store_next(0x1)
+    blockhash_value_slot = storage.store_next(
+        True if call_opcode in [Op.DELEGATECALL, Op.CALLCODE] else False
+    )
+
+    code = (
+        Op.MSTORE(0, 1)
+        + Op.SSTORE(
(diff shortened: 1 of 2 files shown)


IzzyTheBuilder (commented):
Looks great! 🎉

reedsa:
I was tinkering a bit with DELEGATECALL and CALLCODE to match the blockhash returned from the contract. Ended up just verifying that the call is successful.

If verifying the blockhash when using DELEGATECALL or CALLCODE is critical, we can find a way to do so for these call types. The issue was that the returned Op.BLOCKHASH in storage cannot be cross referenced with the return value of the history contract at the separate address. Maybe I'm missing something that would make this possible.

tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py
```diff
@@ -315,6 +315,48 @@ def test_block_hashes_history(
...
+    code = Op.MSTORE(0, 1) + Op.SSTORE(
+        return_code_slot,
+        call_opcode(
+            address=Spec.HISTORY_STORAGE_ADDRESS,
+            args_offset=0,
+            args_size=32,
+            ret_offset=32,
+            ret_size=32,
+        ),
+    )
```
marioevz, line 339: Checking the returned value would be hard because we (currently) don't have a way to compare it against the previous block's real hash, but we could store `Op.ISZERO(Op.MLOAD(32))` as a sanity check that the returned value is not zero, which should actually fail on `call_opcode == Op.DELEGATECALL`.

marioevz (commented):
Just one comment but looks great overall, thanks!

tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py
reedsa: I notice since the DELEGATECALL doesnt touch that slot, we can assume it's zero. I want to be sure that the blockhash being zero for other cases like CALL will not give us false positives. Do you think this will be a problem?

marioevz (approved):
Looks great, thanks for the changes!

PR #1384, opened 2025-04-02
refactor(tests): Merge EOF stack underflow tests

chfast:
## 🗒️ Description
This combines the cases of `test_all_opcodes_stack_underflow()` to `test_all_opcodes_variadic_stack_underflow()`.
By the way, this also improves some details, e.g. avoids instruction sequence as `RETURN STOP`.

## ✅ Checklist

### DIFF
--- a/tests/osaka/eip7692_eof_v1/eip5450_stack/test_code_validation.py
+++ b/tests/osaka/eip7692_eof_v1/eip5450_stack/test_code_validation.py
@@ -2,7 +2,7 @@
 
 import itertools
 from enum import Enum, auto, unique
-from typing import Generator, Tuple
+from typing import Tuple
 
 import pytest
 
@@ -15,6 +15,10 @@ from ethereum_test_vm.bytecode import Bytecode
 
 from .. import EOF_FORK_NAME
 from ..eip3540_eof_v1.test_all_opcodes_in_container import valid_eof_opcodes
+from ..eip7620_eof_create.helpers import (
+    smallest_initcode_subcontainer,
+    smallest_runtime_subcontainer,
+)
 
 REFERENCE_SPEC_GIT_PATH = "EIPS/eip-5450.md"
 REFERENCE_SPEC_VERSION = "f20b164b00ae5553f7536a6d7a83a0f254455e09"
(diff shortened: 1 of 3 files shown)


tests/osaka/eip7692_eof_v1/eip5450_stack/test_code_validation.py
```diff
@@ -479,9 +488,35 @@ def test_all_opcodes_variadic_stack_underflow(
     if not op.terminating:
         code += Op.STOP
 
+    sections = [
+        Section.Code(
+            code,
+            # Set reasonable stack height. Don't rely on automatic calculation,
+            # because we are in the invalid stack height scenario.
+            # max_stack_height=max(spread, stack_height, int(spread >= 0)),
```
chfast, line 496: This is the main difference between the old and this test: if we don't set `max_stack_height` explicitly EEST will compute some value. The proposed `max(spread, stack_height, int(spread >= 0))` is a slightly complicated computation of the value. But there is no "good" value, because we are in the invalid stack height situation.

E.g. for `ADD STOP` EEST computes 2, the above is 0.

pdobacz: why is the `max_stack_height=` line commented out tho?

chfast: To bring the exact effect of the old test (removed in this PR). This is my question: should we keep it calculated by EEST, use the new formula (commented out here), or both by a test param (may be too much tests).

pdobacz: aah, I see. I think either works, leaving it to EEST is simpler so somewhat better. I think we can safely assume implementations will first detect stack underflow and not even look at max_stack_height. Third option, maybe cleanest, to just set it to 0?

chfast: That's the point. We want the max_stack_height to be as correct as possible so that the implementation don't fail on the stack height check in case the stack underflow check is incorrect. But we also report the error, so maybe this is too paranoid. I will use the new formula. With EEST I'm not sure why it calculates the values it calculates.

pdobacz: I see (again), good points :sweat_smile: . But seems to boil down to what you wrote above - there's no good value for max_stack_height, as you'd need to guess what the implementation calculates in the circumstances of stack underflow. If you guess wrong, such exception "shadowing" can still occur.

One idea I have is to approach this the following way (might be lots of work :( ): first generate programs like the ones you're using now, but without stack underflow, with a correct max_stack_height. Use these programs as sanity checks - code must validate for each case. Then provoke the underflow by an additional POP injected in front of the offending instruction. This would preserve the max_stack_height over the code section in the two variants (note that we're using the assumption that no instruction pushes 2 items onto the stack!).

PR #1390, opened 2025-04-03
fix(tests): EIP-7702: use penultimate block number instead of fixing block 0

sonhv0212:
## 🗒️ Description

In the test case that uses setcode tx to call the history contract, the block number is fixed 0, which leads to failure when running in `execute` mode because block number 0 is usually out of range for history contract queries.

This PR did a trick by using the contract code to get the parent block number instead of fixing it as 0 for the test case that set code history contract

### DIFF
--- a/tests/prague/eip7702_set_code_tx/test_set_code_txs.py
+++ b/tests/prague/eip7702_set_code_tx/test_set_code_txs.py
@@ -2656,18 +2656,32 @@ def test_set_code_to_system_contract(
             caller_payload = consolidation_request.calldata
             call_value = consolidation_request.value
         case Address(0x0000F90827F1C53A10CB7A02335B175320002935):  # EIP-2935
-            # Use the penultimate block to ensure it is saved in the history contract
-            caller_payload = Hash(max(pre.get_block_number() - 1, 0))
+            # This payload is used to identify the number of blocks to be subtracted from the
+            # latest block number
+            caller_payload = Hash(1)
             caller_code_storage[call_return_data_size_slot] = 32
         case _:
             raise ValueError(f"Not implemented system contract: {system_contract}")
 
+    # Setup the code to call the system contract
+    match system_contract:
+        case Address(0x0000F90827F1C53A10CB7A02335B175320002935):  # EIP-2935
+            # Do a trick here to get the block number of the penultimate block to ensure it is
+            # saved in the history contract
+            check_block_number = Op.SUB(Op.NUMBER, Op.CALLDATALOAD(0))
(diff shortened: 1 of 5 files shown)


sonhv0212:
cc @marioevz , seems that you're the author of this test case

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -2656,7 +2656,8 @@ def test_set_code_to_system_contract(
             caller_payload = consolidation_request.calldata
             call_value = consolidation_request.value
         case Address(0x0000F90827F1C53A10CB7A02335B175320002935):  # EIP-2935
-            caller_payload = Hash(0)
+            # Use the penultimate block to ensure it is saved in the history contract
+            caller_payload = Hash(max(pre.get_block_number() - 1, 0))
```
winsvega, line 2660: shouldn't it be in env?

sonhv0212: I think it should be in env too, but I can't find how to get block number from node without using `_eth_rpc` of `Alloc`, can you suggest a better solution?

marioevz: I agree that this should be in the `env`, and I think this requires a bit more complex solution.

We probably need to take the same approach as `pre` where it's a global fixture generated by the filler/execute plugin:

In the filler plugin, the fixture result would always be an instance of `Environment` with the default values.

In the execute plugin, the fixture is also an instance of `Environment` but the values are populated using the `return self._eth_rpc...` methods.

spencer-tb: Is this worth considering for then final prague release? https://github.com/ethereum/execution-spec-tests/issues/1508
Or should we move it to an execute backlog label?

marioevz: I did a trick for a similar test in the mainnet tests and this can be avoided:
https://github.com/ethereum/execution-spec-tests/blob/869737e9f02fb26c7ae8c5a1e4755d8866a97a1f/tests/prague/eip2935_historical_block_hashes_from_state/test_eip_mainnet.py#L24-L37

I think we can avoid the rpc and the env changes by applying a similar change in this PR @sonhv0212 .

sonhv0212: I reworked this PR based on your idea, can you review it again @marioevz

marioevz (approved):
Thanks!

PR #1394, opened 2025-04-03
refactor(tests): EIP-7002, EIP-7251: Test block invalidation if system contract is missing

marioevz:
## 🗒️ Description
Requires #1454.

Addresses following EIP changes:
- https://github.com/ethereum/EIPs/pull/9508
- https://github.com/ethereum/EIPs/pull/9582

I have a branch to fill these tests but it's currently failing, this is because we don't have a way of t8n interface to respond to this scenario, all our invalid block tests invalidate the block after a valid response from the t8n, but in this case we rely of t8n returning the exception along with the executed block somehow.

We need this issue addressed before merging these tests.

From the list of checks added to the EIP this PR checks only one at the moment:

* ✅(This PR) The call has a dedicated gas limit of `30_000_000`: `30_000_000` == valid block, `30_000_001` == invalid block.
* ✅(Existing test) Gas consumed by this call does not count against the block’s overall gas usage.
* ✅(Existing test)Both the gas limit assigned to the call and the gas consumed is excluded from any checks against the block’s gas limit.
* ✅(Existing test) The call does not follow [EIP-1559](./eip-1559.md) fee burn semantics — no value should be transferred as part of this call.
* ✅(This PR) If there is no code at `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS`, the corresponding block **MUST** be marked invalid.
* ✅(This PR) If the call to the contract fails or returns an error, the block **MUST** be invalidated.

## ✅ Checklist


winsvega:
here you hit the case where t8n must do some validation logic. I many times ask them to do so. but many blockchain logics are out of t8n scope. and asking t8n to validate something leads to people implementing additional t8n test logic in their t8n. so instead of checking the client we checking t8n's test logic. 

I would suggest we invalidate this logic/blocks on our side and don't necessarily force t8ns to do so. as t8n is used only to generate the test. the end test verification happens on hive or consume direct test runners.

src/ethereum_test_tools/utility/generators.py
```diff
@@ -148,17 +177,40 @@ def wrapper(
...
                     ),
                 ]
-            elif test_type == DeploymentTestType.DEPLOY_AFTER_FORK:
+            elif test_type == DeploymentTestType.DEPLOY_ON_FORK_BLOCK:
                 blocks = [
-                    Block(  # Empty block on fork
-                        txs=[],
+                    Block(  # Deployment on fork block
+                        txs=[deploy_tx],
                         timestamp=15_000,
```
spencer-tb, line 184: Should we make the `15_000` a constant based of the `fork`. We might need to add this in the `forks.py` logic.
Maybe a function from `class Paris` (and in base forks) called `transition_time()` that returns `15_000`. All the other forks (incl. Prague) would inherit this.

We could then use `fork.transition_time() + 1` etc

src/ethereum_test_tools/utility/generators.py
```diff
@@ -63,19 +84,22 @@ def generate_system_contract_deploy_test(
...
-    `deploy_after_fork-nonzero_balance` | after             | True        |
-    `deploy_after_fork-zero_balance`    | after             | False       |
+                                          | before/after fork | has balance | invalid block |
+    --------------------------------------|-------------------|-------------|---------------|
+    `deploy_before_fork-nonzero_balance`  | before            | True        | False         |
+    `deploy_before_fork-zero_balance`     | before            | False       | False         |
+    `deploy_on_fork_block-nonzero_balance`| on fork block     | True        | False         |
+    `deploy_on_fork_block-zero_balance`   | on fork block     | False       | False         |
+    `deploy_after_fork-nonzero_balance`   | after             | True        | True          |
+    `deploy_after_fork-zero_balance`      | after             | False       | True          |
```
spencer-tb, line 102: I think the "invalid block" column assumes `fail_on_empty_code=True`, and doesn't correctly represent what happens for EIP-2935 where `fail_on_empty_code=False`

spencer-tb (commented):
Looks good from my side so. Spent a bit of time today looking into this.

src/ethereum_test_tools/utility/generators.py
marioevz: I think it's a great idea, but might be out of scope for this PR. I can create an issue.

src/ethereum_test_tools/utility/generators.py
marioevz: Fixed the table in the last version of the PR :)

marioevz:
> here you hit the case where t8n must do some validation logic. I many times ask them to do so. but many blockchain logics are out of t8n scope. and asking t8n to validate something leads to people implementing additional t8n test logic in their t8n. so instead of checking the client we checking t8n's test logic.
> 
> I would suggest we invalidate this logic/blocks on our side and don't necessarily force t8ns to do so. as t8n is used only to generate the test. the end test verification happens on hive or consume direct test runners.

It's interesting that this is the first time we hit an issue after all txs have been executed in the t8n, and it was just a particular problem that in EELS t8n implementation, if something failed out of the scope of tx execution, the t8n would error out and not return anything, so there was no way of separate a t8n error on formulating the response from an error during block execution which was an actual protocol check.

marioevz:
PR should be ready for review @spencer-tb, I added a couple more things, including an updated `eels_resolutions.json` which points to this branch: https://github.com/marioevz/execution-specs/tree/forks/prague
Which is basically https://github.com/ethereum/execution-specs/tree/forks/prague, with https://github.com/ethereum/execution-specs/pull/1191, https://github.com/ethereum/execution-specs/pull/1183 and https://github.com/ethereum/execution-specs/pull/1182 merged on top of it.

I could create a separate PR, or we could create a merge commit to preserve the commits in the PR, which I think are cleanly separated as of now 👍

spencer-tb (approved):
LGTM!

Amazing - we just need the changelog! We should add that eels resolutions changed in this PR too within the changelog.

Issue #1176, opened 2025-02-05
bug(tooling): `uv sync` with uv>=0.5.22 modifies `uv.lock`


danceratopz:
EEST's `uv.lock` gets modified upon:
1. Upgrading/using `uv` 0.5.22 (or higher).
2.  Running `uv sync --all-extras`.

The full diff is included below - the changes seem harmless. The change from `platform_system` to `sys_platform` is not mentioned in uv's release notes, but [intentional](https://github.com/astral-sh/uv/issues/10560#issuecomment-2587450034).

Looks like we should ask everyone to upgrade their `uv` versions and bump the version in CI.

Initially reported by @Emirhan-Cavusoglu-sftw - thanks!

### Reproduce the issue
Using main@https://github.com/ethereum/execution-spec-tests/commit/1877a947f02bad7198eff01fef0ec50e8cdbe39e
```
git restore uv.lock; rm -rf .venv; rm -rf ~/.local/bin/uv; curl -LsSf https://astral.sh/uv/0.5.22/install.sh | sh
uv sync --all-extras; git status uv.lock
```
No changes to `uv.lock` with `uv==0.5.21`.

### `uv.lock` diff
```diff
diff --git a/uv.lock b/uv.lock
index 0c843773a0..5763039a09 100644
--- a/uv.lock
+++ b/uv.lock
@@ -224,7 +224,7 @@ name = "click"
 version = "8.1.7"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "colorama", marker = "platform_system == 'Windows'" },
+    { name = "colorama", marker = "sys_platform == 'win32'" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/96/d3/f04c7bfcf5c1862a2a5b845c6b2b360488cf47af55dfa79c98f6a6bf98b5/click-8.1.7.tar.gz", hash = "sha256:ca9853ad459e787e2192211578cc907e7594e294c7ccc834310722b41b9ca6de", size = 336121 }
 wheels = [
@@ -590,7 +590,7 @@ requires-dist = [
     { name = "pytest-metadata", specifier = ">=3,<4" },
     { name = "pytest-xdist", specifier = ">=3.3.1,<4" },
     { name = "pyyaml", specifier = ">=6.0.2,<7" },
[...]

PR #1396, opened 2025-04-03
refactor(exceptions,specs): Support general block exceptions - Pydantic context to parse using mapper

marioevz:
### Pydantic Mapper Parsing

This changes introduces the exception types right in the model when reading external tool responses.

Instead of parsing all exceptions to a `str`, and then using the mapper to manually parse the exception when it's time to use it, the mapper is passed as a context parameter to `model_validate` and picked up by a `BeforeValidator` defined in a field so it is automatically converted to `BlockException`, `EOFException`, `TransactionException` or `UndefinedException`.

If the field is parsed as `UndefinedException` it means that the mapper does not know about the exception and it could not be parsed.

### Mapper Optimizations/Improvements

Instead of defining a `_mapping_data` property, which must be re-computed each time the property is accessed, two new `ClassVar` have been defined and are used to parse exceptions in the mapper.

#### `mapping_substring`

Similar behavior to `_mapping_data` but in the form of a dictionary: the substring is searched within the message returned by the external tool and if found the exception is returned.

#### `mapping_regex`

Dictionary containing regex strings mapped to exceptions: `re.search` is used on the message returned by the external tool and if found the exception is returned.

### `Block.exception` Matched Against Transition Tool Response

`Block.exception` is now matched against a new optional field returned by the transition tool in the `Result` object: `blockException`.

Previously we could produced an invalid block test by one of several methods:
- An invalid transaction included in the block, detected in the `Result.rejected_transactions` field of the returned object by the transition tool.
- Mock a block header or body field by sending a correct set of transactions to the transition tool and then modifying the field in the response before exporting it to the fixture.

None of these methods required the transition tool to inform us that a block was invalid, but with new EIP changes like:
https://github.com/ethereum/EIPs/pull/9508
https://github.com/ethereum/EIPs/pull/9582

There's a block failure type that we cannot mock in the tests: system contract failures.

This new field is used the tests for both of these EIP changes: https://github.com/ethereum/execution-spec-tests/pull/1394

### `reliable` Class Variable in the Exception Mapper

Added a new property to the exception mappers that when set to `False` means that the exception mapper cannot reliably discriminate between types of exceptions.

This needs to be set back to `True` once the EELS's t8n returns different exceptions for each type.

### Global `default_t8n` Fixture for Unit Tests

New global fixture that contains an already-running instance of the default t8n tool, which significantly speeds up unit test runs.

## ✅ Checklist


marioevz:
Hey @gurukamath, I have a branch for the `blockException` field returned by EELS's T8N here: https://github.com/marioevz/execution-specs/commit/d615cb57fc8d662bcd30eb8242a8e1d890ee0a1f

Wanted to ask your opinion on this and whether you think it's a good idea to have this.

gurukamath:
> Wanted to ask your opinion on this and whether you think it's a good idea to have this.

Is there any particular scenario you have in mind? Because the system transactions are supposed to fail silently iirc, so I am not sure you would expect to catch anything here.

marioevz:
> > Wanted to ask your opinion on this and whether you think it's a good idea to have this.
> 
> Is there any particular scenario you have in mind? Because the system transactions are supposed to fail silently iirc, so I am not sure you would expect to catch anything here.

Yes that was the way they worked but with https://github.com/ethereum/EIPs/pull/9508 and https://github.com/ethereum/EIPs/pull/9582 they are no longer supposed to fail silently, so instead the block should be invalid, even when the call does not fail but instead the address contains no code.

The problem as I could debug it is that T8N throws at some point when the block is invalid, and a zero length response is returned instead of the usual JSON response, so we cannot parse anything.

The branch I pointed out uses a try-except block to catch the exception, and when caught it populates a new field in the t8n response to signal a block-wide issue, which is what we parse in this PR. Please feel free to suggest anything else since I'm not super familiar with T8N in EELS so there's potentially a better solution 👍

gurukamath:
Ah. I see. In that case this sounds reasonable to me.

tests/cancun/eip4844_blobs/test_blob_txs.py
```diff
@@ -1020,6 +1020,7 @@ def test_invalid_blob_tx_contract_creation(
                     TransactionException.TYPE_3_TX_CONTRACT_CREATION,
                 ],
                 header_verify=header_verify,
+                skip_exception_verification=True,
```
winsvega, line 1023: why?

are you making an invalid rlp type of test? 
this is a special group of tests. we must think about it

winsvega:
looks like a good approach. 
can you add a unit test where
1) transaction expected exception but it didn't happen
2) transaction expected exception but another exception happen
3) transaction expected exception and it happen, so the validation is green
4) transaction didn't expect exception but it happen

as well as the error strings in case.
there was a helper message in exception. that error string is not defined in mapper. exception not found. 
if error string is found it was also fetching the exception CODE_NAME  (so you know what type of exception you hit)
and if excpected exception CODE_NAME is not defined in mapper it was also hinting where in code you should define the error string. 

just check that this hint logic is not affected.

src/ethereum_test_exceptions/exception_mapper.py
```diff
@@ -1,19 +1,12 @@
 """EEST Exception mapper."""
 
-from abc import ABC, abstractmethod
-from dataclasses import dataclass
+import re
+from abc import ABC
+from typing import Any, ClassVar, Dict, Generic
 
-from bidict import frozenbidict
```
danceratopz, line 6: Nice :) I removed `bidict` as an eest dependency directly in a commit to this PR. Not sure why this is causing CI to fail yet, will look into it.

src/ethereum_test_specs/helpers.py
```diff
@@ -1,71 +1,93 @@
...
+    ExceptionBase,
+    ExceptionWithMessage,
+    TransactionException,
+    UndefinedException,
+)
 from ethereum_test_types import Transaction, TransactionReceipt
 
 
-class TransactionUnexpectedSuccessError(Exception):
+class UnexpectedSuccessError(Exception):
```
danceratopz, line 19: I understand the exception rename, but I think the new name is a little too generic. Made a suggestion in #1404.

src/ethereum_test_specs/helpers.py
```diff
@@ -1,71 +1,93 @@
...
     """Exception used when the transaction expected to fail succeeded instead."""
 
-    def __init__(self, index: int, nonce: int):
-        """Initialize the exception with the transaction index and nonce."""
+    def __init__(self, ty: Literal["Block", "Transaction"], **kwargs):
```
danceratopz, line 22: `ty` is a little too skimpy for my liking, how do you like `execution_context`? I also made a suggestion for this in #1404.

tests/cancun/eip4844_blobs/test_blob_txs.py
danceratopz: Could you add a comment to explain why? :pray:

danceratopz (commented):
Really nice refactor! This is a much nicer approach. Just a couple of comments below, respectively in #1404.

src/ethereum_test_exceptions/exception_mapper.py
danceratopz: The CI fail is definitely a fake positive and reminds me of a `uv` bug we once had that rewrote the lock file #1176. I updated the `uv` version in CI to 0.6.12, perhaps that fixes it in a subsequent run.

tests/cancun/eip4844_blobs/test_blob_txs.py
marioevz: This is actually what we are already doing for EELS filled tests: all exceptions in the mapper there are variations of "Invalid transaction", trimmed to different lengths, so it's ineffective at the moment.

I think we should open a issue to remove this, and another one to EELS to add strings to each transaction exception, it should not be that many.

marioevz:
> looks like a good approach. can you add a unit test where
> 
>     1. transaction expected exception but it didn't happen
> 
>     2. transaction expected exception but another exception happen
> 
>     3. transaction expected exception and it happen, so the validation is green
> 
>     4. transaction didn't expect exception but it happen
> 
> 
> as well as the error strings in case. there was a helper message in exception. that error string is not defined in mapper. exception not found. if error string is found it was also fetching the exception CODE_NAME (so you know what type of exception you hit) and if excpected exception CODE_NAME is not defined in mapper it was also hinting where in code you should define the error string.
> 
> just check that this hint logic is not affected.

Thanks! This unit test actually already existed here but it contained an `xfail` for it to pass:
https://github.com/ethereum/execution-spec-tests/blob/58a2ec149e379e2dfc06d84cc6e1e4651192360d/src/ethereum_test_specs/tests/test_expect.py#L261-L354

And actually the test was in correct but I fixed in this  PR.

Also it was extremely slow, and it turns out that every time we do `ExecutionSpecsTransitionTool()` (instantiating the t8n for an unit test) it took many seconds, and in this test it was doing it once per parametrized value.

So I implemented the `default_t8n` fixture globally and sped up this and many other tests!

@danceratopz I added the comment you requested in the blobs test 👍

src/ethereum_test_exceptions/exception_mapper.py
marioevz: Failures seem to be legitimate discrepancies between the expected exception and the actual exception received.

I think it should be message discrepancies, but will check and update.

danceratopz: This is a nice catch! We were looking at different fails :)

At the point I reviewed the PR, the coverage script was failing at a [later step ](https://github.com/ethereum/execution-spec-tests/actions/runs/14302920742/job/40094387104). 

> The CI fail is definitely a fake positive and reminds me of a uv bug we once had that rewrote the lock file https://github.com/ethereum/execution-spec-tests/issues/1176. I updated the uv version in CI to 0.6.12, perhaps that fixes it in a subsequent run.

Debugging locally via nectors/act, I saw that this fail was indeed due to #1176, the coverage workflow wasn't still using `astral-sh/setup-uv` but rather our local `setup-uv` action which was using an older version of `uv` that would still trigger #1176. This is addressed in #1409, which will fix this CI fail in this PR, which will reappear the coverage check itself passes.

src/ethereum_test_specs/helpers.py
```diff
@@ -91,63 +122,112 @@ def __init__(
...
+        want_exception, got_exception = (
+            self.want_exception,
+            self.got_exception,
+        )
+        if want_exception and not got_exception:
+            raise UnexpectedExecutionSuccessError(
+                execution_context=self.execution_context, **self.context
+            )
+        elif not want_exception and got_exception:
+            assert self.got_message is not None
```
danceratopz, line 171: I think this will be the case if `isinstance(got_exception, UndefinedException)`? 

If that's the case, as `UnexpectedExecutionFailError` is probably the most interesting aspect of the fail and we should avoid raising an (unhelpful) assertion error before reporting it?

src/ethereum_test_specs/helpers.py
```diff
@@ -1,71 +1,102 @@
...
+            f"\n   What: {execution_context.value} unexpectedly failed!"
             f'\n  Error: "{message}" ({exception})'
         )
         super().__init__(message)
 
 
-class TransactionExceptionMismatchError(Exception):
-    """Exception used when the actual transaction error string differs from the expected one."""
+class UndefinedExecutionExceptionError(Exception):
+    """Exception used when the exception is undefined."""
```
danceratopz, line 59:
```suggestion
    """Exception used when a client's exception message isn't present in its `ExceptionMapper`."""
```

danceratopz (commented):
Looks good, just one small question and a possible improvement to `src/ethereum/test_spec/helpers.py::verify()`.

src/ethereum_test_specs/helpers.py
marioevz: This is covered by:
```python
        if self.got_exception is None:
            ...
        else:
            self.got_message = (
                got_exception.message
                if isinstance(got_exception, ExceptionWithMessage)
                else str(got_exception)
            )
```
`self.got_exception` is not `None`, and also it's not `ExceptionWithMessage`, so it will be a `UndefinedException` converted to `str`.

danceratopz (approved):
LGTM!

PR #1286, opened 2025-03-06
bug(fill,types): duplicate `FixtureAuthorizationTuple`'s `v` field as `yParity`

danceratopz:
## 🗒️ Description
Unfortunately, the `yParity` field in authorization lists was named `v` in our state test fixture format, which differs from from EIP-7702:
https://github.com/ethereum/EIPs/blob/aa391ae306408fcd049a1442a5deab47675c0c87/EIPS/eip-7702.md?plain=1#L45

This change will enable erigon to correctly run the 7702 tests (with non-zero `yParity` in authorization lists, this default to `0` as it wasn't found).

Thanks to @holiman for detecting this issue.

## 🔗 Related Issues
- https://github.com/erigontech/erigon/issues/14073
- https://github.com/bluealloy/revm/pull/2150

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).

PR #1296, opened 2025-03-11
chore(ci): use manual x-dist value for self-hosted runners

spencer-tb:
## 🗒️ Description
To speed up the release process I've added a new user to the hive server called `usainbolt`. This user will be used entirely as a self-hosted runner set for the [Build and Package Fixtures](https://github.com/spencer-tb/execution-spec-tests/actions/workflows/fixtures.yaml) workflow step. Note this is only runs when creating new releases.

Currently our releases take a long time finish as fixture builds run sequentially, due to our single self-hosted runner instance. Everything in total currently takes around 5hrs. I've now added 6 self hosted runners.
Screenshot 2025-03-11 at 11 49 51

This will take our entire release process time down to <3hrs. Note we could improve this by finding a way to fill the `slow` marked fixtures faster, ideally giving us a release build time of <30mins.

### Other Changes

- Removed `pectra-devnet-6` as a separate build, Prague tests are within develop now.
- Set `x-dist` to 10 instead of auto. With the self-hosted runners when using auto it will select `n=32` for each build. This will ultimately throttle the EELS resolver and give us timeout errors when filing.

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/pull/1051

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.

PR #1398, opened 2025-04-04
🐞 fix(types): Add missing yParity in Authorization tuple

raxhvl:
## 🗒️ Description
EIP-7702 specs requires the signature component `v` to be called `yParity`.  #1286 addressed this issue for fixtures, this PR extends the fix for transaction payload, by moving the code to the parent class `AuthorizationTupleGeneric`.

## 🔗 Related Issues
#1286 

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_types/types.py
+++ b/src/ethereum_test_types/types.py
@@ -461,6 +461,18 @@ class AuthorizationTupleGeneric(CamelModel, Generic[NumberBoundTypeVar], Signabl
         """
         return self.magic.to_bytes(1, byteorder="big")
 
+    @model_serializer(mode="wrap", when_used="json-unless-none")
+    def duplicate_v_as_y_parity(self, serializer):
+        """
+        Add a duplicate 'yParity' field (same as `v`) in JSON fixtures.
+
+        Background: https://github.com/erigontech/erigon/issues/14073
+        """
+        data = serializer(self)
+        if "v" in data and data["v"] is not None:
+            data["yParity"] = data["v"]
+        return data
+
 
 class AuthorizationTuple(AuthorizationTupleGeneric[HexNumber]):
     """Authorization tuple for transactions."""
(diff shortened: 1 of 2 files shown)


src/ethereum_test_types/types.py
```diff
@@ -467,11 +467,14 @@ class AuthorizationTuple(AuthorizationTupleGeneric[HexNumber]):
 
     signer: EOA | None = None
     secret_key: Hash | None = None
+    yParity: HexNumber | None = None  # noqa: N815
```
marioevz, line 470: I would not add a new field, because these can be problematic sometimes.

I think we should move the `model_serializer` that #1286 added from `FixtureAuthorizationTuple` into `AuthorizationTupleGeneric`, because since now we need the same behavior in `FixtureAuthorizationTuple` and `AuthorizationTuple`, and both inherit `AuthorizationTupleGeneric`.

raxhvl:
Hey @marioevz, I moved the code to the parent class.

winsvega:
so now we gonna have an excessive dublicate field in all of the fixtures, because one client refuses to follow the format?

danceratopz:
> @marioevz so now we gonna have an excessive duplicate field in all of the fixtures, because one client refuses to follow the format?

@winsvega To be fair, the mistake was on our side. We mistakenly named the `yParity` field `v` in our initial Prague devnet fixture releases. We added the duplicate field in #1296 to address the mistake, yet maintain backwards compatibility for other clients. We could consider removing `v` and informing clients, but it feels a bit risky.

danceratopz (commented):
This makes sense to me. The changes to #1286 considered the change only necessary for fixtures; internally it can stay as-is, which was incorrect. Is this required for `execute`?

winsvega:
I remember it was always v in the tests.

marioevz (approved):
Fantastic, just tested out locally and works like a charm. Thanks!

marioevz:
> > @marioevz so now we gonna have an excessive duplicate field in all of the fixtures, because one client refuses to follow the format?
> 
> @winsvega To be fair, the mistake was on our side. We mistakenly named the `yParity` field `v` in our initial Prague devnet fixture releases. We added the duplicate field in #1296 to address the mistake, yet maintain backwards compatibility for other clients. We could consider removing `v` and informing clients, but it feels a bit risky.

Only for state test fixtures, and only for the ones using type-4 transactions, so a small percentage.

PR #1403, opened 2025-04-07
fix(tests): EOF - Change container section size to 4 bytes

gumb0:
## ✅ Checklist


pdobacz (approved):
thinking out loud - is there a container which can be parsed valid in both old and new rules? if yes, would be nice to have it here and ensure that it is parsed "the right way", i.e. with 4-byte container size.

gumb0:
> thinking out loud - is there a container which can be parsed valid in both old and new rules? if yes, would be nice to have it here and ensure that it is parsed "the right way", i.e. with 4-byte container size.

Valid in new rules will always fail with "container with 0 size" in old rules.

tests/osaka/eip7692_eof_v1/eip3540_eof_v1/test_container_validation.py
```diff
@@ -348,22 +348,22 @@ def test_valid_containers(
         ),
         Container(
             name="zero_size_container_section",
-            raw_bytes="ef00 01 01 0004 02 0001 0001 03 0001 0000 ff 0000 00 00800000 00",
+            raw_bytes="ef00 01 01 0004 02 0001 0001 03 0001 00000000 ff 0000 00 00800000 00",
```
chfast, line 351: I think you should add 2 more tests as the above: for the incomplete container sections size.

pdobacz:
> > thinking out loud - is there a container which can be parsed valid in both old and new rules? if yes, would be nice to have it here and ensure that it is parsed "the right way", i.e. with 4-byte container size.
> 
> Valid in new rules will always fail with "container with 0 size" in old rules.

ok, and valid in old rules must unconditionally be too large (leading 15 (?) for the smallest container) in new rules.

OK that sounds good, just ensure we have some "valid, by old rules" containers in the tests, I don't think we have them?

pdobacz (approved):

chfast (approved):

PR #1404, opened 2025-04-07
Suggestions for #1396, refactor(exceptions,specs): Support general block exceptions - Pydantic context to parse using mapper

danceratopz:
## 🗒️ Description
Couple of suggestions for `src/ethereum_test_specs/helpers.py` in #1396:
1. Define an enum class instead of using string literal for the execution context of the exception.
2. Rename `ty` to `execution_context`.
3. Add more specific exception class names.

## 🔗 Related Issues
#1396


marioevz (approved):
Awesome, thanks!

PR #1407, opened 2025-04-07
feat(fill,execute): add new `zkevm` marker

marioevz:
## 🗒️ Description
Adds a marker that we can apply to tests that are relevant to zkEVM by using:

```python
@pytest.mark.zkevm
```

We can run fill for every fork but only tests that use this marker, e.g.:
```
uv run fill --from=Cancun --until=Osaka -m zkevm
```

@kevaundray

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/shared/execute_fill.py
+++ b/src/pytest_plugins/shared/execute_fill.py
@@ -81,6 +81,10 @@ def pytest_configure(config: pytest.Config):
         "markers",
         "execute: Markers to be added in execute mode only.",
     )
+    config.addinivalue_line(
+        "markers",
+        "zkvm: Tests that are relevant to zkVM.",
+    )
 
 
 @pytest.fixture(autouse=True)
--- a/src/pytest_plugins/shared/execute_fill.py
+++ b/src/pytest_plugins/shared/execute_fill.py
@@ -83,7 +83,7 @@ def pytest_configure(config: pytest.Config):
     )
     config.addinivalue_line(
         "markers",
-        "zkvm: Tests that are relevant to zkVM.",
+        "zkevm: Tests that are relevant to zkEVM.",
     )
 
 


marioevz:
@danceratopz Kev also asked me if we export the markers to the filled fixtures, and I'm not sure if we have this feature?

I think we could add markers to the fixtures and the index, it would be very helpful.

src/pytest_plugins/shared/execute_fill.py
```diff
@@ -81,6 +81,10 @@ def pytest_configure(config: pytest.Config):
         "markers",
         "execute: Markers to be added in execute mode only.",
     )
+    config.addinivalue_line(
+        "markers",
+        "zkvm: Tests that are relevant to zkVM.",
```
danceratopz, line 86: @kevaundray in the context of EVM tests, I would have expected a `zkevm` marker. Any strong feelings or reason for `zkvm`? :slightly_smiling_face:  
```suggestion
        "zkevm: Tests that are relevant to zkEVM.",
```

danceratopz (commented):
Nice! Sorry for the pedantic question below @kevaundray! 

About propagating the markers...

> @danceratopz Kev also asked me if we export the markers to the filled fixtures, and I'm not sure if we have this feature?
> 
> I think we could add markers to the fixtures and the index, it would be very helpful.

We don't, but I'd love to add this! It will add quite a lot of data to the `_info` section (and the index), but it'll be well worth it imo. Having a generic solution would solve:
- https://github.com/ethereum/execution-spec-tests/issues/1363

src/pytest_plugins/shared/execute_fill.py
kevaundray: Hey :) zkEVM would be the more correct/specific term

danceratopz (approved):
lgtm

PR #1411, opened 2025-04-08
fix(tests): EIP-7702: send tx of eoa after setcode tx is mined

sonhv0212:
## 🗒️ Description

Currently, the test `test_eoa_tx_after_set_code` is including both setcode tx that sets code for an EOA and tx using EOA as sender in same block. However, it is invalid due to changes introduced in this PR https://github.com/ethereum/go-ethereum/pull/31430. When I try to run this test in `execute` mode, the second tx is rejected with the error `gapped-nonce tx from delegated accounts`

Here is my solution to fix it:
- I split txs into two blocks; the first includes setcode tx that set code for an EOA, and the second is the tx from that EOA
- We need to ensure txs of two different blocks must be included in different block as their order in node, so I refactored `TransactionPost` a little bit to send batches of txs by block.

### DIFF
--- a/src/ethereum_test_execution/transaction_post.py
+++ b/src/ethereum_test_execution/transaction_post.py
@@ -14,7 +14,7 @@ from .base import BaseExecute
 class TransactionPost(BaseExecute):
     """Represents a simple transaction-send then post-check execution format."""
 
-    transactions: List[Transaction]
+    blocks: List[List[Transaction]]
     post: Alloc
 
     format_name: ClassVar[str] = "transaction_post"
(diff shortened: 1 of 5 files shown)


sonhv0212:
cc @marioevz , seems that you're the author of this test case. Can you help me verify this?
Thank you!!! 😄

marioevz (approved):
LGTM, thanks for these changes, I think it's a good idea to make this distinction.

I added a parametrization to fill the same test as it existed before this PR, whilst skipping it for execute, and the new variant you introduced.

PR #1416, opened 2025-04-09
feat(consume): add `ExceptionMapper` support to the consume-engine simulator

marioevz:
## 🗒️ Description

Adds support of the exception mapper to consume-engine:

`consume engine` now checks exceptions returned by the execution clients in their Engine API responses, specifically in the `validationError`field of the `engine_newPayloadVX` method.

Exception mappers of all current mainnet clients have been added to the `ethereum_clis` library.

The feature can be disabled by using `--disable-strict-exception-matching` for specific clients or forks.

## ✅ Checklist

### DIFF
--- a/src/ethereum_clis/clis/ethereumjs.py
+++ b/src/ethereum_clis/clis/ethereumjs.py
@@ -71,6 +71,9 @@ class EthereumJSExceptionMapper(ExceptionMapper):
         TransactionException.TYPE_3_TX_BLOB_COUNT_EXCEEDED: "exceed maximum allowance",
         TransactionException.TYPE_3_TX_ZERO_BLOBS: "tx should contain at least one blob",
         TransactionException.TYPE_3_TX_WITH_FULL_BLOBS: "Invalid EIP-4844 transaction",
+        TransactionException.TYPE_4_TX_CONTRACT_CREATION: (
+            'tx should have a "to" field and cannot be used to create contracts'
+        ),
         TransactionException.TYPE_4_EMPTY_AUTHORIZATION_LIST: (
             "Invalid EIP-7702 transaction: authorization list is empty"
         ),
--- a/src/ethereum_clis/clis/ethereumjs.py
+++ b/src/ethereum_clis/clis/ethereumjs.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import ClassVar, Dict, Optional
 
 from ethereum_test_exceptions import (
+    BlockException,
     EOFException,
     ExceptionBase,
     ExceptionMapper,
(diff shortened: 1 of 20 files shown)


src/ethereum_clis/clis/ethereumjs.py
```diff
@@ -48,30 +49,47 @@ class EthereumJSExceptionMapper(ExceptionMapper):
...
+        ),
+        TransactionException.TYPE_4_EMPTY_AUTHORIZATION_LIST: (
+            "Invalid EIP-7702 transaction: authorization list is empty"
+        ),
         TransactionException.INTRINSIC_GAS_TOO_LOW: "is lower than the minimum gas limit of",
-        TransactionException.INITCODE_SIZE_EXCEEDED: "max initcode size exceeded",
+        TransactionException.INITCODE_SIZE_EXCEEDED: (
+            "the initcode size of this transaction is too large"
+        ),
+        TransactionException.TYPE_4_TX_CONTRACT_CREATION: (
```
jochem-brouwer, line 84: This key is two times in this list? How does this work with mappings? What if the same keys have different values, does it throw? :thinking:

winsvega:
I found there are some issue with refactored exception mapper. perhaps this can be address in here in this PR
https://github.com/ethereum/execution-spec-tests/issues/1440

src/ethereum_test_exceptions/exception_mapper.py
```diff
@@ -48,14 +48,19 @@ def __init__(self) -> None:
...
         }
 
-    def message_to_exception(self, exception_string: str) -> ExceptionBase | UndefinedException:
+    def message_to_exception(
+        self, exception_string: str
+    ) -> List[ExceptionBase] | UndefinedException:
         """Match a formatted string to an exception."""
+        exceptions: List[ExceptionBase] = []
         for exception, substring in self.mapping_substring.items():
             if substring in exception_string:
```
winsvega, line 57: if there are now regex strings should be accounted here?

src/ethereum_test_exceptions/exception_mapper.py
```diff
@@ -76,16 +91,19 @@ def mapper_validator(v: str, info: ValidationInfo) -> Dict[str, Any] | Undefined
     """
     if v is None:
         return v
-    assert isinstance(info.context, dict), f"Invalid context provided: {info.context}"
+    if not isinstance(info.context, dict):
+        return UndefinedException(v, mapper_name="UndefinedExceptionMapper: No context")
     exception_mapper = info.context.get("exception_mapper")
+    if exception_mapper is None:
```
winsvega, line 97: exception_mapper is a property of exception? 
where it supposed to be set?

src/pytest_plugins/consume/hive_simulators/conftest.py
```diff
@@ -40,6 +42,16 @@ def pytest_addoption(parser):
         default=False,
         help="Log the timing data for each test case execution.",
     )
+    consume_group.addoption(
+        "--strict-exception-matching",
```
winsvega, line 46: what is the difference if it is not set to strict?

winsvega (commented):
waut the complexity of exception mapper is growing. I need to study python again.

src/ethereum_clis/clis/ethereumjs.py
marioevz: It was a mistake, I only caught it when it started failing in hive.

Unfortunately python does not care about this:
```
>>> {'a': 'a', 'a': 'b'}
{'a': 'b'}
```
So we should be careful when defining these exceptions to not overwrite.

src/ethereum_test_exceptions/exception_mapper.py
marioevz: Yes in line 56 we go over all regex patterns to try to match any.

src/ethereum_test_exceptions/exception_mapper.py
marioevz: It's part only of `ExceptionWithMessage` which is only used by pydantic on types where we parse an exception from t8n or the client itself.

But the types are currently defined where they are used, see e.g.: https://github.com/ethereum/execution-spec-tests/blob/54275170699538a74729197bffd1b05d6ea97208/src/ethereum_test_rpc/types.py#L89-L94

Perhaps we should bring all these types to this file in order to, 1) not redefine, 2) make usage clearer when viewing this file.

src/pytest_plugins/consume/hive_simulators/conftest.py
marioevz: Yeah, I changed this to `--disable-strict-exception-matching` for it to be enabled by default and in case we need to turn it off for some client we can do so, but ideally we should simply fix the exception messages if that's the case.

src/ethereum_test_exceptions/exception_mapper.py
winsvega: It's confusing. Message is not part of an exception.

There is ExceptionCode
And error message returned by the tool.
And test validator compares exception code against error message returned by the tool using exception mapper

src/pytest_plugins/consume/hive_simulators/conftest.py
marioevz: Merged that one :)

src/pytest_plugins/consume/hive_simulators/conftest.py
```diff
@@ -196,6 +208,33 @@ def buffered_genesis(client_genesis: dict) -> io.BufferedReader:
...
 
+@pytest.fixture(scope="function")
+def client_exception_mapper(
+    client_type: ClientType,
+) -> ExceptionMapper | None:
+    """Return the exception mapper for the client type."""
+    for client in EXCEPTION_MAPPERS:
+        if client in client_type.name:
+            return EXCEPTION_MAPPERS[client]
+    return None
```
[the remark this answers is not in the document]
marioevz: Great optimization IMO :D 

It would be so great if we could scope to `module` where each module is either a test suite or test client, but it's too much pytest hacking.

src/ethereum_test_exceptions/exceptions.py
```diff
@@ -381,6 +383,10 @@ class TransactionException(ExceptionBase):
     """
     Transaction is type 4, but contains an authorization that has an invalid format.
     """
+    TYPE_4_TX_PRE_FORK = auto()
+    """
+    Transaction type 4 included before activation fork.
+    """
```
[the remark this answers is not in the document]
marioevz: And it's already merged :)

src/pytest_plugins/consume/hive_simulators/engine/conftest.py
```diff
@@ -20,8 +21,15 @@ def pytest_configure(config):
...
 @pytest.fixture(scope="function")
-def engine_rpc(client: Client) -> EngineRPC:
+def engine_rpc(client: Client, client_exception_mapper: ExceptionMapper | None) -> EngineRPC:
     """Initialize engine RPC client for the execution client under test."""
+    if client_exception_mapper:
+        return EngineRPC(
+            f"http://{client.ip}:8551",
+            response_validation_context={
+                "exception_mapper": client_exception_mapper,
+            },
```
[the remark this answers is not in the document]
marioevz: Pydantic contexts are too powerful 🔥

danceratopz (approved):
LGTM!

src/ethereum_clis/clis/ethereumjs.py
marioevz: I created a tracking issue for this because there should definitely be a way to disallow this.

Potentially create exception mapper instances instead of subclasses.

src/pytest_plugins/consume/hive_simulators/engine/test_via_engine.py
```diff
@@ -86,6 +88,38 @@ def test_blockchain_via_engine(
                                 "Client failed to raise expected Engine API error code: "
                                 f"{payload.error_code}"
                             )
+                        elif payload_response.status == PayloadStatusEnum.INVALID:
+                            if payload_response.validation_error is None:
+                                raise Exception(
+                                    "Client returned INVALID but no validation error was provided."
+                                )
```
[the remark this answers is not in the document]
marioevz: Refactored this to include a class named `LoggedError` that automatically calls `logger.fail` on init, and is now used throughout the file.

PR #1425, opened 2025-04-09
chore(pytest): temporary consume engine tweak for nimbul-el

spencer-tb:
## 🗒️ Description
Currently `consume-engine` is not compatible with `nimbus-el`. These are due to the following points:

1) For each `forkchoiceUpdate` and `newPayload` request we send, each response should contain the [`PayloadStatusV1`](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#payloadstatusv1) within it. For nimbus `PayloadStatusV1` is missing the `validationError` field for VALID responses (note it exists for INVALID).

2) For nimbus the `forkchoiceUpdate` response doesn't include the [`payloadId`](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#response-1). 

Our pydantic models error out due to the latter. This PR adds a temporary workaround that allows us to run `consume-engine` for nimbus, without breaking other clients.

## 🔗 Related Issues
N/A

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_rpc/types.py
+++ b/src/ethereum_test_rpc/types.py
@@ -1,7 +1,7 @@
 """Types used in the RPC module for `eth` and `engine` namespaces' requests."""
 
 from enum import Enum
-from typing import Any, List
+from typing import Any, List, Optional
 
 from pydantic import AliasChoices, Field, model_validator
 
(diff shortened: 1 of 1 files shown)


spencer-tb:
Assuming this gets approved I will create an issue to revert this change once nimbus updates.

src/ethereum_test_rpc/types.py
```diff
@@ -84,14 +84,40 @@ class PayloadStatus(CamelModel):
 
     status: PayloadStatusEnum
     latest_valid_hash: Hash | None
-    validation_error: str | None
+    # TODO: Temporarily optional to allow nimbus-el to run `consume-engine`
+    validation_error: Optional[str] = None
```
marioevz, line 88: Was there an error with `str | None = None`?

src/ethereum_test_rpc/types.py
```diff
@@ -84,14 +84,40 @@ class PayloadStatus(CamelModel):
...
-    payload_id: Bytes | None
+    # TODO: Temporarily optional to allow nimbus-el to run `consume-engine`
+    payload_id: Optional[Bytes] = None
+
+    # TODO: Temporarily override default pydantic validation for `validation_error` via this class
+    # Enforces that `validation_error` is required when status is INVALID or INVALID_BLOCK_HASH
+    # Allows us to run `consume-engine` for nimbus-el.
+    @model_validator(mode="before")
+    @classmethod
+    def ensure_validation_error(cls, data):
```
marioevz, line 103: Couldn't this validator be added to `PayloadStatus`?

marioevz (commented):
I'm hesitant to add this check to all clients since it is out of spec.

I would incline to add a context-sensitive validator that checks if the client is the nimbus-el, and only then allow this out-of-spec behavior.

I added context to the rpc response validation (for exceptions not client-specific validations) in this PR: https://github.com/ethereum/execution-spec-tests/pull/1416/files

We could cherry-pick the context changes into this PR and then apply these relaxed validation only to nimbus-el.

src/ethereum_test_rpc/types.py
spencer-tb: Ahh will try this

src/ethereum_test_rpc/types.py
spencer-tb: Let me try this as well thanks

PR #1438, opened 2025-04-11
chore(forks): osaka tweak and fix blob schedule logic

spencer-tb:
## 🗒️ Description
Updates the logic in forks, such that we always include the blob schedule within the fixtures for any new fork.

Additionally removes `CancunEIP7692`.

## 🔗 Related Issues
N/A.

## ✅ Checklist

- [x] ~~All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~


src/ethereum_test_forks/forks/forks.py
```diff
@@ -975,7 +975,7 @@ def blob_schedule(cls, block_number: int = 0, timestamp: int = 0) -> BlobSchedul
             max_blobs_per_block=cls.max_blobs_per_block(block_number, timestamp),
             base_fee_update_fraction=cls.blob_base_fee_update_fraction(block_number, timestamp),
         )
-        if last_blob_schedule is None or last_blob_schedule != current_blob_schedule:
+        if parent_fork.name() != cls.name() or last_blob_schedule != current_blob_schedule:
```
marioevz, line 978: If the goal is to add an entry to every fork, even if there's no blob changes between previous fork and current one, then we can remove a lot of the logic, but we also should add a test to https://github.com/ethereum/execution-spec-tests/blob/main/src/ethereum_test_forks/tests/test_forks.py, which should verify that the entry is added to the fork even when it has no blob_schedule changes:
```python
class NewTestFork(Prague):
    ...

assert NewTestFork.blob_schedule() == {"Cancun": ..., "Prague": ..., "NewTestFork": ...}
```

marioevz (commented):
Just one comment, thanks!

marioevz (approved):
LGTM.

Simplified the logic and added a new unit test in the last commit.

PR #1439, opened 2025-04-11
feat(static tests) : do not compile test fillers source code while loading the models

winsvega:
@marioevz  please don't merge there are bullet points to complete:

- [x] When reading Filler.json into pydantic do not compile test source right away. this speeds up test generation
- [x] Support lllc container docker image, so no need to have lllc compiled locally. can import from docker hub: winsvega/lllc:latest
- [x] After all tests generated or on exception fail stop, need to call stop_lllc_containers() function 

_- [x] Make sure exceptions are working correctly after mapper refactoring (too many things to fix)_ https://github.com/ethereum/execution-spec-tests/issues/1440

the test generation is supported with lllc from docker hub. 
https://hub.docker.com/r/winsvega/lllc

it works fast using docker exec. all we need is to pull a docker image in our CI

### DIFF
--- a/src/ethereum_test_specs/static_state/common/common.py
+++ b/src/ethereum_test_specs/static_state/common/common.py
@@ -138,7 +138,7 @@ class CodeInFillerSource:
                 return function_signature + function_parameters
             return function_signature
 
-        # Prase plain code 0x
+        # Parse plain code 0x
         elif self.code_raw.lstrip().startswith("0x"):
             compiled_code = self.code_raw[2:].lower()
 
--- a/src/ethereum_test_specs/static_state/common/common.py
+++ b/src/ethereum_test_specs/static_state/common/common.py
@@ -14,6 +14,7 @@ from typing_extensions import Annotated
 from ethereum_test_base_types import Address, Hash, HexNumber
 
 from .compile_yul import compile_yul
+from .docker import get_lllc_container_id
 
 
 def parse_hex_number(i: str | int) -> int:
@@ -144,11 +145,18 @@ class CodeInFillerSource:
 
         # Prase lllc code
         elif self.code_raw.lstrip().startswith("{") or self.code_raw.lstrip().startswith("(asm"):
-            binary_path = "lllc"
             with tempfile.NamedTemporaryFile(mode="w+", delete=False) as tmp:
                 tmp.write(self.code_raw)
                 tmp_path = tmp.name
-            result = subprocess.run([binary_path, tmp_path], capture_output=True, text=True)
+
+            # using lllc
+            # result = subprocess.run(["lllc", tmp_path], capture_output=True, text=True)
+
+            result = subprocess.run(
+                ["docker", "exec", get_lllc_container_id(), "lllc", tmp_path[5:]],
+                capture_output=True,
+                text=True,
+            )
             compiled_code = "".join(result.stdout.splitlines())
         else:
             raise Exception(f'Error parsing code: "{self.code_raw}"')
--- a/src/ethereum_test_specs/static_state/common/common.py
+++ b/src/ethereum_test_specs/static_state/common/common.py
@@ -14,7 +14,6 @@ from typing_extensions import Annotated
 from ethereum_test_base_types import Address, Hash, HexNumber
 
 from .compile_yul import compile_yul
-from .docker import get_lllc_container_id
 
 
 def parse_hex_number(i: str | int) -> int:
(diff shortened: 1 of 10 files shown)


winsvega:
restored this logic
https://github.com/ethereum/execution-spec-tests/pull/1362/commits/a092ecb07d2d79c58d73799c3f681438da18f507

winsvega:
@marioevz 

```
execution-spec-tests/src/ethereum_test_specs/static_state/common/docker.py
def stop_lllc_containers():
```

needs to be called after filling of the fixtures has finished. or in case of exception and we are exiting. 
where this part of logic is defined in pytest plugin of static test filler?

src/ethereum_test_specs/static_state/common/common.py
```diff
@@ -53,116 +48,142 @@ def parse_args_from_string_into_array(stream: str, pos: int, delim: str = " "):
...
+        """Compile the code from source to bytes."""
+        if isinstance(self.code_raw, int):
+            # Users pass code as int (very bad)
+            hex_str = format(self.code_raw, "02x")
+            return bytes.fromhex(hex_str)
+
+        if not isinstance(self.code_raw, str):
+            raise ValueError(f"parse_code(code: str) code is not string: {self.code_raw}")
+        if len(self.code_raw) == 0:
+            return bytes.fromhex("")
```
marioevz, line 73:
```suggestion
            return b""
```
Tiny improvement.

src/ethereum_test_specs/static_state/common/common.py
```diff
@@ -53,116 +48,142 @@ def parse_args_from_string_into_array(stream: str, pos: int, delim: str = " "):
...
+            compiled_code = self.code_raw[2:].lower()
+
+        # Prase lllc code
+        elif self.code_raw.lstrip().startswith("{") or self.code_raw.lstrip().startswith("(asm"):
+            with tempfile.NamedTemporaryFile(mode="w+", delete=False) as tmp:
+                tmp.write(self.code_raw)
+                tmp_path = tmp.name
+
+            # using lllc
+            # result = subprocess.run(["lllc", tmp_path], capture_output=True, text=True)
```
marioevz, line 153: I think it should be ok to just require this to be installed already.

At the moment, we prepare releases and fill tests using out external runners, and all of our current external runners point to our server that has already `lllc` installed, so it should just work.

Downside would be that we would be unable to add unit tests for state static tests that contain lllc, but I think that should be ok.

src/ethereum_test_specs/static_state/common/common.py
```diff
@@ -53,116 +48,142 @@ def parse_args_from_string_into_array(stream: str, pos: int, delim: str = " "):
...
+        compiled_code = ""
+
+        raw_marker = ":raw 0x"
+        raw_index = self.code_raw.find(raw_marker)
+        abi_marker = ":abi"
+        abi_index = self.code_raw.find(abi_marker)
+        yul_marker = ":yul"
+        yul_index = self.code_raw.find(yul_marker)
+
+        # Prase :raw
```
marioevz, line 84:
```suggestion
        # Parse :raw
```

src/ethereum_test_specs/static_state/common/common.py
```diff
@@ -53,116 +48,142 @@ def parse_args_from_string_into_array(stream: str, pos: int, delim: str = " "):
...
+            with tempfile.NamedTemporaryFile(mode="w+", delete=False) as tmp:
+                tmp.write(native_yul_options + self.code_raw[source_start:])
+                tmp_path = tmp.name
+            compiled_code = compile_yul(
+                source_file=tmp_path,
+                evm_version=options[0] if len(options) >= 1 else None,
+                optimize=options[1] if len(options) >= 2 else None,
+            )[2:]
+
+        # Prase :abi
```
marioevz, line 116:
```suggestion
        # Parse :abi
```

src/ethereum_test_specs/static_state/common/common.py
```diff
@@ -53,116 +48,142 @@ def parse_args_from_string_into_array(stream: str, pos: int, delim: str = " "):
...
+                    ],
+                )
+                return function_signature + function_parameters
+            return function_signature
+
+        # Prase plain code 0x
+        elif self.code_raw.lstrip().startswith("0x"):
+            compiled_code = self.code_raw[2:].lower()
+
+        # Prase lllc code
```
marioevz, line 146:
```suggestion
        # Parse lllc code
```

marioevz (commented):
Looks great, just a few comments, let me know what you think about the docker alternative.

src/ethereum_test_specs/static_state/common/common.py
```diff
@@ -53,116 +47,146 @@ def parse_args_from_string_into_array(stream: str, pos: int, delim: str = " "):
...
+                            if parameter_types[t_index] == "bool"
+                            else False and ValueError("unhandled parameter_types")
+                            for t_index, t in enumerate(tokens[1:])
+                        ]
+                    ],
+                )
+                return function_signature + function_parameters
+            return function_signature
+
+        # Prase plain code 0x
```
marioevz, line 141:
```suggestion
        # Parse plain code 0x
```

marioevz (approved):
I verified and all legacy tests are filling locally, I think we can go ahead and merge this plus #1442.

marioevz:
I marked the last checkbox in the description because we are now using locally built `lllc`, which doesn't require for us to stop any container.

PR #1441, opened 2025-04-11
feat(tests): add increased nonce with invalid nonce tests

danceratopz:
## 🗒️ Description

This PR attempts to add some missing test cases based purely on a coverage analysis of the test source itself. In particular, L269 in `test_gas` was not hit:
https://github.com/ethereum/execution-spec-tests/blob/62d2563dcfd58242f1846bab13f29a64dc558567/tests/prague/eip7702_set_code_tx/test_gas.py#L256-L276

Either of these two test cases fill the line coverage gap, but I am out of my depth with 7702 and would appreciate feedback on the test cases and whether they're reasonable. In particular, the second case fails to fill for all but one parameter combo - it need further restriction on the conditions it runs (ran out of time on this one).
1. The `multiple_invalid_nonce_authorizations_self_sponsored_multiple_signers` test cases fill and add coverage. All clients pass with consume-engine.
2. The `single_invalid_authorization_eoa_authority_multiple_signers` test cases only fill successfully for one parameter combo - this is WIP, needs further debugging.

Perhaps this already helps you understand the fails from 2.:
```
FAILED tests/prague/eip7702_set_code_tx/test_gas.py::test_gas_cost[fork_Prague-state_test-single_invalid_authorization_eoa_authority_multiple_signers] - ethereum_test_base_types.composite_types.Storage.KeyValueMismatchError: incorrect value in address 0x0000000000000000000000000000000000001000 (test_code_address) for key 0x0000000000000000000000000000000000000000000000000000000000000000: want 0x35f81 (dec:221057), got 0x0 (dec:0)
FAILED tests/prague/eip7702_set_code_tx/test_gas.py::test_account_warming[fork_Prague-state_test-single_invalid_authorization_eoa_authority_multiple_signers-check_delegated_account_first_True] - ethereum_test_base_types.composite_types.Storage.KeyValueMismatchError: incorrect value in address 0x0000000000000000000000000000000000001000 (callee_address) for key 0x0000000000000000000000000000000000000000000000000000000000000001: want 0xa28 (dec:2600), got 0xa8c (dec:2700)
FAILED tests/prague/eip7702_set_code_tx/test_gas.py::test_account_warming[fork_Prague-state_test-single_invalid_authorization_eoa_authority_multiple_signers-check_delegated_account_first_False] - ethereum_test_base_types.composite_types.Storage.KeyValueMismatchError: incorrect value in address 0x0000000000000000000000000000000000001000 (callee_address) for key 0x0000000000000000000000000000000000000000000000000000000000000000: want 0xa28 (dec:2600), got 0x1450 (dec:5200)
FAILED tests/prague/eip7702_set_code_tx/test_gas.py::test_intrinsic_gas_cost[fork_Prague-state_test-valid_True-single_invalid_authorization_eoa_authority_multiple_signers] - ethereum_test_specs.helpers.UnexpectedExecutionFailError: Unexpected fail for Transaction ({'index': 0, 'nonce': 0}):
```

Generate coverage for this module with:
```
uv sync --all-extras
fill --until Prague -m state_test
 tests/prague/eip7702_set_code_tx/test_gas.py  --cov tests --cov-report=html:7702-coverage
```

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -25,6 +25,7 @@ Test fixtures for use by clients are available for each release on the [Github r
 ### 🧪 Test Cases
 
 - ✨ [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702): Test precompile case in same transaction as delegation without extra gas in case of precompile code execution; parametrize all call opcodes in existing precompile test ([#1431](https://github.com/ethereum/execution-spec-tests/pull/1431)).
+- ✨ [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702): Add invalid nonce authorizations tests for the case of multiple signers when the sender's nonce gets increased ([#1441](https://github.com/ethereum/execution-spec-tests/pull/1441)).
 - ✨ [EIP-7623](https://eips.ethereum.org/EIPS/eip-7623): Additionally parametrize transaction validity tests with the `to` set to an EOA account (previously only contracts) ([#1422](https://github.com/ethereum/execution-spec-tests/pull/1422)).
 
 ## [v4.2.0](https://github.com/ethereum/execution-spec-tests/releases/tag/v4.2.0) - 2025-04-08
--- a/tests/prague/eip7702_set_code_tx/test_gas.py
+++ b/tests/prague/eip7702_set_code_tx/test_gas.py
@@ -453,6 +453,16 @@ def gas_test_parameter_args(
         ),
         pytest.param(
             {
+                "signer_type": SignerType.MULTIPLE_SIGNERS,
+                "authority_type": AddressType.EOA,
+                "authorization_invalidity_type": AuthorizationInvalidityType.INVALID_NONCE,
+                "self_sponsored": True,
+                "authorizations_count": multiple_authorizations_count,
+            },
+            id="multiple_invalid_nonce_authorizations_self_sponsored_multiple_signers",
+        ),
+        pytest.param(
+            {
                 "signer_type": SignerType.SINGLE_SIGNER,
                 "authorization_invalidity_type": AuthorizationInvalidityType.INVALID_CHAIN_ID,
                 "authorizations_count": multiple_authorizations_count,
--- a/tests/prague/eip7702_set_code_tx/test_gas.py
+++ b/tests/prague/eip7702_set_code_tx/test_gas.py
@@ -437,6 +437,14 @@ def gas_test_parameter_args(
         ),
         pytest.param(
             {
+                "authority_type": AddressType.EOA_WITH_SET_CODE,
+                "signer_type": SignerType.MULTIPLE_SIGNERS,
+                "authorization_invalidity_type": AuthorizationInvalidityType.INVALID_NONCE,
+            },
+            id="single_invalid_authorization_eoa_authority_multiple_signers",
+        ),
+        pytest.param(
+            {
                 "signer_type": SignerType.SINGLE_SIGNER,
                 "authorization_invalidity_type": AuthorizationInvalidityType.INVALID_NONCE,
                 "authorizations_count": multiple_authorizations_count,
--- a/tests/prague/eip7702_set_code_tx/test_gas.py
+++ b/tests/prague/eip7702_set_code_tx/test_gas.py
@@ -92,9 +92,22 @@ class AuthorityWithProperties:
     """Dataclass to hold the properties of the authority address."""
 
     authority: EOA
+    """
+    The address of the authority to be used in the transaction.
+    """
     address_type: AddressType
+    """
+    The type of the address the authority was before the authorization.
+    """
     invalidity_type: AuthorizationInvalidityType | None
-    empty: bool
+    """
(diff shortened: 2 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Test fixtures for use by clients are available for each release on the [Github r
 ### 🧪 Test Cases
 
 - ✨  Test 7702 precompile case in same transaction as delegation without extra gas in case of precompile code execution; parametrize all call opcodes in existing precompile test ([#1431](https://github.com/ethereum/execution-spec-tests/pull/1431)).
+- ✨ [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702): Add tests for the case where Test precompile case in same transaction as delegation without extra gas in case of precompile code execution; parametrize all call opcodes in existing precompile test ([#1431](https://github.com/ethereum/execution-spec-tests/pull/1431)).
```
fselmo, line 23: Why is this copying my bullet from above it 🤔

danceratopz: That's a good question - fixed!

tests/prague/eip7702_set_code_tx/test_gas.py
```diff
@@ -435,6 +435,14 @@ def gas_test_parameter_args(
             },
             id="single_invalid_authorization_invalid_chain_id_single_signer",
         ),
+        pytest.param(
+            {
+                "authority_type": AddressType.EOA_WITH_SET_CODE,
```
fselmo, line 440: Just looking now and had a quick q. I noticed this is not in the `if include_pre_authorized:` test cases. `EOA_WITH_SET_CODE` implies pre-authorized, right? If so, should this be `AddressType.EOA`? This passes all the tests but I want to make sure that's your intended test case.

fselmo: nm, I wasn't looking at the code block in the description. I'll keep that in mind 😅.

fselmo, line 440: I think I know at least some of what is going on. If you don't re-authorize this pre-authorized address, this first authorization is marked as `skip=True`, which, if there are no other authorizations in the transaction will yield `TransactionException.TYPE_4_EMPTY_AUTHORIZATION_LIST` for all the tests that send transactions.

Some of the tests for `test_account_warming` are failing with a `post` state verification mismatch before that even happens.

I'm not quite sure what's best here and still getting used to all the intricacies of this test setup but I think the test may require `"authorizations_count": multiple_authorizations_count` to work well with `MULTIPLE_SIGNERS` and maybe even `"re_authorize": True` to work outside of the `include_pre_authorized` code block.

That still doesn't get the tests filling but I'm not sure if all the conditionals in the test setup are accounting for this test case in the first place. I know that adding the following test under `include_many` does work but I think some version of what I mentioned above should technically work inside the main `cases` block so it may be down to tweaking the test setup conditionals.

```py
if include_many:
    cases += [
        ...,
        pytest.param(
            {
                "authority_type": AddressType.EOA_WITH_SET_CODE,
                "re_authorize": True,
                "signer_type": SignerType.MULTIPLE_SIGNERS,
                "authorization_invalidity_type": AuthorizationInvalidityType.INVALID_NONCE,
                "authorizations_count": multiple_authorizations_count,
            },
            id="pre_authorized_eoa_authority_re_authorized_invalid_nonce_multiple_signers_multiple_auths",
        ),
    ]
```

marioevz:
Created PR #1444 to fix the logic of the test expectation for the two added tests, and added a new case.

Basically, because the delegation had an invalid nonce, the test expected that there was no delegation at the authority address, but that's not true because the type of the authority was `EOA_WITH_SET_CODE`, so the delegation was there even though the authorization failed.

Sorry about the confusing logic to this test, at this point we are close to re-implement the EVM in the tests.

I added some more comments to help understand the test a bit better.

danceratopz:
This fills with the `7702-to-precompile` EELS branch at commit https://github.com/ethereum/execution-specs/pull/1191/commits/bbc469729ab095300b07de8fe323c3f9fac31857 from:  
- https://github.com/ethereum/execution-specs/pull/1191 

I diffed the fixtures against those from `main` as a sanity check and it LGTM. Contract code and storage has changed, as expected due to the use of addresses as storage keys, but the transactions remain unchanged :+1:.

@marioevz I updated `eels_resolver.json` as this EELS branch is the currently the best branch to fill Prague (`forks/prague` doesn't fill due to a rename of a t8n field, which has also been addressed in https://github.com/ethereum/execution-specs/pull/1191). We should update to `forks/prague` as soon as this PR been merged, but execution-specs are having some CI issues at the moment.

Feel free to revert the latest commit, if you'd rather leave the old eels branch.

marioevz (approved):
LGTM, I think the eels_resolutions.json change is ok, and I have this other PR #1454 to update to forks/prague once https://github.com/ethereum/execution-specs/pull/1191 is merged, but this temporary workaround should be ok.

Thanks!

PR #1442, opened 2025-04-11
feat(fill,tests): add all static `GeneralStateTestsFiller` from `ethereum/tests` 

winsvega:
## 🗒️ Description
All ethereum/tests state test fillers

## 🔗 Related Issues

To rebase on:
https://github.com/ethereum/execution-spec-tests/pull/1439

## ✅ Checklist


marioevz (approved):
LGTM!

A couple of things in the commits I made after yours:
- Some of the modules were causing issues due to use of hyphens in the name, so I renamed them to use underscores.
- I marked `stTimeConsuming` as slow, just so we know that these tests take time to fill when using EELS, although in #1491 I recommend we use evmone to fill all legacy tests
- I marked all `tests/legacy/state_tests/Cancun/stEIP4844_blobtransactions/blobhashListBounds*Filler.yml` tests with the `skip` marker, since these were the only tests in the entire set that expected no failures in Cancun, and a failure in Prague, and since these are already covered and parametrized for future forks in ./tests/cancun/eip4844*`, they can be considered redundant.
- Modified several tests to start from block 1, because higher blocks were causing issues, and the block number was irrelevant to the tests.

danceratopz (changes requested):
Can we add the tests at `tests/static` instead of `tests/legacy`? Happy to hear other suggestions, but seems more logical/reads better?
```
fill --until=Prague --fill-static-tests ./tests/static
```

Does the static filler need the `__init__.py` file, btw? If not, we could leave them out? If so, then I would update all the `__init__.py` to remove "legacy" and use "static".

And can we add `./tests/static/README.md` which states the ethereum/tests reference (with link) that these tests were taken from?

chfast:
Are you going to check-in all these files to the repo? This is one of the main pain points for ethereum/tests: there are a lot of commits changing a lot of files in a bulk mode so ethereum/tests is unusable by default `git clone`. You need some fancy git tricks to reduce the download size.

spencer-tb:
I'm realising its okay here with all the files as they will never change, and they are just the fillers not the json fixtures! So from my side its fine without the submodules!

Plus on Dan's comment: legacy -> static

chfast:
But are you going to remove them gradually?

spencer-tb:
Yeah I believe so, once a filler is ported into python the filler will be removed

chfast:
Fine to me, I won't block it.

marioevz:
@danceratopz Change is in, let me know what you think.

danceratopz (approved):
Thanks, LGTM. 

~~Ideally, we'd add an ethereum/tests permalink (with commit hash) from which the files were taken from.~~
Edit: It's there!

And just out of interest, do we need the `__init__.py` files? For the static filler?

danceratopz:
> Ideally, we'd add an ethereum/tests permalink (with commit hash) from which the files were taken from.
Ah sorry, just saw you added this, thanks! So many files!

winsvega:
> > Submodule it to a separate repo? This files only needed to generate a release
> 
> We talked about this and the nice thing about having all of this in the same repository is that we can now have atomic PRs that update a static yml/json filler to python.

I would still double check the coverage with ported ethereum test.txt file

marioevz:
> Thanks, LGTM.
> 
> ~Ideally, we'd add an ethereum/tests permalink (with commit hash) from which the files were taken from.~ Edit: It's there!
> 
> And just out of interest, do we need the `__init__.py` files? For the static filler?

This line fails if we don't add these :(
https://github.com/ethereum/execution-spec-tests/blob/ab9817d3573bfeab93773800ae3df09bf2b7e880/src/pytest_plugins/filler/static_filler.py#L322

I'd rather not add them too and I hope we find a workaround at some point.

winsvega:
> Thanks, LGTM. 
> 
> ~~Ideally, we'd add an ethereum/tests permalink (with commit hash) from which the files were taken from.~~
> Edit: It's there!
> 
> And just out of interest, do we need the `__init__.py` files? For the static filler?

Yes so it will scan Filler files for pydantic scheme.

PR #1456, opened 2025-04-15
zkevm: add bytecode worst case

jsign:
This PR adds a test that creates the current bytecode worst case scenario for provers.

It creates the target bytecodes as part of pre-alloc, which isn't friendly for filling. An alternative would be to generate these bytecodes with txs. That probably requires creating blocks before the attack block since you can't deploy ~300MiB of code in a single block due to gas limits. Sounds doable, but it can take some work.

To fill: `uv run fill --from=Cancun --until=Cancun -m "zkevm and blockchain_test"`

cc @kevaundray @marioevz

### DIFF
--- a/src/ethereum_test_tools/__init__.py
+++ b/src/ethereum_test_tools/__init__.py
@@ -79,6 +79,7 @@ from .code import (
     Conditional,
     Initcode,
     Switch,
+    While,
     Yul,
     YulCompiler,
 )
@@ -149,6 +150,7 @@ __all__ = (
     "TransactionTest",
     "TransactionTestFiller",
     "UndefinedOpcodes",
+    "While",
     "Withdrawal",
     "WithdrawalRequest",
     "Yul",
--- a/src/ethereum_test_tools/code/__init__.py
+++ b/src/ethereum_test_tools/code/__init__.py
@@ -1,16 +1,17 @@
 """Code related utilities and classes."""
 
-from .generators import CalldataCase, Case, CodeGasMeasure, Conditional, Initcode, Switch
+from .generators import CalldataCase, Case, CodeGasMeasure, Conditional, Initcode, Switch, While
 from .yul import Solc, Yul, YulCompiler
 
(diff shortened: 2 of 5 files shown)


marioevz (commented):
I spoke with about the purpose of the test and he explained that it was necessary that the contracts were not part of the pre-alloc since that would make the size of the fixture several gigabytes.

I suggested we create a factory contract that generates random code using keccak with the initcode.

We could:
- use a first transaction to deploy the factory contract.
- use N transactions to call the factory and create the desired amount of contracts.
- lastly send a transaction that uses EXTCODESIZE/HASH on all contracts that were created by the prior txs

raxhvl:
Hey @marioevz could we deploy the factory contract in a separate block? And then call `EXTCODESIZE/HASH`  txs in next block as part of the same blockchain test?

i.e is it possible to persist state between two blocks in a test?

jsign:
The factory creation and calling to the factory should be in as many blocks as needed (due to gas limit). After all is ready, do the "attack" in a single separate block so you have full gas limit.

> My understanding is that the t8n tool only supports a single block after which it writes the post state to disk.

For these kind of tests, [blockchain tests](https://github.com/ethereum/execution-spec-tests/pull/1456/files#diff-cdcef4e8328e54064d53a63c2653db5f0eeab8030d09db345b118bcd87b0cf50R62), you can pass multiple blocks, and the state is kept in the entire blockchain.

raxhvl:
That makes sense. One challenge I think is getting the deployed contract addresses to call in "attack" block.

marioevz:
> That makes sense. One challenge I think is getting the deployed contract addresses to call in "attack" block. We could write the addresses to the factory's storage, i.e we'll spend some non-trivial amount of gas in the attack block reading from storage.

Passing them through calldata is another option, because the created contract addresses are deterministic and known during the test function execution.

marioevz (approved):
Rebased and should be ready to merge :)

PR #1464, opened 2025-04-16
Suggestions for: feat(consume): add ExceptionMapper support to the consume-engine simulator

danceratopz:
## 🗒️ Description
A few suggestions for #1416.

Feel free to pick and choose, as you like.

## 🔗 Related Issues
- #1416

### DIFF
--- a/src/pytest_plugins/consume/hive_simulators/engine/test_via_engine.py
+++ b/src/pytest_plugins/consume/hive_simulators/engine/test_via_engine.py
@@ -23,6 +23,7 @@ def test_blockchain_via_engine(
     engine_rpc: EngineRPC,
     fixture: BlockchainEngineFixture,
     client_strict_exception_matching: bool,
+    fork_strict_exception_matching: bool,
 ):
     """
     1. Check the client genesis block hash matches `fixture.genesis.block_hash`.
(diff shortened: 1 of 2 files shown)


src/pytest_plugins/consume/hive_simulators/engine/test_via_engine.py
```diff
@@ -115,11 +116,15 @@ def test_blockchain_via_engine(
                                         f'got: "{payload_response.validation_error}" '
                                         f'expected: "{payload.validation_error}"'
                                     )
-                                    if client_strict_exception_matching:
+                                    if (
+                                        not client_strict_exception_matching
+                                        or not fork_strict_exception_matching
```
marioevz, line 121: I don't like having to repeat this `or` logic everywhere, case in point, line 104 in this file also requires this logic and was missed.
So I think we should create yet another fixture with client+fork strict exception matching.

marioevz (approved):
Thanks!

src/pytest_plugins/consume/hive_simulators/engine/test_via_engine.py
danceratopz: Thanks for catching that! And yes, this is much nicer / less error prone  :+1:

Issue #1434, opened 2025-04-10
Make max default block gas parametrizable in `fill` command (currently ~30 million)


marioevz:
Regarding the gas increment that is being pushed for Osaka, I think we should make the max-gas of our tests configurable.

We hard-code 30 million gas in many places because we assumed that this is the limit, but I think we should make it a parameter to `fill` with a default number, and we increment this number every time we want to increment the mainnet defaults to generate all tests with an increased gas automatically.

winsvega:
Actually would be nice if it errors on unreasonably high gas to force us to rewrite test sonit won't consume unnecessary large amount and be compatible with execute.

Unless absolutely necessary

marioevz:
> Actually would be nice if it errors on unreasonably high gas to force us to rewrite test sonit won't consume unnecessary large amount and be compatible with execute.
> 
> Unless absolutely necessary

I think the key would be to separate the tests that reach this limit by filling the block with repeated actions as part of the tests, vs the ones that require the operations consuming the gas as part of the test logic. Right now we don't have a clear separation.

PR #1470, opened 2025-04-16
feat(tests): Add flag to configure `max_gas` for fill command

reedsa:
## 🗒️ Description

Add flag to configure `max_gas` for fill command with default of of 30_000_000.
Add fixture for `max_gas` to use in `gas_limit` of a Transaction.

## 🔗 Related Issues

#1434 

## ✅ Checklist


reedsa:
Are tests that use more than 30_000_000 gas expected to be set to the enormous amounts like in [this example](https://github.com/ethereum/execution-spec-tests/blob/3719e9277ca9fdaf97799b8f4084f57e977eb7e4/tests/osaka/eip7692_eof_v1/eip7873_tx_create/test_txcreate.py#L497). Some values of `50_000_000`, etc. in various osaka tests as well.

Should these values remain or should they be set to the `max_gas` value from the configurable fixture?

marioevz:
> @marioevz I like the idea of setting the gas_limit as part of the [`Environment` class](https://github.com/ethereum/execution-spec-tests/blob/3719e9277ca9fdaf97799b8f4084f57e977eb7e4/src/ethereum_test_types/types.py#L347). I assume the initial value should just use the flag value via the test fixture somehow? Perhaps a new method on the class similar to `set_fork_requirements` that is called from [`make_state_test_fixture`](https://github.com/ethereum/execution-spec-tests/blob/3719e9277ca9fdaf97799b8f4084f57e977eb7e4/src/ethereum_test_specs/state.py#L170)?
> 
> I noticed there is a default value of `100_000_000_000_000_000` for the `gas_limit`. Should that value be lowered to the default configurable value of `30_000_000` and allow overrides with the `--gas_limit` flag?
> 
> Once it is in the environment, all tests will change and I'm not certain what effect lowering the value will have. Is this the right approach?

I think the default of `100_000_000_000_000_000` is way overblown, and we should be able to lower it and match it to the value we specify as the block limit, and not that many tests should fail.

Doing the environment-as-a-fixture would require some big changes, like the ones we did for the pre-as-a-fixture.
I.e. The environment changes from being something that gets instantiated by the test function (most tests just do `Environment()`) to something that is passed as a parameter and then read and potentially modified if necessary.

It might not even be necessary that the test function uses the parameter and we could simply pass it from the filler to the spec like we do for the `pre` here: https://github.com/ethereum/execution-spec-tests/blob/b48d1dc81233af6e4d6c7c84e60e8eaa4067a288/src/pytest_plugins/filler/filler.py#L717-L718

If the test function requires to know about its environment, it will be included in the test function as a parameter:
```
def test_something(pre: Alloc, env: Environment ...):
```

And the test could then:
- Read values from its environment, like the `gas_limit` and adapt the test to only use that at most. If the test uses the environment as read-only, the test can be easily filled (`uv run fill`) and executed (`uv run execute`).
- Write values to its environment, like increase the `gas_limit` to a minimum value because the test requires it as a minimum to execute its test logic, and in this case the test could be filled but not executed (because we cannot modify the environment we are running in when it's a live network).

One change I would do to the current implementation of this PR is rename `max_gas` to `block_gas_limit`.

reedsa:
@marioevz Pushed up my latest changes and had some follow up questions for what we've discussed.
1. I was wondering if it would be useful to have the existing `env.gas_limit` only used for transactions and add a new `env.block_gas_limit` variable that is configurable?
2. I am looking at the pytantic models, not totally certain about where validation happens to check that the env has the `block_gas_limit` value set? So far I have: `block_gas_limit: NumberBoundTypeVar = Field(DEFAULT_BLOCK_GAS_LIMIT, alias="currentBlockGasLimit")` as an attribute of `EnvironmentGeneric` with the default set.

marioevz:
>     1. I was wondering if it would be useful to have the existing `env.gas_limit` only used for transactions and add a new `env.block_gas_limit` variable that is configurable?

I normally see `tx.gas_limit` as the limit applied to the transaction and `env.gas_limit` as the total gas limit in the environment (block) where the transaction will execute. If you set `tx.gas_limit=env.gas_limit` it should (IMO) be seen as "the tx should consume all the gas that it can from the environment where it runs".

>     2. I am looking at the pytantic models, not totally certain about where validation happens to check that the env has the `block_gas_limit` value set? So far I have: `block_gas_limit: NumberBoundTypeVar = Field(DEFAULT_BLOCK_GAS_LIMIT, alias="currentBlockGasLimit")` as an attribute of `EnvironmentGeneric` with the default set.

I don't think this validation happens anywhere at the moment, it's simply set to a default max.

As an example, we do a hacky trick to alter the defaults of a transaction by using a `TransactionDefaults` class:
https://github.com/ethereum/execution-spec-tests/blob/c35b42f3d84618e0430124b424c0f362bdbe4ab7/src/ethereum_test_types/types.py#L578-L585

And then in the `fill`/`execute` plugins we import and modify the value to the new defaults for the given execution context:
https://github.com/ethereum/execution-spec-tests/blob/c35b42f3d84618e0430124b424c0f362bdbe4ab7/src/pytest_plugins/execute/execute.py#L192-L200

Maybe this helps for environment too?

marioevz (commented):
I think it looks very good!

After going through the code, I'm more convinced about the separation between `gas_limit` and `block_gas_limit`.

I'll try the changes today locally and potentially merge if it goes well :)

marioevz:
Just to summarize, the two options are:
#### `env` fixture
- Tests would have to define the `env` parameter, and then pass it to `state_test`.
- Basically all state tests are affected

#### `EnvironmentDefaults`
- Only tests that modify gas limit would be affected, and they would need to do:
```python
env = Environment()
...
tx.gas_limit = env.gas_limit  # <- env.gas_limit here automatically contains the --block-max-gas value
```
- We could detect tests that go over this limit by checking `tx.gas_limit > Environment().gas_limit`, but for fill it's irrelevant, and we might only need to check this for execute.

reedsa:
I'm fine with using the `EnvironmentFixture` instead of the fixture. I'll update that shortly!

reedsa:
@marioevz I'm not seeing the flag in the help section when I put in the correct description for the flag group `parser.getgroup("env", "Arguments defining the environment")`. The flag had worked prior when I had it using the same description as the debug group. 😕 

With the `EnvironmentDefaults` approach I am not sure if that's preventing the limit flag from propogating into the fill tests. Maybe I'm missing something or confused about how the flag should work without using a fixture.

marioevz:
> @marioevz I'm not seeing the flag in the help section when I put in the correct description for the flag group `parser.getgroup("env", "Arguments defining the environment")`. The flag had worked prior when I had it using the same description as the debug group. 😕
> 
> With the `EnvironmentDefaults` approach I am not sure if that's preventing the limit flag from propogating into the fill tests. Maybe I'm missing something or confused about how the flag should work without using a fixture.

Have an idea on how to fix it: I think we should move the flag to the `shared` plugin.

I'll push a commit in a bit 👍

src/ethereum_test_types/types.py
```diff
@@ -335,6 +335,18 @@ class Withdrawal(WithdrawalGeneric[HexNumber]):
 
 
 DEFAULT_BASE_FEE = 7
+CURRENT_MAINNET_BLOCK_GAS_LIMIT = 36_000_000
+DEFAULT_BLOCK_GAS_LIMIT = CURRENT_MAINNET_BLOCK_GAS_LIMIT * 2
```
marioevz, line 339: I'm thinking this should be a sane limit of always filling the tests with twice the value we currently have in mainnet, with the possibility of preemptively bumping it before we signal a safe increment to client devs/node operators.

tests/istanbul/eip152_blake2/test_blake2.py
```diff
@@ -532,7 +532,7 @@ def test_blake2b_invalid_gas(
 
 @pytest.mark.valid_from("Istanbul")
 @pytest.mark.parametrize("call_opcode", [Op.CALL, Op.CALLCODE])
-@pytest.mark.parametrize("gas_limit", [100_000_000, 90_000, 110_000, 200_000])
+@pytest.mark.parametrize("gas_limit", [Environment().gas_limit, 90_000, 110_000, 200_000])
```
marioevz, line 535: @reedsa let me know what you think of this since you were the original author of the test.

The idea is that the bump occurs automatically when we bump the default in `CURRENT_MAINNET_BLOCK_GAS_LIMIT` or when we try to fill with a higher block gas limit passed as parameter.

marioevz:
@reedsa Added the changes and it seems to be working now.

The nice thing is that with this change, since the default for `Environment.gas_limit` was drastically reduced, tests that were using excessive gas started failing, and I took the opportunity and reduced the gas limit of many tests where it was unnecessary.

Let me know what you think of the changes.

src/pytest_plugins/filler/filler.py
```diff
@@ -172,6 +173,17 @@ def pytest_addoption(parser: pytest.Parser):
         default=True,
         help="Skip generating an index file for all produced fixtures.",
     )
+    test_group.addoption(
```
reedsa, line 176: Having this in the test group feels better than on it's own 👍 Did you happen to understand why the new test group I added wasn't showing up with a unique description?

src/pytest_plugins/execute/execute.py
```diff
@@ -56,6 +56,18 @@ def pytest_addoption(parser):
             "unless overridden by the test."
         ),
     )
+    execute_group.addoption(
```
reedsa, line 59: Execution tests set the txn gas limit, is this because the tests have no control over the block gas limit on the testnet?

tests/istanbul/eip152_blake2/test_blake2.py
reedsa: Good idea, I think these original values are pretty arbitrary. It's great having the ability to set this in a param!

reedsa (commented):
Great work on these updates! I think the only update I would make is to the help text and I'll go ahead and push up that change.

Any test that currently just uses the default environment could also stop using `env = Environment()` to pass into `state_test` but isn't necessarily hurting anything as is. Looks good to go!

src/pytest_plugins/filler/filler.py
marioevz: It's a bit of a mess the help in our commands 😅

We do some regex matching to see what to include and what not to include in the commands, and I think some cleanup is necessary because the current method is not ideal.

tests/istanbul/eip152_blake2/test_blake2.py
marioevz: I wrote a unit test for this pattern and actually found a bug:
I was overwriting the `EnvironmentDefaults` in a fixture, but if the test uses `Environment().gas_limit` to generate its parametrization like in this example, pytest only resolves fixtures during actual test execution, so too late for this.
I modified the `EnvironmentDefaults` mod to happen during `pytest_config` which happens before collection even.

docs/filling_tests/filling_tests_command_line.md
```diff
@@ -115,82 +115,74 @@ Output:
...
+  --block-gas-limit BLOCK_GAS_LIMIT
+                        Default gas limit used ceiling used for blocks and tests that attempt to consume
+                        an entire block's gas. (Default: 72000000)
 
 Arguments defining debug behavior:
-  --evm-dump-dir EVM_DUMP_DIR, --t8n-dump-dir EVM_DUMP_DIR
-                        Path to dump the transition tool debug output.  (Default: <repo>/logs/evm)
+  --evm-dump-dir, --t8n-dump-dir EVM_DUMP_DIR
+                        Path to dump the transition tool debug output. (Default:
+                        /Users/stu/projects/ethereum/execution-spec-tests/logs/evm)
```
marioevz, line 179:
```suggestion
                        <repo>/logs/evm)
```

whitelist.txt
```diff
@@ -418,6 +418,7 @@ stdout
 stExample
 str
 streetsidesoftware
+stu
```
marioevz, line 421:
```suggestion
```

marioevz (approved):
LGTM, thanks for this!

I'll apply my comments and merge :)

PR #1476, opened 2025-04-17
new(tests): TXCREATE executes validation

pdobacz:
## 🗒️ Description

This is a pretty beefy test which ensures that TXCREATE correctly handles its initcode validation regardless of many circumstances it might find itself in:
- eof/legacy
- static mode
- insufficient balance for endowment
- input
- in top frame or in *call frame
- first or second execution in the transaction
- with/without other invalid/valid initcodes in the tx.initcodes
- with/without an access list entry at the destination address

It is ~3K~ 6K test cases :/, but I've done some reduction of the domain to bring it down to that level. Despite being a lot, I'd argue that ensuring EOF validation runs when it should, regardless of what's happening around it, is warranted to compliment all the EOF validation tests we have

## 🔗 Related Issues

NA

## ✅ Checklist

### DIFF
--- a/tests/osaka/eip7692_eof_v1/eip7873_tx_create/test_txcreate_validates.py
+++ b/tests/osaka/eip7692_eof_v1/eip7873_tx_create/test_txcreate_validates.py
@@ -1,0 +1,229 @@
+"""Test bad TXCREATE cases."""
+
+from enum import Enum, auto, unique
+from typing import Tuple
+
+import pytest
+
+from ethereum_test_base_types import Bytes
+from ethereum_test_base_types.base_types import Address, Hash
+from ethereum_test_tools import (
+    Account,
+    Alloc,
+    Environment,
+    EVMCodeType,
+    StateTestFiller,
+    Transaction,
+    compute_eofcreate_address,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+from ethereum_test_types.eof.v1 import Container, Section
+from ethereum_test_vm.bytecode import Bytecode
+from tests.prague.eip7702_set_code_tx.spec import Spec
+
+from .. import EOF_FORK_NAME
+from ..eip7620_eof_create.helpers import (
+    slot_a,
+    slot_b,
+    slot_code_worked,
+    smallest_initcode_subcontainer,
+    smallest_runtime_subcontainer,
+    value_code_worked,
+)
+from .spec import TXCREATE_FAILURE
+
+REFERENCE_SPEC_GIT_PATH = "EIPS/eip-7873.md"
+REFERENCE_SPEC_VERSION = "1115fe6110fcc0efc823fb7f8f5cd86c42173efe"
+
+pytestmark = pytest.mark.valid_from(EOF_FORK_NAME)
+
+
+@unique
+class ValidatedCode(Enum):
+    """Kinds of valid/invalid EOF."""
+
+    LEGACY = auto()
+    EF = auto()
+    EOFV1_RUNTIME = auto()
+    EOFV1_RUNTIME_INVALID = auto()
+    EOFV1_INITCODE = auto()
+    EOFV2 = auto()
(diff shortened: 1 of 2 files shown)


tests/osaka/eip7692_eof_v1/eip7873_tx_create/test_txcreate_validates.py
```diff
@@ -0,0 +1,229 @@
...
+        """Hash of the bytecode to validate."""
+        bytecode = self.bytecode()
+        if isinstance(bytecode, Bytecode):
+            return bytecode.keccak256()
+        elif isinstance(bytecode, Container):
+            return bytecode.hash
+        elif isinstance(bytecode, Bytes):
+            return bytecode.keccak256()
+        else:
+            raise ValueError("Unknown bytecode type")
```
pdobacz, line 89: I tried hard to not have this, but I failed to return something from `bytecode` which hashes allright and is also fine with mypy. Any Python magic to make this go away?

tests/osaka/eip7692_eof_v1/eip7873_tx_create/test_txcreate_validates.py
```diff
@@ -0,0 +1,229 @@
...
+        salt: int,
+        evm_code_type: EVMCodeType,
+        value: int,
+        input_size: int,
+    ) -> Tuple[Bytecode, Address | None]:
+        """
+        Return snippet to cause TXCREATE to be called along with an address which
+        will end up in the `compute_eofcreate_address`, or None if that would be the snippet
+        itself.
+        """
```
marioevz, line 124:
```suggestion
        """
        if evm_code_type not in [EVMCodeType.LEGACY, EVMCodeType.EOF_V1]:
            raise Exception(f"Test needs to be updated for {evm_code_type}")
```

tests/osaka/eip7692_eof_v1/eip7873_tx_create/test_txcreate_validates.py
```diff
@@ -0,0 +1,229 @@
...
+@pytest.mark.with_all_evm_code_types
+# Subset chosen to limit number of test cases
+@pytest.mark.parametrize("code_a", [ValidatedCode.EOFV1_INITCODE, ValidatedCode.LEGACY])
+@pytest.mark.parametrize("code_b", ValidatedCode)
+# Subset chosen to limit number of test cases
+@pytest.mark.parametrize("factory_a", [Factory.DIRECT, Factory.WITH_CALL])
+@pytest.mark.parametrize("factory_b", Factory)
+@pytest.mark.parametrize("value", [0, 1])
+@pytest.mark.parametrize("input_size", [0, 31])
+def test_txcreate_validates(
```
marioevz, line 173: I think the amount of parametrization is necessary, I would even add with and without access-list key that targets a storage slot in the newly created address.

marioevz (commented):
Thanks, some comments

tests/osaka/eip7692_eof_v1/eip7873_tx_create/test_txcreate_validates.py
pdobacz: yup, I was clearly overthinking this, thx :)!

tests/osaka/eip7692_eof_v1/eip7873_tx_create/test_txcreate_validates.py
pdobacz: this seems further away from TXCREATE execution and its validation, but I may be biased knowing the implementation details.

Maybe this could we well enough served by a separate test, which goes more into initcode doing SSTOREs and SLOADs too?

Okay, actually when I think about testing gas charging, it makes total sense to include here, I'll do that, and the above test I'll add to a todo list of tests to add (for EOFCREATE too), in a separate PR

marioevz (approved):
LGTM, thanks!

PR #1479, opened 2025-04-17
fix(github): Remove `vars` usage in build-fixtures

marioevz:
## 🗒️ Description
Fixture building is currently failing due to the use of `vars`, which is not allowed in that context.

Solution is to accept all required vars as input and then populate the input in the caller workflows.

## 🔗 Related Issues
None

## ✅ Checklist


marioevz:
Bypassing to make the release.

Issue #1453, opened 2025-04-15
Create blockchains that consume full gas on single operations.


marioevz:
Requirement for zkevm testing:

Create a test that exhausts the gas stipend of a full block using a single operation, operation meaning:
- An opcode, namely keccak or maybe other relevant ones.
- A precompile

Potentially also parametrizing whether the tx succeeds or fails (reverts or out-of-gas).

cc'ing @kevaundray to expand or correct the definition of operations

raxhvl:
Looking into this.

kevaundray:
# Task list

##  EVM Opcodes  
- [ ] `00` STOP  
- [ ] `01` ADD  
- [ ] `02` MUL  
- [ ] `03` SUB  
- [ ] `04` DIV  
- [ ] `05` SDIV  
- [ ] `06` MOD  
- [ ] `07` SMOD  
- [ ] `08` ADDMOD  
- [ ] `09` MULMOD  
- [ ] `0A` EXP  
- [ ] `0B` SIGNEXTEND  
- [ ] `10` LT  
- [ ] `11` GT  
- [ ] `12` SLT  
- [ ] `13` SGT  
- [ ] `14` EQ  
- [ ] `15` ISZERO  
- [ ] `16` AND  
- [ ] `17` OR  
- [ ] `18` XOR  
- [ ] `19` NOT  
- [ ] `1A` BYTE  
- [ ] `1B` SHL  
- [ ] `1C` SHR  
- [ ] `1D` SAR  
- [ ] `20` KECCAK256  
- [ ] `30` ADDRESS  
- [ ] `31` BALANCE  
- [ ] `32` ORIGIN  
- [ ] `33` CALLER  
- [ ] `34` CALLVALUE  
- [ ] `35` CALLDATALOAD  
- [ ] `36` CALLDATASIZE  
- [ ] `37` CALLDATACOPY  
- [ ] `38` CODESIZE  
- [ ] `39` CODECOPY  
- [ ] `3A` GASPRICE  
- [ ] `3B` EXTCODESIZE  
- [ ] `3C` EXTCODECOPY  
- [ ] `3D` RETURNDATASIZE  
- [ ] `3E` RETURNDATACOPY  
- [ ] `3F` EXTCODEHASH  
- [ ] `40` BLOCKHASH  
- [ ] `41` COINBASE  
- [ ] `42` TIMESTAMP  
- [ ] `43` NUMBER  
- [ ] `44` PREVRANDAO  
[...]

PR #1483, opened 2025-04-21
✨ feat(tests): zkEVM - Add parametrized opcodes and gas limits 

raxhvl:
## 🗒️ Description
The PR extends #1456 to add dynamic gas_limits and opcodes.

## 🔗 Related Issues
#1453 

### How it works
https://github.com/ethereum/execution-spec-tests/blob/96685834898a03000f9c24f7f6198a4fb48feb38/tests/zkevm/test_proof_size.py#L2-L9

### Configuring gas limits
https://github.com/ethereum/execution-spec-tests/blob/858af44a631813e3c910a2ce32d63d8a707de82c/tests/zkevm/test_proof_size.py#L37

### Configuring opcodes
https://github.com/ethereum/execution-spec-tests/blob/96685834898a03000f9c24f7f6198a4fb48feb38/tests/zkevm/test_proof_size.py#L80-L91

## What's pending
As described in #1456, the pre state size is huge. To solve that, I experimented deploying contracts in the same block, it had the following issues:
1. The code for deploying large contracts gets messy quickly, and it consumes large amounts gas due to high memory usage.
2. I adjusted gas limit to offset the contract deployment costs, but I think it skews the benchmark a bit if we do it in the same block. 

I was wondering if there is a way to persist data between blocks in the test.

If this approach sounds okay, then I can extend this to include precompiles.

Even after deploying contracts in a separate blocks, communicating the deployed addresses to be consumed would need more work.

cc: @jsign

## Output
```bash
tests/zkevm/test_proof_size.py::test_via_opcode[fork_Cancun-blockchain_test-EXTCODEHASH-gas_limit_30000000] FILLED                                               [ 1/16]
tests/zkevm/test_proof_size.py::test_via_opcode[fork_Cancun-blockchain_test-EXTCODEHASH-gas_limit_60000000] FILLED                                               [ 2/16]
tests/zkevm/test_proof_size.py::test_via_opcode[fork_Cancun-blockchain_test-EXTCODEHASH-gas_limit_100000000] FILLED                                              [ 3/16]
tests/zkevm/test_proof_size.py::test_via_opcode[fork_Cancun-blockchain_test-EXTCODEHASH-gas_limit_300000000] FILLED                                              [ 4/16]
tests/zkevm/test_proof_size.py::test_via_opcode[fork_Cancun-blockchain_test-EXTCODESIZE-gas_limit_30000000] FILLED                                               [ 5/16]
tests/zkevm/test_proof_size.py::test_via_opcode[fork_Cancun-blockchain_test-EXTCODESIZE-gas_limit_60000000] FILLED                                               [ 6/16]
tests/zkevm/test_proof_size.py::test_via_opcode[fork_Cancun-blockchain_test-EXTCODESIZE-gas_limit_100000000] FILLED                                              [ 7/16]
tests/zkevm/test_proof_size.py::test_via_opcode[fork_Cancun-blockchain_test-EXTCODESIZE-gas_limit_300000000] FILLED                                              [ 8/16]
```
## ✅ Checklist

### DIFF
--- a/tests/zkevm/test_proof_size.py
+++ b/tests/zkevm/test_proof_size.py
@@ -1,15 +1,36 @@
+"""
+A series of stress tests that assess the ability of a specific
+opcode or precompile to process large proofs in a block.
+
+A large proof is created by deploying a contract with
+maximum allowed bytecode.
+
+This proof is then "ingested" by an opcode or precompile.
+
+These tests first compute the "ingestion cost" for the proof.
+Next, they attempt to perform the maximum number of possible
+ingestions in a block.
+"""
+
 import pytest
 
 from ethereum_test_forks import Fork
-from ethereum_test_tools import (Alloc, Block, BlockchainTestFiller,
-                                 Environment, Transaction)
+from ethereum_test_tools import Alloc, Block, BlockchainTestFiller, Environment, Transaction
 from ethereum_test_tools.vm.opcode import Opcodes as Op
 
 REFERENCE_SPEC_GIT_PATH = "TODO"
 REFERENCE_SPEC_VERSION = "TODO"
 
-MAX_CONTRACT_SIZE = 24 * 1024
+
+##############################
+#                            #
+#          Config            #
+#                            #
+##############################
 GAS_LIMIT = 36_000_000
+
+KiB = 1024
+CONTRACT_BYTECODE_MAX_SIZE = 24 * KiB
 MAX_NUM_CONTRACT_CALLS = (GAS_LIMIT - 21_000) // (3 + 2600)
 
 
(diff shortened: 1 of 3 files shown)


tests/zkevm/test_proof_size.py
```diff
@@ -0,0 +1,126 @@
+"""
+A series of stress tests that assess the ability of a specific
+opcode or precompile to process large proofs in a block.
```
kevaundray, line 3: opcodes/precompiles wouldn't process proofs. 

These tests to be added should be seen as regular blockchain tests that so happen to be edge cases that we want to test in a zkEVM. In that sense, they should not be any different than other blockchain tests

kevaundray:
> The code for deploying large contracts gets messy quickly, and it consumes large amounts gas due to high memory usage.

I think we want to just deploy N distinct contracts over as many blocks as needed, where N is the amount of EXTCODEHASH opcodes that I can fit into a block.

Then in the subsequent blocks, we want to call 1 EXTCODEHASH. Then in the next block, we want to call 2 EXTCODEHASH, all the way upto N EXTCODEHASH.

kevaundray:
EXTCODEHASH requires a "setup" phase. 

This may not be the case for all opcodes/precompiles we want to test.

For the Add opcode for example, we can have block 1, do as many Adds as possible to fill up the block.

tests/zkevm/test_proof_size.py
raxhvl: Agreed. These tests are no different. By "process" I mean making having these opcode / precompile to take the code as input.

jsign:
Uhm, my take is that we should:
- Rename the testing file to `tests_worst_cases.py`
- Leave the original `test_worst_bytecode` as it is (maybe generalizing only gas limit and related)
- Create a separate new test `test_repeated_execution` with this new cases of repeated opcodes or precompiles.

My argument on keeping the second and third cases is exactly because the second requires some complicated setup, compared to opcode/precompile repetition being mostly stateless.

To generalize the setup to avoid big pre-alloc @marioevz suggested and idea [here](https://github.com/ethereum/execution-spec-tests/pull/1456#pullrequestreview-2781860285).

cc @kevaundray @raxhvl

raxhvl:
Thank @jsign. I misunderstood bullets (1) and (2) to be the same test.

PR #1491, opened 2025-04-23
feat(ci): Add `static` feature to produce `fixtures_static`

marioevz:
## 🗒️ Description
Requires #1439 and #1442.

Adds the `legacy` feature to build `fixtures_legacy` on every release.

We are going to use evmone due to speed because of some of the lengthier tests in `stTimeConsuming` that take longer than 60 seconds to execute with EELS, but this could change in the future.

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/.github/configs/evm.yaml
+++ b/.github/configs/evm.yaml
@@ -6,7 +6,7 @@ develop:
   impl: eels
   repo: null
   ref: null
-legacy:
+static:
   impl: evmone
   repo: ethereum/evmone
   ref: master
--- a/.github/configs/evm.yaml
+++ b/.github/configs/evm.yaml
@@ -6,6 +6,11 @@ develop:
   impl: eels
   repo: null
   ref: null
+legacy:
+  impl: evmone
+  repo: ethereum/evmone
+  ref: master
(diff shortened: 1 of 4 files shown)


spencer-tb (approved):
Looks good from my side! Left a comment in the legacy tests PR (approved that too though)

winsvega (approved):
I would prioritize conversion of valid blockchain tests now. as well as transition blockchain tests from ethereum/tests

danceratopz (changes requested):
Thanks, really excited to get this in! 

I'm not a fan of using "legacy" as it's not very descriptive and makes me think of https://github.com/ethereum/legacytests 

I would like to call them `fixtures_ethereum_tests.tar.gz` but as the tests now live in `execution-spec-tests` this is also a bit misleading, so suggest `static`. 

If this is accepted, I'd propose to change `./tests/legacy` to `./tests/static`, too (can be a separate PR).

danceratopz (approved):
LGTM, thanks!!

marioevz:
Successful build action: https://github.com/ethereum/execution-spec-tests/actions/runs/14648955754 🎉

PR #1313, opened 2025-03-13
chore(fill): fix transition fork state test node ids

spencer-tb:
## 🗒️ Description

Currently all state tests marked to be filled from a transition fork like `ShanghaiToCancunAtTime15k` will write the transition fork within the test node id, when it should write `Shanghai`.

As the state test doesn't concern block level logic the transition will remain within `Shanghai`. Only blockchain/engine tests will transition into Cancun, and hence should write the `ShanghaiToCancunAtTime15k` within the test node id.

### TLDR of da PR 💎 
It turn these items within `.meta/index.json`:
```json
  {
    "id": "tests/cancun/eip4844_blobs/test_blob_txs.py::test_blob_type_tx_pre_fork[fork_ShanghaiToCancunAtTime15k-state_test-no_blob_tx]",
    "fixture_hash": "0x3f1eb6aadae06e45277ace8ff1137df988fcf3fd2cbf24e7b98347bdbab3d6fb",
    "fork": "Shanghai",
    "format": "state_test",
    "json_path": "state_tests/cancun/eip4844_blobs/blob_txs/blob_type_tx_pre_fork.json"
  },
  ```
  Into the following:
  ```json
   {
    "id": "tests/cancun/eip4844_blobs/test_blob_txs.py::test_blob_type_tx_pre_fork[fork_Shanghai-state_test-no_blob_tx]",
    "fixture_hash": "0x3f1eb6aadae06e45277ace8ff1137df988fcf3fd2cbf24e7b98347bdbab3d6fb",
    "fork": "Shanghai",
    "format": "state_test",
    "json_path": "state_tests/cancun/eip4844_blobs/blob_txs/blob_type_tx_pre_fork.json"
   },
```

## 🔗 Related Issues
For more context please see #1308.

## ✅ Checklist
- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.

PR #1495, opened 2025-04-25
feat(tests): add eip-2929 precompile address warming tests

danceratopz:
## 🗒️ Description

Idea from @jochem-brouwer - thanks! Please check whether this test is doing what you intended.

Call BALANCE of a precompile address before and after a fork.

According to EIP-2929, when a transaction begins, accessed_addresses is initialized to include:
- tx.sender, tx.to
- and the set of all precompiles

This test verifies that:
1. Precompiles that exist in the predecessor fork are always "warm" (lower gas cost)
2. New precompiles added in a fork are "cold" before the fork and become "warm" after

These would perhaps be better as state tests. Feel free to close and implement a better suggestion, but happy for any feedback if you have time!

Also, this test should live in tests/paris, not tests/frontier, as they're only valid for EIP-2929, but it seemed silly to separate them from the other precompile generator tests. 

The test uses a generator; it generates these tests:
```
fill --fork CancunToPragueAtTime15k tests/frontier/precompiles/test_precompiles.py -k test_precompile_warming --collect-only -q -m blockchain_test
 pytest-regex selected 37 tests to run for regex: .*
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0xb-exists_in_predecessor_False-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0xc-exists_in_predecessor_False-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0xd-exists_in_predecessor_False-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0xe-exists_in_predecessor_False-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0xf-exists_in_predecessor_False-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x10-exists_in_predecessor_False-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x11-exists_in_predecessor_False-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0xa-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x9-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x5-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x6-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x7-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x8-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x1-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x2-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x3-exists_in_predecessor_True-blockchain_test]
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_CancunToPragueAtTime15k-address_0x4-exists_in_predecessor_True-blockchain_test]
```

## 🔗 Related Issues
None

## ✅ Checklist

### DIFF
--- a/tests/berlin/eip2929_gas_cost_increases/test_precompile_warming.py
+++ b/tests/berlin/eip2929_gas_cost_increases/test_precompile_warming.py
@@ -1,0 +1,147 @@
+"""
+abstract: Tests [EIP-2929: Gas cost increases for state access opcodes](https://eips.ethereum.org/EIPS/eip-2929)
+    Test cases for [EIP-2929: Gas cost increases for state access opcodes](https://eips.ethereum.org/EIPS/eip-2929).
+"""  # noqa: E501
+
+from typing import Iterator, Tuple
+
+import pytest
+
+from ethereum_test_forks import (
+    Fork,
+    get_transition_fork_predecessor,
+    get_transition_fork_successor,
+)
+from ethereum_test_tools import (
+    Account,
+    Address,
+    Alloc,
+    Block,
+    BlockchainTestFiller,
+    Transaction,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "EIPS/eip-2929.md"
+REFERENCE_SPEC_VERSION = "0e11417265a623adb680c527b15d0cb6701b870b"
+
+
+def precompile_addresses_in_predecessor_successor(
+    fork: Fork,
+) -> Iterator[Tuple[Address, bool, bool]]:
+    """
+    Yield the addresses of precompiled contracts and whether they existed in the parent fork.
+
+    Args:
+        fork (Fork): The transition fork instance containing precompiled contract information.
+
+    Yields:
+        Iterator[Tuple[str, bool]]: A tuple containing the address in hexadecimal format and a
+            boolean indicating whether the address has existed in the predecessor.
+
+    """
+    predecessor_precompiles = set(get_transition_fork_predecessor(fork).precompiles())
+    successor_precompiles = set(get_transition_fork_successor(fork).precompiles())
+    all_precompiles = successor_precompiles | predecessor_precompiles
+    highest_precompile = int.from_bytes(max(all_precompiles))
+    extra_range = 5
+    extra_precompiles = {
+        Address(i) for i in range(highest_precompile + 1, highest_precompile + extra_range)
+    }
+    all_precompiles = all_precompiles | extra_precompiles
+    for address in sorted(all_precompiles):
+        yield address, address in successor_precompiles, address in predecessor_precompiles
+
+
+@pytest.mark.valid_at_transition_to("Paris", subsequent_forks=True)
+@pytest.mark.parametrize_by_fork(
+    "address,precompile_in_successor,precompile_in_predecessor",
+    precompile_addresses_in_predecessor_successor,
+)
+def test_precompile_warming(
+    blockchain_test: BlockchainTestFiller,
+    fork: Fork,
+    address: Address,
+    precompile_in_successor: bool,
+    precompile_in_predecessor: bool,
+    pre: Alloc,
+):
+    """
+    Call BALANCE of a precompile addresses before and after a fork.
+
+    According to EIP-2929, when a transaction begins, accessed_addresses is initialized to include:
+    - tx.sender, tx.to
+    - and the set of all precompiles
+
+    This test verifies that:
+    1. Precompiles that exist in the predecessor fork are always "warm" (lower gas cost)
+    2. New precompiles added in a fork are "cold" before the fork and become "warm" after
+
+    """
+    sender = pre.fund_eoa()
+    call_cost_slot = 0
+
+    code = (
+        Op.GAS
+        + Op.BALANCE(address)
+        + Op.POP
+        + Op.SSTORE(call_cost_slot, Op.SUB(Op.SWAP1, Op.GAS))
+        + Op.STOP
+    )
+    before = pre.deploy_contract(code, storage={call_cost_slot: 0xDEADBEEF})
+    after = pre.deploy_contract(code, storage={call_cost_slot: 0xDEADBEEF})
+
+    # Block before fork
+    blocks = [
+        Block(
+            timestamp=10_000,
+            txs=[
+                Transaction(
+                    sender=sender,
+                    to=before,
+                    gas_limit=1_000_000,
+                )
+            ],
+        )
+    ]
+
+    # Block after fork
+    blocks += [
+        Block(
+            timestamp=20_000,
+            txs=[
+                Transaction(
+                    sender=sender,
+                    to=after,
+                    gas_limit=1_000_000,
+                )
+            ],
+        )
+    ]
+
+    predecessor = get_transition_fork_predecessor(fork)
+    successor = get_transition_fork_successor(fork)
+
+    def get_expected_gas(precompile_present: bool, fork: Fork) -> int:
+        gas_costs = fork.gas_costs()
+        warm_access_cost = gas_costs.G_WARM_ACCOUNT_ACCESS
+        cold_access_cost = gas_costs.G_COLD_ACCOUNT_ACCESS
+        extra_cost = gas_costs.G_BASE * 2 + gas_costs.G_VERY_LOW
+        if precompile_present:
+            return warm_access_cost + extra_cost
+        else:
+            return cold_access_cost + extra_cost
+
+    expected_gas_before = get_expected_gas(precompile_in_predecessor, predecessor)
+    expected_gas_after = get_expected_gas(precompile_in_successor, successor)
+
+    post = {
+        before: Account(storage={call_cost_slot: expected_gas_before}),
+        after: Account(storage={call_cost_slot: expected_gas_after}),
+    }
+
+    blockchain_test(
+        pre=pre,
+        post=post,
+        blocks=blocks,
+    )
--- a/tests/berlin/eip2929_gas_cost_increases/test_precompile_warming.py
+++ b/tests/berlin/eip2929_gas_cost_increases/test_precompile_warming.py
@@ -44,7 +44,7 @@ def precompile_addresses_in_predecessor_successor(
     successor_precompiles = set(get_transition_fork_successor(fork).precompiles())
     all_precompiles = successor_precompiles | predecessor_precompiles
     highest_precompile = int.from_bytes(max(all_precompiles))
-    extra_range = 5
+    extra_range = 32
     extra_precompiles = {
         Address(i) for i in range(highest_precompile + 1, highest_precompile + extra_range)
     }
--- a/tests/frontier/precompiles/test_precompiles.py
+++ b/tests/frontier/precompiles/test_precompiles.py
@@ -178,8 +178,8 @@ def test_precompile_warming(
         + Op.SSTORE(call_cost_slot, Op.SUB(Op.SWAP1, Op.GAS))
         + Op.STOP
     )
-    before = pre.deploy_contract(code, storage={0: 0xDEADBEEF})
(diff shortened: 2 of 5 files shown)


tests/frontier/precompiles/test_precompiles.py
```diff
@@ -109,3 +115,107 @@ def test_precompiles(
...
+            yield address, True
+        else:
+            yield address, False
+
+
+@pytest.mark.valid_at_transition_to("Paris", subsequent_forks=True)
+@pytest.mark.parametrize_by_fork(
+    "address,exists_in_predecessor", precompile_addresses_in_successor
+)
+def test_precompile_warming(
```
jochem-brouwer, line 151: This currently only tests precompiles which are marked as precompile. So far CancunToPrague it runs up to 0x00..0011. However we also want to include addresses on top of that. For BLS, the MULG1 and MULG2 precompiles were removed, so previously the precompile set ran up to 0x00..0013. We want to test that those also are now cold (NOT warm!)

I suggest for each precompile to just check a full range, for instance up to 0x00.0020

jochem-brouwer (commented):
This is a great start, some comments, because this mainly misses also checking that addresses which could have been precompiles in the past are now "normal" addresses (so should be cold)

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -109,3 +115,107 @@ def test_precompiles(
...
+        fork (Fork): The transition fork instance containing precompiled contract information.
+
+    Yields:
+        Iterator[Tuple[str, bool]]: A tuple containing the address in hexadecimal format and a
+            boolean indicating whether the address has existed in the predecessor.
+
+    """
+    predecessor_precompiles = [
+        hex(int.from_bytes(address, byteorder="big"))
+        for address in get_transition_fork_predecessor(fork).precompiles()
```
marioevz, line 134: Is there an issue with `pytest.mark.parametrize_by_fork` that requires a string or something else behind this?

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -109,3 +115,107 @@ def test_precompiles(
...
+
+    code = (
+        Op.GAS
+        + Op.BALANCE(address)
+        + Op.POP
+        + Op.SSTORE(call_cost_slot, Op.SUB(Op.SWAP1, Op.GAS))
+        + Op.STOP
+    )
+    before = pre.deploy_contract(code, storage={0: 0xDEADBEEF})
+    after = pre.deploy_contract(code, storage={0: 0xDEADBEEF})
```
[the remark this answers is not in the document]
marioevz: It's always a good idea to initialize with something different than zero if zero is one of the possible outcomes of the test, because if there's an out-of-gas situation, the zero from the oog is not confused with an expected outcome.

marioevz, line 177:
```suggestion
    before = pre.deploy_contract(code, storage={call_cost_slot: 0xDEADBEEF})
    after = pre.deploy_contract(code, storage={call_cost_slot: 0xDEADBEEF})
```

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -109,3 +115,107 @@ def test_precompiles(
...
+                    gas_limit=1_000_000,
+                )
+            ],
+        )
+    ]
+
+    if not exists_in_predecessor:
+        expected_gas_before = 2607
+    else:
+        expected_gas_before = 107
```
marioevz, line 210: Can be programmatic with `gas_costs = fork.gas_costs()`, and then the warm cost is `gas_costs.G_WARM_ACCOUNT_ACCESS` and the cold cost is `gas_costs.G_COLD_ACCOUNT_ACCESS`.

The extra 7 is Op.GAS (gas_costs.G_BASE), Op.PUSHx (gas_costs.G_VERY_LOW), and Op.POP (gas_costs.G_BASE).

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -109,3 +115,107 @@ def test_precompiles(
...
+
+    post = {
+        before: Account(storage={call_cost_slot: expected_gas_before}),
+        after: Account(storage={call_cost_slot: 107}),
```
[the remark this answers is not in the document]
marioevz: I think it should be `expected_gas_cold` and `expected_gas_warm`, and then `expected_gas_before` and `expected_gas_after` are equal to either of them depending on whether the precompile was added or removed.

marioevz (commented):
Some comments but overall looks great!

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -109,3 +115,107 @@ def test_precompiles(
...
+    ]
+    successor_precompiles = [
+        hex(int.from_bytes(address, byteorder="big"))
+        for address in get_transition_fork_successor(fork).precompiles()
+    ]
+    for address in successor_precompiles:
+        if address in predecessor_precompiles:
+            yield address, True
+        else:
+            yield address, False
```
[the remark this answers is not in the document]
danceratopz: Thanks should be easy to update for this!

tests/frontier/precompiles/test_precompiles.py
```diff
@@ -109,3 +115,107 @@ def test_precompiles(
...
+            boolean indicating whether the address has existed in the predecessor.
+
+    """
+    predecessor_precompiles = [
+        hex(int.from_bytes(address, byteorder="big"))
+        for address in get_transition_fork_predecessor(fork).precompiles()
+    ]
+    successor_precompiles = [
+        hex(int.from_bytes(address, byteorder="big"))
+        for address in get_transition_fork_successor(fork).precompiles()
```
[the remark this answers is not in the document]
danceratopz: The "fork" is `CancunToPragueAtTime15k`, it's a transition fork:
- predecessor is `Cancun`,
- successor is `Prague`.
That means, in terms of a blockchain test, blocks with a `timestamp<15000` are `Cancun` and `timestamp>=15000` are `Prague`.

Perhaps we can come up with better language? I think you've spotted that this terminology (at best) is only valid for an observer at timestamp=15000 :laughing:

tests/frontier/precompiles/test_precompiles.py
danceratopz: This just returns `Cancun` for `CancunToPragueAtTime15k`, it was introduced in https://github.com/ethereum/execution-spec-tests/pull/1313 where @spencer-tb fixed up how `state_test` gets named in the context of a transition test.

But if we just check a hard-coded range as @jochem-brouwer suggests above, then we don't need this "clever" logic would suggest converting it to a regular fork `state_test`.

tests/frontier/precompiles/test_precompiles.py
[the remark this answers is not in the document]
jochem-brouwer: Aha, this is a very good point, yes I agree let's initialize the storage!

spencer-tb (approved):
LGTM from myside! I moved the test to Berlin btw, where 2929 was introduced, like we do with the other tests.

marioevz:
Created an issue in evmone's repo for the CI failure: https://github.com/ethereum/evmone/issues/1210

We should be able to merge despite of this failure.

tests/berlin/eip2929_gas_cost_increases/test_precompile_warming.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Yields:
+        Iterator[Tuple[str, bool]]: A tuple containing the address in hexadecimal format and a
+            boolean indicating whether the address has existed in the predecessor.
+
+    """
+    predecessor_precompiles = set(get_transition_fork_predecessor(fork).precompiles())
+    successor_precompiles = set(get_transition_fork_successor(fork).precompiles())
+    all_precompiles = successor_precompiles | predecessor_precompiles
+    highest_precompile = int.from_bytes(max(all_precompiles))
```
felix314159, line 46: `from_bytes` requires two arguments, I had to change this line to:
`highest_precompile: int = int.from_bytes(max(all_precompiles), sys.byteorder)`
and import sys to make it work.

spencer-tb: when you say make it work you mean fill?

felix314159: Just having the file opened in an editor shows the problem. But now I know it depends on which Python version you use:
[python 3.11 -> optional parameter](https://docs.python.org/3.11/library/stdtypes.html#int.from_bytes)
[python 3.10 -> mandatory parameter](https://docs.python.org/3.10/library/stdtypes.html#int.from_bytes)

So I think it would be good if we find an exact Python version we enforce in EEST (everyone uses the exact same) to avoid stuff like this in the future.

spencer-tb: https://github.com/ethereum/execution-spec-tests/issues/1516

created an issue for discussion on enforcing the python version.

jochem-brouwer (commented):
Ah I only just re-checked this PR but wanted to note that this looks perfect! :star_struck: 

(besides one comment)

tests/berlin/eip2929_gas_cost_increases/test_precompile_warming.py
```diff
@@ -0,0 +1,147 @@
...
+    extra_range = 32
+    extra_precompiles = {
+        Address(i) for i in range(highest_precompile + 1, highest_precompile + extra_range)
+    }
+    all_precompiles = all_precompiles | extra_precompiles
+    for address in sorted(all_precompiles):
+        yield address, address in successor_precompiles, address in predecessor_precompiles
+
+
+@pytest.mark.valid_at_transition_to("Paris", subsequent_forks=True)
```
jochem-brouwer, line 56: Shouldn't this be Berlin? These tests should start to produce meaningful results once EIP 2929 is introduced. (One could also test before that of course, for forks introducing new precompiles but without 2929 active the gas difference would not depend upon warm/cold precompiles, but rather re-pricing of the opcodes used in the test (which are thus not subject of what we want to test here))

jochem-brouwer: Ah ok, before 2929 (Berlin) precompiles got added in Istanbul and Byzantium but those are thus not subject of the warm/cold tests. So marking it as "Paris" is fine here. Since Berlin, only the KZG (Cancun) and BLS (Prague) were introduced

PR #1497, opened 2025-04-25
zkevm: keccak worst-case

jsign:
This PR adds a test that builds a block to maximize the number of keccak permutations for a given gas limit (considering the max contract size).  

There are two main aspects of the test design.

### Find the optimal input size
The metric we care about is gas per permutation, not simply max opcode calls. Although the dynamic cost of `KECCAK256` is linear in input length, memory expansion is quadratic, so there is a sweet spot for minimal cost per permutation.

Instead of baking magic numbers, the test does a scan for the optimal cost and then does the attack with it. This is much more transparent on how this optimal length is found, plus it sounds also more future-proof.

Another way of understanding the above is by looking at this chart I created from that same script, with a more detailed output:
image

### The “attack” loop  

Once the optimal length is known, I create a loop that drives as many permutations as the block gas limit allows. The loop shouldn't be a tight loop, but quite the opposite -- run as many calls as possible within the max contract size, so the "JUMP"-like gas overhead is amortized as much as possible.

The general structure is:
```text
JUMPDEST
 (PUSH+PUSH+KECCAK256+POP)
 ...
PUSH0
JUMP
```

Note that for 36M you can fit the `(....)` without a loop since the max you can add fits within 24KiB contract limit. But for bigger gas limits, you run short with 24KiB and need the loop!

---

zkVM cycles: 36m gas -> 632 million cycles

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -10,7 +10,8 @@ import math
 import pytest
 
 from ethereum_test_forks import Fork
-from ethereum_test_tools import Alloc, Block, BlockchainTestFiller, Environment, Transaction
+from ethereum_test_tools import (Alloc, Block, BlockchainTestFiller,
+                                 Environment, Transaction)
 from ethereum_test_tools.vm.opcode import Opcodes as Op
 
 REFERENCE_SPEC_GIT_PATH = "TODO"
@@ -26,9 +27,6 @@ KECCAK_RATE = 136
     "gas_limit",
     [
         36_000_000,
-        60_000_000,
-        100_000_000,
-        300_000_000,
     ],
 )
 def test_worst_keccak(
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -85,12 +85,14 @@ def test_worst_keccak(
     #   Denominator = (PUSHN + PUSH1 + KECCAK256 + POP) + PUSH1_DATA + PUSHN_DATA
     # TODO: the testing framework uses PUSH1(0) instead of PUSH0 which is suboptimal for the
     # attack, whenever this is fixed adjust accordingly.
-    max_iters_loop = (MAX_CODE_SIZE - 3) // (4 + 1 + (optimal_input_length.bit_length() + 7) // 8)
+    start_code = Op.JUMPDEST + Op.PUSH20[optimal_input_length]
+    loop_code = Op.POP(Op.SHA3(Op.PUSH0, Op.DUP1))
+    end_code = Op.POP + Op.JUMP(Op.PUSH0)
+    max_iters_loop = (MAX_CODE_SIZE - (len(start_code) + len(end_code))) // len(loop_code)
     code = (
-        Op.JUMPDEST
-        + sum([Op.SHA3(0, optimal_input_length) + Op.POP] * max_iters_loop)
-        + Op.PUSH0
-        + Op.JUMP
+        start_code
+        + (loop_code * max_iters_loop)
+        + end_code
     )
     if len(code) > MAX_CODE_SIZE:
         # Must never happen, but keep it as a sanity check.
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -55,7 +55,8 @@ def test_worst_keccak(
             + math.ceil(i / 32) * gsc.G_KECCAK_256_WORD  # KECCAK256 dynamic cost
             + gsc.G_BASE  # POP
         )
-        # From the available gas, we substract the mem expansion costs considering we know the current input size length i.
+        # From the available gas, we substract the mem expansion costs considering we know the
+        # current input size length i.
         available_gas_after_expansion = max(0, available_gas - mem_exp_gas_calculator(new_bytes=i))
         # Calculate how many calls we can do.
         num_keccak_calls = available_gas_after_expansion // iteration_gas_cost
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -74,8 +74,8 @@ def test_worst_keccak(
     # Now calculate available gas for [attack iteration]:
     #   Numerator = MAX_CODE_SIZE-3. The -3 is for the JUMPDEST, PUSH0 and JUMP.
     #   Denominator = (PUSHN + PUSH1 + KECCAK256 + POP) + PUSH1_DATA + PUSHN_DATA
-    # TODO: the testing framework uses PUSH1(0) instead of PUSH0 which is suboptimal for the attack,
-    # whenever this is fixed adjust accordingly.
+    # TODO: the testing framework uses PUSH1(0) instead of PUSH0 which is suboptimal for the
+    # attack, whenever this is fixed adjust accordingly.
     max_iters_loop = (MAX_CODE_SIZE - 3) // (4 + 1 + (optimal_input_length.bit_length() + 7) // 8)
     code = (
         Op.JUMPDEST
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -1,0 +1,106 @@
+"""
+abstract: Tests zkEVMs worst-case compute scenarios.
+    Tests zkEVMs worst-case compute scenarios.
+
+Tests running worst-case compute opcodes and precompile scenarios for zkEVMs.
+"""
+
+import math
+
+import pytest
+
+from ethereum_test_forks import Fork
+from ethereum_test_tools import Alloc, Block, BlockchainTestFiller, Environment, Transaction
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "TODO"
+REFERENCE_SPEC_VERSION = "TODO"
+
+MAX_CODE_SIZE = 24 * 1024
+KECCAK_RATE = 136
+
+
+@pytest.mark.zkevm
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.parametrize(
+    "gas_limit",
+    [
+        36_000_000,
+        60_000_000,
+        100_000_000,
+        300_000_000,
+    ],
+)
+def test_worst_keccak(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_limit: int,
+):
+    """Test running a block with as many KECCAK256 permutations as possible."""
+    env = Environment(gas_limit=gas_limit)
+
+    # Intrinsic gas cost is paid once.
+    intrinsic_gasc_calculator = fork.transaction_intrinsic_cost_calculator()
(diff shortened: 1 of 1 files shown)


tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,108 @@
...
+    # max_iters_loop contains how many keccak calls can be done per loop.
+    # The loop is as big as possible bounded by the maximum code size.
+    #
+    # The loop structure is: JUMPDEST + [attack iteration] + PUSH0 + JUMP
+    #
+    # Now calculate available gas for [attack iteration]:
+    #   Numerator = MAX_CODE_SIZE-3. The -3 is for the JUMPDEST, PUSH0 and JUMP.
+    #   Denominator = (PUSHN + PUSH1 + KECCAK256 + POP) + PUSH1_DATA + PUSHN_DATA
+    # TODO: the testing framework uses PUSH1(0) instead of PUSH0 which is suboptimal for the attack,
+    # whenever this is fixed adjust accordingly.
```
jsign, line 78: To clarify a bit more here. When doing `Op.SHA3(0, ..)` the generated opcodes include a `PUSH1 0` which could save 1 gas if used `PUSH0`.

Note a huge deal, but just noticed that.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,108 @@
...
+def test_worst_keccak(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_limit: int,
+):
+    """Test running a block with as many KECCAK256 permutations as possible."""
+    env = Environment(gas_limit=gas_limit)
+
+    # Intrisic gas cost is paid once.
```
kevaundray, line 43:
```suggestion
    # Intrinsic gas cost is paid once.
```

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Intrinsic gas cost is paid once.
+    intrinsic_gasc_calculator = fork.transaction_intrinsic_cost_calculator()
+    available_gas = gas_limit - intrinsic_gasc_calculator()
+
+    gsc = fork.gas_costs()
+    mem_exp_gas_calculator = fork.memory_expansion_gas_calculator()
+
+    # Discover the optimal input size to maximize keccak-permutations, not keccak calls.
+    # The complication of the discovery arises from the non-linear gas cost of memory expansion.
```
jsign, line 51: As mentioned in the PR description, the TL;DR here is to find the optimal `input` for the `KECCAK(input)` calls later, so that we have the minimum gas usage per permutation.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,106 @@
...
+    max_keccak_perm_per_block = 0
+    optimal_input_length = 0
+    for i in range(1, 1_000_000, 32):
```
jsign, line 54: The 1_000_000 is arbitrary, should mainly be big enough to know that after some point the memory expansion cost is so big that clearly it won't be an option from that point forward. Thus, the optimal should exist before this point.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,106 @@
...
+        iteration_gas_cost = (
+            2 * gsc.G_VERY_LOW  # PUSHN + PUSH1
+            + gsc.G_KECCAK_256  # KECCAK256 static cost
+            + math.ceil(i / 32) * gsc.G_KECCAK_256_WORD  # KECCAK256 dynamic cost
+            + gsc.G_BASE  # POP
```
jsign, line 59: This is the gas costs of the attack iteration, see L82.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+        available_gas_after_expansion = max(0, available_gas - mem_exp_gas_calculator(new_bytes=i))
```
jsign, line 61: From the available gas, we substract the mem expansion costs considering we know the current input size length `i`.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,106 @@
...
+        num_keccak_calls = available_gas_after_expansion // iteration_gas_cost
```
jsign, line 62: Calculate how many calls we can do.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,106 @@
...
+        num_keccak_permutations = num_keccak_calls * math.ceil(i / KECCAK_RATE)
```
jsign, line 63: In KECCAK you do 1 permutation every 136 bytes.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if num_keccak_permutations > max_keccak_perm_per_block:
+            max_keccak_perm_per_block = num_keccak_permutations
+            optimal_input_length = i
```
jsign, line 67: If we found an input size that is better (reg permutations/gas), then save it.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,106 @@
...
+    #   Denominator = (PUSHN + PUSH1 + KECCAK256 + POP) + PUSH1_DATA + PUSHN_DATA
+    # TODO: the testing framework uses PUSH1(0) instead of PUSH0 which is suboptimal for the
+    # attack, whenever this is fixed adjust accordingly.
+    max_iters_loop = (MAX_CODE_SIZE - 3) // (4 + 1 + (optimal_input_length.bit_length() + 7) // 8)
+    code = (
+        Op.JUMPDEST
+        + sum([Op.SHA3(0, optimal_input_length) + Op.POP] * max_iters_loop)
+        + Op.PUSH0
+        + Op.JUMP
+    )
```
jsign, line 85: TL;DR of this is that we want to have as-big loop as possible. i.e., `attack-contract=JUMPDEST+[attack iteration]+PUSH0 + JUMP`. Said differently one big loop to amortize as much as possible the JUMP-cost.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,103 @@
...
+MAX_CODE_SIZE = 24 * 1024
+KECCAK_RATE = 136
+
+
+@pytest.mark.zkevm
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.parametrize(
+    "gas_limit",
+    [
+        36_000_000,
```
marioevz, line 28: Just as a heads up, we are planning on making this value a parameter to `fill` in this PR: https://github.com/ethereum/execution-spec-tests/pull/1470

The idea is that these tests that are intended to consume the entire block's gas limit can be configured from the script caller.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,103 @@
...
+    pre: Alloc,
+    fork: Fork,
+    gas_limit: int,
+):
+    """Test running a block with as many KECCAK256 permutations as possible."""
+    env = Environment(gas_limit=gas_limit)
+
+    # Intrinsic gas cost is paid once.
+    intrinsic_gasc_calculator = fork.transaction_intrinsic_cost_calculator()
+    available_gas = gas_limit - intrinsic_gasc_calculator()
```
marioevz, line 42:
```suggestion
    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
    available_gas = gas_limit - intrinsic_gas_calculator()
```
Small typo.

tests/zkevm/test_worst_compute.py
marioevz: How long does it currently take to make this calculation?

tests/zkevm/test_worst_compute.py
marioevz: I think these github comments should be added as code comments.

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,103 @@
...
+    # The loop is as big as possible bounded by the maximum code size.
+    #
+    # The loop structure is: JUMPDEST + [attack iteration] + PUSH0 + JUMP
+    #
+    # Now calculate available gas for [attack iteration]:
+    #   Numerator = MAX_CODE_SIZE-3. The -3 is for the JUMPDEST, PUSH0 and JUMP.
+    #   Denominator = (PUSHN + PUSH1 + KECCAK256 + POP) + PUSH1_DATA + PUSHN_DATA
+    # TODO: the testing framework uses PUSH1(0) instead of PUSH0 which is suboptimal for the
+    # attack, whenever this is fixed adjust accordingly.
+    max_iters_loop = (MAX_CODE_SIZE - 3) // (4 + 1 + (optimal_input_length.bit_length() + 7) // 8)
```
marioevz, line 76: I'm beginning to think that, with all these tests that rely so much in the maximum contract size, this value should be a property of the fork, in case we bump the max size at some point, it stays coherent with any max block gas limits we also increase.

`fork.max_contract_size()` perhaps?

tests/zkevm/test_worst_compute.py
```diff
@@ -0,0 +1,103 @@
...
+    code = (
+        Op.JUMPDEST
+        + sum([Op.SHA3(0, optimal_input_length) + Op.POP] * max_iters_loop)
+        + Op.PUSH0
+        + Op.JUMP
+    )
```
marioevz, line 82: I think we can have a tiny improvement by leaving `optimal_input_length` at the top of the stack before starting the `Op.SHA3` loop and then use `Op.DUPx` to fetch it instead of using `Op.PUSHx` over and over.
```suggestion
    start_code = Op.JUMPDEST + Op.PUSH20[optimal_input_length]
    loop_code = Op.POP(Op.SHA3(Op.PUSH0, Op.DUP1))
    end_code = Op.POP + Op.JUMP(Op.PUSH0)
    max_iters_loop = (MAX_CODE_SIZE - (len(start_code) + len(end_code))) // len(loop_code)
    code = (
        start_code
        + (loop_code * max_iters_loop)
        + end_code
    )
```
Or something like that. `Op.PUSH20` could be reduced probably, I just used an arbitrarily large push opcode.

marioevz (commented):
Overall looks good and just have a few comments, thanks!

tests/zkevm/test_worst_compute.py
jsign: Measured now: ~25ms.

tests/zkevm/test_worst_compute.py
jsign: It sounds helpful to know the current one, for sure.
But I wonder if we could still make this dynamic, since many of these tests can help understand how the worst-case scenario evolves if we plan to bump the max contract size.

tests/zkevm/test_worst_compute.py
jsign: Added!

kevaundray (approved):

marioevz (approved):
LGTM, thanks!

PR #1504, opened 2025-04-28
fix: review comments for #1495

marioevz:
## 🗒️ Description
Took the liberty to implement the fixes in the review for #1495 since Dan is OOO.

The IDs are kinda ugly right now:
```
tests/frontier/precompiles/test_precompiles.py::test_precompile_warming[fork_ParisToShanghaiAtTime15k-address_0x0000000000000000000000000000000000000001-precompile_in_successor_True-precompile_in_predecessor_True-blockchain_test]
```

And we could reduce it to only print the address perhaps, but let me know what you think.


spencer-tb (approved):
LGTM, can fix the rest in Dan's PR.

PR #1510, opened 2025-04-29
chore: add osaka engine updates

spencer-tb:
## 🗒️ Description
Following: https://github.com/ethereum/execution-apis/pull/630

I don't think we needed to change much on our end but please double check. I added the `get_blobs` engine rpc method here on the assumption that we can maybe create a test suite that sends blob txs to ELs, then we call `get_blobs` to validate the response. This feels suited towards an extension in `execute`.

## 🔗 Related Issues
N/A.

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_rpc/types.py
+++ b/src/ethereum_test_rpc/types.py
@@ -154,4 +154,4 @@ class GetPayloadResponse(CamelModel):
 class GetBlobsResponse(CamelModel):
     """Represents the response of a get blobs request."""
 
-    result: List[BlobAndProof]
+    result: List[BlobAndProof | None]
--- a/src/ethereum_test_rpc/types.py
+++ b/src/ethereum_test_rpc/types.py
@@ -140,7 +140,8 @@ class BlobAndProof(CamelModel):
     """Represents a blob and proof structure."""
 
     blob: Bytes
-    proofs: List[Bytes]
+    proofs: List[Bytes] | None = None
+    proof: Bytes | None= None
 
 
 class GetPayloadResponse(CamelModel):
--- a/src/ethereum_test_rpc/types.py
+++ b/src/ethereum_test_rpc/types.py
@@ -140,8 +140,9 @@ class BlobAndProof(CamelModel):
     """Represents a blob and proof structure."""
 
     blob: Bytes
-    proofs: List[Bytes] | None = None
-    proof: Bytes | None= None
+    proofs: List[Bytes] | None = None  # >= Osaka (V2)
+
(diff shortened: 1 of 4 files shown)


marioevz:
> test suite that sends blob txs to ELs, then we call `get_blobs` to validate the response. This feels suited towards an extension in `execute`.

This could even replace most of hive simulator tests if possible. The main issue right now is that most of the time we don't have full blobs in our tests but we only put the versioned hashes in the transaction. But perhaps we could put full blobs in the test transactions if we are running in execute mode (we could create a "mode" fixture that is populated with "fill" or "execute" depending on the pytest plugin).

src/ethereum_test_rpc/types.py
```diff
@@ -136,9 +136,22 @@ def blob_versioned_hashes(self) -> List[Hash]:
...
 
     execution_payload: FixtureExecutionPayload
     blobs_bundle: BlobsBundle | None = None
     execution_requests: List[Bytes] | None = None
+
+
+class GetBlobsResponse(CamelModel):
+    """Represents the response of a get blobs request."""
+
+    result: List[BlobAndProof]
```
marioevz, line 157:
```suggestion
    result: List[BlobAndProof | None]
```
Reading from the "or null in case of any missing blobs" in https://github.com/ethereum/execution-apis/blob/main/src/engine/osaka.md#response-1

src/ethereum_test_rpc/types.py
```diff
@@ -136,9 +136,22 @@ def blob_versioned_hashes(self) -> List[Hash]:
         return [Hash(b"\1" + commitment[1:]) for commitment in self.commitments]
 
 
+class BlobAndProof(CamelModel):
+    """Represents a blob and proof structure."""
+
+    blob: Bytes
+    proofs: List[Bytes]
```
marioevz, line 143: We might need to implement `BlobAndProofV1` too to verify Cancun/Prague from this: https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#blobandproofv1

So we can have:
```suggestion
class BlobAndProof(CamelModel):
    """Represents a blob and proof structure."""

    blob: Bytes
    proofs: List[Bytes] | None = None
    proof: Bytes | None= None
```
Or as an alternative we could simply define `class BlobAndProofV1` and `class BlobAndProofV2` each with the expected fields.

marioevz (commented):
Thanks for this! Just a couple of comments.

marioevz (approved):
LGTM, thanks!

PR #1521, opened 2025-05-01
zkEVM: fix max contract size limitation in bytecode attack

jsign:
The previous bytecode-attack test relied on a contiguous PUSH20 + EXTCODESIZE pattern that embedded each 20-byte target address directly in the bytecode. Embedding the addresses with PUSH20 is far cheaper than loading them from storage, so it lets us amplify the attack.

However, contract-size limits cap the number of iterations we can fit in 24 KiB (or even 256 KiB), so the approach runs out of space long before it bumps into the block gas limit. For example, for 36M gas limit, you're limited by the 24KiB max size limit.

This strategy removes that bottleneck:
- Target deployment via CREATE2: The factory now spawns target contracts with CREATE2, incrementing a seed to derive unique addresses.
- Cheap address recalculation: because a CREATE2 address is keccak256(0xff || sender || salt || init_code), the attack contract can recompute each address inside the EVM for only a few gas with the KECCAK256 opcode.
- The attack contract iterates through the seeds, recalculates every target address on the fly, and triggers the attack—all without storing addresses, decoding calldata, or using chained CALLs.

This CREATE2-based loop is both more gas-efficient and less complex than falling back to SLOAD, CALLDATA, or multi-hop calls.

### DIFF
--- a/tests/zkevm/test_worst_bytecode.py
+++ b/tests/zkevm/test_worst_bytecode.py
@@ -172,8 +172,8 @@ def test_worst_bytecode_single_opcode(
         + Op.MSTORE(64, initcode.keccak256())
         # Main loop
         + While(
-            body=Op.EXTCODESIZE(Op.SHA3(32 - 20 - 1, 85)) + Op.MSTORE(32, Op.ADD(Op.MLOAD(32), 1)),
-            condition=Op.PUSH1(1),
+            body=Op.POP(Op.EXTCODESIZE(Op.SHA3(32 - 20 - 1, 85)))
+            + Op.MSTORE(32, Op.ADD(Op.MLOAD(32), 1)),
         )
     )
 
--- a/tests/zkevm/test_worst_bytecode.py
+++ b/tests/zkevm/test_worst_bytecode.py
@@ -8,17 +8,9 @@ Tests for zkEVMs worst-cases scenarios.
 import pytest
 
 from ethereum_test_forks import Fork
-from ethereum_test_tools import (
-    Account,
-    Alloc,
-    Block,
-    BlockchainTestFiller,
-    Environment,
-    Hash,
-    Transaction,
-    While,
-    compute_create_address,
-)
+from ethereum_test_tools import (Account, Alloc, Block, BlockchainTestFiller,
+                                 Environment, Hash, Transaction, While,
+                                 compute_create2_address)
 from ethereum_test_tools.vm.opcode import Opcodes as Op
 
 REFERENCE_SPEC_GIT_PATH = "TODO"
@@ -33,7 +25,6 @@ XOR_TABLE_SIZE = 256
 XOR_TABLE = [Hash(i).sha256() for i in range(XOR_TABLE_SIZE)]
 
 
-# TODO: Parametrize for EOF
 @pytest.mark.zkevm
 @pytest.mark.parametrize(
     "opcode",
@@ -95,12 +86,14 @@ def test_worst_bytecode_single_opcode(
         )
         + Op.MSTORE(
             0,
-            Op.CREATE(
+            Op.CREATE2(
                 value=0,
                 offset=0,
                 size=Op.MSIZE,
+                salt=Op.SLOAD(0),
             ),
         )
+        + Op.SSTORE(0,Op.ADD(Op.SLOAD(0), 1))
         + Op.RETURN(0, 32)
     )
     factory_address = pre.deploy_contract(code=factory_code)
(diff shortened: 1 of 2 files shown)


tests/zkevm/test_worst_bytecode.py
```diff
@@ -95,12 +96,14 @@ def test_worst_bytecode_single_opcode(
...
             0,
-            Op.CREATE(
+            Op.CREATE2(
                 value=0,
                 offset=0,
                 size=Op.MSIZE,
+                salt=Op.SLOAD(0),
             ),
         )
+        + Op.SSTORE(0, Op.ADD(Op.SLOAD(0), 1))
```
jsign, line 106: Use CREATE2 using a seed in storage, to make this predictable for the attacker.

tests/zkevm/test_worst_bytecode.py
```diff
@@ -115,9 +118,18 @@ def test_worst_bytecode_single_opcode(
...
     intrinsic_gas_cost_calc = fork.transaction_intrinsic_cost_calculator()
-    max_number_of_contract_calls = (OPCODE_GAS_LIMIT - intrinsic_gas_cost_calc()) // (
-        gas_costs.G_VERY_LOW + gas_costs.G_BASE + gas_costs.G_COLD_ACCOUNT_ACCESS
+    loop_cost = (
+        gas_costs.G_KECCAK_256  # KECCAK static cost
+        + math.ceil(85 / 32) * gas_costs.G_KECCAK_256_WORD  # KECCAK dynamic cost for CREATE2
+        + gas_costs.G_VERY_LOW * 3  # ~MSTOREs+ADDs
+        + gas_costs.G_COLD_ACCOUNT_ACCESS  # EXTCODESIZE
+        + 30  # ~Gluing opcodes
     )
```
jsign, line 127: This is an ~estimation of gas usage in the loop in L174-L177.
The ~30 is mainly to generate some slack.

tests/zkevm/test_worst_bytecode.py
```diff
@@ -115,9 +118,18 @@ def test_worst_bytecode_single_opcode(
...
+    max_number_of_contract_calls = (
+        # Base available gas = GAS_LIMIT - intrinsic - (out of loop MSTOREs)
+        OPCODE_GAS_LIMIT - intrinsic_gas_cost_calc() - gas_costs.G_VERY_LOW * 4
```
jsign, line 130: The `gas_costs.G_VERY_LOW*4` is from L169-172.

tests/zkevm/test_worst_bytecode.py
```diff
@@ -144,16 +156,27 @@ def generate_deploy_tx(contracts_to_deploy: int):
...
         post[deployed_contract_address] = Account(nonce=1)
         deployed_contract_addresses.append(deployed_contract_address)
 
     opcode_code = (
-        sum(Op.POP(opcode(address=address)) for address in deployed_contract_addresses) + Op.STOP
+        # Setup memory for later CREATE2 address generation loop.
+        Op.MSTORE(0, factory_address)
+        + Op.MSTORE8(32 - 20 - 1, 0xFF)  # 0xFF prefix byte
+        + Op.MSTORE(32, 0)
+        + Op.MSTORE(64, initcode.keccak256())
```
jsign, line 172: Only once we setup the `0xFF+address+salt+hash(initcode)`. The loop will reuse this memory region for generating the CREATE2 addresses.

tests/zkevm/test_worst_bytecode.py
```diff
@@ -144,16 +156,27 @@ def generate_deploy_tx(contracts_to_deploy: int):
...
+        # Main loop
+        + While(
+            body=Op.EXTCODESIZE(Op.SHA3(32 - 20 - 1, 85)) + Op.MSTORE(32, Op.ADD(Op.MLOAD(32), 1)),
```
jsign, line 175: Generate and make the opcode call. After that, modify the specific memory region corresponding to the `seed` with the new value for the next loop.

tests/zkevm/test_worst_bytecode.py
```diff
@@ -144,16 +156,27 @@ def generate_deploy_tx(contracts_to_deploy: int):
...
+            condition=Op.PUSH1(1),
```
jsign, line 176: Run this loop forever using all available gas.
This `condition=` isn't super ideal since `While` does some `JUMPI`.

I could avoid using `While` and do `JUMPDEST+loop+JUMP` but it is a bit messy since I would have to also do `Op.PC` and do some calcs for the right jump target. I think usin `While` is simply much better and understandable.

marioevz (approved):
LGTM, thanks!

I pushed a fix to the `While` generator to allow no condition and jump unconditionally.

marioevz:
I've analyzed the coverage result and it's because of the CREATE -> CREATE2 change.

Merging now.

PR #1524, opened 2025-05-01
zkevm: add ECRECOVER, SHA2-256, IDENTITY and RIPEMD precompiles

jsign:
This PR:
- Adds tests four new precompiles: ECRECOVER, SHA2-256, IDENTITY, and RIPEMD.
- Removes `zkevm` pytest markers since the testing framework now automatically does this.
- Uses `Environment()` configured gas limits, which are expected to be correctly set via the new `--block-gas-limit` filling flag.
- I updated the CI pipeline to use 36M gas limit for filling.

Cycle counts in the new precompiles:
- ECRECOVER: 682 million cycles
- SHA2-256: 5.8 billion cycles
- IDENTITY: 305 million cycles
- RIPEMD: 360 million cycles

The last three are all part of a generic test since these precompiles have the same input parameter structure (i.e., `data`). The generic test automatically discovers the optimal input length for each case to amortize the static cost as much as possible against the quadratic cost of memory expansions. ECRECOVER is a separate test since the parameter input structure is particular.

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -11,6 +11,7 @@ import pytest
 
 from ethereum_test_forks import Fork
 from ethereum_test_tools import (
+    Address,
     Alloc,
     Block,
     BlockchainTestFiller,
@@ -103,6 +104,94 @@ def test_worst_keccak(
         sender=pre.fund_eoa(),
         data=[],
         value=0,
+    )
+
+    blockchain_test(
+        env=env,
+        pre=pre,
+        post={},
+        blocks=[Block(txs=[tx])],
+    )
+
+
+@pytest.mark.zkevm
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.parametrize(
+    "gas_limit",
+    [
+        Environment().gas_limit,
+    ],
+)
+@pytest.mark.parametrize(
+    "address,static_cost,per_word_dynamic_cost,bytes_per_unit_of_work",
+    [
+        pytest.param(0x02, 60, 12, 64, id="SHA2-256"),
+        pytest.param(0x03, 600, 120, 64, id="RIPEMD-160"),
+        pytest.param(0x04, 15, 3, 1, id="IDENTITY"),
+    ],
+)
+def test_worst_precompile_only_data_input(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_limit: int,
+    address: Address,
+    static_cost: int,
+    per_word_dynamic_cost: int,
+    bytes_per_unit_of_work: int,
+):
+    """Test running a block with as many precompile calls which have a single `data` input."""
+    env = Environment(gas_limit=gas_limit)
+
+    # Intrinsic gas cost is paid once.
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+    available_gas = gas_limit - intrinsic_gas_calculator()
+
+    gsc = fork.gas_costs()
+    mem_exp_gas_calculator = fork.memory_expansion_gas_calculator()
+
+    # Discover the optimal input size to maximize precompile calls, not precompile permutations.
+    max_work = 0
+    optimal_input_length = 0
+    for input_length in range(1, 1_000_000, 32):
+        staticcall_parameters_gas = (
+            gsc.G_BASE  # PUSH0 = arg offset
+            + gsc.G_BASE  # PUSH0 = arg size
+            + gsc.G_BASE  # PUSH0 = arg size
+            + gsc.G_VERY_LOW  # PUSH0 = arg offset
+            + gsc.G_VERY_LOW  # PUSHN = address
+            + gsc.G_BASE  # GAS
+        )
+        iteration_gas_cost = (
+            staticcall_parameters_gas
+            + +static_cost  # Precompile static cost
+            + math.ceil(input_length / 32) * per_word_dynamic_cost  # Precompile dynamic cost
+            + gsc.G_BASE  # POP
+        )
+        # From the available gas, we substract the mem expansion costs considering we know the
+        # current input size length.
+        available_gas_after_expansion = max(
+            0, available_gas - mem_exp_gas_calculator(new_bytes=input_length)
+        )
+        # Calculate how many calls we can do.
+        num_calls = available_gas_after_expansion // iteration_gas_cost
+        total_work = num_calls * math.ceil(input_length / bytes_per_unit_of_work)
+
+        # If we found an input size that is better (reg permutations/gas), then save it.
+        if total_work > max_work:
+            max_work = total_work
+            optimal_input_length = input_length
+
+    calldata = Op.CODECOPY(0, 0, optimal_input_length)
+    attack_block = Op.POP(Op.STATICCALL(Op.GAS, address, 0, optimal_input_length, 0, 0))
+    code = code_loop_precompile_call(calldata, attack_block)
+
+    code_address = pre.deploy_contract(code=code)
+
+    tx = Transaction(
+        to=code_address,
+        gas_limit=gas_limit,
+        sender=pre.fund_eoa(),
     )
 
     blockchain_test(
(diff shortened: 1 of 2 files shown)


jsign:
Weirdish [linter failure](https://github.com/ethereum/execution-spec-tests/actions/runs/14784221957/job/41509526179?pr=1524).

tests/zkevm/test_worst_compute.py
```diff
@@ -105,6 +114,94 @@ def test_worst_keccak(
     )
 
 
+@pytest.mark.zkevm
```
danceratopz, line 117: Hey, the `zkevm` marker is now automatically to all tests under `./tests/zkevm`, see:
- #1534
```suggestion
```
It can also be removed below.

jsign: Cool, I'll remove the mark in all existing places. Thank you!

tests/zkevm/test_worst_compute.py
```diff
@@ -105,7 +114,93 @@ def test_worst_keccak(
...
+    [
+        Environment().gas_limit,
+    ],
+)
+@pytest.mark.parametrize(
+    "address,static_cost,per_word_dynamic_cost,bytes_per_unit_of_work",
+    [
+        pytest.param(0x02, 60, 12, 64, id="SHA2-256"),
+        pytest.param(0x03, 600, 120, 64, id="RIPEMD-160"),
+        pytest.param(0x04, 15, 3, 1, id="IDENTITY"),
```
jsign, line 129: `bytes_per_unit_of_work` is mainly a concept of for how many bytes of input the precompile has to do "computational work". In the case of hashes, is related to the block size of the compression function. For identity is "1" since work is linear (i.e., not "chunked")

cc @kevaundray

tests/zkevm/test_worst_compute.py
```diff
@@ -105,7 +114,93 @@ def test_worst_keccak(
...
+    "gas_limit",
+    [
+        Environment().gas_limit,
+    ],
+)
+@pytest.mark.parametrize(
+    "address,static_cost,per_word_dynamic_cost,bytes_per_unit_of_work",
+    [
+        pytest.param(0x02, 60, 12, 64, id="SHA2-256"),
+        pytest.param(0x03, 600, 120, 64, id="RIPEMD-160"),
```
jsign, line 128: I didn't find a way to pull the static and dynamic costs from precompiles. Is there a way to do this? Mostly asking for future-proofness

chfast (commented):
These are not zkevm specific but good general benchmarks for EVM. Can we make this framework for benchmarks a bit more generic? I'd like to contribute my existing benchmarks and more.

jsign:
@marioevz, can you take a look at this PR? Since I'm touching the filling command, I would prefer to have some quick check from you if sounds good!

tests/zkevm/test_worst_compute.py
```diff
@@ -105,22 +100,94 @@ def test_worst_keccak(
...
+        total_work = num_calls * math.ceil(input_length / bytes_per_unit_of_work)
+
+        # If we found an input size that is better (reg permutations/gas), then save it.
+        if total_work > max_work:
+            max_work = total_work
+            optimal_input_length = input_length
+
+    calldata = Op.CODECOPY(0, 0, optimal_input_length)
+    attack_block = Op.POP(Op.STATICCALL(Op.GAS, address, 0, optimal_input_length, 0, 0))
+    code = code_loop_precompile_call(calldata, attack_block)
```
jsign, line 165: The `code_loop_precompile_call` is basically a big loop doing the calls. When this code was written the `While` primitive didn't exist, but actually what `code_loop_precompile_call` is doing is something better since we amortize "loop related" opcodes as much as possible. 

While useful, at some point we can simplify it losing some efficiency if that makes sense.

tests/zkevm/test_worst_compute.py
```diff
@@ -144,24 +211,52 @@ def test_worst_modexp(
...
+    """Test running a block with as many ECRECOVER calls as possible."""
+    env = Environment()
+
+    # Calldata
+    calldata = (
+        Op.MSTORE(0 * 32, 0x38D18ACB67D25C8BB9942764B62F18E17054F66A817BD4295423ADF9ED98873E)
+        + Op.MSTORE(1 * 32, 27)
+        + Op.MSTORE(2 * 32, 0x38D18ACB67D25C8BB9942764B62F18E17054F66A817BD4295423ADF9ED98873E)
+        + Op.MSTORE(3 * 32, 0x789D1DD423D25F0772D2748D60F7E4B81BB14D086EBA8E8E8EFB6DCFF8A4AE02)
+    )
```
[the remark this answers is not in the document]
jsign: BTW, whenever this test is merged many of these existing tests will be generalized in a much simpler setup in https://github.com/ethereum/execution-spec-tests/pull/1581.

But let's merge this PR as it is first, and then I'll rebase that other pr and we can continue there.

tests/zkevm/test_worst_compute.py
raxhvl: We could add it to the gas table similar [keccak](https://github.com/ethereum/execution-spec-tests/blob/328b7f49f56e0a1b05d915f4e6a918e8391d22db/src/ethereum_test_forks/forks/forks.py#L122).

tests/zkevm/test_worst_compute.py
```diff
@@ -144,24 +218,52 @@ def test_worst_modexp(
...
+
+    attack_block = Op.POP(Op.STATICCALL(ECRECOVER_GAS_COST, 0x1, 0, 32 * 4, 0, 0))
```
raxhvl, line 256:
```suggestion
    attack_block = Op.POP(Op.STATICCALL(ECRECOVER_GAS_COST, 0x1, 0, 32 * 4, 0, 32))
```
If we write result of `ecrecover` back to memory hash gets overridden, thus we get fresh input for each call without much complexity.

jsign: Yep, we can try doing that low-hanging fruit and double check the cycles.
For now I'll prefer to merge this PR as is since there's another one buildling on top https://github.com/ethereum/execution-spec-tests/pull/1581.

I can try that change before opening that one.

tests/zkevm/test_worst_compute.py
jsign: That would be quite useful yes.

tests/zkevm/test_worst_compute.py
```diff
@@ -90,7 +92,7 @@ def test_worst_keccak(
 
     tx = Transaction(
         to=code_address,
-        gas_limit=gas_limit,
+        gas_limit=env.gas_limit,
         gas_price=10,
```
marioevz, line 96: Can we take the opportunity to remove "gas_price" from this and next test?

tests/zkevm/test_worst_compute.py
```diff
@@ -144,24 +218,52 @@ def test_worst_modexp(
...
+
+    attack_block = Op.POP(Op.STATICCALL(ECRECOVER_GAS_COST, 0x1, 0, 32 * 4, 0, 0))
+    code = code_loop_precompile_call(calldata, attack_block)
+    code_address = pre.deploy_contract(code=bytes(code))
 
     tx = Transaction(
         to=code_address,
-        gas_limit=gas_limit,
+        gas_limit=env.gas_limit,
+        gas_price=10,
```
marioevz, line 263:
```suggestion
```
Here too.

tests/zkevm/test_worst_compute.py
```diff
@@ -144,24 +218,52 @@ def test_worst_modexp(
...
         sender=pre.fund_eoa(),
+        data=[],
+        value=0,
```
marioevz, line 266:
```suggestion
```
Not really used.

marioevz (approved):
Two minor comments, thanks!

tests/zkevm/test_worst_compute.py
jsign: Yep, I'm doing this in an [upcoming PR](https://github.com/ethereum/execution-spec-tests/pull/1611/files#diff-cdcef4e8328e54064d53a63c2653db5f0eeab8030d09db345b118bcd87b0cf50L39) which has other changes.

All `gas_limit` parametrizations will be gone.

jsign: In your honor, I'll do it this one in particular you pointed out in this PR :D

PR #1525, opened 2025-05-02
chore(tooling): bump ruff and questionary and fix vsc ruff warning and remove non-existent rule SC200

felix314159:
## 🗒️ Description
I think the warning `WARN ruff_server::server::api Received notification $/setTrace which does not have a handler.` is the reason we kept seeing ruff warnings at every vsc startup, this can be ignored so setting the warn level from warn to error resolves this annoyance. I also bumped ruff to the latest version and did the necessary fixes (SC200 rule does not exist), this was spotted via `uv run ruff check --no-fix --show-fixes src tests docs .github/scripts` and manually resolved. Do we even use questionary package? I just saw the comment and bumped it to newer version.

## ✅ Checklist


felix314159:
To clarify commit [e6d6336](https://github.com/ethereum/execution-spec-tests/pull/1525/commits/e6d6336f93184ad4b05f2289066eb45e92e5038c): running `uv run pytest -c ./pytest-framework.ini` fails a few solc related tests on ARM. It ends up trying to use some x86 solc, these adjustments aim to ensure that the same solc binary is used that is used everywhere else.

marioevz:
> To clarify commit [e6d6336](https://github.com/ethereum/execution-spec-tests/pull/1525/commits/e6d6336f93184ad4b05f2289066eb45e92e5038c): running `uv run pytest -c ./pytest-framework.ini` fails a few solc related tests on ARM. It ends up trying to use some x86 solc, these adjustments aim to ensure that the same solc binary is used that is used everywhere else. If you only want to run a subset of tests (e.g. only solc test) use this: `uv run pytest src/pytest_plugins/solc/tests/test_solc.py -c ./pytest-framework.ini`

I still don't quite understand this commit, are changes in src/pytest_plugins/solc/tests/test_solc.py meant to be included or were they debugging changes?

Locally, the test fails for me in arm with `e6d6336f93184ad4b05f2289066eb45e92e5038c`.

cc @danceratopz @spencer-tb I think you are more familiar with these tests than I am.

felix314159:
> > To clarify commit [e6d6336](https://github.com/ethereum/execution-spec-tests/pull/1525/commits/e6d6336f93184ad4b05f2289066eb45e92e5038c): running `uv run pytest -c ./pytest-framework.ini` fails a few solc related tests on ARM. It ends up trying to use some x86 solc, these adjustments aim to ensure that the same solc binary is used that is used everywhere else. If you only want to run a subset of tests (e.g. only solc test) use this: `uv run pytest src/pytest_plugins/solc/tests/test_solc.py -c ./pytest-framework.ini`
> 
> I still don't quite understand this commit, are changes in src/pytest_plugins/solc/tests/test_solc.py meant to be included or were they debugging changes?
> 
> Locally, the test fails for me in arm with `e6d6336f93184ad4b05f2289066eb45e92e5038c`.
> 
> cc @danceratopz @spencer-tb I think you are more familiar with these tests than I am.

It worked on my machine but it is not a robust way of doing things so let's not merge this for now if it does not work for you. I think the root cause of these issues is summarized [here](https://github.com/ethereum/solidity/issues/11351#issuecomment-1066104371) and the best solution is to push solidity team to publish arm linux binaries so that solc-select can be patched to start working as expected on every (commonly-used) platform

pyproject.toml
```diff
@@ -45,12 +45,8 @@ dependencies = [
     "types-pyyaml>=6.0.12.20240917,<7",
     "pytest-json-report>=1.5.0,<2",
     "typing-extensions>=4.12.2,<5",
-    # TODO: bump questionary to a newer release, when it becomes available.
-    # The current release questionary 2.0.1 requires `prompt_toolkit = ">=2.0,<=3.0.36"`.
-    # This conflicts with ipython; while not an EEST dependency, ipython is a very useful tool:
-    # https://eest.ethereum.org/main/dev/interactive_usage/
-    "questionary @ git+https://github.com/tmbo/questionary@ff22aeae1cd9c1c734f14329934e349bec7873bc",
-    "prompt_toolkit>=3.0.48,<4", # ensure we have a new enough version for ipython
+    "questionary>=2.1.0",
```
danceratopz, line 48: Although we lock packages to an explicit version when we use EEST internally (locally, in editable mode), this is not the case when we publish EEST as a package so I think it's best to protect against breaking changes in new major versions by capping the max version:
```suggestion
    "questionary>=2.1.0,<3",
```

Please don't accept the suggest, but rather run:
`uv add "questionary>=2.1.0,<3"`

danceratopz (changes requested):
Good work spotting these problems and thanks a lot for tackling them!

There's too many different types of changes in this PR for my taste. I'd prefer if dependency changes occur in their own dedicated PR when possible. If there's a change in behavior, it then becomes easier to track the source down and it can be easily reverted in an understandable way (via git revert`) without undoing other work.

In particular, the unit test changes don't fit within the scope of the other changes.

Could we just apply the dep changes (and if you like, the ruff config changes) in this PR and move the unit test changes to a separate PR?

felix314159:
All done, moved the yul and solc related stuff out of this PR. Thanks for reviewing!

danceratopz (approved):
Thanks @felix314159! LGTM! 

I just re-generated `uv.lock` with an older version of `uv` to avoid this commit every version field in `uv.lock` getting tainted with new "upload-time" entries, which makes it very difficult to see which versions were actually changed. The `upload-time` field was introduced in `uv` 0.7.0, https://github.com/astral-sh/uv/releases/tag/0.7.0.

danceratopz:
@felix314159 could you please double-check the dependency changes I added to `uv.lock` are as you intended?

PR #1537, opened 2025-05-05
feat(cli,fill): make the eip version checker a separate pytest cli tool

### LINKED ISSUE #533: bug(fill): fix misleading test collection reporting in terminal output

danceratopz:
## 🗒️ Description
- Don't load the `pytest_plugins.spec_version_checker.spec_version_checker` plugin as part of `fill` or `execute`.
- Allow specification of a Github PAT to avoid Github API rate limiting.
- Create a new separate pytest-based CLI `check_eip_versions` that only performs EIP spec version checking in a pytest session.
- Add a new Github Action to run  `check_eip_versions` daily and create an issue upon failure (example issues https://github.com/ethereum/execution-spec-tests/issues/1553 (prague mismatches only) and https://github.com/ethereum/execution-spec-tests/issues/1552 (all forks).

#### Note

This PR unfortunately adds a new pytest ini `pytest-check-eip-versions.ini` which is essentially a duplicate of `fill`'s `pytest.ini`. This is an ugly code duplication, that will likely cause `check_eip_versions` to break if `fill`'s options get changed w/o changing the `check_eip_versions` pytest ini. The issue is that the `pytest_plugins.spec_version_checker.spec_version_checker` needs to be loaded first, which doesn't seem to be possible if used in conjunction with `-c pytest.ini`.

## 🔗 Related Issues
Fixes #533, #1552, #1553
- #533
- #1552 
- #1553

## Todo before merge

- [x] Update docs (adding a PAT, etc).
- [x] Update references.
- [x] Remove "on pull request" / "on push" build triggers in action (testing only).

## ✅ Checklist


.github/workflows/check_eip_versions.yaml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          uv run python .github/scripts/generate_eip_report.py ./reports/eip_check_output.txt ./reports/outdated_eips.md
+
+      - name: Create Issue From File
+        if: steps.check-eip.outputs.exit_code != 0
+        uses: peter-evans/create-issue-from-file@e8ef132d6df98ed982188e460ebb3b5d4ef3a9cd
+        with:
+          title: "chore(tests): eip spec references outdated"
+          content-filepath: ./reports/outdated_eips.md
+          labels: report, automated issue, scope:tests, type:chore
```
marioevz, line 54: Is this going to create a new daily issue once an EIP goes out of date?

If so, maybe we should run this CI workflow weekly instead of daily.

marioevz (commented):
Left just one comment regarding the frequency of the runs for the new CI workflow.

.github/workflows/check_eip_versions.yaml
danceratopz: Good point, daily is too verbose - updated.

spencer-tb:
Tagging this comment, as another alternative to prevent rate limiting! I really think we should use it. Essentially a single endpoint call that gets all the EIP sha's at once! Check it out :)

https://api.github.com/repos/ethereum/EIPs/git/trees/master?recursive=1

https://github.com/ethereum/execution-spec-tests/pull/1357#pullrequestreview-2721217087

danceratopz:
> Tagging this comment, as another alternative to prevent rate limiting! I really think we should use it. Essentially a single endpoint call that gets all the EIP sha's at once! Check it out :)
> 
> https://api.github.com/repos/ethereum/EIPs/git/trees/master?recursive=1
> 
> [#1357 (review)](https://github.com/ethereum/execution-spec-tests/pull/1357#pullrequestreview-2721217087)

That's a nice solution, but I'm out of steam for this topic/PR. For now this (or using a venv) is working nicely:
```
uv run check_eip_versions --github-token=(gh auth token)
```

Tbh, the plugin should be re-written to not collect the evm python test cases at all (it currently collects ~50,000 test cases). I think it should be possible to simply discover test modules (containing "eip") and then create the eip version check test.

Feel free to PR your suggestion, if you fancy. Otherwise, we collect these ideas into a follow-up issue.

spencer-tb:
Created an issue :)

marioevz (approved):
LGTM, thanks!

Issue #1512, opened 2025-04-30
bug(tooling): wrong solc version might be used for static tests


felix314159:
## What is the context of this problem?
When trying to fill certain static tests using e.g. `uv run fill -k "MCOPY_memory_expansion_costFiller" --fill-static-tests --evm-bin=/home/$USER/Documents/evmone/build/bin/evmone-t8n -vv -s` a different solc version might be used than intended. This matters because I can fill these tests with solc 0.8.24 but it fails to fill with solc 0.8.29 (there are good reasons for this but this is not the scope of this issue).

## How to reproduce the issue?
Put the solc 0.8.29 binary to `.venv/bin/solc` and ensure it has +x permission. Now ensure your config is to use 0.8.24 with `uv run solc-select use 0.8.24` (you might have to install it first if you have not done so already). If you run `uv run solc-select versions` it should tell you: `0.8.24 (current, ...)`. Now run the fill command shown above. It will log that solc 0.8.24 is used (both in terminal and in html report) but in fact it uses the 0.8.29 binary, this can be detected e.g. by modifying './src/ethereum_test_specs/static_state/common/compile_yul.py' with sth like 
`
    result2 = subprocess.run(
        ["solc", "--version"],
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        text=True,
        check=True,
    )
    print(f"\nsolc version check: {result2.stdout}")
`

## What should the fix be?
solc-select must be enforced, it should not silently choose the solc binary from `.venv/bin/`. The logging must be consistent with what is actually being used.

danceratopz:
This must have been very annoying! But this seems quite an obscure case, if I understand correctly? As you manually have to overwrite the `solc` binary in the venv? (Btw, the user can instead use the `--solc-bin` flag to specify a different version). 

I'm surprised that solc-select simply assumes the binary it finds there is the correct version. How about, we add our own check to the end of this block to verify that the binary is actually at the version reported by solc-select and exit with error if there is an unexpected version there?
https://github.com/ethereum/execution-spec-tests/blob/624f60bd1330d117524ecaf240397064b7f4519e/src/pytest_plugins/solc/solc.py#L58-L76

felix314159:
> This must have been very annoying! But this seems quite an obscure case, if I understand correctly? As you manually have to overwrite the `solc` binary in the venv? (Btw, the user can instead use the `--solc-bin` flag to specify a different version).
> 
[...]

PR #1556, opened 2025-05-06
fix(solc): don't install `solc` via `solc-select` on ARM; add version sanity check

felix314159:
## 🗒️ Description
Read the [issue](https://github.com/ethereum/execution-spec-tests/issues/1512) for more context

## ✅ Checklist

### DIFF
--- a/src/pytest_plugins/solc/solc.py
+++ b/src/pytest_plugins/solc/solc.py
@@ -1,5 +1,7 @@
 """Pytest plugin for configuring and installing the solc compiler."""
 
+import platform
+import subprocess
 from argparse import ArgumentTypeError
 from shutil import which
 
@@ -63,6 +65,12 @@ def pytest_configure(config: pytest.Config):
         except ArgumentTypeError:
             version = None
         if version != solc_version:
+            # solc-select current does not support ARM linux
+            if platform.system().lower() == "linux" and platform.machine().lower() == "aarch64":
+                error_message = f"Version {version} does not match solc_version {solc_version}\
+ and since solc-select currently does not support ARM linux we can not recover from this problem."
+                pytest.exit(error_message, returncode=pytest.ExitCode.USAGE_ERROR)
+
             if config.getoption("verbose") > 0:
                 print(f"Setting solc version {solc_version} via solc-select...")
             try:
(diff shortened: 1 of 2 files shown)


src/pytest_plugins/solc/solc.py
```diff
@@ -90,6 +100,20 @@ def pytest_configure(config: pytest.Config):
...
+    # test whether solc_version matches actual one
+    solc_version_check_result = subprocess.run(
+        ["solc", "--version"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        text=True,
+        check=True,
+    )
+    solc_version_check_result_string = solc_version_check_result.stdout
+    if str(solc_version_semver) not in solc_version_check_result_string:
```
danceratopz, line 112: I think this is causing the failing unit test in CI (it's a weird one, but could also effect ARM users, I imagine). A more robust check would be to convert `solc_version_check_result_string` to a `semver.Version` using `Version.parse()` and then use `is_compatible()` to check whether the versions match.

See, for example:
https://python-semver.readthedocs.io/en/latest/usage/check-compatible-semver-version.html

src/pytest_plugins/solc/solc.py
```diff
@@ -90,6 +100,20 @@ def pytest_configure(config: pytest.Config):
         )
     config.solc_version = solc_version_semver  # type: ignore
 
+    # test whether solc_version matches actual one
```
danceratopz, line 103: Perhaps you could add a comment why you obtain the version this way? 

I.e. It's due to how the static filler runs `solc`: https://github.com/ethereum/execution-spec-tests/blob/c91549cf31c503d5ea9b3037093449dd0bb4d146/src/ethereum_test_specs/static_state/common/compile_yul.py#L22

danceratopz (commented):
LGTM, just two small comments below!

src/pytest_plugins/solc/solc.py
felix314159: I don't like semver tbh, look how easily it breaks:
```
>>> actualVersion = semver.Version.parse("0.8.25+commit.b61c2a91.Linux.g++")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/admin/Library/Python/3.9/lib/python/site-packages/semver/version.py", line 644, in parse
    raise ValueError(f"{version} is not valid SemVer string")
ValueError: 0.8.25+commit.b61c2a91.Linux.g++ is not valid SemVer string

>>> actualVersion = semver.Version.parse("0.8.25+commit.b61c2a91.Linux.g")
(Success)
```

felix314159: Action now passes but before the most recent commit it would fail due to e.g. sth like `0.8.25+commit.b61c2a91.Linux.g++` not being a substring of `0.8.25+commit.b61c2a91`, so now i only look at the first 10 chars and action now passes

danceratopz: Ups, I guess it's the wrong library for non-semantic versions :slightly_smiling_face: (I didn't realize that `+` isn't allowed!)

src/pytest_plugins/solc/solc.py
```diff
@@ -63,6 +65,12 @@ def pytest_configure(config: pytest.Config):
         except ArgumentTypeError:
             version = None
         if version != solc_version:
+            # solc-select current does not support ARM linux
+            if platform.system().lower() == "linux" and platform.machine().lower() == "aarch64":
+                error_message = f"Version {version} does not match solc_version {solc_version}\
+ and since solc-select currently does not support ARM linux we can not recover from this problem."
```
danceratopz, line 71: Your last commit inadvertently deleted the clarification from the previous commit? 
```suggestion
 and since solc-select currently does not support ARM linux you must manually do the following: \
Build solc from source, and manually move the binary to\
 .venv/.solc-select/artifacts/solc-x.y.z/solc-x.y.z, then run 'uv run solc-select use <x.y.z>'.
```

Also for strings that don't fit in the allowed line-length, please don't use `\`, rather do one of the following:
1. Use brackets that contain (and concatenate) multiple single-line strings. E.g.,
    https://github.com/ethereum/execution-spec-tests/blob/c91549cf31c503d5ea9b3037093449dd0bb4d146/src/pytest_plugins/filler/filler.py#L93-L98
2. `textwrap.dedent()` with a multi-line string:
    https://github.com/ethereum/execution-spec-tests/blob/c91549cf31c503d5ea9b3037093449dd0bb4d146/src/pytest_plugins/consume/hive_simulators/conftest.py#L203-L219

For this case, I'd go for 1.

danceratopz (commented):
LGTM, two last comments.

felix314159:
Done!

danceratopz (approved):
LGTM! Thanks!

I added a changelog to increase visibility for any other ARM users :slightly_smiling_face:

PR #1557, opened 2025-05-06
tooling(ux): only allow running fill command from root folder of repo

felix314159:
## 🗒️ Description
I keep accidentally running fill commands in subfolders, and then it will attempt to fill and after fairly long wait fail (but not without first polluting the subdir with a fixtures folder with a bunch of files. This PR attempts to prevent users from accidentally running a fill command from sth that is not the root dir.

## ✅ Checklist

### DIFF
--- a/src/cli/pytest_commands/consume.py
+++ b/src/cli/pytest_commands/consume.py
@@ -31,6 +31,13 @@ def handle_hive_env_flags(args: List[str]) -> List[str]:
 
 def handle_consume_command_flags(consume_args: List[str], is_hive: bool) -> List[str]:
     """Handle all consume CLI flag pre-processing."""
+    # Ensure the consume command was called in the root folder of execution-spec-tests repo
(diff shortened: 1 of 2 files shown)


danceratopz (commented):
I understand, but I don't think we should try to address the problem in this way. We plan to move `./tests/` to ethereum/execution-specs soonish and this code will need to be removed in order to fill the tests in a different location. To fill tests from another repository, EEST will need to be installed as a package and the commands will have to be able to work from the package in any directory (that contains tests).

Is the error due to not finding the pytest.ini file? 

In order to allow EEST pytest commands to be used from a package, I think we need to move the pytest*.ini files somewhere under `./src/pytest_plugins/`, include them as "data" in the package and have the cli helpers in `./src/cli/pytest_commands` pass them to pytest's `-c` option as absolute paths.

This partially solves your problem, as `fill` shouldn't fail, but it'll still pollute your directories with fixtures files.

PR #1507, opened 2025-04-29
chore(all): move eof from osaka to unscheduled

spencer-tb:
## 🗒️ Description
As upset as I am with the decision, I am ripping the band-aid off fast. I hope we can include EOF in a future fork.

I've created a new folder within `tests/` called `unscheduled`. This will contain tests for features that are not currently CFI'd for future forks. Verkle will sit here in the near future too.

I've created a new fork called `EOFv1` for simplicity. This should be the only breaking change for filling and consuming existing EOF fixtures. Thus clients will need to add this fork. Evmone should additionally add this fork to fill the EOF tests.

A new label is added to EEST called `fork:eof`. I will relabel all the existing EOF PRs and issues with this.

## 🔗 Related Issues
https://github.com/ethereum/EIPs/pull/9703
https://github.com/ethereum/pm/issues/1499

## ✅ Checklist

- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [x] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.

PR #1574, opened 2025-05-07
feat(tests): eip-5920 pay opcode cases

spencer-tb:
## 🗒️ Description
Assumes that EIP-5920 will be in Osaka.

Uses a temp EELS resolution to fill:
- https://github.com/spencer-tb/execution-specs/tree/spec/eip-5920

Still a WIP.

Rebased on #1507 hence requires it.

## 🔗 Related Issues
N/A

## ✅ Checklist


jochem-brouwer:
Linking for completeness: https://github.com/ethereum/EIPs/pull/9590

gurukamath:
Just want to highlight the following

- A new [EELS PR](https://github.com/ethereum/execution-specs/pull/1225) with EIP-5920 implementation for Osaka is up

- I would like to add the following to the list of relevant test cases
    - Trying to transfer non-zero value in static context
    - Trying to transfer zero value in static context
    - Trying to transfer zero value to a non-existent account
    - Trying to transfer higher value than the balance to a non-existent account

jochem-brouwer:
Wanted to drop a quick note here from the EIP (this has been in there for a long time):

`PAY` cannot be implemented on networks with empty accounts (see [EIP-7523](https://eips.ethereum.org/EIPS/eip-7523)).

This is nice, because it means we don't have to deal with "touched" accounts and cleaning up the whole mess a la EIP-161.

The only thing we should strictly check though (in a simple test) is that if you `PAY` with `0` that if this gets used on a non-existing account, it will also leave it as non-existing account (and thus not leave an empty account in the trie).

spencer-tb:
Closing. Can come back to this if PAY gets CFI'd.

Issue #1382, opened 2025-04-02
Missing test cases for ecrecover


chfast:
- `r == 0 and s >= ORDER`
- `r == 0 and s < ORDER`
- `r >= ORDER and s == 0`
- `r >= ORDER and s >= ORDER`

PR #1575, opened 2025-05-07
zkvm: allow using entire gas in bytecode attack

jsign:
This PR proposes a way to fill the bytecode attack, solving two main problems:
- Creating so many 24KiB contracts with 36M gas limit takes a lot of blocks, which bloats the generated fixture in size and generation time.
- Even if so many blocks in the fixture aren't a problem, all the ~300MiB of code appears in the `postState`, which also bloats the fixture. (This PR now uses a new feature introduced in https://github.com/ethereum/execution-spec-tests/pull/1578)

Results:
- I tested in Geth, and it runs exactly as expected.
- Filling this test on my machine with `--block-gas-limit 36M`
  - Only two blocks are needed in the fixture. 
  - Block 1 only has a single tx creating all contracts.
  - Block 2 has the actual attack.
- Takes 2 mins to fill the test, which sounds quite fast considering it creates 300MiB of contracts.
- The fixture json file size is ~35KiB, so very small as wanted.

Fill command: `uv run fill --from=Cancun --until=Cancun -m "zkevm and blockchain_test" --block-gas-limit 36000000 -k test_worst_bytecode`

### DIFF
--- a/tests/zkevm/test_worst_bytecode.py
+++ b/tests/zkevm/test_worst_bytecode.py
@@ -27,9 +27,6 @@ REFERENCE_SPEC_GIT_PATH = "TODO"
 REFERENCE_SPEC_VERSION = "TODO"
 
 MAX_CONTRACT_SIZE = 24 * 1024  # TODO: This could be a fork property
-BLOCK_GAS_LIMIT = 36_000_000  # TODO: Parametrize using the (yet to be implemented) block gas limit
-# OPCODE_GAS_LIMIT = BLOCK_GAS_LIMIT  # TODO: Reduced in order to run the test in a reasonable time
-OPCODE_GAS_LIMIT = 100_000
 
 XOR_TABLE_SIZE = 256
 XOR_TABLE = [Hash(i).sha256() for i in range(XOR_TABLE_SIZE)]
@@ -41,12 +38,19 @@ XOR_TABLE = [Hash(i).sha256() for i in range(XOR_TABLE_SIZE)]
         Op.EXTCODESIZE,
     ],
 )
+@pytest.mark.parametrize(
+    "attack_gas_limit",
+    [
+        Environment().gas_limit,
+    ],
+)
 @pytest.mark.valid_from("Cancun")
 def test_worst_bytecode_single_opcode(
     blockchain_test: BlockchainTestFiller,
     pre: Alloc,
     fork: Fork,
     opcode: Op,
+    attack_gas_limit: int,
 ):
     """
     Test a block execution where a single opcode execution maxes out the gas limit,
@@ -60,7 +64,9 @@ def test_worst_bytecode_single_opcode(
     The test is performed in the last block of the test, and the entire block gas limit is
     consumed by repeated opcode executions.
     """
-    env = Environment(gas_limit=BLOCK_GAS_LIMIT)
+    # We use 100G gas limit to be able to deploy a large number of contracts in a single block,
+    # avoiding bloating the number of preparing blocks in the test.
+    env = Environment(gas_limit=100_000_000_000)
 
     # The initcode will take its address as a starting point to the input to the keccak
     # hash function.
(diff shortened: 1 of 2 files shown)


src/ethereum_test_specs/blockchain.py
```diff
@@ -300,6 +300,7 @@ class BlockchainTest(BaseTest):
     genesis_environment: Environment = Field(default_factory=Environment)
     verify_sync: bool = False
     chain_id: int = 1
+    exclude_post_state: bool = False
```
jsign, line 303: This is a new flag to signal that we want to output an empty `postState` in the fixture, to avoid the mentioned bloat.

This is an easy proposal, I would prefer experts in the testing team to asses if this is okay or is another prefered option. :)

src/ethereum_test_specs/blockchain.py
```diff
@@ -652,7 +654,7 @@ def make_fixture(
             blocks=fixture_blocks,
             last_block_hash=head,
             pre=pre,
-            post_state=alloc,
+            post_state=alloc if not self.exclude_post_state else {},
```
jsign, line 657: Initially, I did it with `None` instead of `{}`. This means the field didn't exist in the output.

While Geth was okay with it, Reth now is a bit more strict and find that problematic since it always expects the field to be set.

So the less invasive way is always have the field, but empty.

tests/zkevm/test_worst_bytecode.py
```diff
@@ -27,7 +27,7 @@
 REFERENCE_SPEC_VERSION = "TODO"
 
 MAX_CONTRACT_SIZE = 24 * 1024  # TODO: This could be a fork property
-OPCODE_GAS_LIMIT = 100_000
+OPCODE_GAS_LIMIT = 36_000_000
```
jsign, line 30: Here's the main contention point I think.

Ideally, I would like to put `Environment().gas_limit`. But I'm already using the `--block-gas-limit` for the actual gas limit of the block, not the attack.

If the conclusion is that we can hardcode the gas limit at the test level to 100GGas, then I could use `Environment().gas_limit` here without problems, and it will fit what it is done in #1382

src/ethereum_test_specs/blockchain.py
marioevz: I think this alone should be a separate PR to have a better discussion on it, could you create another PR independent just for this pls?

jsign: Sure, I'll do that.

jsign: @marioevz, created it here https://github.com/ethereum/execution-spec-tests/pull/1578.
After we see how that goes, I can rebase this PR.

tests/zkevm/test_worst_bytecode.py
marioevz: I think this is ok, `Environment().gas_limit` is just a recommendation to constrain tests to not go above this limit to maintain the testing realistic, but if the setup goes above that limit it should be ok.

If we go this route, one consideration is that the test is not going to be executable via `uv run execute`, but we could introduce a fixture that lets the test know the mode it is currently running in and limit the gas consumed by its setup. Just an idea for now, such a fixture does not exist yet.

jsign: Cool, so I'll change this `OPCODE_GAS_LIMIT` `36m` to `Environment().gas_limit` (so behaves coherently with whatever `--block-gas-limit` is provided via CLI) and hardcode the 100GGas for gas limit.

jsign: Done! 

Let's wait to merge #1578 first, and then I'll rebase this PR and we can do a final check before merging.

tests/zkevm/test_worst_bytecode.py
```diff
@@ -171,8 +177,7 @@ def generate_deploy_tx(contracts_to_deploy: int):
         + Op.MSTORE(64, initcode.keccak256())
         # Main loop
         + While(
-            body=Op.POP(Op.EXTCODESIZE(Op.SHA3(32 - 20 - 1, 85)))
```
jsign, line 174: Just fixed the hardcoded `Op.EXTCODESIZE` with the parameter `opcode`.

marioevz (approved):
LGTM, hope that the rebase helps with the CI timeout :)

PR #1576, opened 2025-05-08
tooling(fix): use --version instead of -v for version checking t8n tool

felix314159:
## 🗒️ Description
Geth's `evm` supports both `-v` and `--version` but Nethermind's `nethtest` only supports `--version`. This does not mean that we can consume direct with nethtest now, but it's one of multiple fixes we must make. Next thing to work on would be `assert cls.detect_binary_pattern is not None -> AttributeError: type object 'CollectOnlyFixtureConsumer' has no attribute 'detect_binary_pattern'` to eventually enabling consume direct via nethtest

## ✅ Checklist

### DIFF
--- a/src/ethereum_clis/ethereum_cli.py
+++ b/src/ethereum_clis/ethereum_cli.py
@@ -41,7 +41,7 @@ class EthereumCLI:
     binary: Path
     default_binary: Path
     detect_binary_pattern: Pattern
-    version_flag: str = "-v"
+    version_flag: str = "--version"  # geth's evm supports both -v and --version but nethtest only supports --version  # noqa: E501
     cached_version: Optional[str] = None
 
     def __init__(self, *, binary: Optional[Path] = None):


src/ethereum_clis/ethereum_cli.py
```diff
@@ -41,7 +41,7 @@ class EthereumCLI:
     binary: Path
     default_binary: Path
     detect_binary_pattern: Pattern
-    version_flag: str = "-v"
+    version_flag: str = "--version"  # geth's evm supports both -v and --version but nethtest only supports --version  # noqa: E501
```
marioevz, line 44: I don't think this is the issue because the `Nethtest` subclass modifies the property here:
https://github.com/ethereum/execution-spec-tests/blob/244852ffe01025a9a8e97747822f6d4592eb7a86/src/ethereum_clis/clis/nethermind.py#L26

In my case, the version returned nethtest does not match the regex:
Regex:
https://github.com/ethereum/execution-spec-tests/blob/244852ffe01025a9a8e97747822f6d4592eb7a86/src/ethereum_clis/clis/nethermind.py#L25
Output:
```
Using launch settings from /home/marioevz/Development/Eth/nethermind/src/Nethermind/Nethermind.Test.Runner/Properties/launchSettings.json...\n1.31.9+98c5dca4696ea1fb882ea7e341ab6e64f5adaa69
```

I modified the regex to `\d+\.\d+\.\d+[\-\+][a-zA-Z0-9]+(\+[a-f0-9]{40})?$` and seems to be working.

PR #1578, opened 2025-05-08
Add optional field to exclude including postState

jsign:
This PR is a requirement for #1575.

### DIFF
--- a/src/ethereum_test_specs/blockchain.py
+++ b/src/ethereum_test_specs/blockchain.py
@@ -301,6 +301,11 @@ class BlockchainTest(BaseTest):
     verify_sync: bool = False
     chain_id: int = 1
     exclude_post_state: bool = False
+    """
+    Exclude the post state from the fixture output.
+    
+    In this case, the state verification is only performed based on the state root.
+    """
 
     supported_fixture_formats: ClassVar[Sequence[FixtureFormat | LabeledFixtureFormat]] = [
         BlockchainFixture,
--- a/src/ethereum_test_specs/blockchain.py
+++ b/src/ethereum_test_specs/blockchain.py
@@ -300,6 +300,7 @@ class BlockchainTest(BaseTest):
     genesis_environment: Environment = Field(default_factory=Environment)
     verify_sync: bool = False
     chain_id: int = 1
+    exclude_post_state: bool = False
 
     supported_fixture_formats: ClassVar[Sequence[FixtureFormat | LabeledFixtureFormat]] = [
         BlockchainFixture,
@@ -643,7 +644,8 @@ class BlockchainTest(BaseTest):
                     t8n, t8n_state=alloc, expected_state=block.expected_post_state
                 )
         self.check_exception_test(exception=invalid_blocks > 0)
-        self.verify_post_state(t8n, t8n_state=alloc)
+        if not self.exclude_post_state:
+            self.verify_post_state(t8n, t8n_state=alloc)
         network_info = BlockchainTest.network_info(fork, eips)
         return BlockchainFixture(
             fork=network_info,
@@ -652,7 +654,7 @@ class BlockchainTest(BaseTest):
             blocks=fixture_blocks,
             last_block_hash=head,
             pre=pre,
-            post_state=alloc,
+            post_state=alloc if not self.exclude_post_state else {},
             config=FixtureConfig(
                 fork=network_info,
                 blob_schedule=FixtureBlobSchedule.from_blob_schedule(fork.blob_schedule()),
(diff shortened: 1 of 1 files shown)


src/ethereum_test_specs/blockchain.py
```diff
@@ -652,7 +654,7 @@ def make_fixture(
             blocks=fixture_blocks,
             last_block_hash=head,
             pre=pre,
-            post_state=alloc,
+            post_state=alloc if not self.exclude_post_state else {},
```
jsign, line 657: Initially, I did it with None instead of {}. This means the field didn't exist in the output.

While Geth was okay with it, Reth now is a bit more strict and find that problematic since it always expects the field to be set.

So the less invasive way is always have the field, but empty.

src/ethereum_test_specs/blockchain.py
```diff
@@ -300,6 +300,7 @@ class BlockchainTest(BaseTest):
     genesis_environment: Environment = Field(default_factory=Environment)
     verify_sync: bool = False
     chain_id: int = 1
+    exclude_post_state: bool = False
```
marioevz, line 303:
```suggestion
    exclude_post_state: bool = False
    """
    Exclude the post state from the fixture output.
    
    In this case, the state verification is only performed based on the state root.
    """
```
Or similar to keep track of the usage of this field.

src/ethereum_test_specs/blockchain.py
marioevz: I think having the field as `{}` is a bit misleading, as compared to removing it.

I feel like there should be some logic in the client consumer that acknowledges the field is missing and then proceeds to only check the state root.

Cc'ing @danceratopz since he has reached out to reth team in the past and they have been responsive on these changes to the test formats iirc.

marioevz (commented):
Some comments, thanks!

src/ethereum_test_specs/blockchain.py
jsign: Sharing here the relevant current reth logic about this:
image

cc @kevaundray  since he was around this code some days ago!

kevaundray: Happy to make this change!

Since it doesn't change any of the tests, it would mainly be a case of: not returning an error if the post state is missing and just returning true.

jsign: @marioevz, I talked with @kevaundray and he'll do the reth change in a fork that we can use for a while before upstreaming the change. 

So, I'll switch to use `None` instead as you prefered. Whenever the reth fork is ready I'll test again just in case, and I can ping you again to potentially merge the pr.

jsign: @marioevz, done:
- Did the change to `None` here.
- Tested in the new with the new https://github.com/eth-applied-research-group/zkevm-benchmark-workload Reth fork with the fix, and is all good.

jsign:
@marioevz is this mergable?

src/ethereum_test_specs/blockchain.py
```diff
@@ -643,7 +648,8 @@ def make_fixture(
                     t8n, t8n_state=alloc, expected_state=block.expected_post_state
                 )
         self.check_exception_test(exception=invalid_blocks > 0)
-        self.verify_post_state(t8n, t8n_state=alloc)
+        if not self.exclude_post_state:
+            self.verify_post_state(t8n, t8n_state=alloc)
```
marioevz, line 652:
```suggestion
        self.verify_post_state(t8n, t8n_state=alloc)
```
After thinking about it, maybe this skipping this check is not necessary because, for example, we could have a huge state because of the size of the contracts, but as a test writer I would still like to know that some code ended up in the addresses I expect them to, so I could add a check in the post for the nonces of each contract

src/ethereum_test_specs/blockchain.py
```diff
@@ -300,6 +300,11 @@ class BlockchainTest(BaseTest):
     genesis_environment: Environment = Field(default_factory=Environment)
     verify_sync: bool = False
     chain_id: int = 1
+    exclude_post_state: bool = False
```
marioevz, line 303:
```suggestion
    exclude_full_post_state_in_output: bool = False
```
Just a nitpick on the naming of this field, just to make it explicit because otherwise I feel like it could be misinterpreted as excluding the post state in the input?

marioevz (commented):
Some last minute comments, thanks!

src/ethereum_test_specs/blockchain.py
jsign: Yes, and I also think this makes more clear the previous change. As in, the post-state assertions are still done, and this flag is only for avoiding putting it into the output.

marioevz (approved):
LGTM, thanks!

PR #1581, opened 2025-05-09
zkevm: add remaining precompiles

jsign:
This PR adds coverage for all remaining precompiles:
- bn128_add
- bn128_mul
- bn128_pairing
- blake2f
- point_evaluation
- All seven new BLS precompiles activated in Pectra.

I created a generic setup, so adding other precompile cases (if anyone thinks other inputs create even worse-case scenarios) is very simple.

I ran this in the SP1 for counting cycles:
- bls12_fp_to_g1 -> 2.3 billion cycles
- bls12_fp_to_g2 -> 2.4 billion cycles
- bls12_g1add -> 2.4 billion cycles
- bls12_g1msm -> Crash
- bls12_g2add -> 2.5 billion cycles
- bls12_g2msm -> Crash
- bls12_pairing -> 2.4 billion cycles
- bn128_add -> 18.5 billion cycles
- bn128_mul -> 28.5 billion cycles
- bn128_pairing -> 18.6 billion cycles
- point_evaluation -> 6.8 billion cycles
- blake2f -> 21 billion cycles


jsign:
The CI is complaining about some `E   TimeoutError: timed out`. Maybe related to these test being super-heavy for EELS as filler (locally I use Geth).

tests/zkevm/test_worst_compute.py
```diff
@@ -144,24 +238,212 @@ def test_worst_modexp(
...
+    [
+        pytest.param(
+            0x01,
+            [
+                "38D18ACB67D25C8BB9942764B62F18E17054F66A817BD4295423ADF9ED98873E",
+                "000000000000000000000000000000000000000000000000000000000000001B",
+                "38D18ACB67D25C8BB9942764B62F18E17054F66A817BD4295423ADF9ED98873E",
+                "789D1DD423D25F0772D2748D60F7E4B81BB14D086EBA8E8E8EFB6DCFF8A4AE02",
+            ],
+            32 * 4,
```
chfast, line 277: Can't you take the length as `len(input)`?

tests/zkevm/test_worst_compute.py
```diff
@@ -144,24 +238,212 @@ def test_worst_modexp(
...
+            0x0B,
+            [
+                "000000000000000000000000000000000572CBEA904D67468808C8EB50A9450C9721DB309128012543902D0AC358A62AE28F75BB8F1C7C42C39A8C5529BF0F4E00000000000000000000000000000000166A9D8CABC673A322FDA673779D8E3822BA3ECB8670E461F73BB9021D5FD76A",
+                "4C56D9D4CD16BD1BBA86881979749D280000000000000000000000000000000009ECE308F9D1F0131765212DECA99697B112D61F9BE9A5F1F3780A51335B3FF981747A0B2CA2179B96D2C0C9024E522400000000000000000000000000000000032B80D3A6F5B09F8A84623389C5F80C",
+            ],
+            256,
+            id="bls12_g1add",
+        ),
+        pytest.param(
+            0x0C,
```
chfast, line 358: The worst input is probably the one for 128 pairs. Please try to input imputs from the existing tests instead of using raw hex values here.

jsign: By "existing tests" you mean here https://eips.ethereum.org/assets/eip-2537/test-vectors ? There the inputs are in raw hex.

chfast: For BLS precompiles IDs you can also use these constants: https://github.com/ethereum/execution-spec-tests/blob/main/tests/prague/eip2537_bls_12_381_precompiles/spec.py#L112-L119

jsign: Oh, amazing. Thanks for the pointers, I'll do.

tests/zkevm/test_worst_compute.py
```diff
@@ -234,23 +235,167 @@ def test_worst_modexp(
...
+            id="bls12_g2add",
+        ),
+        pytest.param(
+            bls12381_spec.Spec.G2MSM,
+            [
+                # TODO: the //2 is required due to a limitation of the max contract size limit.
+                # In a further iteration we can insert the inputs as calldata or storage and avoid
+                # having to do PUSHes which has this limtiation. This also applies to G1MSM.
+                (bls12381_spec.Spec.P2 + bls12381_spec.Scalar(bls12381_spec.Spec.Q))
+                * (len(bls12381_spec.Spec.G2MSM_DISCOUNT_TABLE) // 2),
```
jsign, line 343: We can improve on this in a further PR. With this style the RISCV emulator already crashes.

tests/zkevm/test_worst_compute.py
```diff
@@ -234,23 +235,173 @@ def test_worst_modexp(
...
+                "8F59A8D2A1A625A17F3FEA0FE5EB8C896DB3764F3185481BC22F91B4AAFFCCA25F26936857BC3A7C2539EA8EC3A952B7",
+                "873033E038326E87ED3E1276FD140253FA08E9FC25FB2D9A98527FC22A2C9612FBEAFDAD446CBC7BCDBDCD780AF2C16A",
+            ],
+            id="point_evaluation",
+        ),
+        pytest.param(
+            bls12381_spec.Spec.G1ADD,
+            [
+                bls12381_spec.Spec.G1,
+                bls12381_spec.Spec.P1,
```
jsign, line 316: This is why we accept `BytesConcatenation`.

tests/zkevm/test_worst_compute.py
```diff
@@ -234,23 +235,173 @@ def test_worst_modexp(
...
+            ],
+            id="bls12_g1add",
+        ),
+        pytest.param(
+            bls12381_spec.Spec.G1MSM,
+            [
+                (bls12381_spec.Spec.P1 + bls12381_spec.Scalar(bls12381_spec.Spec.Q))
```
jsign, line 323: This is why we accept also `bytes`.

jsign:
@chfast, I applied your suggestions if you'd like to take another look.
Cycles remained the same for everything, but now both MSM make the SP1 RISCV emulator crash :)

tests/zkevm/test_worst_compute.py
```diff
@@ -237,23 +243,173 @@ def test_worst_modexp(
...
 
 @pytest.mark.valid_from("Cancun")
-def test_worst_ecrecover(
+@pytest.mark.parametrize(
+    "precompile_address,parameters",
+    [
+        pytest.param(
+            0x01,
+            [
+                "38D18ACB67D25C8BB9942764B62F18E17054F66A817BD4295423ADF9ED98873E",
```
chfast, line 252: Please add one line comment for each of these hex inputs where are they from and why they are worst cases.

tests/zkevm/test_worst_compute.py
```diff
@@ -237,23 +243,173 @@ def test_worst_modexp(
...
+                "1A87B0584CE92F4593D161480614F2989035225609F08058CCFA3D0F940FEBE3",
+                "1A2F3C951F6DADCC7EE9007DFF81504B0FCD6D7CF59996EFDC33D92BF7F9F8F6",
+                "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF",
+            ],
+            id="bn128_mul",
+        ),
+        pytest.param(
+            0x08,
+            [
+                # First pairing
```
chfast, line 281: Have you check what is the worst?
- 2 pairings: result is 1, no exp amortization,
- 1 pairing: result is 0, even less exp amortization,
- a lot of pairings: exp amortization.

chfast (approved):
I left two questions/comments I'd like to see address. But overall looks good to me.

tests/zkevm/test_worst_compute.py
jsign: This is a regular signature. Do you already know some quirks for other input be a better candidates? We can further iterate on it.

I'll add a comment explaining this is a regular signature and add a TODO if we can find a better input.

tests/zkevm/test_worst_compute.py
jsign: It is not, but mainly because for the full amortization we need a separate test than this generic one to deal with the total input size not to blow up the max contract size coming from PUSHes for MSTORE. Probably using CALLDATA.

I'll add a TODO that this case can be iterated further. As an extra dimension to improve might be particular pairing cases that could be worse.

tests/zkevm/test_worst_compute.py
```diff
@@ -237,23 +243,180 @@ def test_worst_modexp(
...
 
 @pytest.mark.valid_from("Cancun")
-def test_worst_ecrecover(
+@pytest.mark.parametrize(
+    "precompile_address,parameters",
+    [
+        pytest.param(
+            0x01,
+            [
+                # TODO: below inputs are covering a regular signature. Worth exploring
```
chfast, line 252: For any kind of cryptography verification (signature verification, pairing check), the positive case is the worst case. If the input is invalid it means it has failed one of the verification steps.

jsign: Cool, I'll remove that part of the comment then.

chfast (approved):

PR #1587, opened 2025-05-12
fix(static_tests): adjusted gasLimit to only allow gasLimit of up to 30mil as enforced in eip-7825

felix314159:
## 🗒️ Description
For now this is only for static_tests we have. Currently filling and executing to see if it all still works. Please thoroughly review this PR. These changes were done via a Python script that has no notion of what any of this means, next thing we should do is within our framework at fill check whether gasLimit is in allowed range

## ✅ Checklist

### DIFF
--- a/tests/static/state_tests/Cancun/stEIP1153_transientStorage/transStorageOKFiller.yml
+++ b/tests/static/state_tests/Cancun/stEIP1153_transientStorage/transStorageOKFiller.yml
@@ -810,7 +810,7 @@ transStorageOK:
     - :label static_call :abi staticCall()
 
     gasLimit:
-    - 80000000
+    - 30000000
     gasPrice: 10
     nonce: 1
     to: cccccccccccccccccccccccccccccccccccccccc
--- a/tests/static/state_tests/Cancun/stEIP1153_transientStorage/transStorageResetFiller.yml
+++ b/tests/static/state_tests/Cancun/stEIP1153_transientStorage/transStorageResetFiller.yml
@@ -247,7 +247,7 @@ transStorageReset:
 
 
     gasLimit:
-    - 80000000
+    - 30000000
     gasPrice: 10
     nonce: 1
     to: cccccccccccccccccccccccccccccccccccccccc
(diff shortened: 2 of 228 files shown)


winsvega:
does test filling work  without errors?

jochem-brouwer:
Shouldn't this just mark the tests which have a gas limit higher than 30M invalid/not available from the fork which includes EIP-7825? Merging this would also require explicit checks that the test itself does not depend on having more than 30M gas (one use case for this "huge" gas limit would be benchmarks/stress tests, especially for precompiles)

winsvega:
I think most of the test have high gas limit just in case and copied from template. if the test filling works we can cap it.

winsvega (changes requested):
Here some fixes
https://github.com/felix314159/execution-spec-tests/pull/1

Need to fix all the failures. 

- [ ] stShift/ shift combinations needs to be reworked. it consumes more 30M gas for all the combinations
- [ ] lets merge test scenarios it converts all stBadOpcode/opcDiffPlaces

felix314159:
After rebase I filled again with `uv run fill ./tests/static --fill-static-tests --evm-bin=/home/$USER/Documents/evmone/build/bin/evmone-t8n -n auto -vv` and now we are down to 36 failed tests:
```
* stCreateTest/createLargeResultFiller.yml
* stCreateTest/CREATE_FirstByte_loopFiller.yml
* stCreateTest/CreateOOGafterMaxCodesizeFiller.yml
* stShift/shiftCombinationsFiller.yml
* stSelfBalance/diffPlacesFiller.yml
```

winsvega:
stShift/shiftCombinationsFiller.yml  has too many combinations better to rewrite in .py
stCreateTest/CREATE_FirstByte_loopFiller.yml  this test eats too much gas. it can be optimized when converted .py
stCreateTest/CreateOOGafterMaxCodesizeFiller  have large count contract interactions, need to rewrite it for prague limiting number of contract interactions
stSelfBalance/diffPlacesFiller.yml   can be rewritten as a scenario for merged test scenarios. need see into the cases. scenario like CREATE/CREATE2  then CALL/CALLCODE/DELEGATE/STATIC  created contract is not implemented.

spencer-tb:
As these are yml files, I'm wondering if we could set the gas limit as an env var that is set during fill time. That way it can be dynamic and we don't need to change it again if the gas limit changes etc. Something like:
```yml
gasLimit: ${GAS_LIMIT}
```

PR #1595, opened 2025-05-13
fix(github): Use evmone for zkevm release

marioevz:
## 🗒️ Description
Modifies zkevm github action to build with evmone due to these being gas intenstive tests.

Long term solution is to dynamically increase the EELS timeout with a ceiling depending on how much gas the test is requesting.


spencer-tb (approved):
LGTM! Thanks

PR #1597, opened 2025-05-14
chore(tooling): bump `uv.lock` revision

danceratopz:
## 🗒️ Description

`uv` `0.6.15` bumped `uv.lock` version to 2 and added a new field to dependency `url` fields: `upload_time`, which was renamed to `upload-time` in `0.7.0`. Generating `uv.lock` with a version `>=0.6.15` creates a lot of noise. Search for "upload" in https://github.com/astral-sh/uv/releases.

This PR uses `uv==0.7.3` to update the uv.lock and add these new fields, without any dependency changes to make it easier to parse the uv.lock in the case of a true dependency change.

In the future, we should only make dependency changes with `uv>=0.7.0` to avoid switching back and forth between formats; this is noted in the docs.

## 🔗 Related Issues
None

## ✅ Checklist

- [x] All: ~~Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~ skipped

### DIFF
--- a/docs/dev/deps_and_packaging.md
+++ b/docs/dev/deps_and_packaging.md
@@ -2,6 +2,8 @@
 
 EEST uses [`uv`](https://docs.astral.sh/uv/) to manage and pin its dependencies.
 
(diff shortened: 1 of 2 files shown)


felix314159 (approved):
Ty!

PR #1609, opened 2025-05-15
feat(vm): Memory variable

marioevz:
## 🗒️ Description

Introduces the `MemoryVariable` abstraction to make contracts more readable when there's heavy use of the EVM memory as variable storage location.

To use, simply declare a variable with an unique offset that is not used by any other variable, and then the variable can be used in-place to read the value from memory:

```python
v = MemoryVariable(128)

bytecode = Op.ADD(v, Op.CALLDATASIZE())
```

The previous example is equivalent to:

```python
bytecode = Op.ADD(Op.MLOAD(offset=128), Op.CALLDATASIZE())
```

The variable also contains methods to generate bytecode to modify the variable in memory.

```python
v = MemoryVariable(128)

bytecode = (
    v.store(0xff)
    + v.add(1)
    + v.return_value()
)
```

The previous example is equivalent to:

```python
bytecode = (
    Op.MSTORE(offset=128, value=0xff)
    + Op.MSTORE(offset=128, value=Op.ADD(Op.MLOAD(offset=128), 1))
    + Op.RETURN(offset=128, size=32)
)
```

## ✅ Checklist

### DIFF
--- a/tests/frontier/scenarios/common.py
+++ b/tests/frontier/scenarios/common.py
@@ -220,8 +220,9 @@ def make_gas_hash_contract(pre: Alloc) -> Address:
         + Conditional(
             condition=Op.ISZERO(current_byte),
             if_true=(
-                # Increase the calldata byte offset, and if it's greater than the calldata size,
-                # return, otherwise jump to the calldata copy code and read the next byte.
+                # Increase the calldata byte offset, and if it's greater than
+                # the calldata size, return, otherwise jump to the calldata
+                # copy code and read the next byte.
                 byte_offset.add(1)
                 + Conditional(
                     condition=Op.GT(byte_offset, Op.CALLDATASIZE()),
--- a/src/ethereum_test_vm/helpers.py
+++ b/src/ethereum_test_vm/helpers.py
@@ -8,7 +8,8 @@ class MemoryVariable(Bytecode):
     """
     Variable abstraction to help keep track values that are stored in memory.
 
-    To use, simply declare a variable with an unique offset that is not used by any other variable.
+    To use, simply declare a variable with an unique offset that is not used
+    by any other variable.
 
     The variable then can be used in-place to read the value from memory:
 
@@ -24,7 +25,8 @@ class MemoryVariable(Bytecode):
     bytecode = Op.ADD(Op.MLOAD(offset=128), Op.CALLDATASIZE())
     ```
 
-    The variable also contains methods to add and subtract values from the memory offset.
+    The variable also contains methods to add and subtract values from the
+    memory offset.
 
     ```python
     v = MemoryVariable(128)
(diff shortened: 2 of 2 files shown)


marioevz:
@winsvega I thought of this when reading the code to `make_gas_hash_contract` and I thought that the hardest part of following the logic was mentally tracking the memory offsets and what they meant, so I modified the python code in that function to use this to see if it makes it more readable.

Let me know what you think.

tests/frontier/scenarios/common.py
```diff
@@ -190,20 +190,40 @@ def make_gas_hash_contract(pre: Alloc) -> Address:
...
+    offset_calldata_copy = len(initialize_code)
+    offset_conditional = offset_calldata_copy + len(calldata_copy)
+
+    # Deploy contract
     gas_hash_address = pre.deploy_contract(
-        code=Op.MSTORE(0, 0)
-        + Op.JUMPDEST
-        + Op.CALLDATACOPY(63, Op.MLOAD(0), 1)
-        + Op.JUMPDEST
+        code=initialize_code
```
winsvega, line 211: I prefere bot to subfunc the code for readability

tests/frontier/scenarios/common.py
```diff
@@ -190,20 +190,40 @@ def make_gas_hash_contract(pre: Alloc) -> Address:
...
+        + calldata_copy  # offset_calldata_copy
```
winsvega, line 212: This makes it less readable

winsvega (commented):
To be honest I read plain evm opcodes much better. The fact that I don't know what's going on in this optimisations makes me go back and forth in defenitions.

Let me read it again tomorrow

winsvega:
I was thinking the other night. maybe it will be even better for readability if Op will support memory variables? 
so you can call it like

```
call(gas, address, balance,    input_var,  output_var) 
add(variable_name,  variable_name)
add(variable_name, value) 
```

tests/frontier/scenarios/common.py
```diff
@@ -190,20 +190,40 @@ def make_gas_hash_contract(pre: Alloc) -> Address:
...
     we at least could spend unique gas amount.
     """
+    # EVM memory variables
+    variable_byte_offset = MemoryVariable(0)
+    variable_current_byte = MemoryVariable(32)
+
+    # Code for memory initialization
+    initialize_code = variable_byte_offset.store(0)
+    calldata_copy = Op.JUMPDEST + Op.CALLDATACOPY(
+        dest_offset=variable_current_byte.offset + 32 - 1,
```
winsvega, line 200: 32, -1  magic numbers?

winsvega (changes requested):
I think it can be made more readable. 
maybe Op.Mstore(var, value)  can be supported? 

but the most important. is to have some unit test for this logic. 
to check that before and after refactoring it perfroms the same result. 

and if you find it usefull it can be moved to a global namespace so we can reuse it in other tests.

tests/frontier/scenarios/common.py
marioevz: It was 63 before, we were putting it in the last byte of the variable.
So the offset, plus its size (32), minus the one byte.

PR #1612, opened 2025-05-16
fix(fill): fix `--collect-only` after introducing `FixtureOutput` class

danceratopz:
## 🗒️ Description
Follow-up bug fix to:
- #1471 

Adds a `CliRunner`-based module test to catch the bug, which is fixed by creating the `FixtureOutput` object before exiting early if running with `--collect-only`.

## 🔗 Related Issues
Introduced in:
- #1471.

## ✅ Checklist

- [x] ~~All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~ skipped


marioevz (approved):
LGTM, thanks for the unit test :)

PR #1614, opened 2025-05-16
feat(tests): ckzg functions for peerdas tests

felix314159:
## 🗒️ Description
Added an eest wrapper for ckzg functions we will probably use in our tests.
A few open questions:
* Their code relies on `trusted_setup.txt`, I just import in the functions that need it and it assumes it is in the same folder. Is this compatible with our pytest framework or do we need to change how this file is accessed
* I currently don't understand why I can only create blobs that hold values 0x73 or lower, otherwise calling `compute_cells` complains
* It seems to me that every cell holds the exact same commitment. So you commit to your blob once and then copy paste the 48 byte commitment into every cell? Seems odd/wasteful to me, but could be the case

Next steps:
* Use the functions in here to prepare a few static examples of blob/commitment/cells/proofs and store them as JSON (yml and me don't talk no more) in our repo. Then we can re-use these static examples in our tests instead of having to repeatedly create them at runtime.

## ✅ Checklist


kevaundray:
> I currently don't understand why I can only create blobs that hold values 0x73 or lower, otherwise calling compute_cells complains

Blobs are not entirely opaque sequence of bytes. You can think of a blob as being 32 byte chunks where each 32 byte must represent an integer less than q, where q is the bls12-381 scalar field. 

The "easier" way to always have it pass is to generate a 31 byte integer and then serialize that into 32 bytes. This is because every 31 byte integer will be less than q since q is roughly a 32 byte integer.

> It seems to me that every cell holds the exact same commitment. So you commit to your blob once and then copy paste the 48 byte commitment into every cell? Seems odd/wasteful to me, but could be the case

Yeah, in the original API it was deduplicated, but then you need another index to index the deduplicated blob because the verify function can take in multiple commitments -- I found this to be more confusing though

tests/osaka/eip7594_peerdas/spec.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    FIELD_ELEMENTS_PER_BLOB = 4096
+    BYTES_PER_FIELD_ELEMENT = 32
+    BYTES_PER_BLOB = FIELD_ELEMENTS_PER_BLOB * BYTES_PER_FIELD_ELEMENT
+
+    # below are mine, not sure if they are defined in specs somewhere
+    KZG_PROOF_LENGTH = 48
+    KZG_COMMITMENT_LENGTH = 48
+    CELL_LENGTH = 2048
```
jtraglia, line 31: There are constants for these in the specs! Please see:

* [`BYTES_PER_COMMITMENT`](https://github.com/ethereum/consensus-specs/blob/cc6996c22692d70e41b7a453d925172ee4b719ad/specs/deneb/polynomial-commitments.md?plain=1#L83)
* [`BYTES_PER_PROOF`](https://github.com/ethereum/consensus-specs/blob/cc6996c22692d70e41b7a453d925172ee4b719ad/specs/deneb/polynomial-commitments.md?plain=1#L84)
* [`BYTES_PER_CELL`](https://github.com/ethereum/consensus-specs/blob/cc6996c22692d70e41b7a453d925172ee4b719ad/specs/fulu/polynomial-commitments-sampling.md?plain=1#L101)

These are not exposed to the bindings, but maybe they should be. I can do this later if you want.

tests/osaka/eip7594_peerdas/helper_functions.py
```diff
@@ -0,0 +1,240 @@
...
+    if len(hex_byte) != 2:
+        print("You should have passed a singular hex byte like e.g. fe to this function")
+        return None
+
+    if int(hex_byte, 16) > 115:
+        # TODO: figure out why this happens
+        print("For some reason bytes larger than this will lead to errors later in compute_cells(), so it's not allowed for now.")  # noqa: E501
+        return None
+
+    hex_byte = hex_byte.lower()
```
jtraglia, line 36: Yes, this is because of [`BLS_MODULUS`](https://github.com/ethereum/consensus-specs/blob/cc6996c22692d70e41b7a453d925172ee4b719ad/specs/deneb/polynomial-commitments.md?plain=1#L82). The uint256 value of each field element must be less than this. Given that Python supports large integers, I think it would be best if this function took a seed value, then generated `FIELD_ELEMENTS_PER_BLOB` random field elements and joined them together to make the blob. Something like:

```python
import random

seed = 42
random.seed(seed)
BLS_MODULUS = 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001
ints = [random.randrange(BLS_MODULUS) for _ in range(FIELD_ELEMENTS_PER_BLOB)]
encoded = [i.to_bytes(BYTES_PER_FIELD_ELEMENT, "big") for i in ints]
blob = b"".join(encoded)
```

tests/osaka/eip7594_peerdas/helper_functions.py
```diff
@@ -0,0 +1,240 @@
...
+    blob: bytes = bytes_from_hex(blob_string)
+
+    return blob
+
+
+def eest_blob_to_kzg_commitment(blob: bytes) -> bytes:
+    """Take a blob and returns a cryptographic commitment to it. Note: Each cell seems to hold a copy of this commitment."""  # noqa: E501
+    # sanity check
+    expected_blob_length = 2**int(math.log2(Spec.BYTES_PER_BLOB))
+    assert len(blob) == expected_blob_length, f"Expected blob of length {expected_blob_length} but got blob of length {len(blob)}"  # noqa: E501
```
jtraglia, line 56: This feels a strange. Why don't we just do `assert len(blob) == Spec.BYTES_PER_BLOB`?

tests/osaka/eip7594_peerdas/helper_functions.py
```diff
@@ -0,0 +1,240 @@
...
+
+    # calculate commitment
+    ts = ckzg.load_trusted_setup("trusted_setup.txt", 0)
```
jtraglia, line 59: This is a slow operation. It would be best if this were only done once. Maybe in some setup function?

tests/osaka/eip7594_peerdas/helper_functions.py
```diff
@@ -0,0 +1,240 @@
...
+
+def eest_compute_cells(blob: bytes) -> list[int]:
+    """Take a blob and returns a list of cells that are derived from this blob."""
+    ts = ckzg.load_trusted_setup("trusted_setup.txt", 0)
+
+    cells = ckzg.compute_cells(blob, ts)
+
+    assert len(cells) == 128
+
+    return cells
```
jtraglia, line 75: Several of these wrapper functions don't appear to be very useful. Could we not just call ckzg directly?

tests/osaka/eip7594_peerdas/helper_functions.py
```diff
@@ -0,0 +1,240 @@
...
+    # calculate commitment
+    ts = ckzg.load_trusted_setup("trusted_setup.txt", 0)
+    commitment = ckzg.blob_to_kzg_commitment(blob, ts)
+
+    assert len(commitment) == Spec.KZG_COMMITMENT_LENGTH, f"Expected {Spec.KZG_COMMITMENT_LENGTH} resulting commitments but got {len(commitment)} commitments"  # noqa: E501
+
+    return commitment
+
+
+def eest_compute_cells(blob: bytes) -> list[int]:
```
jtraglia, line 67: I believe this return type isn't correct. It doesn't return a list of ints. It returns a list of bytes.

jtraglia (commented):
Hey @felix314159, nice work. I would suggest only loading the trusted setup once and removing the wrapper functions. I don't believe these really add anything, right? Ckzg will do these checks too.

tests/osaka/eip7594_peerdas/helper_functions.py
marioevz: We should add the trusted setup as a module constant at the beginning of the file:
```python
from os.path import realpath
from pathlib import Path

TRUSTED_SETUP_FILE_NAME = "trusted_setup.txt"
TRUSTED_SETUP_PATH = Path(realpath(__file__)).parent / TRUSTED_SETUP_FILE_NAME
TRUSTED_SETUP = ckzg.load_trusted_setup(str(TRUSTED_SETUP_PATH), 0)
```

tests/osaka/eip7594_peerdas/helper_functions.py
marioevz: Agreed, great suggestion, and the `seed` can be an int to replace the `hex_byte: str` in the original function.

tests/osaka/eip7594_peerdas/helper_functions.py
```diff
@@ -0,0 +1,240 @@
...
+    for i in range(len(recovered_cells)):
+        assert cells[i] == recovered_cells[i], f"Failed to correctly restore missing cells. At index {i} original cell was {cells[i]} but reconstructed cell does not match: {recovered_cells[i]}"  # noqa: E501
+        assert proofs[i] == recovered_proofs[i], f"Failed to correctly restore missing proofs. At index {i} original proof was {proofs[i]} but reconstructed proof does not match: {recovered_proofs[i]}"  # noqa: E501
+
+    # print("Successful reconstruction")
+
+class PersistentBlobGenerator:
+    # PersistentBlobGenerator("4a") creates blob_4a.json in cwd and contains blob, commitment, cells and proofs
+    encoding: str = "utf-8"
+
```
marioevz, line 146: Pydantic can be of excellent help here:
```python
from typing import List
from pydantic import BaseModel
from ethereum_test_base_types import Bytes

class Blob(BaseModel):
    name: str
    blob: Bytes
    commitments: List[Bytes]
    cells: List[Bytes]
    proofs: List[Bytes]
```

And this can be then converted to json using `model_dump` or `model_dump_json`, and the `to_json` and `from_json` methods can be removed.

marioevz (commented):
This is excellent, thanks for putting this together!

Just a couple of comments to make some parts of it more pythonic, just overall we should try to avoid to use strings when other more appropriate primitive python types can be used.

uv.lock
```diff
@@ -1,10 +1,6 @@
 version = 1
```
danceratopz, line 1: Hey @felix314159, as we discussed, can you please revert the updates to all packages and only modify the required packages. 

Instructions available here:
https://eest.ethereum.org/main/dev/deps_and_packaging/

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+                / would_be_filename
+            )
+        return would_be_static_blob_path
+
+    @staticmethod
+    def NewBlob(fork: Fork, seed: int = 0, timestamp: int = 0) -> "Blob":  # noqa: N802
+        """Construct Blob instances. Fork-specific logic is encapsulated within nested functions."""  # noqa: E501
+        assert fork.engine_get_blobs_version() is not None, (
+            f"You provided fork {fork.name()} but it does not support blobs!"
+        )  # TODO: why does mypy say '"type[BaseFork]" has no attribute "engine_get_blobs_version"'?
```
marioevz, line 101: Can instead use `supports_blobs` here:
```suggestion
        assert fork.supports_blobs(), (
            f"Provided fork {fork.name()} but it does not support blobs!"
        )
```

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+            return Bytes(blob)
+
+        def get_versioned_hash(commitment: Bytes, version: int = 1) -> Hash:
+            """Calculate versioned hash for a given blob."""
+            # TODO: is this in specs or made up by us? do we need it?
+            return Hash(bytes([version]) + sha256(commitment).digest()[1:])
+
+        def get_name(fork_str: str, seed: int, timestamp: int) -> str:
+            """Derive blob name from the seed that generates its data."""
+            return "blob_" + fork_str + "_" + str(seed) + "_" + str(timestamp)
```
marioevz, line 135: Seemingly does the same thing as `Blob.get_filepath`?

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+        elif fork == "cancun":
+            would_be_static_blob_path = (
+                eest_root
+                / "tests"
+                / "cancun"
+                / "eip4844_blobs"
+                / "static_blobs"
+                / would_be_filename
+            )
+        return would_be_static_blob_path
```
marioevz, line 94: After seeing the amount of files this commits to the repo, and that these files will increase fork by fork, I think we should fallback to use local cache and bite the bullet that is the resource utilization when using the blobs locally for the first time.

We could use this trick to determine a local cache folder: https://github.com/ethereum/execution-spec-tests/blob/3810d22f84f206866f66f4f6bf856187c2893ab1/src/pytest_plugins/consume/consume.py#L25-L27

We'll also need a file lock to access the file only once per process/thread (useful for example when filling tests with xdist).

File lock example: https://github.com/ethereum/execution-spec-tests/blob/3810d22f84f206866f66f4f6bf856187c2893ab1/src/ethereum_test_fixtures/file.py#L60-L70

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+    cells: List[Bytes] | None  # None (in json: null)  < Osaka, List[Bytes] >= Osaka
+
+    versioned_hash: Hash
+    name: str  # blob_<fork>_<seed>
+    fork: str
+    seed: int
+    timestamp: int
+
+    @staticmethod
+    def get_filepath(fork: str, seed: int, timestamp: int):
```
marioevz, line 67: I think we can improve this even further by, instead of passing the fork name, we instead pass the characteristics that make the blob unique, e.g.
```suggestion
    def get_filepath(kzg_cell_proofs: int, seed: int):
```
Now Cancun and Prague blobs will resolve to the same file name because they both use `kzg_cell_proofs = 0`, and Osaka and future forks use `kzg_cell_proofs = 128`.

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+            commitment = ckzg.blob_to_kzg_commitment(data, TRUSTED_SETUP)
+
+            assert len(commitment) == BYTES_PER_COMMITMENT, (
+                f"Expected {BYTES_PER_COMMITMENT} resulting commitments but got {len(commitment)} commitments"  # noqa: E501
+            )
+
+            return commitment
+
+        def get_proof(data: Bytes) -> List[Bytes] | Bytes:
+            if fork_str in ["cancun", "prague"]:
```
marioevz, line 154: We need to future proof this by abstracting this characteristic at the fork level.

Add a new abstract class method to the [base fork class](https://github.com/ethereum/execution-spec-tests/blob/3810d22f84f206866f66f4f6bf856187c2893ab1/src/ethereum_test_forks/base_fork.py#L146):
```python
    @classmethod
    @abstractmethod
    def blob_kzg_cell_proofs(cls, block_number: int = 0, timestamp: int = 0) -> int:
        """Return the amount of KZG cell proofs that the blobs must contain for a given fork, zero meaning that the fork does not support cell proofs."""
        pass
```

And then return the proper value for the `Frontier`, `Cancun` and `Osaka` forks:
```python
class Frontier:
    ...
    @classmethod
    def blob_kzg_cell_proofs(cls, block_number: int = 0, timestamp: int = 0) -> int:
        """Frontier does not support blobs."""
        raise Exception("Frontier does not support blobs")
```

```python
class Cancun:
    ...
    @classmethod
    def blob_kzg_cell_proofs(cls, block_number: int = 0, timestamp: int = 0) -> int:
        """Cancun supports blobs but they do not contain cell proofs."""
        return 0
```

```python
class Osaka:
    ...
    @classmethod
    def blob_kzg_cell_proofs(cls, block_number: int = 0, timestamp: int = 0) -> int:
        """Osaka supports blobs and cell proofs."""
        return 128
```

Then you can do `fork.blob_kzg_cell_proofs()` in here and modify the function's behavior accordingly.

src/ethereum_test_forks/__init__.py
```diff
@@ -98,3 +100,23 @@
...
+BYTES_PER_PROOF = 48
+BYTES_PER_COMMITMENT = 48
+KZG_ENDIANNESS: Literal["big"] = "big"
+
+#       eip-7691
+MAX_BLOBS_PER_BLOCK_ELECTRA = 9
+TARGET_BLOBS_PER_BLOCK_ELECTRA = 6
+MAX_BLOB_GAS_PER_BLOCK = 1179648
+TARGET_BLOB_GAS_PER_BLOCK = 786432
+BLOB_BASE_FEE_UPDATE_FRACTION_PRAGUE = 5007716
```
marioevz, line 122:
```suggestion
```
Usually we put constants in `src/ethereum_test_forks/forks/forks.py` as return values of the methods for each fork, but even so, I think the only ones we need from this bunch is the kzg cell proof count at the moment, so we can remove the rest (see previous comments about the blob kzg cell proof methods).

src/ethereum_test_forks/forks/forks.py
```diff
@@ -543,6 +543,11 @@ def pre_allocation_blockchain(cls) -> Mapping:
         """
         return {}
 
+    @classmethod
+    def engine_get_blobs_version(cls, block_number: int = 0, timestamp: int = 0) -> Optional[int]:
+        """At genesis, blobs cannot be retrieved through the engine API."""
+        return None
+
```
marioevz, line 550: We can remove `engine_get_blobs_version` entirely from this PR as it is not really used and it's already part of #1644.

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+            cells=cells,
+            versioned_hash=versioned_hash,
+            name=name,
+            fork=fork_str,
+            seed=seed,
+            timestamp=timestamp,
+        )
+
+    @staticmethod
+    def LoadBlobFromFile(file_path: str) -> "Blob":  # noqa: N802
```
marioevz, line 201:
```suggestion
    def from_file(file_path: str) -> "Blob":  # noqa: N802
```

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+            detected_timestamp = parts[3].removesuffix(".json")
+            assert detected_seed.isdigit(), (
+                f"Failed to extract seed from filename. Ended up with seed {detected_seed} given filename {file_path}"  # noqa: E501
+            )
+            assert detected_timestamp.isdigit(), (
+                f"Failed to extract timestamp from filename. Ended up with timestamp {detected_timestamp} given filename {file_path}"  # noqa: E501
+            )
+            file_path = Blob.get_filepath(
+                detected_fork, int(detected_seed), int(detected_timestamp)
+            )
```
marioevz, line 224: Feels unnecessary, is it just for debugging purposes?

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+            print(f"Blob {output_location} already exists. It will be overwritten.")
+
+        with open(output_location, "w", encoding="utf-8") as f:  # overwrite existing
+            f.write(json_str)
+
+    def verify_cell_kzg_proof_batch(self, cell_indices: list) -> bool:  # noqa: E501
+        """Check whether all cell proofs are valid and returns True only if that is the case."""
+        assert self.fork in ["osaka"], (
+            f"verify_cell_kzg_proof_batch() is not available for fork: {self.fork}"
+        )
```
marioevz, line 252:
```suggestion
        assert self.fork.blob_kzg_cell_proofs() > 0, (
            f"verify_cell_kzg_proof_batch() is not available for fork: {self.fork}"
        )
```

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+
+        assert self.cells is not None, "..."  # TODO: write text
+
+        assert isinstance(self.proof, list), (
+            "This function only works when self.proof is a list, but it seems to be of type bytes (not a list)"  # noqa: E501
+        )
+
+        assert len(self.cells) == 128, (
+            f"You are supposed to pass a full cell list with 128 elements to this function, but got list of length {len(self.cells)}"  # noqa: E501
+        )
```
marioevz, line 291:
```suggestion
        assert len(self.cells) == self.fork.blob_kzg_cell_proofs(), (
            f"You are supposed to pass a full cell list with {self.fork.blob_kzg_cell_proofs()} elements to this function, but got list of length {len(self.cells)}"  # noqa: E501
        )
```
Same for the rest of the function, it can be adapted in the same way.

src/ethereum_test_types/eest_types.py
marioevz, line 1: I really don't think this rename is necessary, were you getting an error somehow with the original name? If so, could you share so we can help debug pls?

tests/cancun/eip4844_blobs/static_blobs/blob_cancun_0_0.json
marioevz, line 1: When we take the local cache approach, we should remove these static blob files from the branch.

marioevz (commented):
Left a lot of comments but I think these are necessary to future-proof the new code introduced 👍 

Thanks!

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+    # TARGET_BLOBS_PER_BLOCK_ELECTRA,
+    Cancun,
+    Fork,
+    Osaka,
+    Prague,
+)
+
+TRUSTED_SETUP_FILE_NAME = "blob_trusted_setup.txt"
+TRUSTED_SETUP_PATH = Path(realpath(__file__)).parent / TRUSTED_SETUP_FILE_NAME
+TRUSTED_SETUP = ckzg.load_trusted_setup(str(TRUSTED_SETUP_PATH), 0)
```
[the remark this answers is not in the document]
felix314159: Good idea but I will make this function void, no need to return the trusted setup if it is already accessible

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+                / "tests"
+                / "cancun"
+                / "eip4844_blobs"
+                / "static_blobs"
+                / would_be_filename
+            )
+        return would_be_static_blob_path
+
+    @staticmethod
+    def NewBlob(fork: Fork, seed: int = 0, timestamp: int = 0) -> "Blob":  # noqa: N802
```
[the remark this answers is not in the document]
felix314159: Yes we should use snake case I agree. However, I strongly disagree calling it `from_seed` as the seed is just an optional parameter. You probably wanted to name it `from_fork`.

src/ethereum_test_types/blob.py
```diff
@@ -0,0 +1,469 @@
...
+        # if this blob already exists then load from file
+        blob_location: Path = Blob.get_filepath(fork_str, seed, timestamp)
+        if blob_location.exists():
+            print(f"Blob exists already, reading it from file {blob_location}")
+            return Blob.LoadBlobFromFile(str(blob_location))
+
+        def generate_blob_data(rng_seed: int = 0) -> Bytes:
+            """Calculate blob data deterministically via provided seed."""
+            # apply RNG seed
+            random.seed(rng_seed)
```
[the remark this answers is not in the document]
felix314159: Very good catch, thanks

src/ethereum_test_types/blob.py
felix314159: We are not the same: `get_name` only gets the name (e.g. blob_0_0) and `get_filepath` determine the relative filepath to the blob file (e.g. ./tests/osaka/eip7594_peerdas/static_blobs/blob_0_0.json). I think having one line of almost duplicated logic (.json vs no .json) is okay, we aren't going to constantly change this naming scheme.

src/ethereum_test_types/blob.py
felix314159: What problem does this solve? I intentionally named the blobs so that from the filename alone I can tell which fork we created them for. Yes Cancun and Prague use the same blob format/rules but it is not harmful to have the info which blob they were created for in the filename. I feel like a common theme of our repository is having the same name re-used in different contexts (e.g. types.py in multiple locations with different contents) which makes everything harder to understand.

I do agree that the resulting .json files should not be pushed to the repository, so the local duplication (e.g. there might be a case where you generate the same blob for Cancun and Prague and the only difference in their fork name) is not an issue IMO.

src/ethereum_test_types/blob.py
felix314159: I feel like this negatively affects the readability of the code. We already have `fork.supports_blobs()` and now if we add `fork.blob_kzg_cell_proofs()` we have a second way of determining whether a fork supports blobs (and if it supports blobs it also supports proofs of some kind) or not. Yes, it is true that your proposed `blob_kzg_cell_proofs()` returns more info (which kind of proofs are supported), but we still would have to manually add `blob_kzg_cell_proofs()` to every new fork that happens in the future (just like in the current design we have to add a singular string (fork name) to the list to handle the new fork). 

What does it even mean to receive value `128`? I see what you are trying to do but I am not convinced by this approach. I think would rather have:
```
if fork_str in ["cancun", "osaka"]:
    # logic that is relevant for the forks clearly described in the line above
elif fork_str in ["osaka", "futureForkAddedHere"]:
    # logic that is relevant for the forks clearly described in the line above
```
than
```
magicInt = fork.blob_kzg_cell_proofs()
if magicInt > someInt:
    # implicitely handles fork ? (it is unclear what someInt means)
elif magicInt > someOtherInt:
    # implicitely handles fork ? (it is unclear what someOtherInt means)
```

src/ethereum_test_types/blob.py
felix314159: If the test writer tries to import an existing blob via filename this code ensures that the name of the blob file is valid (as in matches the expected format). I would prefer to keep this, if someone tries to import an invalidly named blob this way we at least get a descriptive error message

tests/cancun/eip4844_blobs/static_blobs/blob_cancun_0_0.json
felix314159: Yes, let's remove them

src/ethereum_test_forks/__init__.py
felix314159: I'm not sure what is the best way to do this, we need to find a decent middleground between no abstraction and having a function for every single constant that might exist in our fork class. Maybe we can create a dictionary (constant_name: value) in the forks class called BLOB_CONSTANTS and assign all values. we also create the function `get_constant()`. now in every fork after cancun (including cancun) we do the following:
```
CONSTANTS = {
        **Fork.BLOB_CONSTANTS,  # Inherit all constants
        "MY_CONSTANT_3": 123,  # Override only variable 3 with this value
    }
```
if no constant has to be changed for this fork you just put `pass`.
if you want to access a constant you do fork.get_constant("MY_CONSTANT_3"). this means we only have to define one construct for each blob fork, not have a new function for every single constant. i still think this can be improved, will think about it more

src/ethereum_test_types/blob.py
marioevz: It solves duplication of work because timestamp and fork indirectly affect the output (timestamp only in the case of transition forks), but we can reduce the amount of produced files by using the variables that directly affect the output, namely the cell proof count.

So for example if you have two tests and both reference the blobs with seeds 1 and 2:
```python
@pytest.mark.valid_at_transition_to("Osaka")
def test_1(fork: Fork):
    blob_1 = Blob(fork=fork, seed=1, timestamp=14_999)
    blob_0 = Blob(fork=fork, seed=2, timestamp=15_000)

@pytest.mark.valid_from("Prague")
def test_2(fork: Fork):
    blob_1 = Blob(fork=fork, seed=1, timestamp=1000)
    blob_0 = Blob(fork=fork, seed=2, timestamp=2000)
```
With the fork-name/timestamp/seed approach you will generate files:
```
blob_PragueToOsakaAtTime15k_1_14999.json
blob_PragueToOsakaAtTime15k_2_15000.json
blob_Prague_1_1000.json
blob_Prague_2_2000.json
blob_Osaka_1_1000.json
blob_Osaka_2_2000.json
```
And the cell-proofs/seed approach you will generate:
```
blob_1_cell_proofs_0.json
blob_2_cell_proofs_0.json
blob_1_cell_proofs_128.json
blob_2_cell_proofs_128.json
```
So it's going to grow with the amount of forks and the amount of tests (due to the timestamp).

src/ethereum_test_types/eest_types.py
marioevz: Good catch :D I think this is particular to using `uv run python ./src/...` but still I think it should be ok to finally split types.py into different files.

But I think that one deserves its own PR, wdyt?

src/ethereum_test_types/blob.py
felix314159: Ok, I am now implementing storing blobs in the local cache

src/ethereum_test_forks/__init__.py
```diff
@@ -1,5 +1,7 @@
 """Ethereum test fork definitions."""
 
+from typing import Literal
+
```
marioevz, line 4: Seems unnecessary.
```suggestion
```

tests/cancun/eip4844_blobs/test_blob_txs_full.py
```diff
@@ -237,17 +243,13 @@ def generate_full_blob_tests(
...
         pytest.param(
             [  # Txs
                 [  # Blobs per transaction
-                    Blob(
-                        data=bytes(blob_size),
-                        kzg_commitment=INF_POINT,
-                        kzg_proof=INF_POINT,
-                    ),
+                    # you can provide the blob via file name
+                    Blob.from_file("blob_cancun_0_0"),
```
marioevz, line 252: We should only use the more-abstract from_fork method in tests.

src/ethereum_test_types/blob_trusted_setup.txt
marioevz, line 1: Minor nit-pick but could we rename this file to "kzg_trusted_setup.txt"?

src/ethereum_test_types/blob_types.py
```diff
@@ -1,19 +1,514 @@
...
-    kzg_commitment: Bytes
-    kzg_proof: Bytes | None = None
-    kzg_cell_proofs: List[Bytes] | None = None
+    commitment: Bytes
+    proof: List[Bytes] | Bytes  # Bytes < Osaka, List[Bytes] >= Osaka
+    cells: List[Bytes] | None  # None (in json: null)  < Osaka, List[Bytes] >= Osaka
+
+    versioned_hash: Hash
+    name: str  # blob_<fork>_<seed>
+    fork: str
```
marioevz, line 52:
```suggestion
    fork: Fork
```
After #1686, and can also delete all `str`<->`Fork` conversions everywhere in this file.

felix314159: A side-effect of this change is that pydantic now struggles to serialize and deserialize Blob object (when we sticked to basic python types `Blob.model_validate_json(json_str)` and `self.model_dump_json()` 'just worked' but now we have to add hard-to-read logic to tell pydantic how to handle the Fork type at serialization and deserialization:

```
    @field_validator("fork", mode="before")
    @classmethod
    def validate_fork(cls, v):
        """
        When reading JSON file and trying to go back from fork string to fork object we must
        tell pydantic how to do this.
        """
        if isinstance(v, str):
            return fork_string_to_object(v)
        return v

    @field_serializer("fork")
    def serialize_fork(self, fork: Fork) -> str:
        """
        When trying to serialize a Blob object into a JSON file we must
        tell pydantic how to do this.
        """
        return fork.name()
```

tests/cancun/eip4844_blobs/test_blob_txs_full.py
```diff
@@ -237,17 +241,12 @@ def generate_full_blob_tests(
...
     return [
         pytest.param(
             [  # Txs
                 [  # Blobs per transaction
-                    Blob(
-                        data=bytes(blob_size),
-                        kzg_commitment=INF_POINT,
-                        kzg_proof=INF_POINT,
-                    ),
+                    Blob.from_fork(Cancun),
```
marioevz, line 249:
```suggestion
                    Blob.from_fork(fork),
```
We can just pass the fork here to avoid constraining the test to a hard-coded fork.

tests/cancun/eip4844_blobs/test_blob_txs_full.py
```diff
@@ -256,11 +255,7 @@ def generate_full_blob_tests(
         pytest.param(
             [  # Txs
                 [  # Blobs per transaction
-                    Blob(
-                        data=bytes(blob_size),
-                        kzg_commitment=INF_POINT,
-                        kzg_proof=INF_POINT,
-                    )
+                    Blob.from_fork(Cancun),
```
marioevz, line 258:
```suggestion
                    Blob.from_fork(fork),
```
Same here.

tests/cancun/eip4844_blobs/test_blob_txs_full.py
```diff
@@ -270,11 +265,7 @@ def generate_full_blob_tests(
...
```
marioevz, line 268:
```suggestion
                    Blob.from_fork(fork),
```

tests/cancun/eip4844_blobs/test_blob_txs_full.py
```diff
@@ -8,7 +8,9 @@
 
 import pytest
 
+from ethereum_test_base_types.base_types import Hash
 from ethereum_test_forks import Fork
+from ethereum_test_forks.forks.forks import Cancun
```
marioevz, line 13:
```suggestion
```
If the rest of the comments are applied.

marioevz (approved):
LGTM!

Just a couple of comments but we can merge after those 👍

tests/cancun/eip4844_blobs/test_blob_txs_full.py
felix314159: done

src/ethereum_test_types/blob_types.py
```diff
@@ -1,19 +1,449 @@
...
+            # sanity check
+            field_elements: int = cast(int, fork.get_blob_constant("FIELD_ELEMENTS_PER_BLOB"))
+            bytes_per_field: int = cast(int, fork.get_blob_constant("BYTES_PER_FIELD_ELEMENT"))
+            assert len(data) == field_elements * bytes_per_field, (
+                f"Expected blob of length "
+                f"{field_elements * bytes_per_field} but got blob of length {len(data)}"
+            )
+
+            # calculate commitment
+            commitment = ckzg.blob_to_kzg_commitment(data, Blob._trusted_setup)
```
marioevz, line 165:
```suggestion
            commitment = ckzg.blob_to_kzg_commitment(data, Blob.trusted_setup())
```
Always use the `Blob.trusted_setup()` outside of that function, and that function is the only one that gets to access `Blob._trusted_setup` directly, because otherwise we have the risk of this code-path being reached first and the variable not being initialized.

src/ethereum_test_types/transaction_types.py
```diff
@@ -652,36 +655,58 @@ class NetworkWrappedTransaction(CamelModel, RLPSerializable):
...
+        rlp([tx_payload_body,                   blobs, commitments, proofs])
+
+    >= Osaka:
+        rlp([tx_payload_body, wrapper_version,  blobs, commitments, cell_proofs])
     """
 
     tx: Transaction
-    wrapper_version: Literal[1] | None = None
-    blobs: Sequence[Blob]
+    blob_objects: Sequence[Blob]
```
marioevz, line 667: Is the `blob_objects` -> `blobs` rename really necessary? I feel like we are modifying code without any gains.

src/ethereum_test_types/transaction_types.py
```diff
@@ -692,10 +717,53 @@ def get_rlp_fields(self) -> List[str]:
...
+        #       Proofs      []kzg4844.Proof
+        #   }
+
+        #   type versionedBlobTxWithBlobs struct {
+        #       BlobTx      *BlobTx
+        #       Version     byte
+        #       Blobs       []kzg4844.Blob
+        #       Commitments []kzg4844.Commitment
+        #       Proofs      []kzg4844.Proof
+        #   }
```
marioevz, line 747:
```suggestion
```
We can remove this comment.

src/ethereum_test_types/transaction_types.py
```diff
@@ -692,10 +717,53 @@ def get_rlp_fields(self) -> List[str]:
...
+            "commitments",
+            *rlp_proofs,
+            *rlp_cell_proofs,
+        ]
+
+        assert ("proofs" in rlp_fields) or ("cell_proofs" in rlp_fields), (
+            "Neither proofs nor cell_proofs are in rlp_fields. Critical error!"
+        )
+
+        logger.debug(f"Ended up with this rlp field list: {rlp_fields}")
```
marioevz, line 764:
```suggestion
```
Can also be removed.

marioevz (commented):
More comments but hopefully this is one of the last iterations, thanks again for putting this together!

src/ethereum_test_types/blob_types.py
felix314159: I don't think this could have happened but I agree that we should apply your suggestion anyway, it would never double load so I agree that this is safer at close to 0 performance cost

src/ethereum_test_base_types/conversions.py
```diff
@@ -47,6 +52,16 @@ def to_bytes(input_bytes: BytesConvertible) -> bytes:
             input_bytes = "0" + input_bytes
         return bytes.fromhex(input_bytes)
 
+    # handle wrapper_version field for >= osaka
```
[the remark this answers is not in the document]
felix314159: I don't understand what this comment means but I think it's fine to leave things as is. It's safe to assume that the wrapper_version will always be stored as 1 byte (ethereum won't have that many different wrapper versions that we arrive at 256 in our lifetimes, there is no reason for them to suddenly choose a wrapper version consisting of multiple bytes). Also like in any other popular language, Python stdlib functions like `to_bytes` will rarely ever see backwards-incompatible changes, and if they do change the function signature of this it would be trivial to spot with mypy

src/ethereum_test_types/transaction_types.py
felix314159: I renamed `blobs` into `blob_objects`, not the other way around. And yes I think we greatly benefit from this rename because we have a function `blobs()` that dynamically creates a field named `blobs` which holds not blobs but only the data of blobs. With this rename we have access to `blob_objects` (the actual objects) and `blobs` (their data, accessed via e.g. b.data). If you had argued that we should rename 'blobs' (the function) to something else I would have agreed, but now I do not want to mess with working code ;)

src/ethereum_test_types/transaction_types.py
felix314159: Yeah true, this is one of the comments that seem helpful now but will be harmful when geth changes their code (which we will not notice) and a wrong/misleading comment is worse than no comment at all

src/ethereum_test_types/transaction_types.py
felix314159: Sure we can remove it, but generally when the logging level is set to debug and something can be helpful for debugging (here: if the RLP field format changes in some future fork) it should be left turned on IMO. This is a subjective take, but I do think that there is an argument to be made for leaving verbose logging turned on even in a production setting, as long as it has to manually be toggled on by setting the logging level to DEBUG

src/ethereum_test_types/blob_types.py
```diff
@@ -1,19 +1,449 @@
...
+            data=data,
+            commitment=commitment,
+            proof=proof,
+            cells=cells,
+            versioned_hash=versioned_hash,
+            name=name,
+            fork=fork,
+            seed=seed,
+            timestamp=timestamp,
+        )
```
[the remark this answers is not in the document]
felix314159: Good idea, i added filelock to reading and writing blobs

felix314159:
It can be merged now

src/ethereum_test_base_types/conversions.py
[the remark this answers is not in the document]
marioevz: The argument is that `Bytes` (which uses `to_bytes` internally) is used throughout the code base in pydantic models that do arbitrary length of bytes, e.g. [`Transaction.data`](https://github.com/ethereum/execution-spec-tests/blob/851f9f5bc88d25f869a7fcd7798b10b9c5bceb7e/src/ethereum_test_types/transaction_types.py#L175), [`Header.extra_data`](https://github.com/ethereum/execution-spec-tests/blob/851f9f5bc88d25f869a7fcd7798b10b9c5bceb7e/src/ethereum_test_specs/blockchain.py#L115), so it doesn't quite make sense that this single-byte constrain applies to them.

Getting into details as to why `Bytes` does not accept integers, take the example of [`Transaction.data`](https://github.com/ethereum/execution-spec-tests/blob/851f9f5bc88d25f869a7fcd7798b10b9c5bceb7e/src/ethereum_test_types/transaction_types.py#L175), and what `Transaction(data=0)` should mean,  `0x`, `0x00`, or even `0x0000000000..00`?

Since the data of the transaction is normally important for tests, this forces the user to always pass a value where the byte-length is strictly defined.

We have the [`FixedSizeBytes`](https://github.com/ethereum/execution-spec-tests/blob/851f9f5bc88d25f869a7fcd7798b10b9c5bceb7e/src/ethereum_test_base_types/base_types.py#L282) base class, which accepts integers for instantiation because it always has a strictly defined byte-length, so if we really want this to be a single byte type, we can create a type in [`src/ethereum_test_base_types/base_types.py`](https://github.com/ethereum/execution-spec-tests/blob/main/src/ethereum_test_base_types/base_types.py):
```python
class Byte(FixedSizeBytes[1]):  # type: ignore
    """Class that helps represent a single byte in tests."""

    pass
```

and use that as the field type.

src/ethereum_test_types/transaction_types.py
```diff
@@ -652,36 +655,58 @@ class NetworkWrappedTransaction(CamelModel, RLPSerializable):
...
+
+    >= Osaka:
+        rlp([tx_payload_body, wrapper_version,  blobs, commitments, cell_proofs])
     """
 
     tx: Transaction
-    wrapper_version: Literal[1] | None = None
-    blobs: Sequence[Blob]
+    blob_objects: Sequence[Blob]
+    wrapper_version: Union[Bytes | None] = None  # only exists in >= osaka
```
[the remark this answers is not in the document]
marioevz: See previous comment.

marioevz (changes requested):

src/ethereum_test_types/blob_types.py
marioevz: Yet to be fixed in the latest commit.

src/ethereum_test_types/transaction_types.py
marioevz, line 668: Thanks for removing the changes to `src/ethereum_test_base_types/conversions.py`, but this still needs a fix, run this in the python cli to see what I mean:
```python
from ethereum_test_types import NetworkWrappedTransaction, Transaction
NetworkWrappedTransaction(tx=Transaction(), blob_objects=[], wrapper_version=1)
```

My suggestion is to just roll it back to an `int` or `None`
```suggestion
    wrapper_version: Literal[1] | None = None  # only exists in >= osaka
```

Or the other approaches could be `Bytes` or `None` and use a pydantic validator, or change `tx_wrapper_version` in `tests/osaka/eip7594_peerdas/test_get_blobs.py` to return `b"\1"`.

marioevz (approved):
Thanks!

PR #1616, opened 2025-05-16
feat(tests/zkevm): Test jump and jumpdest

codygunton:
## 🗒️ Description
Replay of https://github.com/ethereum/execution-spec-tests/pull/1528 which died because it targeted a feature branch.

Add two `JUMP`-heavy workloads to see how zkVMs handle this. Cycle counts from [the Reth fork](https://github.com/kevaundray/reth/tree/kw/stateless-experiment-zkvm-benchmarks):

### JUMP-heavy test 
```
[crates/zkvm-benchmarks/host/src/main.rs:57:9] &reports = [
    WorkloadMetrics {
        name: "tests/zkevm/test_worst_compute.py::test_worst_jumps[fork_Cancun-blockchain_test]-1",
        total_num_cycles: 700358751,
        region_cycles: {
            "validation": 449377028,
            "read_input": 250977795,
            "verify-witness": 342781,
        },
    },
]
 ```

## JUMPDEST-heavy test
```
[crates/zkvm-benchmarks/host/src/main.rs:57:9] &reports = [
    WorkloadMetrics {
        name: "tests/zkevm/test_worst_compute.py::test_worst_jumpdests[fork_Cancun-blockchain_test]-1",
        total_num_cycles: 1053017248,
        region_cycles: {
            "read_input": 250976958,
            "verify-witness": 543260,
            "validation": 802036362,
        },
    },
]
```

## 🔗 Related Issues
N/A

## ✅ Checklist

- [ ] All: Set appropriate labels for the changes. <- I seem to not have an abillity to set these, else I would use the same as https://github.com/ethereum/execution-spec-tests/pull/1514
- [X] All: Considered squashing commits to improve commit history.
- [ ] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md). <- I don't see precedents for adding individual tests.
- [X] Tests: Included the type and version of evm t8n tool used to locally execute test cases:  e.g., ref with commit hash or geth 1.13.1-stable-3f40e65. <- Just using `ExecutionSpecsTransitionTool`

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -18,6 +18,7 @@ from ethereum_test_tools import (
     Bytecode,
     Environment,
     Transaction,
+    While,
 )
 from ethereum_test_tools.vm.opcode import Opcodes as Op
 
@@ -282,3 +283,79 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
         raise ValueError(f"Code size {len(code)} exceeds maximum code size {MAX_CODE_SIZE}")
 
     return code
+
+
+@pytest.mark.zkevm
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.slow
+def test_worst_jumps(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+):
+    """Test running a JUMP-intensive contract."""
+    env = Environment()
+
+    def jump_seq():
+        return Op.JUMP(Op.ADD(Op.PC, 1)) + Op.JUMPDEST
+
+    bytes_per_seq = len(jump_seq())
+    seqs_per_call = MAX_CODE_SIZE // bytes_per_seq
+
+    # Create and deploy the jump-intensive contract
+    jumps_code = sum([jump_seq() for _ in range(seqs_per_call)])
+    jumps_address = pre.deploy_contract(code=jumps_code)
+
+    # Call the contract repeatedly until gas runs out.
+    caller_code = While(body=Op.POP(Op.CALL(address=jumps_address)))
+    caller_address = pre.deploy_contract(caller_code)
+
+    txs = [
+        Transaction(
+            to=caller_address,
+            gas_limit=env.gas_limit,
+            sender=pre.fund_eoa(),
+        )
+    ]
+
+    blockchain_test(
+        genesis_environment=env,
+        pre=pre,
+        post={},
+        blocks=[Block(txs=txs)],
+    )
+
+
+@pytest.mark.zkevm
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.slow
+def test_worst_jumpdests(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+):
+    """Test running a JUMPDEST-intensive contract."""
+    env = Environment()
+
+    # Create and deploy a contract with many JUMPDESTs
+    jumpdests_code = sum([Op.JUMPDEST] * MAX_CODE_SIZE)
+    jumpdests_address = pre.deploy_contract(code=jumpdests_code)
+
+    # Call the contract repeatedly until gas runs out.
+    caller_code = While(body=Op.POP(Op.CALL(address=jumpdests_address)))
+    caller_address = pre.deploy_contract(caller_code)
+
+    txs = [
+        Transaction(
+            to=caller_address,
+            gas_limit=env.gas_limit,
+            sender=pre.fund_eoa(),
+        )
+    ]
+
+    blockchain_test(
+        genesis_environment=env,
+        pre=pre,
+        post={},
+        blocks=[Block(txs=txs)],
+    )
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -304,7 +304,7 @@ def test_worst_jumps(
 
     # Create and deploy the jump-intensive contract
     jumps_code = sum([jump_seq() for _ in range(seqs_per_call)])
-    jumps_address = pre.deploy_contract(code=jumps_code)
+    jumps_address = pre.deploy_contract(code=bytes(jumps_code))
 
     # Call the contract repeatedly until gas runs out.
     caller_code = While(body=Op.POP(Op.CALL(address=jumps_address)))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by marioevz — APPROVED** [review]
LGTM, thanks and again sorry for the auto-close of the previous PR :)

**Comment by jsign** [comment]
I'll take a further look in some mins!

File: tests/zkevm/test_worst_compute.py [hunk 1]
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
     return code
+
+
+@pytest.mark.zkevm
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.slow
+@pytest.mark.parametrize(
+    "gas_limit",
+    [36_000_000],
+)
```
jsign, line 294:
```suggestion
```
See next-next comment.

tests/zkevm/test_worst_compute.py
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
+def test_worst_jumps(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_limit: int,
```
jsign, line 299:
```suggestion

```

tests/zkevm/test_worst_compute.py
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
+):
+    """Test running a JUMP-intensive contract."""
+    env = Environment(gas_limit=gas_limit)
```
jsign, line 302:
```suggestion
    env = Environment()
```
Since the testing-framework added support for `--block-gas-limit` we will parametrize using that. For example, [the CI filler](https://github.com/ethereum/execution-spec-tests/blob/main/.github/configs/feature.yaml#L16) already does this.

If we want to generate the suite for a higher gas limit, we can change the value in the CLI.

tests/zkevm/test_worst_compute.py
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
+    gsc = fork.gas_costs()
+
+    def jump_seq(i):
+        return Op.PUSH2(4 + 5 * i) + Op.JUMP + Op.JUMPDEST
```
jsign, line 306: We could do also:
```suggestion
        return Op.JUMP(Op.ADD(Op.PC, 1)) + Op.JUMPDEST
```
that sounds more robust, but will eat some extra byte and gas (but not that much I think). 

If the cycle count of that change isn't acceptable, we can simplify a bit too with:
```suggestion
        return Op.JUMP(4 + 5 * i) + Op.JUMPDEST
```
which should be equivalent to your code.

tests/zkevm/test_worst_compute.py
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
+
+    gas_per_seq = gsc.G_VERY_LOW + gsc.G_MID + gsc.G_JUMPDEST
+    bytes_per_seq = 5
```
jsign, line 309: Maybe:
```suggestion
    bytes_per_seq = len(jump_seq(1))
```
or something similar?

tests/zkevm/test_worst_compute.py
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
+    seqs_per_call = MAX_CODE_SIZE // bytes_per_seq
+    gas_per_call = gas_per_seq * seqs_per_call
+
+    # Create and deploy the jump-intensive contract
+    jumps_code = sum(jump_seq(i) for i in range(seqs_per_call))
+    jumps_address = pre.deploy_contract(code=jumps_code)
+
+    # Determine how many calls to the jumps contract to make.
```
jsign, line 317: Ah, I usually have a `While` without a condition (that's supported), and let the tx run until it runs out of gas.

So, no need to calculate the exact limit since the prover will have to prove everything anyway. :)

Maybe you want to try it and see if you simplify it that way, you should get the same cycles (or even a tiny more).

tests/zkevm/test_worst_compute.py
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
+    )
+
+
+@pytest.mark.zkevm
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.slow
+@pytest.mark.parametrize(
+    "gas_limit",
+    [36_000_000],
+)
```
jsign, line 364: Ditto

tests/zkevm/test_worst_compute.py
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
+    pre: Alloc,
+    fork: Fork,
+    gas_limit: int,
+):
+    """Test running a JUMPDEST-intensive contract."""
+    env = Environment(gas_limit=gas_limit)
+    gsc = fork.gas_costs()
+
+    # Create and deploy a contract with many JUMPDESTs
+    jumpdests_code = Op.PUSH1(3) + Op.JUMP + sum(Op.JUMPDEST for _ in range(MAX_CODE_SIZE - 3))
```
jsign, line 376: You can remove the `Op.JUMP`, it is fine to directly execute `Op.JUMPDEST` contracts since are noops.

tests/zkevm/test_worst_compute.py
```diff
@@ -282,3 +283,135 @@ def code_loop_precompile_call(calldata: Bytecode, attack_block: Bytecode):
...
+    jumpdests_address = pre.deploy_contract(code=jumpdests_code)
+    gas_per_call = gsc.G_VERY_LOW + gsc.G_MID + (MAX_CODE_SIZE - 3) * gsc.G_JUMPDEST
+
+    # Determine how many calls to the jumps contract to make.
```
jsign, line 380: Ditto.

jsign (commented):
Left some comments for your consideration!

tests/zkevm/test_worst_compute.py
codygunton: Right, this makes sense--the prover needs to prove the full tx even to show it reverted. I don't know much about how these clients work so I was a little worried that perhaps are some shortcuts for identifying if a tx runs out of gas, but I see by manually checking the cycle counts are similar that there is no issue there, so I just max out gas as you suggest.

jsign (approved):

Issue #1571, opened 2025-05-07
Create Blockchain tests that contain arithmetic opcodes that consume the whole block.


kevaundray:
Note: @chfast noted that we can make this more generic by allowing one to specify how much of the block will be filled, and the zkEVM fixtures would just use the max for now.

## General context

Generally the structure of these tests are split into two:

- Setup phase
- Benchmarking block

**Setup phase**

The setup phase encompasses the pre-state and blocks that are needed in order to setup the block full of a specific block.
An example of this would be a "block full of EXTCODESIZE" where the setup phase would consist of N different contracts being deployed, where N is the number of EXTCODESIZE opcdes we want to put in the final block. 

**Benchmarking block**

For benchmarking, we assume that the final block is the one with the block full of the particular opcode we want to benchmark

A subtask of #1453 

## Arithmetic opcodes

[ ] ADD
[ ] MUL
[ ] SUB
[ ] DIV
[ ] SDIV
[ ] MOD
[ ] SMOD
[ ] ADDMOD
[ ] MULMOD
[ ] EXP
[ ] SIGNEXTEND

kevaundray:
Noting here that some of them like LT, GT are completed -- its just that they are not worse case -- this is okay and I think we just need to specify this in the linked issue and when everything is complete, we can create a longstanding issue to optimize them

PR #1625, opened 2025-05-19
feat(tests): add benchmarks with mass shift opcodes

chfast:
## 🗒️ Description
This adds the benchmarks for EVM execution focused on shift instructions. Shift instructions are executed in left-right pairs and the shift amount is taken from the value pool on the stack.

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/1571

## ✅ Checklist

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -6,6 +6,7 @@ Tests running worst-case compute opcodes and precompile scenarios for zkEVMs.
 """
 
 import math
+import random
 
 import pytest
 
@@ -21,6 +22,7 @@ from ethereum_test_tools import (
     While,
 )
 from ethereum_test_tools.vm.opcode import Opcodes as Op
+from ethereum_test_vm import Opcode
 
 REFERENCE_SPEC_GIT_PATH = "TODO"
 REFERENCE_SPEC_VERSION = "TODO"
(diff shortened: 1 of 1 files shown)


tests/zkevm/test_worst_compute.py
```diff
@@ -519,3 +521,91 @@ def test_worst_binop_simple(
...
+        while True:
+            index = rng.randint(0, len(shift_amounts) - 1)
+            sh = shift_amounts[index]
+            new_v = shift_fn(v, sh) % 2**256
+            if new_v != 0:
+                return new_v, index
+
+    def make_dup(i):
+        """Create a DUP instruction to get the i-th shift amount constant from the stack."""
+        return Opcode(0x80 + (len(shift_amounts) - i))
```
chfast, line 565: Is there a nicer way of creating DUP opcodes based on index?

tests/zkevm/test_worst_compute.py
```diff
@@ -519,3 +521,91 @@ def test_worst_binop_simple(
...
+        return to_unsigned(to_signed(x) >> s)
+
+    def select_shift_amount(shift_fn, v):
+        """Select a shift amount that will produce a non-zero result."""
+        while True:
+            index = rng.randint(0, len(shift_amounts) - 1)
+            sh = shift_amounts[index]
+            new_v = shift_fn(v, sh) % 2**256
+            if new_v != 0:
+                return new_v, index
```
jsign, line 561: Nice idea.

tests/zkevm/test_worst_compute.py
jsign: Good question, not that I know.

tests/zkevm/test_worst_compute.py
```diff
@@ -519,3 +521,91 @@ def test_worst_binop_simple(
...
+    code_suffix = Op.POP + Op.JUMP(len(shift_amounts) * 2)
+    code_body_len = MAX_CODE_SIZE - len(code_prefix) - len(code_suffix)
+
+    code_body = Bytecode()
+    v = initial_value
+    while len(code_body) <= code_body_len - 4:
+        v, i = select_shift_amount(shl, v)
+        code_body += make_dup(i) + Op.SHL
+        v, i = select_shift_amount(shift_right_fn, v)
+        code_body += make_dup(i) + shift_right
```
jsign, line 592: So this loop will converge to some value that only has one bit set somewhere, no? From that point forward, the remaining loops before OOG should always have the same args for shifts, no? (since `select_shift_amount` will try to keep that remaining bit)

If that's the case, I think that's fine, just double checking with you.

jsign (approved):
LGTM. Left a question to double-check with you.

Cycles:
```
tests/zkevm/test_worst_compute.py::test_worst_shifts[fork_Cancun-blockchain_test-shift_right_SHR]-1     1760587131
tests/zkevm/test_worst_compute.py::test_worst_shifts[fork_Cancun-blockchain_test-shift_right_SAR]-1     1910438088
```

marioevz (approved):
Just two comments, but looks good to me in general.

tests/zkevm/test_worst_compute.py
chfast: That doesn't work or I don't understand it. I need two things here: get the opcode number of the `DUP2` opcode (e.g. `int(Op.DUP2)` or `Op.DUP2.opcode`) and then to construct another DUP having the `int` index number, e.g. having `index=4` I need to "create" `DUP4` instruction.

tests/zkevm/test_worst_compute.py
chfast: Yes, at some point (~1000 iterations) this converges to the value with single bit set. But the value will not remain constant, the program will move this bit around till the end.

tests/zkevm/test_worst_compute.py
chfast: I added a TODO for this and I will propose something later.

PR #1633, opened 2025-05-20
fix(t8n): include system contracts in alloc for besu generated state tests

danceratopz:
## 🗒️ Description
This change includes the `pre_allocation_blockchain()` for the fork in the pre-alloc for the `state_test` format for the Besu `t8n` tool (`evmtool`). See related issues.

If the system contracts aren't present, then filing fails with an error similar to:
```
    transition_tool_output = t8n.evaluate(
src/ethereum_clis/clis/besu.py:166: in evaluate
    response.raise_for_status()  # exception visible in pytest failure output
.venv/lib/pypy3.10/site-packages/requests/models.py:1024: in raise_for_status
    raise HTTPError(http_error_msg, response=self)
E   requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://localhost:34805/
```

This is a funny one, as the generated `state_test` fixtures won't pass when executed against any other clients, so I'm not sure if we should include it. But perhaps the commit is useful to the Besu Team while trying to fill Osaka tests.

## 🔗 Related Issues
These seem to be the changes that broke `fill`:
- https://github.com/hyperledger/besu/commit/7cd5407451c0797e594eea0260e31d5c8cc45ddd

from:
- https://github.com/hyperledger/besu/pull/8573

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_specs/state.py
+++ b/src/ethereum_test_specs/state.py
@@ -7,7 +7,7 @@ from typing import Any, Callable, ClassVar, Dict, Generator, List, Optional, Seq
 import pytest
 from pydantic import Field
 
-from ethereum_clis import TransitionTool
+from ethereum_clis import BesuTransitionTool, TransitionTool
 from ethereum_test_base_types import HexNumber
 from ethereum_test_exceptions import BlockException, EngineAPIError, TransactionException
 from ethereum_test_execution import (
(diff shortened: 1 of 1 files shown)


marioevz:
I don't think we should merge this.

Without this, the worst case is that besu can only be used to fill blockchain tests, which for the most part is good enough.

If we add this, it means that we generate state tests but they are incompatible with any other state test consumer, even Besu perhaps?

danceratopz:
> I don't think we should merge this.
> 
> Without this, the worst case is that besu can only be used to fill blockchain tests, which for the most part is good enough.
> 
> If we add this, it means that we generate state tests but they are incompatible with any other state test consumer, even Besu perhaps?

Agreed, spoke to Besu today, they will revert the behavior on their side for state test generation.

PR #1638, opened 2025-05-21
fix(tests): EIP-2537: Make pairing test dynamic to env.gas_limit

marioevz:
## 🗒️ Description
Fixes multi-inf tests in EIP-2537 to make them dynamic according to the set evn.gas_limit.

Test was failing with fill parameter `--block-gas-limit=36000000` due to the hard-coded number of pairings.

The test now uses a look to search for the maximum number of pairings that can fit within limit.

## 🔗 Related Issues
None

## ✅ Checklist

- [ ] ~~All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~

### DIFF
--- a/tests/prague/eip2537_bls_12_381_precompiles/conftest.py
+++ b/tests/prague/eip2537_bls_12_381_precompiles/conftest.py
@@ -90,7 +90,7 @@ def call_succeeds(
 @pytest.fixture
 def call_contract_code(
     precompile_address: int,
-    precompile_gas: int,
+    precompile_gas: int | None,
     precompile_gas_modifier: int,
     expected_output: bytes | SupportsBytes,
     call_succeeds: bool,
@@ -126,12 +126,18 @@ def call_contract_code(
     assert call_opcode in [Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL]
     value = [0] if call_opcode in [Op.CALL, Op.CALLCODE] else []
 
+    precompile_gas_value_opcode: int | Op
+    if precompile_gas is None:
+        precompile_gas_value_opcode = Op.GAS
+    else:
+        precompile_gas_value_opcode = precompile_gas + precompile_gas_modifier
+
     code = (
         Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE())
         + Op.SSTORE(
             call_contract_post_storage.store_next(call_succeeds),
             call_opcode(
-                precompile_gas + precompile_gas_modifier,
+                precompile_gas_value_opcode,
                 precompile_address,
                 *value,  # Optional, only used for CALL and CALLCODE.
                 0,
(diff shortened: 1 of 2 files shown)


tests/prague/eip2537_bls_12_381_precompiles/test_bls12_pairing.py
```diff
@@ -133,6 +135,57 @@ def test_valid(
...
+    while True:
+        precompile_gas = pairing_gas(len(input_data + inf_data))
+        new_tx_gas_limit = (
+            extra_gas
+            + intrinsic_gas_cost_calculator(calldata=input_data + inf_data)
+            + memory_expansion_gas_calculator(new_bytes=len(input_data + inf_data))
+            + precompile_gas
+        )
+        if new_tx_gas_limit > environment_gas_limit:
+            break
```
jsign, line 171: :+1:

tests/prague/eip2537_bls_12_381_precompiles/test_bls12_pairing.py
```diff
@@ -235,6 +284,57 @@ def test_invalid(
...
+        tx_gas_limit = new_tx_gas_limit
+        input_data = inf_data + input_data
```
jsign, line 322: Should this be extracted so can be reused also in the prev test to avoid repetition? Or maybe it is fine, you're the boss :smile:

jsign (approved):

tests/prague/eip2537_bls_12_381_precompiles/conftest.py
```diff
@@ -126,12 +126,18 @@ def call_contract_code(
     assert call_opcode in [Op.CALL, Op.CALLCODE, Op.DELEGATECALL, Op.STATICCALL]
     value = [0] if call_opcode in [Op.CALL, Op.CALLCODE] else []
 
+    precompile_gas_value_opcode: int | Op
+    if precompile_gas is None:
+        precompile_gas_value_opcode = Op.GAS
+    else:
+        precompile_gas_value_opcode = precompile_gas + precompile_gas_modifier
```
jsign, line 133: Was this required for the fix we needed? Or an coherency improvement?

marioevz: The problem was that the pytest fixtures try to automatically calculate the precompile_gas needed for every precompile gas.

But in this case it was done in reverse so we needed the call contract without knowing the gas, so this is a fix to allow creation of the contract without knowing the gas by simply passing the full available gas to the precompile in the call.

tests/prague/eip2537_bls_12_381_precompiles/test_bls12_pairing.py
marioevz: Yeah, it was simply done in a hurry 😛 

Should be ok.

Issue #1565, opened 2025-05-06
tracker(packaging): make eest a publishable package


danceratopz:
Tracking issue for packaging related issues leading up to moving `./tests` to ethereum/execution-specs:
- [x] ethereum/execution-spec-tests#1560 
- [x] ~~Package the resolver as a wheel? We won't need the resolver going forward, but this is required in order to distribute eest as a wheel.~~
- [x] ethereum/execution-spec-tests#1561  
- [ ] ethereum/execution-specs#1451 
- [ ] (not required to move `./tests/`!) ethereum/execution-specs#1452

richardgreg:
Sure, I'm interested :)

danceratopz:
As all of EEST moves to execution-specs, this is no longer required; all packaging changes will happen in EELS.

Issue #1560, opened 2025-05-06
chore(packaging): publish `hive.py` as a package


danceratopz:
We can't publish EEST as a package with Github spec'd deps; they must also be published packages.

I'd suggesting renaming the package to `ethereum-hive`.

PR #1645, opened 2025-05-23
docs: some ideas on moving `./tests` to ethereum/execution-specs

danceratopz:
## 🗒️ Description

Just some thoughts about what we envision "The Merge" (or depending on your POV, the "The Move" :laughing:)  should look like and what it requires. 

Not complete by any means, just thought it could help serve as a basis for discussion! 

Feel free to add comments to this PR or in Discord!

### Background

For those unfamiliar with the discussion: The plan is to move the EL client test cases in `./tests/` to ethereum/execution-specs.

Here's the main motivation for this change (these are overlapping): 

1. Spec and tests versioned together. Less ambiguous.
2. EIP authors/implementers:
    i. Need only concern themselves with one repo.
    ii. Can directly & very easily add basic tests to sanity check their implementations.
3. This removes the chicken & egg problem when testing the spec: To test an EELS branch you need an EEST release which needs an EELS branch :recycle:. 
4. Simplifies tooling and debugging: Removes the need for [petertdavies/ethereum-spec-evm-resolver](https://github.com/petertdavies/ethereum-spec-evm-resolver), which currently allows us to fill multiple EELS implementations from multiple repos/branches within one test session.

## 🔗 Related Issues

Current sub-tasks:
- #1560 
- #1565 

**There's no plan to merge this PR - but we can update the doc with comments to gain a shared understanding!**

### DIFF
--- a/docs/internal/moving_tests_to_execution_specs.md
+++ b/docs/internal/moving_tests_to_execution_specs.md
@@ -1,0 +1,129 @@
+# Moving `./tests/` to `execution-specs`
+
+## Requirements
+
+1. Test implementers can continue working on tests from Day 1.
+   i. Familiar CI checks enabled in Github Actions.
+   ii. TBD: Test implementers can continue using familiar tooling: `uv`, `ruff`, `tox`.
+2. The execution-spec-tests git history of `./tests` is maintained in ethereum-specs.
+3. Consistent pytest versions across repos.
+4. Consistent minimum Python versions across repos.
+5. The "[Test Case Reference](https://eest.ethereum.org/main/tests/)" documentation is published on ethereum-specs.
+
+## Main Steps
+
+### Phase I
+
+For the initial move.
+
+1. EEST: Make EEST an installable package, tracking issue: [ethereum/execution-spec-tests#1565](https://github.com/ethereum/execution-spec-tests/issues/1565).
+2. EEST: Remove last dependencies on execution-specs to remove any circular dependencies:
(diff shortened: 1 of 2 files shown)


winsvega:
So the execution-spec-tests will be used as a package in execution-specs ?

danceratopz:
> So the execution-spec-tests will be used as a package in execution-specs ?

Yes.

spencer-tb (commented):
Thanks for making a start on this! Some of my thoughts on the "merge" below.

My understanding of the overarching goals across the repos:
- Make EELS the defacto single source of truth for the EL spec, tests and fixtures!
- Use EEST as the testing framework for filling the fixtures, and consuming them, long term its maybe preferable to split all the pytest plugins into seperate repos.
- Either a) maintain `ethereum/tests` as an additional source of tests fixtures, updating it every hardfork, adding the json fixture files too. Or b) archiving once appropriate.

Once the tests have moved from EEST to EELS, as far as I know we can execute the tests against the spec without requiring filled fixtures, but additionally fill from EELS using EESTs fill as a dependency for when we want to release tests to clients.
- The primary issue I can see here is having to make changes to the EEST framework types/packages for specific EIPs. This would be annoying and defeats the purpose of an easy self contained repo. The only solution I can think of is to move all of these EEST packages to EELS but then I think we can **only** fill from the EELS repo. To me this is a tradeoff worth having.

Another tricky part is aligning on releases. Are we okay adding pre-releases in EELS as we do in EEST, to me this a must have. This might end up looking more clean with the EELS branch structure compared to EESTs current approach using main.

Should we still refill the static tests from EELS? My assumption is yes but I don't think we would be able to execute them directely against EELS without filling them into fixtures first. Until we can do the latter to me it doesn't make sense to move them into EELS.

We should make sure we can still fill from every transition tool from within eels.

Issue #1509, opened 2025-04-29
Osaka CFI'd Tracker Issue


marioevz:
- [ ] [EIP-5920](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5920.md): PAY opcode
- [ ] RIP-7212: Precompile for secp256r1 Curve Support
- [ ] [EIP-7762](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7762.md): Increase MIN_BASE_FEE_PER_BLOB_GAS
- [ ] [EIP-7823](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7823.md): Set upper bounds for MODEXP
- [ ] [EIP-7825](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7825.md): Transaction Gas Limit Cap
- [ ] [EIP-7883](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7883.md): ModExp Gas Cost Increase
- [ ] [EIP-7907](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7907.md): Meter Contract Code Size And Increase Limit
- [ ] [EIP-7917](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7917.md): Deterministic proposer lookahead
- [ ] [EIP-7918](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7918.md): Blob base fee bounded by execution cost

PR #1649, opened 2025-05-25
feat(tests): Update benchmark test to query newly introduced `max_code_size()` from fork config

jochem-brouwer:
## 🗒️ Description
This PR expands the benchmark test which queries as many other accounts (`EXTCODE*` and `CALL*` operations) by also providing a max code size setting in the fork.

- Adds `max_code_size()` to forks
- Adds `max_initcode_size()` to forks
- Edits benchmark test to calculate a correct gas limit w.r.t. deploying the to-be-attacked contracts vs. the actual contract

Note that the deposit costs for the EIP-7907 are much higher (relative) than the original costs. This includes the 200 gas/byte code deposit cost, but also the quadratic memory expansion cost.

The cost of querying one such contract (the "loop cost" of the benchmark) is calculated as 2687 gas/contract. The deposit however costs at least 52M gas due to the 200 gas/byte deposit cost of the 0x40000 bytes. So the cost to deposit such contract vs. querying it is about `20_000` times higher :eyes: 

## 🔗 Related Issues
- https://github.com/ethereum/execution-spec-tests/pull/1646 can use the `max_initcode_size()` as well
- ~~Expands the benchmark to also include the bumped sizes in https://eips.ethereum.org/EIPS/eip-7907~~
- The benchmark will read the max initcode/code size from fork config, so to add bumped limits like EIP-7907 the `max_initcode_size` or `max_code_size` for Osaka has to be manually edited to return that value
- Part of Osaka tracker https://github.com/ethereum/execution-spec-tests/issues/1509

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_forks/base_fork.py
+++ b/src/ethereum_test_forks/base_fork.py
@@ -473,6 +473,18 @@ class BaseFork(ABC, metaclass=BaseForkMeta):
 
     @classmethod
     @abstractmethod
+    def max_code_size(cls) -> int | None:
+        """Return the maximum code size to be deployed in a CREATE-frame."""
+        pass
+
+    @classmethod
+    @abstractmethod
+    def max_initcode_size(cls) -> int | None:
+        """Return the maximum initcode size when using a CREATE frame."""
+        pass
+
+    @classmethod
+    @abstractmethod
     def call_opcodes(
         cls, block_number: int = 0, timestamp: int = 0
     ) -> List[Tuple[Opcodes, EVMCodeType]]:
--- a/src/ethereum_test_forks/base_fork.py
+++ b/src/ethereum_test_forks/base_fork.py
@@ -474,7 +474,7 @@ class BaseFork(ABC, metaclass=BaseForkMeta):
     @classmethod
     @abstractmethod
     def max_code_size(cls) -> int | None:
-        """Return the maximum code size to be deployed in a CREATE-frame."""
+        """Return the maximum code size allowed to be deployed in a contract creation."""
         pass
 
     @classmethod
--- a/src/ethereum_test_forks/base_fork.py
+++ b/src/ethereum_test_forks/base_fork.py
@@ -480,7 +480,7 @@ class BaseFork(ABC, metaclass=BaseForkMeta):
     @classmethod
     @abstractmethod
     def max_initcode_size(cls) -> int | None:
-        """Return the maximum initcode size when using a CREATE frame."""
+        """Return the maximum initcode size allowed to be used in a contract creation."""
         pass
 
     @classmethod
--- a/src/ethereum_test_forks/base_fork.py
+++ b/src/ethereum_test_forks/base_fork.py
@@ -473,13 +473,13 @@ class BaseFork(ABC, metaclass=BaseForkMeta):
 
     @classmethod
     @abstractmethod
-    def max_code_size(cls) -> int | None:
+    def max_code_size(cls) -> int:
         """Return the maximum code size allowed to be deployed in a contract creation."""
         pass
 
     @classmethod
     @abstractmethod
-    def max_initcode_size(cls) -> int | None:
+    def max_initcode_size(cls) -> int:
         """Return the maximum initcode size allowed to be used in a contract creation."""
         pass
 
--- a/tests/zkevm/test_worst_bytecode.py
+++ b/tests/zkevm/test_worst_bytecode.py
@@ -27,8 +27,6 @@ from ethereum_test_tools.vm.opcode import Opcodes as Op
 REFERENCE_SPEC_GIT_PATH = "TODO"
 REFERENCE_SPEC_VERSION = "TODO"
 
-MAX_CONTRACT_SIZE = 24 * 1024  # TODO: This could be a fork property
-
 XOR_TABLE_SIZE = 256
 XOR_TABLE = [Hash(i).sha256() for i in range(XOR_TABLE_SIZE)]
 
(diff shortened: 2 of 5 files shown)


tests/zkevm/test_worst_bytecode.py
```diff
@@ -65,10 +63,45 @@ def test_worst_bytecode_single_opcode(
...
+    )
+    # Calculate the number of contracts to be targeted
+    num_contracts = (
+        # Base available gas = GAS_LIMIT - intrinsic - (out of loop MSTOREs)
+        attack_gas_limit - intrinsic_gas_cost_calc() - gas_costs.G_VERY_LOW * 4
+    ) // loop_cost
+
+    # Set the block gas limit to a relative high value to ensure the code deposit tx
+    # fits in the block (there is enough gas available in the block to execute this)
+    # TODO: verify this, I think we can just do `code_deposit_gas_minimum * 2 * num_contracts`?
```
jsign, line 102: > TODO: verify this, I think we can just do `code_deposit_gas_minimum * 2 * num_contracts`?

I think so, but I would suggest to debug this test as is and see that there's a big buffer. This is why I prefered originally to set up simply a 100Ggas... trying to do some math gymnastics to later do `2*` sounds like overcomplicating the setup.

We can even put 1 trillion gas or something and call it a day to cover an attack scenario that is much bigger than today. (Note that if is 1 trillion or tighter to the actual need, it doesn't make filling faster or similar).

jsign:
> @jsign I'm coming to this test from the EIP-7907 discussions (not zk related). I'm wondering how this benchmark should be ran? It thus includes the pre-setup of deploying these large contracts. It is clear that the final block contains the actual data we care about, but I'm not sure if test executors have a direct way to measure this. Should I just hack in a performance timer to measure the time it takes to execute the last block?

AFAIK, today test executors never have a special consideration for benchmark-like tests, so I think if you want to measure wall-clock time I think you prob have to do a hack yes.

src/ethereum_test_forks/base_fork.py
```diff
@@ -462,6 +462,18 @@ def evm_code_types(cls, block_number: int = 0, timestamp: int = 0) -> List[EVMCo
         """Return list of EVM code types supported by the fork."""
         pass
 
+    @classmethod
+    @abstractmethod
+    def max_code_size(cls) -> int | None:
```
marioevz, line 467: Should we also introduce an `account_load_cost_calc` to provide a way to dynamically calculate the costs of loading an account depending on the fork and code size?

marioevz (commented):
LGTM, it just needs a rebase to fix CI and conflicts.

Thanks for implementing this :)

tests/zkevm/test_worst_bytecode.py
jochem-brouwer: Oh right, good point. No idea why I did not take this super high constant but instead lowered it :sweat_smile:

I thought of adding this as assertion, but this is not necessary, since post state has the check that the nonces of the predeployed contracts are 1 (so contract is deployed) :smile: :+1:

jochem-brouwer:
Have also edited the other tests which now depend on the max code size to read it from the fork.

Could, once this is merged, all other PRs which have the `MAX_CODE_SIZE` now also read this from the `fork`?

@jsign also see my comment on the "changed" gas limit here: https://github.com/ethereum/execution-spec-tests/pull/1649#discussion_r2116789864. I think it is more clear to make this a formula, but I am also fine by hardcoding a very high gas limit.

@marioevz Typecheck correctly complains that my `max_code_size()` returns `int | None` and that thus arithmetics like `>` do not work. I'm not sure, because Frontier for instance has no code size limit and also forks before Shanghai do not have an initcode limit. To hardcode a super high constant is a possibility, but it feels incorrect. Thoughts? Feel free to push the update here also. (I'll mark it ready to review here since this seems to be the last thing to fix here)

marioevz:
> I want to get this PR merged, but I need to solve this `int | None` problem: marking Frontier/Chainstart to have a code size limit is incorrect, but marking it as `None` will make the typecheck complain (because it's not a number). Should we handle this in tests to default to some reasonable value? For instance make all code sizes default to the current code size of EIP-170 (24 KiB)

Makes sense to me, default from frontier should be 24 KiB, and every test that uses this parameter should take this into account, although I doubt it will even be a problem.

jochem-brouwer:
I removed the EIP 7907 config from Osaka also, for experimenting this is fine but I don't think we want to activate it by default (?)

jochem-brouwer:
Definitely not! I do not have time to look at it today, but will pick this up tomorrow.

What filler are you using? Standard EELS? How many tests does this pick up?

The gas limit is set as 36M which is reasonable this might therefore also have found a performance degradation in the filler?

However my hunch says it has to do with the tests I raised the gas limit above the setting you provided. Maybe I messed up the calculation and it generates way too many contracts and because of the raised gas limit this is thus possible? Likely has to do with a test with elevated gas limits.

jsign:
> Definitely not! I do not have time to look at it today, but will pick this up tomorrow.
> 
> What filler are you using? Standard EELS? How many tests does this pick up?
> 

I'm using geth as a filler, but in theory shouldn't matter since I also used it for the `main` filling.

> However my hunch says it has to do with the tests I raised the gas limit above the setting you provided. Maybe I messed up the calculation and it generates way too many contracts and because of the raised gas limit this is thus possible? Likely has to do with a test with elevated gas limits.

Yep, I suspect the same. Probably related with the only test that now calculates differently the number of contracts and maybe something odd there!

tests/zkevm/test_worst_bytecode.py
```diff
@@ -65,10 +63,43 @@ def test_worst_bytecode_single_opcode(
...
-    # We use 100G gas limit to be able to deploy a large number of contracts in a single block,
-    # avoiding bloating the number of preparing blocks in the test.
-    env = Environment(gas_limit=100_000_000_000)
-    attack_gas_limit = Environment().gas_limit
+    # The attack gas limit is the gas limit which the target tx will use
+    # The test will scale the block gas limit to setup the contracts accordingly to be
+    # able to pay for the contract deposit. This has to take into account the 200 gas per byte,
+    # but also the quadratic memory expansion costs which have to be paid each time the
+    # memory is being setup
+    attack_gas_limit = 100_000_000
```
marioevz, line 71:
```suggestion
    attack_gas_limit = Environment().gas_limit
```
This is the issue with the slow test, `Environment().gas_limit != env.gas_limit`, since the test is targeted to use the default block-gas-limit only for the attack (last transaction), and as much as possible for the contract deployment.

jsign (approved):
Compared cycles and all looks good :+1:

Issue #1657, opened 2025-05-26
Create Blockchain tests that contain 1-parameter set opcodes that consume the whole block & misc


kevaundray:
## General context

Generally the structure of these tests are split into two:

- Setup phase
- Benchmarking block

**Setup phase**

The setup phase encompasses the pre-state and blocks that are needed in order to setup the block full of a specific block.
An example of this would be a "block full of EXTCODESIZE" where the setup phase would consist of N different contracts being deployed, where N is the number of EXTCODESIZE opcdes we want to put in the final block. 

**Benchmarking block**

For benchmarking, we assume that the final block is the one with the block full of the particular opcode we want to benchmark

# 1-Parameter set opcodes

## Environment

- [x] CALLDATALOAD

## Blobs

- [x] BLOBHASH

## Stack & Memory
- POP will be covered indirectly in https://github.com/ethereum/execution-spec-tests/issues/1651
- [x] MLOAD
- [x] TLOAD
- [x] TSTORE
- [x] MSTORE
- [x] MSTORE8

Last two aren't 1-parameter but are still tracked here.

## Contract Lifecycle

- [x] SELFDESTRUCT

## Note

- This is a subtask of #1453

PR #1658, opened 2025-05-26
zkevm: add CALLDATALOAD benchmark

jsign:
This PR adds a benchmark for `CALLDATALOAD` covering both existinga and absent referenced calldata word.

Cycles:
```
tests/zkevm/test_worst_compute.py::test_worst_calldataload[fork_Cancun-blockchain_test_from_state_test-empty]-1 2015522043
tests/zkevm/test_worst_compute.py::test_worst_calldataload[fork_Cancun-blockchain_test_from_state_test-one-loop]-1      2363234163
tests/zkevm/test_worst_compute.py::test_worst_calldataload[fork_Cancun-blockchain_test_from_state_test-zero-loop]-1     2423217008
```

Targets #1657

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -1069,14 +1069,18 @@ def _generate_bn128_pairs(n: int, seed: int = 0):
 
 
 @pytest.mark.parametrize(
-    "empty_calldata",
-    [False, True],
+    "calldata",
+    [
+        pytest.param(b"", id="empty"),
+        pytest.param(b"\x00", id="zero-loop"),
+        pytest.param(b"\x00" * 31 + b"\x20", id="one-loop"),
+    ],
 )
 def test_worst_calldataload(
     state_test: StateTestFiller,
     pre: Alloc,
     fork: Fork,
-    empty_calldata: bool,
+    calldata: bytes,
 ):
     """Test running a block with as many CALLDATALOAD as possible."""
     env = Environment()
@@ -1091,7 +1095,7 @@ def test_worst_calldataload(
 
     tx = Transaction(
         to=pre.deploy_contract(code=code),
-        data=[] if empty_calldata else [0x00],
+        data=calldata,
         gas_limit=env.gas_limit,
         sender=pre.fund_eoa(),
     )
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -1080,17 +1080,17 @@ def test_worst_calldataload(
     """Test running a block with as many CALLDATALOAD as possible."""
     env = Environment()
 
-    code_prefix = Op.JUMPDEST
-    code_suffix = Op.PUSH0 + Op.JUMP
+    code_prefix = Op.PUSH0 + Op.JUMPDEST
+    code_suffix = Op.PUSH1(1) + Op.JUMP
     code_body_len = MAX_CODE_SIZE - len(code_prefix) - len(code_suffix)
-    code_loop_iter = Op.POP(Op.PUSH0 + Op.CALLDATALOAD)
+    code_loop_iter = Op.CALLDATALOAD
     code_body = code_loop_iter * (code_body_len // len(code_loop_iter))
     code = code_prefix + code_body + code_suffix
     assert len(code) <= MAX_CODE_SIZE
 
     tx = Transaction(
         to=pre.deploy_contract(code=code),
-        data=[] if empty_calldata else [0x42] * 32,
+        data=[] if empty_calldata else [0x00],
         gas_limit=env.gas_limit,
         sender=pre.fund_eoa(),
     )
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -1027,6 +1027,7 @@ def test_amortized_bn128_pairings(
         to=code_address,
         gas_limit=env.gas_limit,
         data=_generate_bn128_pairs(optimal_per_call_num_pairings, 42),
+        sender=pre.fund_eoa(),
     )
 
     state_test(
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -1086,7 +1086,7 @@ def test_worst_calldataload(
     code_loop_iter = Op.POP(Op.PUSH0 + Op.CALLDATALOAD)
     code_body = code_loop_iter * (code_body_len // len(code_loop_iter))
     code = code_prefix + code_body + code_suffix
-    assert len(code) == MAX_CODE_SIZE
+    assert len(code) <= MAX_CODE_SIZE
 
     tx = Transaction(
         to=pre.deploy_contract(code=code),
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -1027,7 +1027,6 @@ def test_amortized_bn128_pairings(
         to=code_address,
         gas_limit=env.gas_limit,
         data=_generate_bn128_pairs(optimal_per_call_num_pairings, 42),
-        sender=pre.fund_eoa(),
     )
 
     state_test(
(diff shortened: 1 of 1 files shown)


tests/zkevm/test_worst_compute.py
```diff
@@ -955,3 +956,40 @@ def test_empty_block(
...
+):
+    """Test running a block with as many CALLDATALOAD as possible."""
+    env = Environment()
+
+    code_prefix = Op.JUMPDEST
+    code_suffix = Op.PUSH0 + Op.JUMP
+    code_body_len = MAX_CODE_SIZE - len(code_prefix) - len(code_suffix)
+    code_loop_iter = Op.POP(Op.PUSH0 + Op.CALLDATALOAD)
+    code_body = code_loop_iter * (code_body_len // len(code_loop_iter))
+    code = code_prefix + code_body + code_suffix
```
jochem-brouwer, line 980:
```suggestion
    code = code_prefix + code_body + Op.JUMPDEST * (code_body_len % len(code_loop_iter)) + code_suffix
```

(just stumbled on this PR, could fill with JUMPDEST to pass the `MAX_CODE_SIZE` assertion)

tests/zkevm/test_worst_compute.py
```diff
@@ -955,3 +956,40 @@ def test_empty_block(
...
+    fork: Fork,
+    empty_calldata: bool,
+):
+    """Test running a block with as many CALLDATALOAD as possible."""
+    env = Environment()
+
+    code_prefix = Op.JUMPDEST
+    code_suffix = Op.PUSH0 + Op.JUMP
+    code_body_len = MAX_CODE_SIZE - len(code_prefix) - len(code_suffix)
+    code_loop_iter = Op.POP(Op.PUSH0 + Op.CALLDATALOAD)
```
jochem-brouwer, line 978: For more CALLDATALOAD, use empty calldata or calldata filled with zeros (arbitray length), then CALLDATALOAD will push 0 to stack in all cases and PUSH0/POP combo is not necessary :smile: :+1:

jochem-brouwer: So:

`PUSH0 JUMPDEST CALLDATALOAD CALLDATALOAD .. .. .. PUSH1 0x01 JUMP` would fill with more `CALLDATALOAD` iff calldata is consists only of zero bytes :)

jochem-brouwer: So:

`PUSH0 JUMPDEST CALLDATALOAD CALLDATALOAD .. .. .. CALLDATALOAD CALLDATALOAD PUSH1 0x01 JUMP` would fill with more `CALLDATALOAD` iff calldata is consists only of zero bytes :)

tests/zkevm/test_worst_compute.py
jsign: Ah, I think I prefer to change the `==` to `<=`, just to avoid extra unrelated gas.

tests/zkevm/test_worst_compute.py
jochem-brouwer: Actually, just realized, if you want calldata you could even make it "loop" over it by using above code (without using explicit `PUSH0 CALLDATALOAD POP`) and use (for instance):
```
0x0000..0020
0x0000..0040
0x0000..0000
```

To make it `CALLDATALOAD(32) CALLDATALOAD(64) CALLDATALOAD(0)`

jochem-brouwer: And for loop balance to avoid stack overflows insert a POP before the JUMP

`PUSH0 JUMPDEST CALLDATALOAD CALLDATALOAD .. .. .. CALLDATALOAD CALLDATALOAD POP PUSH1 0x01 JUMP`

This would work (?) in all cases I think (based on max code size and your calldata input). Would avoid theoretical stack overflows

jsign: That would use some extra gas from tx calldata, no? 
Is that POP before the `PUSH1`, needed?

I switched now to you first suggestion:
```
tests/zkevm/test_worst_compute.py::test_worst_calldataload[fork_Cancun-blockchain_test_from_state_test-empty_calldata_True]-1    2015521819
tests/zkevm/test_worst_compute.py::test_worst_calldataload[fork_Cancun-blockchain_test_from_state_test-empty_calldata_False]-1   2423216794
```

jsign:
Rebased since it had conflicts with another merged PR!

tests/zkevm/test_worst_compute.py
```diff
@@ -1066,3 +1066,39 @@ def _generate_bn128_pairs(n: int, seed: int = 0):
...
+    code_suffix = Op.PUSH1(1) + Op.JUMP
+    code_body_len = MAX_CODE_SIZE - len(code_prefix) - len(code_suffix)
+    code_loop_iter = Op.CALLDATALOAD
+    code_body = code_loop_iter * (code_body_len // len(code_loop_iter))
+    code = code_prefix + code_body + code_suffix
+    assert len(code) <= MAX_CODE_SIZE
+
+    tx = Transaction(
+        to=pre.deploy_contract(code=code),
+        data=[] if empty_calldata else [0x00],
```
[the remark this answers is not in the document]
jsign: The cycles (mentioned in the PR description) show that both cases produce a relevant difference.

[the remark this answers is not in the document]
jochem-brouwer: Ah ok! From an EVM point of view this would thus not make a difference but from the ZK-EVM it does. I was not aware of this.

Would the loop also make a difference? So the calldata would be the 

```
0x00..0020
```

This would loop `CALLDATLOAD(32) CALLDATALOAD(0)`. (For the loop in EVM to correct loop back to the JUMPDEST would thus either mean code starts with JUMPDEST and calldata loop finishes with 0, or there is this explicit `POP PUSH <jumpdest target> JUMP` at the end of the loop). 

I'm thus not aware of how these zk cycles work and what would impact it, but could imagine "looping" over the calldata could provide additional cycles.

Note that the calldata cost is only paid once (upfront) and especially with high limits (even considering tx gas limits of 30M if proposed as in https://eips.ethereum.org/EIPS/eip-7825) this gas cost is negligible compared to the execution cost.

The calldata of `0x00..0020` would cost `31 * 4 + 1 * 16` = 140 gas (or `35 tokens * 4 = 140`). Note that calldata floor cost per https://eips.ethereum.org/EIPS/eip-7623 is not relevant here, since execution cost is much higher than the calldata floor cost (of 35 tokens * 10 = 350 gas compared to pre-EIP-7623 cost of the 140 gas)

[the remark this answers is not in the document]
jochem-brouwer: (assuming a tx gas limit of 30M (EIP 7825) here, so `30_000_000 >> 21_000 (intrinsic tx cost) + 350` -> can thus disregard calldata cost for this analysis (? I think? Unless this calldata cost has some added zk cycles which I am not aware of and would not expect?)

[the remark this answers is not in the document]
jsign: Tomorrow I'll change the PR to receive some byte slice so we can try `[]` and `[0x00]` (as now), and add this other `0x00...20` case and see what happens.

tests/zkevm/test_worst_compute.py
jochem-brouwer: Yes you are right! This would indeed work. I got a bit confused in my head with either stack imbalance (when jumping back to the JUMPDEST adding extra items) or having an invalid JUMP (based on the CALLDATALOAD jump on that). But the pattern you just provided should indeed work for all `calldata` patterns :smile: :+1:

tests/zkevm/test_worst_compute.py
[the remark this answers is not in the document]
jochem-brouwer: Cool! I also realize I must really get myself up-to-date with the zk-world, I have almost no concept of this and no idea what these cycles are (my first guess would be some count of operations to either execute it or generate a proof (?)). If you have some pointers for basic reading on that let me know :smile: :+1: 

I think for now from the EVM side these tests abuse CALLDATALOAD the "best" we can :stuck_out_tongue_winking_eye:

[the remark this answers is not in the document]
jsign: Honestly, I would say it is even better the less you know. How zkEVM behaves on tests depends on which guest program (i.e. stateless node) is used, plus which zkEVM (SP1, Risc0, or many other variants).

So, the fewer assumptions on how things work, I think, for now is the best. This is why we're trying to cover as much ground as possible since it isn't entirely obvious where they might be inefficient.

I think after these "black box" approach, we could maybe dive into details but that probably is more zkVM or guest program specific.

As a first approach to zkVMs in general, you can skim a bit [SP1 docs](https://docs.succinct.xyz/docs/sp1/introduction).

marioevz (approved):
LGTM, thanks!

tests/zkevm/test_worst_compute.py
```diff
@@ -1066,3 +1066,43 @@ def _generate_bn128_pairs(n: int, seed: int = 0):
...
+
+
+@pytest.mark.parametrize(
+    "calldata",
+    [
+        pytest.param(b"", id="empty"),
+        pytest.param(b"\x00", id="zero-loop"),
+        pytest.param(b"\x00" * 31 + b"\x20", id="one-loop"),
+    ],
+)
```
jochem-brouwer, line 1078: Should this test not be marked `valid_from` since it uses PUSH0? (Shanghai)

This will OOG before Shanghai but it will thus not be seen directly from test that it is failing (but yes, zkEVM will report a super low amount of cycles!!)

tests/zkevm/test_worst_compute.py
```diff
@@ -1066,3 +1066,43 @@ def _generate_bn128_pairs(n: int, seed: int = 0):
...
+    calldata: bytes,
+):
+    """Test running a block with as many CALLDATALOAD as possible."""
+    env = Environment()
+
+    code_prefix = Op.PUSH0 + Op.JUMPDEST
+    code_suffix = Op.PUSH1(1) + Op.JUMP
+    code_body_len = MAX_CODE_SIZE - len(code_prefix) - len(code_suffix)
+    code_loop_iter = Op.CALLDATALOAD
+    code_body = code_loop_iter * (code_body_len // len(code_loop_iter))
```
jochem-brouwer, line 1092: I'll raise this on an open PR but I think it would be helpful to have an utility method to help constructing these loops, with a loop_init, loop and loop_suffix (and a potential code offset to account for the JUMPDEST location)

jochem-brouwer (commented):
Can't re-approve but one small comment regarding marking test to be valid `>= Shanghai` and a proposal to add a loop-creation helper in utility methods :smile: :+1:

PR #1661, opened 2025-05-27
fix(github/coverage): Fix coverage script on PRs that do not convert tests

marioevz:
Fixes the coverage script on PRs that convert no tests by adding `--ported-from-output-file` to the fill command.

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -120,10 +120,9 @@ jobs:
           new_sources=$(echo "$CHANGED_TEST_FILES" | tr ',' '\n')
           echo "Changed or new test files: $new_sources"
 
-          # uv run returns an error code if no tests detected in changed files. ignore it by setting || true.
-          files=$(uv run fill $new_sources --show-ported-from --clean --quiet --links-as-filled | grep .json || true)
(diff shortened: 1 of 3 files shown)


winsvega:
Lets see how the error ignore works.
I think no need for more options. Empty discovered files are not an error.

spencer-tb:
> Looks fine from my side! Although I don't understand why we don't see this echo before the failure:
> 
> https://github.com/ethereum/execution-spec-tests/blob/59b705c48e06877dd0a56ee7051b8cf6f25a10d5/.github/workflows/coverage.yaml#L121
> 
> Within the failing CI: https://github.com/ethereum/execution-spec-tests/actions/runs/15274199357/job/42956513753?pr=1660

ahh I understand! the echo is from another PR that just got merged :)

spencer-tb (approved):

marioevz:
> Lets see how the error ignore works. I think no need for more options. Empty discovered files are not an error.

It did work, but I think this improvement wouldn't harm:
- Return code was always non-zero, even when it did find files, and this change fixes it.
- It also silences a warning in mypy in function `pytest_collection_modifyitems` which now has the correct types for the parameters.

Will proceed to merge.

PR #1670, opened 2025-05-28
feat(tests): add EIP-7951 for secp256r1 curve

LouisTsai-Csie:
## 🗒️ Description

This PR implements EIP-7951 precompile to support `secp256r1` curve (based on RIP-7212).

### Running the test locally

Since the current EELS repository does not support this functionality, I submitted an EELS [PR](https://github.com/ethereum/execution-specs/pull/1249) to implement the `P256VERIFY` precompile. This precompile relies on cryptographic operations not currently supported by the default Python packages, so two additional packages are required:

- [`cryptography`](https://pypi.org/project/cryptography/)
- [`pycryptodome`](https://pypi.org/project/pycryptodome/)

Additionally the EELS resolver must have these 2 packages added its `setup.cfg`.

#### To test this PR locally, follow the steps below.

1. Update the EELS resolutions file to the following:

```
 "Osaka": {
        "git_url": "https://github.com/spencer-tb/execution-specs.git",
        "branch": "forks/osaka-devnet-2",
        "commit": "49e601dfba8248dcbb6a5685c59edb5a02a89589"
    }
```

2. Update the EELS resolver to use the following commit within `pyproject.toml`:
```
ethereum-spec-evm-resolver = { git = "https://github.com/spencer-tb/ethereum-spec-evm-resolver", rev = "ee273e7344e24a739ebfbf0ea1f758530c4d032b" }
```

### Source of the test vector
In this PR, we convert the [Wycheproof](https://github.com/C2SP/wycheproof) dataset into secp256r1_test.json and import the test cases for validation.

While this approach is commonly used across various Layer 2 implementations (see analysis [here](https://hackmd.io/MGL2RNe-S6WjUEX3AJ8Ycg#Test-Cases)), adding additional test cases is still necessary to ensure broader coverage.

### Implementation Note
In both EIP-7951 & RIP-7212 we could not depend on the output data length to determine whether the low-level call is successful, this might be different from the behavior of `BLS` precompile (EIP-2537). 

Consider the following two scenarios:
1. A staticcall to the precompile with an invalid signature should return true, with empty return data.
2. If the staticcall is made with a valid signature but insufficient gas, it should return false, and the return data should also be empty.

## 🔗 Related Issues
https://github.com/ethereum/execution-specs/pull/1249

## ✅ Checklist

### DIFF
--- a/tests/frontier/precompiles/test_precompiles.py
+++ b/tests/frontier/precompiles/test_precompiles.py
@@ -16,7 +16,7 @@ from ethereum_test_tools import (
 from ethereum_test_tools.vm.opcode import Opcodes as Op
 
 
-def precompile_addresses(fork: Fork) -> Iterator[Tuple[str, bool]]:
+def precompile_addresses(fork: Fork) -> Iterator[Tuple[Address, bool]]:
     """
     Yield the addresses of precompiled contracts and their support status for a given fork.
 


spencer-tb (commented):
Really nice! Just wondering where you got the vectors from? Are they the same as these and if so do we know how they are generated: https://github.com/google/wycheproof/blob/master/testvectors/ecdsa_secp256r1_sha256_test.json

I think in a future PR we should add some additional edge cases manually. As for BLS (EIP-2537) we initially relied heavily on the external test vectors and missed some important coverage points.

Although the curve itself is extremely well tested the point of these manual cases would be to find divergence between the libraries that clients use or how they integrate the library (or simply a client implementation). Some cases I can think of just now:
- Zero signature component, `s=0` should trigger divide by zero on libraries, `r=0` too.
- All possible point at infinity input cases, `(x, y) = (0, 0)` is the obvious case.
- Some/all point at infinity output cases, I think when the hash is 0.
- Boundary conditions, that aim to trigger overflows. Cases where `x, y >= p` or `r, s >= n`, or where they are close to `p-1`, `n-1`. The RIP follows NIST so `1 <= s <= n-1`, but all ethereum signatures thus far are `s <= n/2`, adding a case for this would be nice.
- Input length or encoding cases, i.e < 160 bytes, > 160 bytes, 31 byte components etc.

tests/osaka/eip7212_secp256r1_precompiles/test_secp256r1_precompile_module.py
```diff
@@ -0,0 +1,116 @@
...
+            Spec.INVALID_RETURN_VALUE,
+            -1,
+            False,
+            id="insufficient_gas",
+        ),
+    ],
+)
+@pytest.mark.parametrize("precompile_address", [Spec.P256VERIFY], ids=[""])
+@pytest.mark.eip_checklist("new_precompile/test/value_transfer/fee/over")
+@pytest.mark.eip_checklist("new_precompile/test/value_transfer/fee/under")
```
marioevz, line 59: Should be:
```suggestion
@pytest.mark.eip_checklist("new_precompile/test/gas_usage/constant/exact")
@pytest.mark.eip_checklist("new_precompile/test/gas_usage/constant/oog")
```

tests/osaka/eip7212_secp256r1_precompiles/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+@pytest.fixture
+def call_contract_address(pre: Alloc, call_contract_code: Bytecode) -> Address:
+    """Address where the test contract will be deployed."""
+    return pre.deploy_contract(call_contract_code)
+
+
+@pytest.fixture
+def sender(pre: Alloc) -> EOA:
+    """Sender of the transaction."""
+    return pre.fund_eoa(1_000_000_000_000_000_000)
```
marioevz, line 124:
```suggestion
    return pre.fund_eoa()
```
The default amount should be enough, otherwise we can take a look if any of the tests is using too much gas.

marioevz (commented):
Awesome work! I've left a couple of comments to be resolved before proceeding 👍

tests/osaka/eip7212_secp256r1_precompiles/test_secp256r1_precompile_module.py
marioevz: I think we should add an explicit `exact` case here anyway just for the sake of completeness and readability, even though I understand it's implicitly checked in many other tests.

The `over` case we can leave the case but unmarked because it's not part of the checklist (we could add it to the checklist later).

tests/osaka/eip7212_secp256r1_precompiles/test_secp256r1_precompile_module.py
```diff
@@ -0,0 +1,116 @@
...
+]
+
+
+@pytest.mark.parametrize(
+    "input_data,expected_output,vector_gas_value",
+    vectors_from_file("secp256r1_test.json"),
+)
+@pytest.mark.parametrize("precompile_address", [Spec.P256VERIFY], ids=[""])
+@pytest.mark.eip_checklist("new_precompile/test/call_contexts/normal")
+def test_valid(state_test: StateTestFiller, pre: Alloc, post: dict, tx: Transaction):
```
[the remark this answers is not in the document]
LouisTsai-Csie: I've updated the invalid test case, please let me know if there is anything more that needs to be added!

tests/osaka/eip7212_secp256r1_precompiles/test_secp256r1_precompile_module.py
```diff
@@ -0,0 +1,28 @@
...
+
+pytestmark = [
+    pytest.mark.valid_from("Osaka"),
+    pytest.mark.parametrize("precompile_address", [Spec.P256VERIFY], ids=[""]),
+]
+
+
+@pytest.mark.parametrize(
+    "input_data,expected_output,vector_gas_value",
+    vectors_from_file("secp256r1_test.json"),
```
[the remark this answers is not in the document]
spencer-tb: Could we add the link here: https://github.com/C2SP/wycheproof/blob/main/testvectors/ecdsa_secp256r1_sha256_test.json

Stating something like "vectors generated from project wycheproof" etc

spencer-tb:
Amazing! Looks good from myside! 

Could we get these 2 comments in:
- https://github.com/ethereum/execution-spec-tests/pull/1670#discussion_r2135478739
- https://github.com/ethereum/execution-spec-tests/pull/1670#discussion_r2135636284

And maybe update the PR description/name to be 7951, stating that its based on 7212?

spencer-tb:
Rebasing onto main will fix the evmone coverage report fail in ci :)

PR #1671, opened 2025-05-28
zkevm: add temp memory opcodes

jsign:
This PR adds tests for `TLOAD` and `TSTORE` under different scenarios.

Cycles:
```
tests/zkevm/test_worst_compute.py::test_worst_tload[fork_Cancun-blockchain_test_from_state_test-val_mut_False-key_mut_True]-1   102655021
tests/zkevm/test_worst_compute.py::test_worst_tload[fork_Cancun-blockchain_test_from_state_test-val_mut_False-key_mut_False]-1  105422512
tests/zkevm/test_worst_compute.py::test_worst_tload[fork_Cancun-blockchain_test_from_state_test-val_mut_True-key_mut_True]-1    372113852
tests/zkevm/test_worst_compute.py::test_worst_tload[fork_Cancun-blockchain_test_from_state_test-val_mut_True-key_mut_False]-1   376385880
tests/zkevm/test_worst_compute.py::test_worst_tstore[fork_Cancun-blockchain_test_from_state_test-dense_val_mut_False-key_mut_False]-1   521616177
tests/zkevm/test_worst_compute.py::test_worst_tstore[fork_Cancun-blockchain_test_from_state_test-dense_val_mut_False-key_mut_True]-1    524215317
tests/zkevm/test_worst_compute.py::test_worst_tstore[fork_Cancun-blockchain_test_from_state_test-dense_val_mut_True-key_mut_False]-1    1023777662
tests/zkevm/test_worst_compute.py::test_worst_tstore[fork_Cancun-blockchain_test_from_state_test-dense_val_mut_True-key_mut_True]-1     1028785489
```

Targets #1657

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -9,7 +9,7 @@ import math
 import operator
 import random
 from enum import Enum, auto
-from typing import Optional, cast
+from typing import cast
 
 import pytest
 from py_ecc.bn128 import G1, G2, multiply


tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,105 @@ def test_worst_unop(
...
+
+    # If `val_mut` is True, we mutate the value that will be returned by the nex iteration of
+    # TLOADs.
+    code_val_mut = (
+        Op.DUP2  # MUL current value
+        + Op.PUSH32(2**256 - 5)  # MUL constant factor
+        + Op.MUL  # Calculate new value
+        + Op.SWAP2  # Save it
+        + Op.POP  # Cleanup
+        + Op.TSTORE(Op.DUP2, Op.DUP2)
```
jsign: I prefer to do linear `+` instead of `opcode(..., ...)` in cases where `Op.DUPX` is used since feels more clear what's happening in the stack.

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,105 @@ def test_worst_unop(
...
+    """Test running a block with as many TLOAD calls as possible."""
+    env = Environment()
+
+    start_key = 41
+    start_val = 43
+    code_prefix = Op.PUSH1(start_val) + Op.PUSH1(start_key) + Op.JUMPDEST
+
+    # If `key_mut` is True, we mutate the key on every iteration of the big loop by mutiplying by
+    # a constant that should produce a long chain of distinct values.
+    code_key_mut = Op.PUSH1(3) + Op.MUL if key_mut else Bytecode()
```
jsign, line 762: To mutate keys or values, I usually multiply by `3`. This is to have a long chain of different results mod 2**256. [Proof](https://math.stackexchange.com/a/2460162).

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,105 @@ def test_worst_unop(
...
+
+    # If `val_mut` is True, we mutate the value that will be returned by the nex iteration of
+    # TLOADs.
+    code_val_mut = (
+        Op.DUP2  # MUL current value
+        + Op.PUSH1(3)  # MUL constant factor
+        + Op.MUL  # Calculate new value
+        + Op.SWAP2  # Save it
+        + Op.POP  # Cleanup
+        + Op.TSTORE(Op.DUP2, Op.DUP2)
```
jsign, line 772: I prefer to do linear + instead of opcode(..., ...) in cases where Op.DUPX is used since feels more clear what's happening in the stack.

Also have the feeling this could be simplified. Open to suggestions!

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,105 @@ def test_worst_unop(
...
+):
+    """Test running a block with as many TSTORE calls as possible."""
+    env = Environment()
+
+    init_key = 42
+    code_prefix = Op.PUSH1(init_key) + Op.JUMPDEST
+
+    # If `key_mut` is True, we mutate the key on every iteration of the big loop by mutiplying by
+    # a constant that should produce a long chain of distinct values.
+    code_key_mut = Op.PUSH1(3) + Op.MUL if key_mut else Bytecode()
```
jochem-brouwer, line 818: Even cheaper than using `PUSH MUL` (3 + 5 = 8 gas) would be to use the `GAS` opcode (2 gas) which strictly decreases during the call frame :smile: :+1: The downside is that this only produces "low" numbers, while the `MUL` variant will eventually write to "high" numbers in the transient storage.

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,105 @@ def test_worst_unop(
...
+    code_suffix = code_key_mut + Op.JUMP(len(code_prefix) - 1)
+
+    # If `dense_val_mut` is set, we use PC as a cheap way of always storing a different value than
+    # the previous one.
+    loop_iter = Op.TSTORE(Op.DUP2, Op.PC if dense_val_mut else Op.DUP1)
+
+    code_body_len = (MAX_CODE_SIZE - len(code_prefix) - len(code_suffix)) // len(loop_iter)
+    code_body = loop_iter * code_body_len
+    code = code_prefix + code_body + code_suffix
+
```
jochem-brouwer, line 828: Suggestion to make the assertion here that `code` is smaller than or equal to `MAX_CODE_SIZE`

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,105 @@ def test_worst_unop(
...
+    tx = Transaction(
+        to=pre.deploy_contract(code),
+        gas_limit=env.gas_limit,
+        sender=pre.fund_eoa(),
+    )
```
jochem-brouwer, line 833: We know that the tx OOGs due to the infinite loop, but it would also be nice to somehow detect that the test is being ran correctly. The condition would be that the loop gets ran at least 3 times (so 3 times running the JUMPDEST opcode at the start).

This verifies:

- The loop has run once and thus the code for the loop is correct
- Running the loop a second time succesfully verifies that the stack is correctly setup at the re-entering of the loop (no stack underflows)

Note, if stack pushes too much items before re-entering the loop (JUMPDEST) then at some point we will stack overflow thus then the loop limit is bound by the stack size, not by the gas limit (which we want)

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,105 @@ def test_worst_unop(
...
+        + Op.TSTORE(Op.DUP2, Op.DUP2)
+        if val_mut
+        else Bytecode()
+    )
+
+    # Note that both potential `code_key_mut` and `code_value_mut` are done at the end of
+    # big-loops to avoid creating too much influence in the benchmark results for measuring TLOADs.
+    code_suffix = code_key_mut + code_val_mut + Op.JUMP(len(code_prefix) - 1)
+
+    loop_iter = Op.POP(Op.TLOAD(Op.DUP1))
```
jochem-brouwer, line 781: Note, for more TLOAD (but will always push 0 to stack due to non-initialized storage) use `GAS` (2 gas) instead of `DUP1` (3 gas). This will also load different transient storage on each call.

jochem-brouwer (commented):
Stumbled upon this one, some comments on how to possibly squeeze even more operations out of it! :smile: :+1:

tests/zkevm/test_worst_compute.py
jsign: I agree and is a known fact about all zkevm tests in general, not only this one. The testing-framework can only detect post-state invariants and we usually try to squeeze gas to avoid storage writes.

What I thought was creating some CI tool that allows to catch this via low cycle counts in tests, to also detect if some refactor or similar borke it.

Do you have any alternative idea?

tests/zkevm/test_worst_compute.py
jsign: I think I'm fine with lower numbers, so I'll switch to GAS and see if makes as diff!

tests/zkevm/test_worst_compute.py
jochem-brouwer: `SWAP1 PUSH1(3) MUL SWAP1`?

This moves stack (leftmost top stack item) `[key, value]` to `[value, key]`, performs MUL on the `value`, and then fixes the structure again by `SWAP1` to ensure stack is `[key, value]` again. (Also saves gas :smile: :+1: )

tests/zkevm/test_worst_compute.py
jsign: Pushed for TLOAD. Didn't change cycles much but I still prefer it since is shorter :)
```
tests/zkevm/test_worst_compute.py::test_worst_tload[fork_Cancun-blockchain_test_from_state_test-val_mut_False-key_mut_False]-1  100314705
tests/zkevm/test_worst_compute.py::test_worst_tload[fork_Cancun-blockchain_test_from_state_test-val_mut_False-key_mut_True]-1   100318477
tests/zkevm/test_worst_compute.py::test_worst_tload[fork_Cancun-blockchain_test_from_state_test-val_mut_True-key_mut_False]-1   368631899
tests/zkevm/test_worst_compute.py::test_worst_tload[fork_Cancun-blockchain_test_from_state_test-val_mut_True-key_mut_True]-1    372113890
tests/zkevm/test_worst_compute.py::test_worst_tstore[fork_Cancun-blockchain_test_from_state_test-dense_val_mut_False-key_mut_False]-1   521616009
tests/zkevm/test_worst_compute.py::test_worst_tstore[fork_Cancun-blockchain_test_from_state_test-dense_val_mut_False-key_mut_True]-1    526989619
tests/zkevm/test_worst_compute.py::test_worst_tstore[fork_Cancun-blockchain_test_from_state_test-dense_val_mut_True-key_mut_False]-1    1026842295
tests/zkevm/test_worst_compute.py::test_worst_tstore[fork_Cancun-blockchain_test_from_state_test-dense_val_mut_True-key_mut_True]-1     1034676775
```

tests/zkevm/test_worst_compute.py
jsign: I think I would have a problem to make the value change work, right? Since I can't easily reproduce GAS on the TSTORE.

tests/zkevm/test_worst_compute.py
jochem-brouwer: One idea, but this will add code to the EVM which will thus limit the amount of opcodes we want to squeeze out of it, is to edit the loop to not-be an infinite loop but instead `JUMPI` based on if `GAS` is higher than some constant. Also add an item to the loop which increments each time on the loop. This code would then `RETURN` that value. The test then itself first calls into a helper contract, which `CALLs` into this actual loop contract (note, this obviously already limits the test a lot because it will chop 1/64 gas of the available gas for the loop). This would then in the end SSTORE the loop count into a storage slot (which we can then verify that it is > 2).

The alternative would be a helper contract which we call into a different tx in a different block and instead use the target block to fill it with the desired loop to figure out these zk cycles. (This thus means the entire test consists of two blocks which thus bloats the cycles if we directly read the cycles after executing these blocks. Plus, I am not sure, but the entire block handling might also add cycles where we want to isolate the EVM cycles here?). This would also only work if the zkEVM supports block tests, or state tests with multiple txs (I think this is not supported in general).

Another alternative would indeed be to test the code inside `execution-spec-tests` for instance running it in EELS (with tracing on (is this supported in EELS?)) and use this trace to verify the test runs correctly.

jsign: Ah, that would only cover broken tests in structure. 
If it is broken due to EVM errors, it will OOG anyway.

jsign: Your first paragraph solution is the most obvious one, but I feel it will be very complicated to do exact measurements of number of loops compared to today's strategy.

> Another alternative would indeed be to test the code inside execution-spec-tests for instance running it in EELS (with tracing on (is this supported in EELS?)) and use this trace to verify the test runs correctly.

That sounds interesting.

tests/zkevm/test_worst_compute.py
jochem-brouwer: For a stable value you could use `CALLVALUE` (2 gas) and store the to-be-pushed var in the callvalue?

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,101 @@ def test_worst_unop(
     )
 
 
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.parametrize("key_mut", [True, False])
+@pytest.mark.parametrize("val_mut", [True, False])
```
jochem-brouwer, line 746: I think it would be helpful to add comments here, since it is not directly clear why `TSTORE` is used in this `TLOAD`  "stress test" when `val_mut = True`. `key_mut` would thus modify the key at the end of each loop (such that the new loop queries a new key) and `val_mut` modifies the value in the transient storage (at least, this is how I understand it from reading the code -> explicitly that this is thus only done at the end of each loop, not at each loop "item")

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,101 @@ def test_worst_unop(
...
+def test_worst_tload(
+    state_test: StateTestFiller,
+    pre: Alloc,
+    key_mut: bool,
+    val_mut: bool,
+):
+    """Test running a block with as many TLOAD calls as possible."""
```
jochem-brouwer, line 753: As many TLOADs is only when key/value mutations are false, I think here a mention to figure out the zk cycles necessary for each scenario where each of these scenarios is also a stress test (mutating keys/values or not) would be fine (although the zk cycles stuff is likely implicit because this is in the zkevm dir?)

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,101 @@ def test_worst_unop(
...
+    env = Environment()
+
+    start_key = 41
+    code_key_mut = Bytecode()
+    code_val_mut = Bytecode()
+    if key_mut and val_mut:
+        code_prefix = Op.PUSH1(start_key) + Op.JUMPDEST
+        code_key_mut = Op.POP + Op.GAS
+        code_val_mut = Op.TSTORE(Op.DUP2, Op.GAS)
+        loop_iter = Op.POP(Op.TLOAD(Op.DUP1))
```
jochem-brouwer, line 763: Ah ok wait this is somewhat confusing, so in this case the key is only edited after each loop, while at the `key_mut = True` `val_mut = False` the key is actually edited each loop "item"

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,101 @@ def test_worst_unop(
...
+    if key_mut and not val_mut:
+        code_prefix = Op.JUMPDEST
+        loop_iter = Op.POP(Op.TLOAD(Op.GAS))
+    if not key_mut and val_mut:
+        code_prefix = Op.JUMPDEST
+        code_val_mut = Op.TSTORE(Op.CALLVALUE, Op.GAS)  # CALLVALUE configured in the tx
+        loop_iter = Op.POP(Op.TLOAD(Op.CALLVALUE))
+    if not key_mut and not val_mut:
+        code_prefix = Op.JUMPDEST
+        loop_iter = Op.POP(Op.TLOAD(Op.CALLVALUE))
```
jochem-brouwer, line 773: For the case of `start_key = 0` we could `code_prefix = Op.PC + Op.JUMPDEST` and `loop_iter = Op.TLOAD()`. Since `TLOAD` will always push 0 we thus dont need the `CALLVALUE` + `POP` opcodes.

tests/zkevm/test_worst_compute.py
jochem-brouwer: For readability: I would put `loop_iter` directly after `code_prefix` :thinking:

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,101 @@ def test_worst_unop(
...
+    """Test running a block with as many TLOAD calls as possible."""
+    env = Environment()
+
+    start_key = 41
+    code_key_mut = Bytecode()
+    code_val_mut = Bytecode()
+    if key_mut and val_mut:
+        code_prefix = Op.PUSH1(start_key) + Op.JUMPDEST
+        code_key_mut = Op.POP + Op.GAS
+        code_val_mut = Op.TSTORE(Op.DUP2, Op.GAS)
```
jochem-brouwer, line 762: Had to check, but this looks correct: this will store `Op.GAS` at the key which we just mutated at `code_key_mut` (so the key written to is the key we are going to also `TLOAD` in the loop and the value will be a different value than the key)

tests/zkevm/test_worst_compute.py
jochem-brouwer: It seems to me: `value=start_key if not key_mut and val_mut else 0,` that this value is thus 0 for this specific part of the test? It can thus be changed to the suggestion above.

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,101 @@ def test_worst_unop(
...
+    env = Environment()
+
+    init_key = 42
+    code_prefix = Op.PUSH1(init_key) + Op.JUMPDEST
+
+    # If `key_mut` is True, we mutate the key on every iteration of the big loop.
+    code_key_mut = Op.POP + Op.GAS if key_mut else Bytecode()
+    code_suffix = code_key_mut + Op.JUMP(len(code_prefix) - 1)
+
+    # If `dense_val_mut` is set, we use PC as a cheap way of always storing a different value than
```
jochem-brouwer, line 816:
```suggestion
    # If `dense_val_mut` is set, we use GAS as a cheap way of always storing a different value than
```

tests/zkevm/test_worst_compute.py
```diff
@@ -741,6 +741,101 @@ def test_worst_unop(
...
+    # the previous one.
+    loop_iter = Op.TSTORE(Op.DUP2, Op.GAS if dense_val_mut else Op.DUP1)
```
jochem-brouwer, line 818:
```suggestion
    loop_iter = Op.TSTORE(Op.DUP1, Op.GAS if dense_val_mut else Op.DUP1)
```

This should be the same, right? I find it easier to reason about the behavior like this (but `DUP2` also fine :smile: :+1: )

jochem-brouwer (commented):
Some small comments :smile: :+1:

tests/zkevm/test_worst_compute.py
jsign: Yes, your interpretation is correct. I'll add comments to help future readers.

tests/zkevm/test_worst_compute.py
jsign: > Ah ok wait this is somewhat confusing, so in this case the key is only edited after each loop, while at the key_mut = True val_mut = False the key is actually edited each loop "item"

No no, `code_key_mut` and `code_val_mut` are always appended in `code_suffix` -- thus both are only applied once per big-loop. (This applies to all cases)

Per iteration we use Op.DUP1 since is the place for the current key value.

tests/zkevm/test_worst_compute.py
jsign: Not sure I understand your comment. There's no case where `start_key=0`.
This case is about a stable key. This is passed as a `value` in the tx, and here used with `CALLVALUE` (since is a fixed value).

tests/zkevm/test_worst_compute.py
jsign: Nope, `Op.DUP1` isn't equivalent as `Op.DUP2` for this case. 

`Op.DUP1` would be duplicating whatever we decided to insert in the value, but this isn't correct. That would only be true if `key_mut=True` right? If `key_mut=False` using the value would be wrong if `val_mut=True`.

tests/zkevm/test_worst_compute.py
jochem-brouwer: I'm talking about this:

```python
    if key_mut and val_mut:
        code_prefix = Op.PUSH1(start_key) + Op.JUMPDEST
        loop_iter = Op.POP(Op.TLOAD(Op.DUP1))
        code_key_mut = Op.POP + Op.GAS
        code_val_mut = Op.TSTORE(Op.DUP2, Op.GAS)
    if key_mut and not val_mut:
        code_prefix = Op.JUMPDEST
        loop_iter = Op.POP(Op.TLOAD(Op.GAS))
```

In the first code block, `TLOAD` is the same key for each loop "item" right (because it `TLOAD(Op.DUP1)`? In the second code block, it is different for each loop "item", however (because it `TLOAD(Op.GAS)`)?

tests/zkevm/test_worst_compute.py
jochem-brouwer: Yup you are right! I got a bit confused here, but the behavior indeed differs based on `dense_val_mut`. So indeed, keep `DUP2` :smile:

tests/zkevm/test_worst_compute.py
jochem-brouwer: Yup, sounds good!

tests/zkevm/test_worst_compute.py
jsign: Ahh, I see what you mean. You're correct. I think the way it is now is the best since sounds more dynamic, but then I pushed a commit now to fix the comment in the parameters of the test (which I think was related to other comment you did).

jsign:
I rebased again, only changes are using the new `fork.max_code_size()`.

tests/zkevm/test_worst_compute.py
```diff
@@ -1031,6 +1031,107 @@ def test_worst_unop(state_test: StateTestFiller, pre: Alloc, opcode: Op, fork: F
...
+    init_key = 42
+    code_prefix = Op.PUSH1(init_key) + Op.JUMPDEST
+
+    # If `key_mut` is True, we mutate the key on every iteration of the big loop.
+    code_key_mut = Op.POP + Op.GAS if key_mut else Bytecode()
+    code_suffix = code_key_mut + Op.JUMP(len(code_prefix) - 1)
+
+    # If `dense_val_mut` is set, we use GAS as a cheap way of always storing a different value than
+    # the previous one.
+    loop_iter = Op.TSTORE(Op.DUP2, Op.GAS if dense_val_mut else Op.DUP1)
```
jochem-brouwer, line 1114: This always writes to the same key. Would we also want a test which writes to different keys (thus filling up the tstorage?). Basically the TLOAD tests with `key_mut`, `value_mut` but now for TSTORE.

jochem-brouwer (approved):
Will approve, but I also somewhat think we should add the `key_mut`, `value_mut` for the TSTORE tests.

tests/zkevm/test_worst_compute.py
jsign: If `key_mut` it doesn't write to the same key, or maybe I'm misunderstanding your comment?

PR #1677, opened 2025-05-29
 fix(tests,eof): Specify Env(gas_limit)

pdobacz:
## 🗒️ Description

Minor fixup to eof tests done in order to make a final release with the `EOFv1` pseudo-fork. This may make it easier to maintain EOF impl for teams which decide to.

## 🔗 Related Issues

NA

## ✅ Checklist

### DIFF
--- a/.github/actions/build-fixtures/action.yaml
+++ b/.github/actions/build-fixtures/action.yaml
@@ -31,23 +31,28 @@ runs:
       id: evm-builder
       with:
         type: ${{ steps.properties.outputs.evm-type }}
+    - name: Generate fixtures using fill
+      shell: bash
+      run: |
+        uv run fill -n ${{ steps.evm-builder.outputs.x-dist }} --evm-bin=${{ steps.evm-builder.outputs.evm-bin }} --solc-version=${{ steps.properties.outputs.solc }} --skip-evm-dump ${{ steps.properties.outputs.fill-params }} --output=fixtures_${{ inputs.release_name }}.tar.gz --build-name ${{ inputs.release_name }}
     - name: Wrap ethereum/tests fixtures with eofwrap tool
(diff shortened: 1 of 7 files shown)


.github/actions/build-fixtures/action.yaml
```diff
@@ -31,23 +31,28 @@ runs:
...
+        rm -rf fixtures
       env:
         tests_url: https://github.com/ethereum/tests/archive/refs/tags/v
         tests_version: 14.1
         output_path: blockchain_tests/unscheduled/eofwrap
-    - name: Generate fixtures using fill
-      shell: bash
-      run: |
-        uv run fill -n ${{ steps.evm-builder.outputs.x-dist }} --evm-bin=${{ steps.evm-builder.outputs.evm-bin }} --solc-version=${{ steps.properties.outputs.solc }} --skip-evm-dump ${{ steps.properties.outputs.fill-params }} --output=fixtures_${{ inputs.release_name }}.tar.gz --build-name ${{ inputs.release_name }}
     - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02
```
pdobacz, line 56: I needed to move the main `fill` step to before eofwrap runs, because now the output dir must be empty, so we cannot perform the trick to include eofwrap outputs there beforehand. Instead, the `eofwrap` step repackages the tarball including the result of eofwrapping. 

The `fill` step hasn't been modified at all. If the way this is done is raising any concerns, I guess we can also ditch eofwrap as an alternative. It is fairly unlikely it will be used again.

marioevz (approved):
LGTM. If we were keeping eofwrap I would suggest modifying the static test generator as another option, but this should be ok as a one-time solution.

PR #1691, opened 2025-05-31
feat(tooling,ci): add changelog formatting validation

### LINKED ISSUE #1183: feat(tooling,ci): add sanity checks for changelog format
For example, to verify that fulls stops are applied consistently in Changelog items 🙂 
See https://github.com/ethereum/execution-spec-tests/pull/1180#discussion_r1943297623

richardgreg:
## 🗒️ Description

- Add CI check using grep to validate bullet points end with periods
- Fix 31 existing changelog entries missing periods
- 
## 🔗 Related Issues

Fixes #1183

## ✅ Checklist


richardgreg:
Will need to fix the case of it flagging colons

tox.ini
```diff
@@ -37,7 +37,10 @@ commands = python -c "import src.cli.tox_helpers; src.cli.tox_helpers.pyspelling
 [testenv:markdownlint]
 description = Lint markdown files (markdownlint)
 extras = docs
-commands = python -c "import src.cli.tox_helpers; src.cli.tox_helpers.markdownlint()"
+allowlist_externals = bash
+commands =
```
danceratopz, line 41: - Ooh, this is a bit wordy in the terminal. Could you add a function to `src/cli/tox_helpers.py` and use this in tox.ini? 
- Also please improve the error message: colon ":" is also accepted as a line ending.
- Additionally, I don't think the regex will catch nested lists due to `^-`.

danceratopz (commented):
Few comments below, thanks!

danceratopz (approved):
Thanks @richardgreg, especially for adding this to `tox_helpers` with clear output. LGTM!

I ran the checks locally and saw they were failing, but that they didn't fail in CI. This was because the `markdownlint` job uses a dedicated action to handle the `mardownlint-cli` dependency and doesn't map 1-to-1 to the tox env (perhaps you realized this, but I'd forgotten!). 

I decided to resolve this by adding back the `changelog` job to tox_verify (that you had originally added). 

I also added the new check to our precommit config and updated the docs for the new tox check.

Thanks!

Issue #1690, opened 2025-05-30
Create Blockchain tests for the remaining opcodes in #1453 that consume the whole block


kevaundray:
## General context

Generally the structure of these tests are split into two:

- Setup phase
- Benchmarking block

**Setup phase**

The setup phase encompasses the pre-state and blocks that are needed in order to setup the block full of a specific block.
An example of this would be a "block full of EXTCODESIZE" where the setup phase would consist of N different contracts being deployed, where N is the number of EXTCODESIZE opcdes we want to put in the final block. 

**Benchmarking block**

For benchmarking, we assume that the final block is the one with the block full of the particular opcode we want to benchmark

# Opcodes

## Environment 

- [ ]  CALLDATACOPY  
- [ ]  CODECOPY 
- [ ]  RETURNDATACOPY 

## Stack & Memory
 
- [ ]  MCOPY

## Control Flow

- [ ] JUMPI

## Note

- This is a subtask of #1453

jsign:
> Hello [@kevaundray](https://github.com/kevaundray) , I've created this [PR](https://github.com/ethereum/execution-spec-tests/pull/1693) for `JUMPI`, but I do not know how to use `Reth` to calculate the cycle counts, could you please share any resource for instruction? Thanks

Here are some steps:
- Fill the test that you want to measure, so you have the json files.
- Clone https://github.com/eth-applied-research-group/zkevm-benchmark-workload
- In that repo create `zkevm-fixtures/fixtures/blockchain_tests` folder structure and paste the fixtures in that last folder
- Run `cargo run -p ere-hosts --release -- --zkvm sp1`
- The results will be in `zkevm-metrics`.

I will soon be adding a feat to allow passing a flag for the folder path so the third step isn't necessary.

jsign:
@LouisTsai-Csie, want me to take care of `CALLDATACOPY`, `CODECOPY`, `RETURNDATACOPY` and `MCOPY`, or you were planning to also work on those? LMK!

LouisTsai-Csie:
@jsign Could you help with `CALLDATACOPY`, `CODECOPY`, `RETURNDATACOPY` and `MCOPY` opcode, I am working on porting Nethermind tests, especially the `MODEXP` precompile. It would be my main focus this week as some of the tests are complicate

jsign:
[...]

PR #1693, opened 2025-06-02
feat(benchmark): add jumpi-intense test case

LouisTsai-Csie:
## 🗒️ Description

Add a JUMPI-intensive operation for ZKVm as test case.

Currently 

## 🔗 Related Issues

Update issue #1690 

## ✅ Checklist

### DIFF
--- a/tests/benchmark/test_worst_compute.py
+++ b/tests/benchmark/test_worst_compute.py
@@ -1192,7 +1192,7 @@ def test_worst_jumpi_fallthrough(
     )
 
     state_test(
-        genesis_environment=env,
+        env=env,
         pre=pre,
         post={},
         tx=tx,
@@ -1219,7 +1219,7 @@ def test_worst_jumpis(
     )
 
     state_test(
-        genesis_environment=env,
+        env=env,
         pre=pre,
         post={},
         tx=tx,


kevaundray:
CC either @jsign  @jochem-brouwer @chfast

tests/zkevm/test_worst_compute.py
```diff
@@ -515,6 +515,44 @@ def jump_seq():
     )
 
 
+@pytest.mark.zkevm
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.slow
+def test_worst_jumpis(
```
chfast, line 521:
```suggestion
def test_worst_jumpi_fallthough(
```

tests/zkevm/test_worst_compute.py
```diff
@@ -515,6 +515,44 @@ def jump_seq():
...
+    state_test: StateTestFiller,
+    pre: Alloc,
+):
+    """Test running a JUMPI-intensive contract."""
+    env = Environment()
+
+    def jumpi_seq():
+        return Op.JUMPI(Op.ADD(Op.PC, 1), Op.PUSH0)
```
chfast, line 529: I'd avoid the `PC+1` sequence. It's better to just use `PUSH` with a pre-computed code offset. This will increase the `JUMPI` density.

tests/zkevm/test_worst_compute.py
```diff
@@ -515,6 +515,44 @@ def jump_seq():
...
+
+    bytes_per_seq = len(jumpi_seq())
+    seqs_per_call = MAX_CODE_SIZE // bytes_per_seq
+
+    # Create and deploy the jump-intensive contract
+    jumpis_code = sum([jumpi_seq() for _ in range(seqs_per_call)])
+    jumpis_address = pre.deploy_contract(code=bytes(jumpis_code))
+
+    # Call the contract repeatedly until gas runs out.
+    caller_code = While(body=Op.POP(Op.CALL(address=jumpis_address)))
```
chfast, line 539: We should avoid using a `CALL` here. Just fill the `While` body with the `JUMPI`s.

chfast (changes requested):

tests/zkevm/test_worst_compute.py
```diff
@@ -515,6 +515,44 @@ def jump_seq():
...
+    env = Environment()
+
+    def jumpi_seq():
+        return Op.JUMPI(Op.ADD(Op.PC, 1), Op.PUSH0)
+
+    bytes_per_seq = len(jumpi_seq())
+    seqs_per_call = MAX_CODE_SIZE // bytes_per_seq
+
+    # Create and deploy the jump-intensive contract
+    jumpis_code = sum([jumpi_seq() for _ in range(seqs_per_call)])
```
jochem-brouwer, line 535:
```suggestion
    jumpis_code = jumpi_seq() * seqs_per_call
```

I believe this also works :smile: :+1:

jochem-brouwer (commented):
Agree with the comments from @chfast, also have a small style suggestion, plus a suggestion for a test where the `JUMPI` performs the jump (instead of "falling through" and thus not performing the jump).

tests/zkevm/test_worst_compute.py
LouisTsai-Csie: @jochem-brouwer I run the fixture and I think the test case contains incorrect jump destination.

This is the screenshot from evm.codes, as you can see in the first iteration, the `JUMP` operation would go to `0x03` (See the top stack value), but the destination is `ADD` operation, not the `JUMPDEST`. 
This makes the execution stops here, all the remaining opcode would not run as expected.

Screenshot 2025-06-03 at 4 48 06 PM

LouisTsai-Csie: This is the result from the EEST traces.
Section A corresponds to the `CALL` operation within the `While` loop body.
Section B begins the `JUMP` operation tests. As shown at the end of section B, an `InvalidJumpDestError` occurs.

Screenshot 2025-06-03 at 5 07 25 PM

I will create a PR for this ASAP!

jsign: I don't recall of any intentions of jumping to an invalid destination, so probably wasn't intended. cc @codygunton

LouisTsai-Csie: Here is the [PR](https://github.com/ethereum/execution-spec-tests/pull/1710), please help me review the changes, thanks

tests/zkevm/test_worst_compute.py
```diff
@@ -515,6 +515,86 @@ def jump_seq():
...
+@pytest.mark.slow
+def test_worst_jumpis(
+    state_test: StateTestFiller,
+    pre: Alloc,
+):
+    """Test running a JUMPI-intensive contract."""
+    env = Environment()
+
+    def jumpi_seq(dest: int):
+        return Op.JUMPI(Op.PUSH2(dest), Op.DUP1) + Op.JUMPDEST
```
jochem-brouwer, line 567: This should not use PUSH2, because if `dest` is larger than `2^16-1` it will not fit in PUSH2 anymore. If you however just use `dest` (without explicit `Op.PUSH2`) then the framework will automagically pick the correct push value for you. 

However, I believe this is not necessary.

I believe the entire code of this should be:

`Op.JUMPDEST + Op.JUMPI(Op.PUSH0, 1)`

We want to benchmark JUMPI here, and since we are already jumping (in the "happy" case so not the "fallthrough" case) we do not need extra while loops here.

To further optimize, use

`Op.JUMPDEST + Op.JUMPI(Op.PUSH0, Op.NUMBER)`

This changes the `1` (`PUSH1` (note: this tests "one byte", not necessarily "1", could be anything 0-255)) to `Op.NUMBER`. This is the `JUMPI` condition which will jump if the item is nonzero. So for any nonzero block number (genesis is the 0 block and does not contain transactions) this condition will be truthy. `PUSH1` costs 3 gas and `NUMBER` costs 2.

tests/zkevm/test_worst_compute.py
```diff
@@ -515,6 +515,86 @@ def jump_seq():
...
+
+    bytes_per_seq = len(jumpi_seq())
+    seqs_per_call = MAX_CODE_SIZE // bytes_per_seq
+
+    # Create and deploy the jumpi-intensive contract
+    jumpis_code = jumpi_seq() * seqs_per_call
+    jumpis_address = pre.deploy_contract(code=bytes(jumpis_code))
+
+    # Call the contract repeatedly until gas runs out.
+    caller_code = While(body=Op.POP(Op.CALL(address=jumpis_address)))
```
jochem-brouwer, line 539: For this test, the `CALL` should be removed and a while loop should be used for the gas optimization.

tests/zkevm/test_worst_compute.py
codygunton: Not intended, thanks!

LouisTsai-Csie:
@jochem-brouwer Thanks for the review, update accordingly!

marioevz:
#1804 got merged! Please rebase the branch and move the tests to the appropriate folder.

tests/zkevm/test_worst_compute.py
```diff
@@ -792,6 +792,73 @@ def jump_seq():
...
+    jumpis_address = pre.deploy_contract(code=bytes(jumpis_code))
+
+    tx = Transaction(
+        to=jumpis_address,
+        gas_limit=env.gas_limit,
+        sender=pre.fund_eoa(),
+    )
+
+    state_test(
+        genesis_environment=env,
```
marioevz, line 828:
```suggestion
        env=env,
```
Reason being that the environment here is passed to:
- State test environment as-is for state tests.
- Block 1 environment for blockchain tests

If we force the genesis environment we have the risk of forcing this test on a different pre-allocation group when filling the engine-x tests, so I would rather not specify the genesis environment unless absolutely necessary.

tests/zkevm/test_worst_compute.py
```diff
@@ -792,6 +792,73 @@ def jump_seq():
...
+    jumpi_address = pre.deploy_contract(jumpi_code)
+
+    tx = Transaction(
+        to=jumpi_address,
+        gas_limit=env.gas_limit,
+        sender=pre.fund_eoa(),
+    )
+
+    state_test(
+        genesis_environment=env,
```
marioevz, line 855:
```suggestion
        env=env,
```

tests/zkevm/test_worst_compute.py
```diff
@@ -792,6 +792,73 @@ def jump_seq():
...
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.slow
+def test_worst_jumpis(
+    state_test: StateTestFiller,
+    pre: Alloc,
+):
+    """Test running a JUMPI-intensive contract."""
+    env = Environment()
+
+    jumpi_code = Op.JUMPDEST + Op.JUMPI(Op.PUSH0, Op.NUMBER)
```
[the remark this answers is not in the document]
marioevz: I seems correct to me:
- Second argument is the condition, and `Op.NUMBER` is guaranteed to never be zero, so it will always perform the jump.
- It's different from `test_worst_jumpi_fallthrough` because that one never performs the jump.

PR #1710, opened 2025-06-03
fix(tests/zkevm): update jump seq for valid dest

LouisTsai-Csie:
## 🗒️ Description

There is an issue with the original test case: the jump target is not a valid `JUMPDEST` operation.
More details could be found in the [discussion](https://github.com/ethereum/execution-spec-tests/pull/1693#discussion_r2122967972). 

The updated version precomputes valid jump destinations to improve the density of `JUMP` operations.

The repeated pattern is structured as follows:
```
0x00 PUSH2 DEST
0x03 JUMP
0x04 JUMPDEST
...
```
The first `JUMPDEST` is at offset `0x04`. Since each pattern occupies 5 bytes, subsequent `JUMPDEST` locations can be calculated as `4 + 5 * i`, where i is the pattern index. The total code size must remain within the allowed limit.

## ✅ Checklist

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -764,14 +764,14 @@ def test_worst_jumps(state_test: StateTestFiller, pre: Alloc, fork: Fork):
     env = Environment()
     max_code_size = fork.max_code_size()
 
-    def jump_seq():
-        return Op.JUMP(Op.ADD(Op.PC, 1)) + Op.JUMPDEST
+    def jump_seq(dest: int):
+        return Op.JUMP(Op.PUSH2(dest)) + Op.JUMPDEST
 
     bytes_per_seq = len(jump_seq())
     seqs_per_call = max_code_size // bytes_per_seq
 
     # Create and deploy the jump-intensive contract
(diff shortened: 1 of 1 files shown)


tests/zkevm/test_worst_compute.py
```diff
@@ -757,14 +757,14 @@ def test_worst_jumps(
     """Test running a JUMP-intensive contract."""
     env = Environment()
 
-    def jump_seq():
-        return Op.JUMP(Op.ADD(Op.PC, 1)) + Op.JUMPDEST
+    def jump_seq(dest: int):
+        return Op.JUMP(Op.PUSH2(dest)) + Op.JUMPDEST
```
jochem-brouwer, line 761: I think this test should be written as `Op.JUMPDEST + Op.JUMP(0)` (no need for a loop) :smile: :+1:

jochem-brouwer (changes requested):

jochem-brouwer (commented):
This looks good to me, but CI is failing :sweat_smile:

jochem-brouwer (approved):
LGTM. Not sure why CI fails.

PR #1714, opened 2025-06-04
feat(forks): Enforce keyword-only arguments in fork methods

### LINKED ISSUE #1709: Force fork methods to use keyword arguments
Currently, all fork methods use positional arguments for the `block_number` and `timestamp`, e.g. https://github.com/ethereum/execution-spec-tests/blob/351c4a1fd10a4b241e4ea1c7301d91783ebc4553/src/ethereum_test_forks/base_fork.py#L237-L246

But this is a footgun because the tester could pass do this:

```
ShanghaiToCancunAt15k.calldata_gas_calculator(15_000)
```

and expect the function to return the value of Cancun, but it will return Shanghai because we are specifying `block_number=15_000` instead of the expected `timestamp=15_000`.

We should change the abstract definitions and actual definitions to force keyword arguments:
```python
    @classmethod
    @abstractmethod
    def calldata_gas_calculator(
        cls, *, block_number: int = 0, timestamp: int = 0
    ) -> CalldataGasCalculator:
        """
        Return callable that calculates the transaction gas cost for its calldata
        depending on its contents.
        """
        pass
```
Note the `*` before the arguments which denotes keyword arguments must always be used.

akafle01:
## 🗒️ Description

This PR modifies the `calldata_gas_calculator` method in the base and derived fork classes to enforce keyword-only arguments for `block_number` and `timestamp` using `*`.

All method calls across the codebase have been updated to explicitly use keyword arguments.

I wasn't able to run the test suite locally due to a missing plugin error (`pytest_plugins.concurrency`), but the changes were made carefully and consistently. Please let me know if you'd like help resolving test setup.

## 🔗 Related Issues

Fixes #1709

## ✅ Checklist

### DIFF
--- a/src/ethereum_test_fixtures/blockchain.py
+++ b/src/ethereum_test_fixtures/blockchain.py
@@ -103,7 +103,9 @@ class HeaderForkRequirement(str):
 
     def required(self, fork: Fork, block_number: int, timestamp: int) -> bool:
         """Check if the field is required for the given fork."""
-        return getattr(fork, f"header_{self}_required")(block_number, timestamp)
+        return getattr(fork, f"header_{self}_required")(
+            block_number=block_number, timestamp=timestamp
+        )
 
     @classmethod
     def get_from_annotation(cls, field_hints: Any) -> "HeaderForkRequirement | None":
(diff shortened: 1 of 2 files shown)


marioevz (commented):
Thanks for taking this.

My main concern is that this code could break if we convert only one of the fork methods at a time:
https://github.com/ethereum/execution-spec-tests/blob/d6252aafec3553afe978432672971d7eadbeb8fe/src/ethereum_test_forks/transition_base_fork.py#L64-L81

I think we need to transform all the abstract method in `BaseFork` at the same time, which could be a major chore.

marioevz (approved):
LGTM, thanks for this!

Took the liberty to rebase, and make a couple more fixes to get this merged.

PR #1717, opened 2025-06-05
fix(tests): Fix tiny error in test_call_and_callcode_gas_calculation.

evgenyzdanovich:
## 🗒️ Description

Fixes the tiny bug in `test_call_and_callcode_gas_calculation`.
- the test is using the same _hardcoded_ address for `CALL` and contains exact gas calculation.
- as a result, the calculation was incorrect on subsequent executions (because of `empty_account_cost`)

The fix is easy - to generate empty address on every test execution to make sure the `CALL` address is indeed empty. 

## ✅ Checklist

### DIFF
--- a/tests/frontier/opcodes/test_call_and_callcode_gas_calculation.py
+++ b/tests/frontier/opcodes/test_call_and_callcode_gas_calculation.py
@@ -63,7 +63,7 @@ CALLCODE_SUFFICIENT_GAS = CALLCODE_GAS + CALLEE_INIT_STACK_GAS
 
 
 @pytest.fixture
-def callee_code(callee_opcode: Op) -> Bytecode:
+def callee_code(pre: Alloc, callee_opcode: Op) -> Bytecode:
     """
     Code called by the caller contract:
         PUSH1 0x00 * 4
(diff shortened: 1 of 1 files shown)


tests/frontier/opcodes/test_call_and_callcode_gas_calculation.py
```diff
@@ -72,7 +73,10 @@ def callee_code(callee_opcode: Op) -> Bytecode:
         GAS <- value doesn't matter
         CALL/CALLCODE.
     """
-    return callee_opcode(Op.GAS(), 0xFEFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF, 1, 0, 0, 0, 0)
+    # The address needs to be random,
+    # otherwise the calculations (empty_account_cost) are incorrect.
+    address = f"0x{random.getrandbits(160):040x}"
```
marioevz, line 78: We try to avoid using the `random` library in order to guarantee that the fixtures are unchanged from one fill to the next.

The best approach here is to add `pre: Alloc` to the parameters of the fixture and then get an address using `pre.empty_account()`. This account is guaranteed to be empty but it's also calculated deterministically, so it will be the same over different fillings of the same test.

evgenyzdanovich: Thanks, done.

evgenyzdanovich:
gentle ping, please?

evgenyzdanovich:
@marioevz

marioevz (approved):
LGTM, thanks for this, and sorry for the merge delay!

Issue #1734, opened 2025-06-10
tracker(nethermind): Porting Nethermind tests


LouisTsai-Csie:
# Summary

The Nethermind team is currently working on benchmark tests to support raising the gas limit. This issue serves as a tracker for porting tests from [this PR](https://github.com/NethermindEth/nethermind/pull/7266/files).

If the test case already exists in EEST, there’s no need to port it from Nethermind, please provide a link to the relevant EEST code snippet. If the test needs to be added, kindly share the PR link.

# Category
## block

- Warmup
  - [x] Warmup - [Done](https://github.com/ethereum/execution-spec-tests/pull/1742), [Discussion](https://github.com/ethereum/execution-spec-tests/issues/1734#issuecomment-2984506576)
- ETH transfers: [Link](https://github.com/NethermindEth/nethermind/blob/ceb8d57b8530ce8181d7427c115ca593386909d6/tools/EngineRequestsGenerator/TestCases/Transfers.cs)
  - [x] ETH transfers - https://github.com/ethereum/execution-spec-tests/pull/1742
- Transactions: [Link](https://github.com/NethermindEth/nethermind/blob/ceb8d57b8530ce8181d7427c115ca593386909d6/tools/EngineRequestsGenerator/TestCases/TxDataZero.cs)
  - [x] Tx with big zero data - https://github.com/ethereum/execution-spec-tests/pull/1820
  
### Cryptographic
- KECCAK256 (`0x20`): [Link](https://github.com/NethermindEth/nethermind/blob/ceb8d57b8530ce8181d7427c115ca593386909d6/tools/EngineRequestsGenerator/TestCases/Keccak256.cs) - [Done](https://github.com/ethereum/execution-spec-tests/blob/6e272b28e72635e6c0493be10426ecfce94472d5/tests/zkevm/test_worst_compute.py#L339), [Discussion](https://github.com/ethereum/execution-spec-tests/issues/1734#issuecomment-2983646419)
  - [x] Keccak256 from 1 byte
  - [x] Keccak256 from 8 bytes
  - [x] Keccak256 from 32 bytes

### Stateful Opcodes
- SELFBALANCE (`0x47`) - https://github.com/ethereum/execution-spec-tests/pull/1771
  - [x] SimpleInstructionTwoContracts (Endlessly pushing self balance to stack (1000 per 1 contract))

### Environment
- CALLER (`0x33`)
[...]

PR #1698, opened 2025-06-02
zkevm: add 0-param opcodes coverage

jsign:
This PR adds coverage for many 0-param opcodes.

Cycles:
```
tests/zkevm/test_worst_compute.py::test_worst_zero_param[fork_Cancun-blockchain_test_from_state_test-opcode_GAS]-1      927313344
tests/zkevm/test_worst_compute.py::test_worst_zero_param[fork_Cancun-blockchain_test_from_state_test-opcode_TIMESTAMP]-1        945280386
tests/zkevm/test_worst_compute.py::test_worst_zero_param[fork_Cancun-blockchain_test_from_state_test-opcode_GASLIMIT]-1 945298712
tests/zkevm/test_worst_compute.py::test_worst_zero_param[fork_Cancun-blockchain_test_from_state_test-opcode_NUMBER]-1   945298968
tests/zkevm/test_worst_compute.py::test_worst_callvalue[fork_Cancun-blockchain_test_from_state_test-from_origin_False-non_zero_value_True]-1    983444990
tests/zkevm/test_worst_compute.py::test_worst_callvalue[fork_Cancun-blockchain_test_from_state_test-from_origin_False-non_zero_value_False]-1   983643412
tests/zkevm/test_worst_compute.py::test_worst_returndatasize[fork_Cancun-blockchain_test_from_state_test-return_data_style_ReturnDataStyle.IDENTITY-non_zero_size_True]-1      999200881
tests/zkevm/test_worst_compute.py::test_worst_returndatasize[fork_Cancun-blockchain_test_from_state_test-return_data_style_ReturnDataStyle.REVERT-non_zero_size_False]-1       999224296
tests/zkevm/test_worst_compute.py::test_worst_returndatasize[fork_Cancun-blockchain_test_from_state_test-return_data_style_ReturnDataStyle.RETURN-non_zero_size_False]-1       999224436
tests/zkevm/test_worst_compute.py::test_worst_returndatasize[fork_Cancun-blockchain_test_from_state_test-return_data_style_ReturnDataStyle.IDENTITY-non_zero_size_False]-1     999243170
tests/zkevm/test_worst_compute.py::test_worst_returndatasize[fork_Cancun-blockchain_test_from_state_test-return_data_style_ReturnDataStyle.REVERT-non_zero_size_True]-1        999255038
[...]

PR #1737, opened 2025-06-11
refactor(benchmark): optimize the zero param opcode case

LouisTsai-Csie:
## 🗒️ Description

For the test case involving a zero-parameter stack-producing opcode, as proposed in [PR ethereum/execution-spec-tests#1698](https://github.com/ethereum/execution-spec-tests/pull/1698), the following code pattern is used (taking `ADDRESS` 0x30 as an example):

```
JUMPDEST
ADDRESS
POP
ADDRESS
POP
... (repeat the push-pop pattern)
ADDRESS
POP
PUSH0
JUMP
```

This pattern pushes a stack-producing opcode with zero parameters (`ADDRESS`) followed by a POP, and repeats this sequence until the code size approaches the maximum code size. At the end, the code performs a `JUMP` back to the beginning to re-execute the sequence in a loop.

Each zero-parameter opcode like `ADDRESS` consumes 2 gas units, and each `POP` also consumes 2 gas units. Additionally, the control flow instructions at the end cost: `JUMPDEST` (1 gas), `PUSH0` (1 gas), and `JUMP` (8 gas). Given the maximum allowed code size of 24,576 bytes and that each `ADDRESS`/`POP` pair takes 4 bytes, we can fit (24576 - 3) // 4 = 6143 sequences. Given a 60 million gas limit, this allows: 60,000,000 // (6143 * 4 + 3) = 2411 loop iterations

Each loop uses 12286 `ADDRESS` opcodes at 2 gas each, so the total gas spent on `ADDRESS` execution alone is 2411 * 6143 * 2 = 29,633,746 gas.

This test can be optimized using a `STATICCALL` loop. The idea is:
- Deploy a helper contract containing 1024 `ADDRESS` opcodes (limited by the maximum stack depth).
- In the main contract, use a loop to repeatedly `STATICCALL` the helper contract.

The optimized loop would look like:

```
JUMPDEST
PUSH0
PUSH0
PUSH0
PUSH20 <address>
PUSH2 0xFFFF
STATICCALL
POP
PUSH0
JUMP
```
Breakdown of gas cost per loop iteration:
- JUMPDEST (1)
- PUSH0 ×3 (3 × 2 = 6)
- PUSH20 (3)
- PUSH2 (3)
- STATICCALL (100)
- POP (2)
- PUSH0 (2)
- JUMP (8)
- External contract execution: 1024 * 2 = 2048 gas
Total per iteration: 1 + 6 + 3 + 3 + 100 + 2 + 2 + 8 + 2048 = 2173 gas, accounting for a small static cost (e.g. 2,500 gas for contract warmup), the number of iterations is (60,000,000 - 2,500) // 2175 = 27,586. Total gas spent on `ADDRESS` execution via `STATICCALL` is 27586 iterations × 1024 opcodes × 2 gas = 56,496,128 gas

By switching from direct inline execution to a `STATICCALL`-based pattern, the gas utilization becomes significantly more efficient: ~56.5M gas used for `ADDRESS` in the optimized version vs. ~29.7M gas in the original.

This would affect issue https://github.com/ethereum/execution-spec-tests/issues/1734

## 🔗 Related Issues

Issue ethereum/execution-spec-tests#1734 

## ✅ Checklist

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -84,6 +84,7 @@ def test_worst_zero_param(
     state_test: StateTestFiller,
     pre: Alloc,
     opcode: Op,
+    fork: Fork,
 ):
     """Test running a block with as many zero-parameter opcodes as possible."""
     env = Environment()
@@ -91,11 +92,13 @@ def test_worst_zero_param(
     opcode_sequence = opcode * MAX_STACK_HEIGHT
     target_contract_address = pre.deploy_contract(code=opcode_sequence)
 
-    caller_contract_code = While(body=Op.POP(Op.STATICCALL(address=target_contract_address)))
-    caller_contract_address = pre.deploy_contract(code=caller_contract_code)
+    calldata = Bytecode()
+    attack_block = Op.POP(Op.STATICCALL(Op.GAS, target_contract_address, 0, 0, 0, 0))
+    code = code_loop_precompile_call(calldata, attack_block, fork)
+    code_address = pre.deploy_contract(code=code)
 
     tx = Transaction(
-        to=caller_contract_address,
+        to=code_address,
         gas_limit=env.gas_limit,
         sender=pre.fund_eoa(),
     )


tests/zkevm/test_worst_compute.py
```diff
@@ -82,22 +83,19 @@ def make_dup(index: int) -> Opcode:
...
-    code_prefix = Op.JUMPDEST
-    iter_loop = Op.POP(opcode)
-    code_suffix = Op.PUSH0 + Op.JUMP
-    code_iter_len = (max_code_size - len(code_prefix) - len(code_suffix)) // len(iter_loop)
-    code = code_prefix + iter_loop * code_iter_len + code_suffix
-    assert len(code) <= max_code_size
+    opcode_sequence = opcode * MAX_STACK_HEIGHT
+    target_contract_address = pre.deploy_contract(code=opcode_sequence)
+
+    caller_contract_code = While(body=Op.POP(Op.STATICCALL(address=target_contract_address)))
```
marioevz, line 94:
```suggestion
    caller_contract_code = (
        Op.JUMPDEST + Op.JUMPDEST + Op.STATICCALL(address=target_contract_address) + Op.JUMP
    )
```
Tiny optimization here because we can actually use the return code of the staticcall as the jump destination.

marioevz (approved):
Great analysis, I think this is a nice optimization!

Just one small suggestion.

tests/zkevm/test_worst_compute.py
jsign: There's another optimization possible, which is what we do in many other tests -- put as many iterations of STATICCALL as possible considering the max contract size and jump to the beginning at the end.

As in:
```
JUMPDEST
POP(STATICCALL)
POP(STATICCALL)
...
...
JUMP 0
```

So we amortize the 8 gas of JUMP in as many STATICCALL as possible. It is basically recovering the deleted lines in this PR but framing it for the STATICCALL instead of `opcode`.

jsign (approved):
Nice work! I only left a comment for you to consider since I think we could squeeze a bit more with some known trick.

tests/zkevm/test_worst_compute.py
LouisTsai-Csie: I like the idea of using `DUP1` instead of a `PUSH` operation.

Using `PUSH20` to push the address consumes 20 bytes of immediate value and costs 3 gas, whereas `DUP1` only costs 2 gas and requires no immediate data.

@marioevz  I think this could further be optimized in the current `While` implementation, so I might not change my implementation here, wdyt.

LouisTsai-Csie:
@jochem-brouwer Thank you so much for the explanation, it really helped me understand the inner workings of the test better. I’ve updated accordingly, and this PR is now ready for review. It would be great if you could take another look. Thanks again!

tests/zkevm/test_worst_compute.py
jochem-brouwer: Nvm I need to get some coffee and then I should revisit above post, because it is not correct.

LouisTsai-Csie: Thank you for the feedback! I’ve summarized our discussion in [this comment](https://github.com/ethereum/execution-spec-tests/issues/1704#issuecomment-2975867921). I was thinking we could merge this PR first, and then follow up with a separate one to finalize the implementation and update all related usages?

LouisTsai-Csie: I found an existing helper function that is more optimized, let me update the code with the helper

tests/zkevm/test_worst_compute.py
```diff
@@ -40,6 +40,7 @@
 REFERENCE_SPEC_GIT_PATH = "TODO"
 REFERENCE_SPEC_VERSION = "TODO"
 
+MAX_STACK_HEIGHT = 1024
```
jochem-brouwer, line 43: I think that this constant should be moved to the fork config (such that if we change max stack height at some point we can edit it in the fork schedule :smile: :+1). So in `src/ethereum_test_forks/forks`. Will nevertheless approve (ok to cleanup in another PR).

jochem-brouwer (approved):
:tada:

tests/zkevm/test_worst_compute.py
LouisTsai-Csie: I’ve created an issue to track this: https://github.com/ethereum/execution-spec-tests/issues/1757. I’ll pick it up in a separate PR! Thanks

PR #1738, opened 2025-06-12
fix(filler): Fix shared-pre-state on xdist

marioevz:
## 🗒️ Description
Performs two fixes:
- Use correct name for function `model_post_init`
- Don't raise exception on duplicated accounts in the pre for multiple xdist workers. Reason being that we normally have consistent contracts across tests, for example system contracts.

## ✅ Checklist


danceratopz (approved):
Thanks! LGTM!

PR #1741, opened 2025-06-13
fix(tests): Fix test_worst_selfdestruct_created

marioevz:
## 🗒️ Description
Fixes `test_worst_selfdestruct_created` and `test_worst_selfdestruct_initcode` to be filled in higher block gas limits.

## 🔗 Related Issues
None.

## ✅ Checklist

### DIFF
--- a/tests/zkevm/test_worst_stateful_opcodes.py
+++ b/tests/zkevm/test_worst_stateful_opcodes.py
@@ -515,7 +515,6 @@ def test_worst_selfdestruct_existing(
         + gas_costs.G_SELF_DESTRUCT
         + 63  # ~Gluing opcodes
     )
-    assert loop_cost == 7720
     final_storage_gas = (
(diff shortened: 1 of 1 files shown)


jsign (approved):
Thanks!

The tricky part about this is that we didn't realize about this until we create a tag for 150m gas. I guess testing this in CI itself might be too heavy? (since we'd have to do the filling for all 36m, 60m, etc (all known configs)). Worth doing this or maybe not?

marioevz:
> Thanks!
> 
> The tricky part about this is that we didn't realize about this until we create a tag for 150m gas. I guess testing this in CI itself might be too heavy? (since we'd have to do the filling for all 36m, 60m, etc (all known configs)). Worth doing this or maybe not?

I think we could go just for the largest one in CI, 150M, and use `--skip-evm-dump` to make it run a bit faster.

PR #1742, opened 2025-06-13
feat(zkevm): create pure ether transfer worst case

LouisTsai-Csie:
## 🗒️ Description

In Nethermind’s test case, a block is filled with simple Ether transfer transactions, each transferring 1 wei. The purpose of this PR is to reproduce a specific scenario and expand test coverage. Four scenarios are included, as outlined below:
1. Account A sends Ether to itself
2. Account A sends Ether to Account B
3. Account A sends Ether to multiple different accounts
4. Multiple different accounts send Ether to Account A

## 🔗 Related Issues

Issue https://github.com/ethereum/execution-spec-tests/issues/1734

## ✅ Checklist

### DIFF
--- a/tests/zkevm/test_worst_blocks.py
+++ b/tests/zkevm/test_worst_blocks.py
@@ -10,6 +10,7 @@ import pytest
 from ethereum_test_forks import Fork
 from ethereum_test_tools import (
     Account,
+    Address,
     Alloc,
     Block,
     BlockchainTestFiller,
@@ -21,7 +22,7 @@ from ethereum_test_tools import (
 @pytest.fixture
 def iteration_count(eth_transfer_cost: int):
     """Calculate the number of iterations based on the gas limit and intrinsic cost."""
-    return min(5, Environment().gas_limit // eth_transfer_cost)
+    return Environment().gas_limit // eth_transfer_cost
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


tests/zkevm/test_worst_blocks.py
```diff
@@ -0,0 +1,137 @@
...
+    Environment,
+    Transaction,
+)
+
+
+@pytest.fixture
+def iteration_count(fork: Fork):
+    """Calculate the number of iterations based on the gas limit and intrinsic cost."""
+    intrinsic_cost = fork.transaction_intrinsic_cost_calculator()
+    return Environment().gas_limit // intrinsic_cost()
```
marioevz, line 25:
```suggestion
def iteration_count(eth_transfer_cost: int):
    """Calculate the number of iterations based on the gas limit and intrinsic cost."""
    return Environment().gas_limit // eth_transfer_cost```

File: tests/zkevm/test_worst_blocks.py [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+
+    senders = [pre.fund_eoa() for _ in range(iteration_count)]
+    receivers = [pre.fund_eoa(0) for _ in range(iteration_count)]
+
+    single_sender = pre.fund_eoa()
+    single_receiver = pre.fund_eoa(0)
+
+    if case_id == "a_to_a":
+        """Sending to self."""
+        total_balance = pre._eoa_fund_amount_default - eth_transfer_cost * iteration_count * 0x0A
```
marioevz, line 60: It's tricky because `_eoa_fund_amount_default` changes in execute.

tests/zkevm/test_worst_blocks.py
```diff
@@ -0,0 +1,137 @@
...
+    pre: Alloc,
+    iteration_count: int,
+    transfer_amount: int,
+    eth_transfer_cost: int,
+):
+    """Generate the test parameters based on the case ID."""
+    case_id = request.param
+
+    senders = [pre.fund_eoa() for _ in range(iteration_count)]
+    receivers = [pre.fund_eoa(0) for _ in range(iteration_count)]
```
marioevz, line 53: I feel like these could be an iterator each.

marioevz (commented):
Looks good, some comments.

LouisTsai-Csie:
@marioevz Thanks for the review, I had some update and left a few questions.

tests/zkevm/test_worst_blocks.py
```diff
@@ -0,0 +1,153 @@
...
+@pytest.fixture
+def eth_transfer_cost(fork: Fork):
+    """Transaction gas limit."""
+    intrinsic_cost = fork.transaction_intrinsic_cost_calculator()
+    return intrinsic_cost()
+
+
+def get_distinct_sender_list(pre: Alloc, iteration_count: int):
+    """Get a list of distinct sender accounts."""
+    return [pre.fund_eoa() for _ in range(iteration_count)]
```
marioevz, line 42: You can modify this function and `get_distinct_receiver_list`, `get_single_sender_list` and `get_single_receiver_list` to be generators.
```suggestion
def get_distinct_sender_list(pre: Alloc):
    """Get a list of distinct sender accounts."""
    while True:
        yield pre.fund_eoa()
```

And then in `ether_transfer_case`, you can just call the generators into a variable:
```python
if case_id == "a_to_a":
    senders = get_single_sender_list()
    receivers = senders
elif case_id == "a_to_b":
    senders = get_single_sender_list()
    receivers = get_single_receiver_list()
elif case_id == "diff_acc_to_b":
    senders = get_distinct_sender_list()
    receivers = get_single_receiver_list()
elif case_id == "a_to_diff_acc":
    senders = get_single_sender_list()
    receivers = get_distinct_receiver_list()
elif case_id == "diff_acc_to_diff_acc":
    senders = get_distinct_sender_list()
    receivers = get_distinct_receiver_list()
return senders, receivers
```

tests/zkevm/test_worst_blocks.py
```diff
@@ -0,0 +1,153 @@
...
+
+def get_distinct_receiver_list(pre: Alloc, iteration_count: int):
+    """Get a list of distinct receiver accounts."""
+    return [pre.fund_eoa(0) for _ in range(iteration_count)]
+
+
+def get_single_sender_list(pre: Alloc, iteration_count: int):
+    """Get a list of single sender accounts."""
+    sender = pre.fund_eoa()
+    return [sender for _ in range(iteration_count)]
```
marioevz, line 53:
```suggestion
def get_single_sender_list(pre: Alloc):
    """Get a list of single sender accounts."""
    sender = pre.fund_eoa()
    while True:
        yield sender
```
Similar to previous suggestion.

tests/zkevm/test_worst_blocks.py
marioevz: Yeah, it's not simple, but I also feel like checking the post here is not worth it, in the rest of the cases is easy, so I feel like we should only check the post in cases that are not "a_to_a" and that's it.

LouisTsai-Csie:
@marioevz Thanks for your explanation, I've updated with
1. Change the sender/receiver function to iterator
2. Remove the check for the case `a_to_a`

tests/zkevm/test_worst_blocks.py
```diff
@@ -0,0 +1,160 @@
...
+    request,
+    pre: Alloc,
+):
+    """Generate the test parameters based on the case ID."""
+    case_id = request.param
+
+    if case_id == "a_to_a":
+        """Sending to self."""
+        senders = get_single_sender_list(pre)
+        receivers = get_single_receiver_list(pre)
```
marioevz, line 78:
```suggestion
        senders = get_single_sender_list(pre)
        receivers = get_single_receiver_list(pre)
```
```suggestion
        senders = get_single_sender_list(pre)
        receivers = senders
```
I think we have to return the exact same generator twice here, otherwise is going to return two different addresses.

PR #1744, opened 2025-06-14
feat(zkevm): add p256verify to worst compute tests

LouisTsai-Csie:
## 🗒️ Description

Adding benchmark test for the `P256VERIFY` precompile contract.

## 🔗 Related Issues

Issue: https://github.com/ethereum/execution-spec-tests/issues/1734

## ✅ Checklist

### DIFF
--- a/tests/zkevm/test_worst_compute.py
+++ b/tests/zkevm/test_worst_compute.py
@@ -706,6 +706,9 @@ def test_worst_precompile_fixed_cost(
     """Test running a block filled with a precompile with fixed cost."""
     env = Environment()
 
+    if precompile_address not in fork.precompiles():
+        pytest.skip("Precompile not enabled")
+
     concatenated_bytes: bytes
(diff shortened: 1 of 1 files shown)


tests/zkevm/test_worst_compute.py
```diff
@@ -680,6 +682,17 @@ def test_worst_modexp(state_test: StateTestFiller, pre: Alloc, fork: Fork):
             ],
             id="bls12_fp_to_g2",
         ),
+        pytest.param(
```
marioevz, line 685: Test is going to pass but the test is only going to test empty account calls because there's no precompile here prior to Prague.

We should test whether `precompile_address` is in `fork.precompiles()`, and if not just `pytest.skip("Precompile not enabled")` at the start of the function.

marioevz (commented):
Excellent, just one comment overall.

Issue #1689, opened 2025-05-30
Create Blockchain tests for Contract Lifecycle and Logging opcodes that consume the whole block


kevaundray:
## General context

Generally the structure of these tests are split into two:

- Setup phase
- Benchmarking block

**Setup phase**

The setup phase encompasses the pre-state and blocks that are needed in order to setup the block full of a specific block.
An example of this would be a "block full of EXTCODESIZE" where the setup phase would consist of N different contracts being deployed, where N is the number of EXTCODESIZE opcdes we want to put in the final block. 

**Benchmarking block**

For benchmarking, we assume that the final block is the one with the block full of the particular opcode we want to benchmark

# Opcodes

## Logging 

- [x]  LOG0  
- [x]  LOG1  
- [x]  LOG2  
- [x]  LOG3  
- [x]  LOG4  

## Contract Lifecycle 
- [x]  CREATE  
- [x]  RETURN  
- [x]  CREATE2  
- [x]  REVERT  

## Note

- This is a subtask of #1453

LouisTsai-Csie:
Could I take this one for the Logging operation?

kevaundray:
> Could I take this one for the Logging operation?

Yes please

jochem-brouwer:
I think for `RETURN` / `REVERT` since the base cost is 0 (I think) it still allocates memory, so we could test to fill the return buffer with large sizes.

For CREATE:

I think the coverage here is already good. To deploy initcode for `code_size` can use `EXTCODESIZE` to ensure all is copied. For the benchmark tx especially for CREATE2 we can re-use the returned create address from CREATE2 as salt (so no need to POP that). This also ensures that the target address is unique (and thus deploys something). I think this is the easy way to ensure that the salt differs each time, otherwise you have to manipulate the salt in some way (e.g. `ADD`). 

For fancy constructions do we also want to call `CREATE` within `CREATE`? Not sure if we want to test that, but maybe a `CREATE` tower where we keep init-ing `CREATE` sounds like a good idea. (This will actually deploy contracts, since at the lowest level the CREATEs will OOG, this will return to the previous frame which has gas left, which will thus succeed deploying (at least the empty contract).

Should we also test CREATE and especially CREATE2 with address collisions? (Attempting to deploy a contract when there is already code there).

LouisTsai-Csie:
[...]

PR #1745, opened 2025-06-15
feat(zkevm): add log opcode worst case

LouisTsai-Csie:
## 🗒️ Description

Create Benchmark for `LOG` Opcode

This PR introduces a benchmark for the `LOG` opcode in the EVM. The `LOG` operation includes multiple fields: `offset`, `size`, and up to `N` topics depending on the variant (LOG0 through LOG4).

Gas Cost Calculation

Using `LOG0` as an example, the gas cost for the `LOG` operation is computed as:
- Static gas: `375`
- Dynamic gas:`dynamic_gas = 375 * topic_count + 8 * size + memory_expansion_cost`

To minimize gas costs:
- Use zero size to avoid dynamic gas from memory access.
- Use `PUSH0` (which costs 2 bytes) for pushing values to the stack, as it’s cheaper than other operation like `PUSH1` (3 bytes with an immediate) or `DUP1` (3 bytes).

Using `PUSH0` is also more space-efficient, since it avoids including immediate values and reduces bytecode size.

Benchmark Bytecode Pattern

A minimal loopable pattern for benchmarking LOGN looks like this:
```
JUMPDEST
PUSH0      ; offset
PUSH0      ; size
...        ; PUSH0 repeated N times for each topic
LOGN
...        ; repeatable pattern
PUSH0
JUMP
```

Using `PUSH0` is also more space-efficient, since it avoids including immediate values and reduces bytecode size. This makes it preferable over `DUP1` for setting up the stack for topics or memory offset/size.

## 🔗 Related Issues

Issue #1689 

## ✅ Checklist


jochem-brouwer (commented):
This test looks fine to me, it tests all LOG opcodes and runs those all as much as possible. However, I think we should also add these situations:

- Log data length is nonzero
- Log data length + data is nonzero (here we have to initialize the memory to something non-zero)
- The topics are nonzero

I am not sure because I am not deep into zkEVM, is calculating the `logsBloom` part of the scope of the EVM (and would thus add extra logic/cycles) @jsign? If this is indeed part of it (either `logsBloom` or calculating `receiptTrie` or both) then those tests should be added for zkEVM (can be done in a different PR).

Those tests are also nice benchmark tests for non-zkEVM also :smile: :+1: 

Will not explicitly approve for now.

LouisTsai-Csie:
Thanks for comment, I do not know `logsBloom` and `receiptTrie` before, i will review some documentation for them

tests/zkevm/test_worst_compute.py
```diff
@@ -1703,3 +1703,47 @@ def test_worst_calldataload(
         post={},
         tx=tx,
     )
+
+
+@pytest.mark.parametrize(
+    "opcode,topic_count",
```
jsign, line 1709: Maybe we can remove the `topic_count` parameter, and derive it from `opcode - Op.LOG0`?
Mainly because combinations like `Op.LOG0, 1` or `Op.LOG3, 4` would never make sense?

jsign (approved):
Looking good! Reg @jochem-brouwer suggestion to cover non-zero scenarios (which I agree are useful), I'll let you decide if those might be worked in this PR or in a separate one -- for now leaving an approval if it is the latter.

Just left one comment for your consideration.

tests/zkevm/test_worst_compute.py
jochem-brouwer: I like this suggestion :smile:

tests/zkevm/test_worst_compute.py
```diff
@@ -1703,3 +1703,47 @@ def test_worst_calldataload(
...
+    pre: Alloc,
+    fork: Fork,
+    opcode: Opcode,
+    topic_count: int,
+):
+    """Test running a block with as many LOG opcodes as possible."""
+    env = Environment()
+    max_code_size = fork.max_code_size()
+
+    iter_code = Op.PUSH0 * (2 + topic_count) + opcode
```
marioevz, line 1729:
```suggestion
    iter_code = Op.PUSH0 * len(opcode.kwargs) + opcode
```
And you can remove the `topic_count` parameter.

marioevz (commented):
LGTM, just one suggestion that is in line with the comments from previous reviewers, thanks!

tests/zkevm/test_worst_compute.py
```diff
@@ -2173,3 +2173,75 @@ def test_worst_push(
...
+    fixed_offset: bool,
+):
+    """Test running a block with as many LOG opcodes as possible."""
+    env = Environment()
+    max_code_size = fork.max_code_size()
+
+    calldata = (
+        Op.PUSH0
+        if empty_value
+        else Op.PUSH32(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF)
```
LouisTsai-Csie, line 2218: I think this could be further optimized for readability

LouisTsai-Csie:
@jsign @jochem-brouwer I’ve updated the test case and added the following scenarios:
- Topics are zero and non-zero
- Log data has a non-zero length

Could you please help me review again? And i am asking why the CI is failing

marioevz (commented):
Thanks! Just some minor comments.

I think we can move this new test function into a new file, perhaps `tests/zkevm/test_worst_opcode.py` ?

We can start by moving only this new function for now in this PR, I have an idea on how to fix the coverage script and we can move the rest once that's fixed.

LouisTsai-Csie:
@marioevz Thanks for review. My original problem for evmone coverage report is that the CI never ends, but after I rebase again, the issue changed. 

After checking the coverage report, the root cause is from BLS signature, which has nothing to do with this benchmark. I move the test case to `tests/zkevm/test_worst_opcode.py`, let's see if this fix the CI issue

@jsign @jochem-brouwer could you please also review this benchmark, let me know if I miss any cases!

tests/zkevm/test_worst_opcode.py
```diff
@@ -0,0 +1,79 @@
...
+    env = Environment()
+    max_code_size = fork.max_code_size()
+
+    calldata = Op.PUSH0 if empty_topic else Op.PUSH32(2**256 - 1)
+
+    topic_count = len(opcode.kwargs or []) - 2
+
+    offset = Op.PUSH0 if fixed_offset else Op.MOD(Op.GAS, 7)
+
+    code_sequence = Op.DUP1 * topic_count + Op.PUSH32(size) + offset + opcode
```
jsign, line 61: Could use `PUSH3` to avoid some bytecode bloating.

jsign, line 61: The `(offset, size)` is always pointing to zeroed memory, no?

We could add another parameter of `non_zero_data : [True, False]`. If `False` do what is done now, and if `True`, maybe in `calldata` add a cheapish `CODECOPY` with size so we get some non-zeroed data?

jsign (approved):
Looks good, I left some comments for your consideration.

marioevz:
The coverage script was broken and I created a PR to fix it: https://github.com/ethereum/execution-spec-tests/pull/1817

I pushed commit [1510245](https://github.com/ethereum/execution-spec-tests/pull/1745/commits/1510245592349a022f0276afe824945fb0d614f9) on top of your branch to verify the fix, and we can remove it once we merge #1817 to main and rebase this branch.

marioevz (approved):
Fixed the coverage issue by rebasing on top of main again, so we can address @jsign  comments and then proceed to merge.

tests/zkevm/test_worst_opcode.py
LouisTsai-Csie: I first push the size via PUSH3 onto the stack, and then access it later via DUP operation.

tests/zkevm/test_worst_opcode.py
```diff
@@ -39,6 +40,7 @@
 )
 @pytest.mark.parametrize("empty_topic", [True, False])
 @pytest.mark.parametrize("fixed_offset", [True, False])
+@pytest.mark.parametrize("non_zero_data", [True, False])
```
jsign, line 43: I thin we should probably remove this independent parameter and bundle it with `size`.
As in, doing `"size,non_zero_data"` with values `[(0, False), (1MiB, True), (1MiB, False)]`.

So we can avoid the current `0 size + non_zero_data` combination that doesn't make sense, and would be burning the all contract `CODECOPY` without any gain.

jsign (approved):
LGTM, I think there's only one easy detail left but leaving it approved.

tests/zkevm/test_worst_opcode.py
```diff
@@ -0,0 +1,95 @@
...
+        pytest.param(Op.LOG2, id="log2"),
+        pytest.param(Op.LOG3, id="log3"),
+        pytest.param(Op.LOG4, id="log4"),
+    ],
+)
+@pytest.mark.parametrize(
+    "size,non_zero_data",
+    [
+        pytest.param(0, False, id="0_bytes_zero_data"),
+        pytest.param(1024 * 1024, False, id="1_MiB_zero_data"),  # 1 MiB
```
marioevz, line 38:
```suggestion
        pytest.param(0, False, id="0_bytes_data"),
        pytest.param(1024 * 1024, False, id="1_MiB_zeros_data"),  # 1 MiB
```

tests/zkevm/test_worst_opcode.py
```diff
@@ -0,0 +1,95 @@
...
+        pytest.param(1024 * 1024, True, id="1_MiB_non_zero_data"),  # 1 MiB
+    ],
+)
+@pytest.mark.parametrize("empty_topic", [True, False])
```
marioevz, line 42:
```suggestion
@pytest.mark.parametrize(
    "zeros_topic", [pytest.param(True, id="zeros_topic"), pytest.param(False, id="non_zero_topic")]
)
```

tests/zkevm/test_worst_opcode.py
```diff
@@ -0,0 +1,95 @@
...
+@pytest.mark.parametrize("fixed_offset", [True, False])
+def test_worst_log_opcodes(
+    state_test: StateTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    opcode: Opcode,
+    empty_topic: bool,
```
marioevz, line 49:
```suggestion
    zeros_topic: bool,
```

tests/zkevm/test_worst_opcode.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # For non-zero data, load  into memory.
+    if non_zero_data:
+        calldata += Op.CODECOPY(dest_offset=0, offset=0, size=Op.CODESIZE)
+
+    # Push the size value onto the stack and access it using the DUP opcode.
+    calldata += Op.PUSH3(size)
+
+    # For non-empty topic, push a non-zero value for topic.
+    calldata += Op.PUSH0 if empty_topic else Op.PUSH32(2**256 - 1)
```
marioevz, line 68:
```suggestion
    # For non-zeros topic, push a non-zero value for topic.
    calldata += Op.PUSH0 if zeros_topic else Op.PUSH32(2**256 - 1)
```

marioevz (approved):
LGTM, thanks!

Just some minor nit-picks.

LouisTsai-Csie:
Thanks @marioevz

PR #1751, opened 2025-06-16
fix(gentest): update imports to fix template

jochem-brouwer:
## 🗒️ Description
When playing around with `uv run gentest` it generated a file which could not be ran directly since it imported stuff which was not the method which was being used. Keeping this as draft, because we might want to update some things for `gentest`:

- [ ] The generated file was invalid and could not be ran. We should add a test for this to ensure the output runs out-of-the-box
- [ ] The tracer works using `debug_traceCall`. We should change this to `debug_traceTransaction`, because my endpoint does not support `debug_traceCall`. I also think that `traceCall` produces incorrect results if your tx is not in index 0, because it will run the call on top of the block and not after executing the previous txs. `debug_traceTransaction` also supports the prestate tracer and it works out of the box.

For the change in `debug_traceTransaction` I have this local fix (in `src/ethereum_test_rpc/rpc.py`):

```
    def trace_call(self, tr: dict[str, str], block_number: str):
        """`debug_traceCall`: Returns pre state required for transaction."""
        return self.post_request(
            "traceTransaction",
            "<hardcoded tx hash>",
            {"tracer": "prestateTracer"},
        )
```

This `gentest` tool is extremely helpful to sandbox a tx and use it to check against clients to see if something funny is happening there! :smile: :+1: 

CC @raxhvl, feel free to take over this PR :smile: :+1: 

## ✅ Checklist

### DIFF
--- a/src/cli/gentest/templates/blockchain_test/transaction.py.j2
+++ b/src/cli/gentest/templates/blockchain_test/transaction.py.j2
@@ -8,7 +8,13 @@ from typing import Dict
 
 import pytest
 
-from ethereum_test_tools import Account, Environment, StateTestFiller, Storage, Transaction
+from ethereum_test_rpc.types import TransactionByHashResponse
+from ethereum_test_tools import (
+    Account,
+    Environment,
+    StateTestFiller,
+    Storage,
+)
 
 REFERENCE_SPEC_GIT_PATH = "N/A"
 REFERENCE_SPEC_VERSION = "N/A"


danceratopz:
I just wanted to check the import fix to see if we can get this merged real quick. 

But this fails even worse on `main` :sob: so we need to fix that first:
- #2259

Also, this is a tough example, but on this branch:

```
uv run gentest 0x79510fa288a2d4d1bbe3794e3a079d2bcff0eb4550bc67e1ab6915b263979877 out.py
```

also fails mypy and ruff checks. So this all needs quite a bit of love as of right now.

PR #1756, opened 2025-06-17
chore(tooling): better mypy config

felix314159:
## 🗒️ Description
We should be explicit about which folders should be scanned [`src`, `tests`] and also about which folders can be ignored for performance reasons (via exclude list). On my machine, when I delete `.mypy_cache` and run `time uv run mypy .` from eest root then i get a time of 9 seconds. Any run after this will be cached and take less than 0.9 sec on my machine. Even if our CI can't make use of caching (here we should ask ourselves if maybe we can make this possible if it is not done already) I think 9 seconds is reasonable and I would like to avoid to pivot to `ty` until that tool is proven to work better than mypy and all of its to-be-expected initial flaws are fixed. I would like to avoid to abandon a tool that has been developed and improved for more than 10 years (Van Rossum himself has been working on mypy) for some rust clickbait alternative.

## ✅ Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,7 +130,6 @@ files = ["src", "tests"]
 exclude = [
     '^\.cache/',
     '^\.git/',
-    '^\.github/',
     '^\.pytest_cache/',
     '^\.ruff_cache/',
     '^\.tox/',


felix314159:
> I think it makes sense. Does it impact performance in any way (positively I mean)?
> 
> Also it seems like the CI is failing now, do you see this in your local runs of tox?

Good catch, what I think happened: The github action seems to make use of the tox config and the [tox config](https://github.com/ethereum/execution-spec-tests/blob/ad7fc35deeb056bd9afc6d543691ec02b469c6da/tox.ini#L18) explicitely specifies the folders: `python_source_dirs = src tests .github/scripts` but since this PR excludes the .github folder it can't find the Python files in it and apparently a lack of python files leads to pipeline failure. 

Easiest fix is just allowing mypy to check the `.github` folder it doesn't contain that many files so it's fine to always have it checked even when it is rarely modified.

PR #1746, opened 2025-06-15
feat(filler): Add `coverage_missed_reason` keyword arguemtn to `ported_from` marker

marioevz:
## 🗒️ Description

Adds a `coverage_missed_reason` keyword arguemtn to `ported_from` marker, which allows the test to be skipped from coverage comparison when the coverage drop has already been manually reviewed and deemed acceptable or not applicable to the test.

## 🔗 Related Issues

None

## ✅ Checklist

- [x] All: Set appropriate labels for the changes.
- [x] All: Considered squashing commits to improve commit history.
- [ ] All: Added an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.

PR #1761, opened 2025-06-17
fix(tests): Add `coverage_missed_reason` to more ported tests.

marioevz:
## 🗒️ Description

This PR will add the `coverage_missed_reason` to more ported tests after it was added by #1746.

## ✅ Checklist


.github/scripts/fill_prepatched_tests.sh
```diff
@@ -0,0 +1,53 @@
...
+git checkout main
+PREV_COMMIT=$(git rev-parse HEAD)
+echo "Checkout head $PREV_COMMIT"
+
+# Take only those files that exist in the filesystem (ignore newly created files)
+files_fixed=$(echo "$files" | tr ' ' '\n' | while read file; do
+    if [ -f "$file" ]; then
+        echo "$file"
+    fi
+done | tr '\n' ' ')
```
danceratopz, line 29: 1. Removed the intermediate `files` variable entirely
2. Direct `jq` parsing from `$CHANGED_TEST_FILES`
3. Removed the final `| tr '\n' ' '` (unnecessary for bash word splitting)

```suggestion
git checkout main
PREV_COMMIT=$(git rev-parse HEAD)
echo "Checkout head $PREV_COMMIT"

# Take only those files that exist in the filesystem (ignore newly created files)
files_fixed=$(echo "$CHANGED_TEST_FILES" | jq -r '.[]?' | while read file; do
    if [ -f "$file" ]; then
        echo "$file"
    fi
done)
```

.github/scripts/parse_ported_tests.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+# Parse ported_from markers from introduced .py tests
+# Usage: parse_ported_tests.sh <changed_test_files> <workspace_path>
```
danceratopz, line 3:
```suggestion
# Usage: parse_ported_tests.sh <changed_test_files> <workspace_path>
# Requirements: jq
```

.github/scripts/parse_ported_tests.sh
```diff
@@ -0,0 +1,77 @@
...
+#   0 - Found converted tests, continue processing
+#   1 - No converted tests found, but updates detected
+
+set -e
+
+CHANGED_TEST_FILES="$1"
+WORKSPACE_PATH="${2:-$GITHUB_WORKSPACE}"
+
+# Convert comma-separated list to newline-separated
+new_sources=$(echo "$CHANGED_TEST_FILES" | tr ',' '\n')
```
danceratopz, line 14:
```suggestion
new_sources=$(echo "$CHANGED_TEST_FILES" | jq -r '.[]?')
```

.github/scripts/fill_prepatched_tests.sh
danceratopz: Actually this (note: `tests_all_modified_files` instead of `tests_all_changed_files`)
https://github.com/ethereum/execution-spec-tests/blob/5865d02c55c0835a7bcc07c6785dc799f303a734/.github/workflows/coverage.yaml#L136

should mean that we need to check for newly created files. But it doesn't help to be safe here, I suppose.

.github/workflows/coverage.yaml
```diff
@@ -11,6 +11,10 @@ on:
 jobs:
   evmone-coverage-diff:
     runs-on: ubuntu-latest
+    
+    env:
+      BLOCK_GAS_LIMIT: "36000000"
+      FILL_UNTIL: "Cancun"
```
danceratopz, line 17: Just something to bear in mind.... I wonder if this will bite us at some point? I mean, if the default fill gas block limit is changed elsewhere, we might forget to change it here. Perhaps they should be defined via a config or global repo ci var.

danceratopz: About config, I think we could start using the config classes to ensure that these variables are consistent a specified version of the repo (@raxhvl introduced these btw). Not easy to use in bash scripts though :sob: 
https://github.com/ethereum/execution-spec-tests/blob/5865d02c55c0835a7bcc07c6785dc799f303a734/src/config/app.py#L15-L32

.github/workflows/coverage.yaml
```diff
@@ -117,189 +120,27 @@ jobs:
...
-              fi
-
-              if [ $file_found -eq 0 ]; then
-                echo "Error: Failed to find the test file $file in test repo"
-                exit 1
-              fi
-          done
+          ./.github/scripts/parse_ported_tests.sh "$CHANGED_TEST_FILES" "${{ github.workspace }}" || true
 
       # This command diffs the .py scripts introduced by a PR
```
danceratopz, line 125:
```suggestion
      # This command checks and fills python test sources introduced by a PR
```

danceratopz (approved):
Thanks for doing this, makes it so much easier to understand! LGTM!

But, we have to be careful to not depend on evmone too heavily going forward, unless it finds a maintainer :cry: Post merge, we should be able to use EELS directly much more easily for coverage. So I would recommend not optimizing this evmone-based flow any more after this PR.

About the suggestions... Solid PR, so these are just ideas...

- Don't have a strong opinion on `jq`, but it does help readability/robustness. 

- Otherwise, small comment about the hard-coded block gas limit... feel free to ignore, but it's something we need to keep in mind and find a better way to manage. Follow-up issue?

- And some v small formatting fixes in `coverage_missed_reason` vals.

.github/scripts/check_changed_tests.sh
```diff
@@ -0,0 +1,28 @@
...
+CHANGED_TEST_FILES="$1"
+
+# Check if any tests changed
+if [ -z "$CHANGED_TEST_FILES" ] || [ "$CHANGED_TEST_FILES" == "[]" ]; then
+    echo "No python files were changed in ./tests/ - no action necessary"
+    exit 1
+fi
+
+# Convert JSON array to space-separated list
+files=$(echo "$CHANGED_TEST_FILES" | tr ',' '\n' | tr -d '[]"' | grep -v '^$')
```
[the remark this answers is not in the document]
marioevz: I think claude got me in this one, this file isn't really used because `"$CHANGED_TEST_FILES"` is always going to be empty if `steps.changed-tests.outputs.tests_any_changed != 'true'`, which is the condition to trigger the script, and makes zero sense.

I'll skip the jq comments because the output in `"$CHANGED_TEST_FILES"` is a comma separated string, and not json at all :/

.github/workflows/coverage.yaml
marioevz: > Just something to bear in mind.... I wonder if this will bite us at some point? I mean, if the default fill gas block limit is changed elsewhere, we might forget to change it here. Perhaps they should be defined via a config or global repo ci var.

I don't think this block gas limit should go up at the same rate as the benchmarking, since the coverage in a benchmark test that uses 1M gas vs one that uses 100M should be exactly the same (It's just 99M gas of the same operations).

.github/scripts/fill_prepatched_tests.sh
marioevz: All good, I took a deep dive into the https://github.com/tj-actions/changed-files?tab=readme-ov-file#outputs- docs, and also github workflow step outputs, so it's much clearer now IMO.

PR #1780, opened 2025-06-20
tests(benchmark): expand worst case modexp benchmarks

LouisTsai-Csie:
## 🗒️ Description

The test cases are divided into two parts: the first tests the upper bound of modexp based on [EIP-7823](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7823.md), and the second ports cases from Nethermind. For the latter one, I’ve added a reference to each individual case.

Script for converting the test case: https://gist.github.com/648207a8651d3319ea6dd06e9c977422.git

This PR ports the modexp test cases from Nethermind, you can find the corresponding [code](https://github.com/NethermindEth/nethermind/blob/ceb8d57b8530ce8181d7427c115ca593386909d6/tools/EngineRequestsGenerator/TestCases/ModexpVulnerability.cs#L122) here.

I wrote a script to generate test cases, this is my process:
1. I port all the test case from Nethermind's codebase to this text [file](https://gist.github.com/LouisTsai-Csie/e747930b5f2480643e78f1c3f9b9be1c).
2. Then I use this [script](https://gist.github.com/LouisTsai-Csie/d4762eb91359d02118c97003def1f042) to create the correct format we use in EEST, please put your script and the text file in the same folder and run the script: `python script.py`
3. It will create a file: `output.txt` and this is the [result](https://gist.github.com/LouisTsai-Csie/8d3a9b11f05067c85d7bf75f1e95aecd).

## 🔗 Related Issues

Issue https://github.com/ethereum/execution-spec-tests/issues/1453

## ✅ Checklist


marioevz:
> @kevaundray I’ve checked with the Nethermind team regarding the ported test. Could you help confirm whether we can change the ending from "00" to "02" for the even case?
> 
> Based on our previous discussion, using "00" as the ending might get stripped by `calldata = bytes(mod_exp_input).rstrip(b"\x00")`, which could cause issues.
> 
> Example:
> 
> https://github.com/ethereum/execution-spec-tests/blob/ead99d164f9acb647013d3ea4c88dbb32013d5dd/tests/zkevm/test_worst_compute.py#L509
> 
> https://github.com/ethereum/execution-spec-tests/blob/ead99d164f9acb647013d3ea4c88dbb32013d5dd/tests/zkevm/test_worst_compute.py#L815

I don't think this is a problem because modexp is supposed to right-pad the input with zeros if the sum of the lengths provided is greater than the calldata input.

It might be better to rollback this change IMO.

marioevz:
I could not make the conversion script work or rather I don't know what the inputs should be, could you share more info on the input data?

LouisTsai-Csie:
Thanks @marioevz, I've updated my script and the description, you could find the details in the PR description now.

PR #1781, opened 2025-06-20
feat(static): implement address tag resolution for static test fillers

### LINKED ISSUE #1750: Allow static tests to have dynamic addresses in `pre`
- Static tests are here: https://github.com/ethereum/execution-spec-tests/tree/main/tests/static, there are fillers in yaml and also in json (older tests)
- Issue is that these tests don't play well with the rest of the tests because they reuse almost always the same addresses, or deploy to addresses that are impossible to deploy on live networks, or some combination of these two.
- `pre` is what we want to change in most fillers, and looks mostly like this in almost all tests: https://github.com/ethereum/execution-spec-tests/blob/b8da4af651c2832099641e821c80b88b2b452c38/tests/static/state_tests/stBugs/staticcall_createfailsFiller.json#L30-L59
- Idea is to change all keys in that structure to tags that can be used in other fields in order for the static test filler to assign addresses during the fill process.
- There are mostly only two types of accounts: EOAs and contracts.
- We could separate the tags for each type as:
  - EOAs: "eoa_1", "eoa_2", ...
  - Contracts: Any other string
- Places where the tags could appear (not exhaustive):
  - Other contract source code
  - tx.to
  - expect.result
  - contract data, abi
- If a test is doing something more complex than this, e.g. introduces a circular dependency (e.g. "contract_1" has tag "contract_2" inside of its source code, and "contract_2" has tag "contract_1" inside of its source code), or is referencing a contract that is created during the tests, it might be worth it to simply update the test to python.
- Current code to parse all of this tests can be found here: https://github.com/ethereum/execution-spec-tests/blob/main/src/ethereum_test_specs/static_state/state_static.py

fselmo:
## 🗒️ Description

---

Note: [WIP] This is very much still a work in progress. A lot of tests are already working well but I'm putting up a PR to get the conversation going around it.

---

Hard-coded address conversion in `.yml` and `.json` static fillers:
  - Add `convert_addresses.py` script to automate tag conversion
  - The correct way to run this is with the ``CONVERT_COINBASE`` flag set to ``False`` as this allows the same coinbase for all tests (just as python tests do). If we decide we want to handle the coinbase setting on the python side, we can turn this flag on and hard-code on the python side... but the currect approach seems correct.
  - Convert static test YAML/JSON files to use address tags

(Python) Generate deterministic addresses from tags coming from static test fillers:
  - Resolve tags to deterministic addresses in the same way python tests do - via pytest static filler plugin
  - Add ``BlockchainEngineXFixture`` support for pre-allocation groups

This enables static tests to use symbolic address tags instead of hardcoded addresses, minimizing muddied context across tests when running via pre-alloc sharing.

## 🔗 Related Issues

closes #1750

## ✅ Checklist


src/pytest_plugins/filler/static_filler.py
```diff
@@ -206,7 +207,39 @@ def collect(self: "FillerFile") -> Generator["FillerTestItem", None, None]:
...
+                            if not (
+                                format_item is BlockchainEngineXFixture
+                                or (
+                                    isinstance(format_item, LabeledFixtureFormat)
+                                    and format_item.format is BlockchainEngineXFixture
+                                )
+                            )
+                        ]
+
+                    for format_with_or_without_label in filtered_formats:
```
danceratopz, line 242: I think this was copied verbatim from `filler.py`. Perhaps we can add a helper function, but perhaps it's not worth it?

danceratopz (commented):
Excited about this! Just a few minor comments after a fleeting review. I can try it out and have a closer look tomorrow!

fselmo:
I've managed to fine tune quite a few fillers that were not passing the tagging system. There are still edge cases I think can be cleaned up and I think some of the `stTransactionTest` cases I can still fine tune. All tests are passing if ran with `-k "not stTransactionTest"` as I still think I can turn more of those on but this is ready for another set of eyes. Commits will be cleaned up as well. These are the stats on tagged vs not tagged, excluding `/stTransactionTest/` (so for all passing tests):

```
Total filler files in tests/static (excluding stTransactionTest): 2,444
   Successfully tagged files: 1,751 (71.6%)
   Untagged files: 693 (28.4%)
```

My hope is there are more tests that can be turned on as well for tagging but examining all the tests and figuring out general approaches was my first attempt and I wanted to get this to a good first pass since it has taken up a lot of dev time.

The list of currently incompatible tests is [here](https://github.com/ethereum/execution-spec-tests/pull/1781/files#diff-194b0e902b30bd863543c0a9c34a3f41377070eb4137e1b439319608778a9bd4R63-R329). This could be converted to a list in the documentation somewhere once it's finalized for this first pass.

convert_addresses.py
```diff
@@ -0,0 +1,475 @@
...
+from pathlib import Path
+
+# Known secret key that maps to sender address
+KNOWN_SECRET_KEY = '45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8'
+KNOWN_SENDER_ADDRESS = 'a94f5374fce5edbc8e2a8697c15331677e6ebf0b'
+
+CONVERT_COINBASE = False
+
+# Ethereum precompile addresses (0x01 through 0x11)
+PRECOMPILE_ADDRESSES = {
```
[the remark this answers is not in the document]
fselmo: This is a good approach I can try to add this instead 👍🏼

src/pytest_plugins/filler/static_filler.py
fselmo: I can look into DRYing this up before we're ready to merge 👍🏼

fselmo:
Hmm it looks like running with `-m "not_slow and tagged" --generate-pre-alloc-groups -n 20` has a few failures, a handful of them being read timeouts and I'm not sure if it's a domino effect or what is happening there. Will need to investigate.

marioevz (approved):
Amazing work! 🎉

This paves the way to executing every single Ethereum test in history in less than an hour compared to tens of hours!

PR #1800, opened 2025-06-24
test(benchmark): add coverage for CALLDATACOPY, CODECOPY, RETURNDATACOPY and MCOPY

jsign:
## 🗒️ Description
This PR adds benchmark coverage for `CALLDATACOPY`, `CODECOPY`, `RETURNDATACOPY` and `MCOPY`.

Cycles:
```
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-1MiB-transaction]-1        125092548
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-1MiB-transaction]-1       126159065
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-100KiB-transaction]-1      134499078
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-100KiB-transaction]-1     137217303
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-1MiB-call]-1        140491870
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-10KiB-transaction]-1       150131958
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-1MiB-transaction]-1 158964435
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-1MiB-transaction]-1        159926649
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-10KiB-transaction]-1      175553835
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-1MiB-call]-1       250364532
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_True-1MiB]-1  255470228
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-1MiB-call]-1       256252199
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-100KiB-call]-1      274666727
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-100KiB-transaction]-1       279043357
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-100KiB-call]-1     285408182
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-1KiB-transaction]-1        288423058
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_True-100KiB]-1        289997797
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-max code size]-1   295109348
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-0.75x max code size]-1     297259608
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-10KiB-call]-1       298718943
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-10KiB-call]-1      298940375
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-0.50x max code size]-1     301460118
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-10KiB-transaction]-1        303028063
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_True-10KiB]-1 303997756
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-0.25x max code size]-1     313942765
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-1MiB-call]-1      392174602
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-1KiB-call]-1       429529132
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_False-1MiB]-1 434913651
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-1KiB-call]-1        438984582
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-100KiB-call]-1    449030956
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-max code size]-1  468315261
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-100KiB-transaction]-1      470248705
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-0.75x max code size]-1    473178700
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-1KiB-transaction]-1       477810980
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-10KiB-call]-1      480022603
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-10KiB-call]-1     480925492
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-0.50x max code size]-1    482887507
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-10KiB-transaction]-1       487246263
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_False-100KiB]-1       495723508
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-0.25x max code size]-1    512144671
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_False-10KiB]-1        518387302
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-1MiB]-1       541304303
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-100KiB-call]-1     546194847
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-100KiB]-1     577292038
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-10KiB]-1      587724282
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-1KiB]-1       684513499
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-1KiB-transaction]-1 696306261
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-1MiB]-1      696523057
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_True-1KiB]-1  697928414
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-1KiB-call]-1      728230622
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_False-1KiB]-1 738653325
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-1KiB-call]-1       738950780
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-100KiB]-1    743235890
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-10KiB]-1     767542764
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-100 bytes-transaction]-1   816395697
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-1KiB-transaction]-1        846794662
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-100 bytes-call]-1  903973263
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-1KiB]-1      956820255
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-100 bytes-call]-1   969365781
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-100 bytes]-1  994860035
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-100 bytes-transaction]-1    1050445283
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_True-100 bytes]-1     1072272349
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-0 bytes]-1    1167061696
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-0 bytes]-1 1173093333
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-10 bytes-transaction]-1    1202497630
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-100 bytes-transaction]-1  1229823144
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_False-100 bytes]-1    1265486319
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_True-10 bytes]-1   1288792240
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-10 bytes-call]-1   1301693268
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-0 bytes-call]-1    1314321612
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_True-0 bytes-transaction]-1     1318847738
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-100 bytes-call]-1 1321986838
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-10 bytes-transaction]-1     1370503008
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-100 bytes-call]-1  1373916005
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-100 bytes]-1 1379633636
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-100 bytes-transaction]-1   1421850691
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_True-10 bytes]-1      1431026421
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_True-0 bytes]-1       1432705603
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_True-10 bytes-call]-1    1437577686
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-10 bytes-transaction]-1   1557798027
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-0 bytes]-1   1582416203
tests/zkevm/test_worst_memory.py::test_worst_codecopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-0 bytes]-1        1585211480
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-10 bytes-call]-1  1590621573
tests/zkevm/test_worst_memory.py::test_worst_mcopy[fork_Cancun-blockchain_test_from_state_test-fixed_src_dst_False-10 bytes]-1  1599488993
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_False-10 bytes]-1     1604711952
tests/zkevm/test_worst_memory.py::test_worst_returndatacopy[fork_Cancun-blockchain_test_from_state_test-fixed_dst_False-0 bytes]-1      1640962382
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-0 bytes-call]-1   1646835084
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_False-fixed_src_dst_False-0 bytes-transaction]-1    1665837321
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-10 bytes-call]-1   1668885796
tests/zkevm/test_worst_memory.py::test_worst_calldatacopy[fork_Cancun-blockchain_test_from_state_test-non_zero_data_True-fixed_src_dst_False-10 bytes-transaction]-1    1684366602
```

## 🔗 Related Issues
https://github.com/ethereum/execution-spec-tests/issues/1690

## ✅ Checklist

- [X] All: Set appropriate labels for the changes.


tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+@pytest.mark.valid_from("Cancun")
+@pytest.mark.parametrize(
+    "origin",
+    [
+        pytest.param(CallDataOrigin.TRANSACTION, id="transaction"),
+        pytest.param(CallDataOrigin.CALL, id="call"),
+    ],
+)
```
jsign, line 37: Cover both cases of `CALLDATACOPY` reading from tx data or CALL.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,263 @@
...
+    fixed_dst: bool,
+):
+    """Test running a block filled with CALLDATACOPY executions."""
+    env = Environment()
+    max_code_size = fork.max_code_size()
+
+    # We create the contract that will be doing the CALLDATACOPY multiple times.
+    jumpdest = Op.JUMPDEST
+    jump_back = Op.JUMP(Op.PUSH0)
+    dst = 0 if fixed_dst else Op.MOD(Op.GAS, 7)
```
jsign, line 66: The `fixed_dst` parameter is an idea used in this file's new tests.

The goal is to cover two cases regarding the destination offset of the memory operation:
- Fixed: in this case, the dest offset is zero.
- Non-fixed: I "rotate" the destination modulo 7 (which is a prime number so that's nice to avoid fixed cycles), leveraging `GAS` as a random-ish source.

I wanted to include this "Non-fixed" case since always copying to the same destination was maybe too simple. I do this "destination rotation" in a way that I think usually shouldn't be that costly nor produce too much "memory expansion overhead".

Interestingly, the "Non-fixed" destination cases looks like make zkVMs work harder by a significant amount.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,263 @@
...
+    iter_block = Op.CALLDATACOPY(dst, Op.PUSH0, Op.CALLDATASIZE)
+    max_iters_loop = (max_code_size - len(jumpdest) - len(jump_back)) // len(iter_block)
+    code = jumpdest + sum([iter_block] * max_iters_loop) + jump_back
+    if len(code) > max_code_size:
+        raise ValueError(f"Code size {len(code)} exceeds maximum code size {max_code_size}")
+    code_address = pre.deploy_contract(code=code)
+
+    tx_target = code_address
```
jsign, line 74: Regarding `origin`:
- If `CallDataOrigin.Transaction` we call the contract directly
- Otherwise, we use a proxy contract to call it.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,263 @@
...
+        # target contract with the calldata.
+        code = Op.CALLDATACOPY(Op.PUSH0, Op.PUSH0, Op.CALLDATASIZE) + Op.STATICCALL(
+            address=code_address, ret_size=Op.CALLDATASIZE
+        )
+        tx_target = pre.deploy_contract(code=code)
+
+    tx = Transaction(
+        to=tx_target,
+        gas_limit=env.gas_limit,
+        data=b"\xff" * size,
```
jsign, line 86: We could also use `0x00`, but I preferred a non-zero value. 
We could also parameterize it in another dimension, but a fourth dimension might end up blowing up stuff, maybe too much. Open to feedback if anyone has a strong opinion on including it.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,263 @@
...
+@pytest.mark.parametrize(
+    "max_code_size_ratio",
+    [
+        pytest.param(0, id="0 bytes"),
+        pytest.param(0.25, id="0.25x max code size"),
+        pytest.param(0.50, id="0.50x max code size"),
+        pytest.param(0.75, id="0.75x max code size"),
+        pytest.param(1.00, id="max code size"),
+    ],
+)
```
jsign, line 108: For `CODECOPY`, I wanted to do a range scan, but was tied to the `max_contract_size`. But considering it is a runtime value, I came up with this strategy of defining the cases as a ratio, so we cover: 25%, 50%, 75% and 100% of the max contract size.

This provides a good understanding of the measurements for all the relevant ranges and auto-adjusts to `--block-gas-limit` configurations.

LMK if anyone knows a better way to do it!

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,263 @@
...
+    pre: Alloc,
+    fork: Fork,
+    max_code_size_ratio: float,
+    fixed_dst: bool,
+):
+    """Test running a block filled with CODECOPY executions."""
+    env = Environment()
+    max_code_size = fork.max_code_size()
+
+    size = int(max_code_size * max_code_size_ratio)
```
jsign, line 121: Note that the "attacker" contract we build is always `max_code_size`, so there's no need to do anything extra to support any `size` target.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,262 @@
...
+    code = (
+        Op.MSTORE8(0, Op.GAS)
+        + Op.MSTORE8(size // 2, Op.GAS)
+        + Op.MSTORE8(size - 1, Op.GAS)
+        + Op.RETURN(0, size)
+    )
+    helper_contract = pre.deploy_contract(code=code)
+
+    # We create the contract that will be doing the RETURNDATACOPY multiple times.
+    code_prefix = Op.STATICCALL(address=helper_contract) if size > 0 else Bytecode()
```
jsign, line 187: For `size == 0` let's avoid the extra overhead, plus the above `MSTORE8`-3-byte-strategy doesn't make sense either.

tests/zkevm/test_worst_memory.py
LouisTsai-Csie: `0xff` works perfect here, but it costs 16 gas per byte, while `0x00` only costs 4. My major consideration is that using `0x00` would allow us to run more opcodes within the same gas limit.

jsign: Oh yeah, I was aware that 0x00 was cheaper :smile: . My main concern was blowing up the number of generated combinations. @marioevz do you have any stance on this?

tests/zkevm/test_worst_memory.py
jsign: Yep, I think I can use a similar strategy if a non-fixed source sounds interesting. I'll wait a day to see if there's any pushback from other reviewers (mainly reg blowup variations), and if that isn't the case I'll do it for all tests.

LouisTsai-Csie: Cool, I will also add it in the LOG opcode benchmark test

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,235 @@
...
+    attack_block = Op.CALLDATACOPY(dst, Op.PUSH0, Op.CALLDATASIZE)
+    code = code_loop_precompile_call(Bytecode(), attack_block, fork)
+    code_address = pre.deploy_contract(code=code)
+
+    tx_target = code_address
+    if origin == CallDataOrigin.CALL:
+        # If the origin is CALL, we need to create a contract that will call the
+        # target contract with the calldata.
+        code = Op.CALLDATACOPY(Op.PUSH0, Op.PUSH0, Op.CALLDATASIZE) + Op.STATICCALL(
+            address=code_address, ret_size=Op.CALLDATASIZE
```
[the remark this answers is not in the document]
jsign: > This `STATICCALL`s with no data in it. Except, it will allocate the return buffer of this frame to contain `Op.CALLDATASIZE` bytes. I believe we want this however:
> 
> This will instead forward the tx calldata to the STATICCALL call frame.

Yes, sorry. This was me mistyping the argument name.

tests/zkevm/test_worst_memory.py
jsign: Fair trick, but this relies on the assumption of the initcode and max code size relationship, which I'm not sure will always hold.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,235 @@
...
+    """Test running a block filled with CODECOPY executions."""
+    env = Environment()
+    max_code_size = fork.max_code_size()
+
+    size = int(max_code_size * max_code_size_ratio)
+
+    code_prefix = Op.PUSH32(size)
+    dst = 0 if fixed_dst else Op.MOD(Op.GAS, 7)
+    attack_block = Op.CODECOPY(dst, Op.PUSH0, Op.DUP1)
+    code = code_loop_precompile_call(code_prefix, attack_block, fork)
```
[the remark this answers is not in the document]
jsign: I'm tempted to do the renaming in this PR but I prefer to add it in https://github.com/ethereum/execution-spec-tests/issues/1704. I can do a PR right after. 

I prefer any change doing wide renamings to be short-lived to avoid conflicts with other PRs. I can do that right after we merge this one.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,235 @@
...
+
+    tx = Transaction(
+        to=pre.deploy_contract(code=code),
```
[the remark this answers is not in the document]
jsign: Note that we don't use `CODESIZE` anywhere in this test. But your suggestion is still helpful if, for some reason, we want to avoid zero-bytes for the missing bytes from the attack-loop construction. Not sure if it makes a difference, but it feels better to do it since it sounds more explicit.

tests/zkevm/test_worst_memory.py
jsign: OK, I'll add it and revert if there's any pushback after :)

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,235 @@
...
+    origin: CallDataOrigin,
+    size: int,
+    fixed_dst: bool,
+):
+    """Test running a block filled with CALLDATACOPY executions."""
+    env = Environment()
+
+    # We create the contract that will be doing the CALLDATACOPY multiple times.
+    dst = 0 if fixed_dst else Op.MOD(Op.GAS, 7)
+    attack_block = Op.CALLDATACOPY(dst, Op.PUSH0, Op.CALLDATASIZE)
```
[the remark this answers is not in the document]
jsign: For the sake of making progress, I'll go with the `dst_src_fixed` approach to bundle non-fixed in both dimensions. At the end of the day we want to have enough signal on worst-case scenarios, so I think that would be enough to surface it.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,235 @@
...
+    size: int,
+    fixed_dst: bool,
+):
+    """Test running a block filled with MCOPY executions."""
+    env = Environment()
+
+    mem_init = Op.MSTORE8(0, Op.GAS) + Op.MSTORE8(size // 2, Op.GAS) + Op.MSTORE8(size - 1, Op.GAS)
+    code_prefix = mem_init if size > 0 else Bytecode()
+    dst = 0 if fixed_dst else Op.MOD(Op.GAS, 7)
+    attack_block = Op.MCOPY(dst, Op.PUSH0, size)
```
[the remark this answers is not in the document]
jsign: > I had to give this some thought because I was assuming you would overwrite the bytes at size // 2 and size - 1 with zeros if dst is nonzero. But I actually think this is not the case (but we should verify with a memory trace).

I'll debug the test to double check. If I find this happened we can think of some workaround.

> This is then actually a very fun test because this alters the memory each time (given dst changes) it is called 😄 👍

Yep, makes a mess with memory which is cooled :smile:

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,235 @@
...
+    fork: Fork,
+    origin: CallDataOrigin,
+    size: int,
+    fixed_dst: bool,
+):
+    """Test running a block filled with CALLDATACOPY executions."""
+    env = Environment()
+
+    # We create the contract that will be doing the CALLDATACOPY multiple times.
+    dst = 0 if fixed_dst else Op.MOD(Op.GAS, 7)
```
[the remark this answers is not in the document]
jochem-brouwer: I like the assertion approach. If we get EELS merged in this repo then we likely also get some helper methods to determine gas cost (or it would be possible to make) so we could write these assertions in such way that they also don't use "magic numbers" (one has to assume that the copied gas cost is correct, e.g. opcode X costs 3 gas. And this could also differ per fork)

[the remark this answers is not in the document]
jsign: @jochem-brouwer I realized that the `assert` strategy is more complex than I imagined. The reason is quite simple: the gas used is dynamic since depends on parameters. :/

Reconsidering the double `MOD` again ha

tests/zkevm/test_worst_memory.py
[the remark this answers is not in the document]
jsign: Note that the "treating as a blackbox" should not only apply for zkEVMs but for ELs, since these benchmarks should be meaningful for those.

Even for EL clients you shouldn't assume you know how they work, since maybe the way EthereumJS, Geth, and Reth works is different reg boundaries too, so you can't assume anything :)

I feel quite hard sometimes to think if these slight combinational differences are worth it, since I don't want to have 1 million fixtures to prove where only 50% might actually make any real difference.

So.. no strong conclusion here, but if we are strict about testing all combinations I think probably all the tests are somewhat incomplete since we always have some kind of assumption that might be wrong since could affect zkEVMs or ELs.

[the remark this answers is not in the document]
jsign: I'll try to make a non-combined variant, see cycles and check if I see any difference. If that isn't the case, I'll keep them bundled to avoid unnecessary blowup. If I see a difference, I'll separate them.

tests/zkevm/test_worst_memory.py
```diff
@@ -49,37 +50,71 @@ class CallDataOrigin:
...
     ],
 )
-@pytest.mark.parametrize("fixed_dst", [True, False])
+@pytest.mark.parametrize(
+    "fixed_src_dst",
+    [
+        True,
+        False,
+    ],
+)
```
jsign, line 59: I made tha non-fixed variant imply non-fixed both in src and dst at the same time. While it can be debatable that might worth separating in dimensions, I think this provides good coverage without blowing up this (and all other tests below) by 2x.

I'm starting to get worried about being able to prove all the benchmark tests in zkVMs already. I prefer to leave it this way until I can do a full run of the whole benchmark suite doing real proving (not only counting cycles), and we can see if there's variability in some cases to justify more fine grained explorations.

For you to have an idea: the fastest block real proving time in our prover is around 4m. But is normal that many blocks take 15m, 30m or in worst cases hours.

tests/zkevm/test_worst_memory.py
```diff
@@ -49,37 +50,71 @@ class CallDataOrigin:
...
+@pytest.mark.parametrize(
+    "non_zero_data",
+    [
+        True,
+        False,
+    ],
+)
```
jsign, line 66: Yeah, more blowup but I think this was more justified (trust me bro! :smile:)

tests/zkevm/test_worst_memory.py
```diff
@@ -49,37 +50,71 @@ class CallDataOrigin:
...
     size: int,
-    fixed_dst: bool,
+    fixed_src_dst: bool,
+    non_zero_data: bool,
 ):
     """Test running a block filled with CALLDATACOPY executions."""
     env = Environment()
 
+    if size == 0 and non_zero_data:
+        pytest.skip("Non-zero data with size 0 is not applicable.")
```
jsign, line 80: Trying to cut combinations that don't add any info.

tests/zkevm/test_worst_memory.py
```diff
@@ -49,37 +50,71 @@ class CallDataOrigin:
...
-    code = code_loop_precompile_call(Bytecode(), attack_block, fork)
+    #
+    # If `non_zero_data` is True, we leverage CALLDATASIZE for the copy length. Otherwise, since we
+    # don't send zero data explicitly via calldata, PUSH the target size and use DUP1 to copy it.
+    prefix = Bytecode() if non_zero_data or size == 0 else Op.PUSH32(size)
+    src_dst = 0 if fixed_src_dst else Op.MOD(Op.GAS, 7)
+    attack_block = Op.CALLDATACOPY(
+        src_dst, src_dst, Op.CALLDATASIZE if non_zero_data or size == 0 else Op.DUP1
+    )
+    code = code_loop_precompile_call(prefix, attack_block, fork)
```
jsign, line 91: Two things:
- `non_zero_data` now means being more careful about assuming `CALLDATASIZE` is the right "size" value.
- For `size==0` is a special^2 case (since even for `not non_zero_data` it doesn't make sense to do `Op.PUSH32(size)` thus we can leverage `CALLDATASIZE` (which is 0) -- hope that isn't confusing and maybe I'm squeezing too much ha)

tests/zkevm/test_worst_memory.py
```diff
@@ -49,37 +50,71 @@ class CallDataOrigin:
...
+    # If `non_zero_data` is True, we fill the calldata with deterministic random data.
+    # Note that if `size == 0` and `non_zero_data` is a skipped case.
+    data = Bytes([i % 256 for i in range(size)]) if non_zero_data else Bytes()
+
     tx = Transaction(
         to=tx_target,
         gas_limit=env.gas_limit,
-        data=b"\xff" * size,
+        data=data,
         sender=pre.fund_eoa(),
```
jsign, line 118: Suggestion from @jochem-brouwer

tests/zkevm/test_worst_memory.py
```diff
@@ -117,14 +158,21 @@ def test_worst_codecopy(
...
+    attack_block = Op.CODECOPY(src_dst, src_dst, Op.DUP1)  # DUP1 copies size.
     code = code_loop_precompile_call(code_prefix, attack_block, fork)
 
+    # The code generated above is not guaranteed to be of max_code_size, so we pad it since
+    # a test parameter targets CODECOPYing a contract with max code size. Padded bytecode values
+    # are not relevant.
+    code = code + Op.INVALID * (max_code_size - len(code))
+    assert len(code) == max_code_size, (
+        f"Code size {len(code)} is not equal to max code size {max_code_size}."
+    )
```
jsign, line 171: Suggestion by @jochem-brouwer

tests/zkevm/test_worst_memory.py
```diff
@@ -117,14 +158,21 @@ def test_worst_codecopy(
...
+
     tx = Transaction(
         to=pre.deploy_contract(code=code),
         gas_limit=env.gas_limit,
-        data=b"\xff" * size,
```
jsign: This was indeed not necessary.

tests/zkevm/test_worst_memory.py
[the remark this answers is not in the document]
jochem-brouwer: I'm not sure I follow, the `CALLDATACOPY` size is constant in this test. There is indeed dynamic pricing for the amount of bytes we copy, but this is thus constant and thus static in this context. 
Note that if you read out of bounds, you still "copy" these bytes and EVM just reads/copies it as zero bytes. So you won't get a "refund" of the to-be-copied bytes or something. (Also not sure what word-alignment is relevant here :sweat_smile: Don't completely follow :smile: )

tests/zkevm/test_worst_memory.py
```diff
@@ -149,7 +197,13 @@ def test_worst_codecopy(
...
     ],
 )
-@pytest.mark.parametrize("fixed_dst", [True, False])
+@pytest.mark.parametrize(
+    "fixed_dst",
+    [
+        True,
+        False,
+    ],
+)
```
jsign, line 206: Note that I did the `fixed_src_dst` change, but I later realized when double-checking runs that was incorrect. `RETURNDATACOPY` [**does not allow** being out of bounds](https://github.com/ethereum/go-ethereum/blob/v1.15.11/core/vm/instructions.go#L334-L336).

So basically I decided to keep the source at offset 0. I _could_ have made the "size" to compensate for the offset, but it felt to me this was against what the `size` parameter means since it wouldn't be true that we're doing a `RETURNDATACOPY` for `size` bytes.

tests/zkevm/test_worst_memory.py
```diff
@@ -172,10 +227,36 @@ def test_worst_returndatacopy(
...
+    code = (
+        returndata_gen
+        + jumpdest
+        + sum([attack_iter] * max_iters_loop)
+        + returndata_gen
+        + jump_back
+    )
+    assert len(code) <= max_code_size, (
+        f"Code size {len(code)} is not equal to max code size {max_code_size}."
+    )
```
jsign, line 259: This was a suggestiong by @jochem-brouwer. After each loop we get some fresh returned data since GAS changed, so sounds like a nice thing to do that gets amortized.

tests/zkevm/test_worst_memory.py
```diff
@@ -204,22 +285,41 @@ def test_worst_returndatacopy(
...
+
+    jumpdest = Op.JUMPDEST
+    jump_back = Op.JUMP(len(mem_touch))
+    max_iters_loop = (max_code_size - 2 * len(mem_touch) - len(jumpdest) - len(jump_back)) // len(
+        attack_block
+    )
+    code = mem_touch + jumpdest + sum([attack_block] * max_iters_loop) + mem_touch + jump_back
+    assert len(code) <= max_code_size, (
+        f"Code size {len(code)} is not equal to max code size {max_code_size}."
+    )
```
jsign, line 322: @jochem-brouwer, when double checking the question of "are offsets 0, size/2 and size-1" erased... the answer was not for many iterations, but actually any zero-byte in memory eventually ends up slowly erasing other memory :smile: 

So after so many iterations non-zero bytes count becomes smaller and smaller up to some points where the whole memory gets erased.

I started experimenting with many things and later realized I could do the same idea of the RETURNDATACOPY and inject non-zero bytes at the end of this loop. So I avoid this long-lasting effect of the memory being completely erased.

I checked the memory when running these tests and this is good enough to never leave with zeroed-memory because of the effect I described before -- and I think this is still "cheap enough" to not interfere with the test main intention.

I even try doing `CODECOPY` instead of these `MSTORE8` as a way to put more random data in memory, but even a small amount of 0x00 bytecodes end up erasing stuff in the long term. Quite funny, but makes sense in hindsight!

tests/zkevm/test_worst_memory.py
jochem-brouwer: Oh wow, yes good point, I had not realized that the return buffer is treated differently :sweat_smile:

tests/zkevm/test_worst_memory.py
jochem-brouwer: > but even a small amount of 0x00 bytecodes end up erasing stuff in the long term. Quite funny, but makes sense in hindsight!

Aha, so we need to generate code without zeros in it :wink: 

Ok interesting regarding this memory erasure... I'm trying to visualize this. Is it because if dst=0 and src=nonzero else then if the value at key 0 is nonzero, then it will be overwritten by zero if the src key is zero. I guess it indeed makes sense that in the end everything gets overwritten by zeros. But I guess it also depends on the initial state of the memory how quickly it is erased :thinking: (yes just "thinking out loud" here :sweat_smile: )

I think the `mem_touch` is a great way to solve this problem. Although, the loop is quite long, so I would not be surprised that halfway the memory is already cleared? Not sure :thinking: This is a nice puzzle itself to think about on how to slow down this memory clearing or to figure out if there is an easy way to "fix" this with some cool trick :thinking:

tests/zkevm/test_worst_memory.py
jochem-brouwer: Hmm this looks confusing, because there is now a stack imbalance at the prefix. That either initializes the stack with 0 or 1 items depending on the parameters. AH! Wait, I see now, for `non_zero_data or size == 0` items we use Op.CALLDATASIZE which thus does not have the Op.DUP1 stack requirement of having an item.

For the 0 case for clarity could also use PUSH0 for the `size==0` case. 

We should take a look if there is a builtin method to pick the right PUSH here (since PUSH32 seems overkill). If you would just use `size` instead of `Op.PUSH32(size)` would it then not pick the "smallest" PUSH?

tests/zkevm/test_worst_memory.py
jochem-brouwer: Interesting results! This makes me curious what the difference between those blocks are (because a 15x diff in time is very big). While I'm still treating this all as a black box this also makes me excited to start studying this a bit more.

tests/zkevm/test_worst_memory.py
```diff
@@ -0,0 +1,335 @@
...
+        )
+        arg_size = Op.CALLDATASIZE if non_zero_data else size
+        code = prefix + Op.STATICCALL(
+            address=code_address, args_offset=Op.PUSH0, args_size=arg_size
+        )
+        tx_target = pre.deploy_contract(code=code)
+
+    # If `non_zero_data` is True, we fill the calldata with deterministic random data.
+    # Note that if `size == 0` and `non_zero_data` is a skipped case.
+    data = Bytes([i % 256 for i in range(size)]) if non_zero_data else Bytes()
```
jochem-brouwer, line 112: I was about to comment that `non_zero_data=False` always yields empty bytes, but this is handled by `size` so we dont "waste" gas on calldata. I like it! :smile: :+1:

jochem-brouwer (approved):
I went through all comments and (quickly) went over the file again - LGTM! Exciting tests juggling around with memory bytes :star_struck:

tests/zkevm/test_worst_memory.py
jsign: I'll share soon with you some preliminary results about these tests run in provers :)

tests/zkevm/test_worst_memory.py
jsign: Changed to `PUSH24` under the logic of a `size` of 10MiB (10x than the current max one, just to be safe):
image

jochem-brouwer: Ah no, I meant "just use the size" as in remove the `Op.PUSH32(size)` and just keep `size`. I think the test formatter will pick the appropriate PUSH value.

For 10 MiB you can actually store it in PUSH3 (24 bits) :wink:

jsign: @jochem-brouwer , you mean doing this?
`prefix = Bytecode() if non_zero_data or size == 0 else size`
I don't think that would work.

I think ideally we should have a way to use [this private func](https://github.com/ethereum/execution-spec-tests/blob/jsign-mem-opcodes/src/ethereum_test_vm/opcode.py#L33).

tests/benchmark/test_worst_memory.py
```diff
@@ -0,0 +1,336 @@
...
+@pytest.mark.parametrize(
+    "size",
+    [
+        pytest.param(0, id="0 bytes"),
+        pytest.param(10, id="10 bytes"),
+        pytest.param(100, id="100 bytes"),
+        pytest.param(1 * 1024, id="1KiB"),
+        pytest.param(10 * 1024, id="10KiB"),
+        pytest.param(100 * 1024, id="100KiB"),
+        pytest.param(1024 * 1024, id="1MiB"),
```
jsign, line 51: @marioevz, I saw your comment in another PR that you might prefer to reduce these range. Is that the case also for this PR?

Maybe since this is a more memory-focused PR instead of 0 and 1MiB we can reduce to 4 cases (e.g., 0, 100, 10KiB and 1MiB) or three (0, 1KiB, 1MiB).

LMK.

jsign:
> #1804 got merged! Please rebase the branch and move the tests to the appropriate folder.

Rebased!

tests/benchmark/test_worst_memory.py
marioevz: I think 4 sounds like a good sweet-spot!

jsign: :+1:

Issue #1778, opened 2025-06-19
Include all test formats in benchmark releases


marioevz:
Currently, the benchmark releases only include the new Blockchain EngineX (prev Engine ReOrg) format, but this format is not easy to consume to zk provers.

We need to fill these releases including state and blockchain test formats too.

Cc @jsign

jsign:
Including some extra explanation on how benchmark tests are consumed for zkEVM benchmarking:
- We use a tool called [zkevm-benchmark-workload](https://github.com/eth-act/zkevm-benchmark-workload) for the benchmarking
- This tool allows zkEVM to prove blocks coming from two main sources: RPC (if we want to benchmark mainnet/testnet/devnet chains) or tests (for EEST tests/benchmarks).
- In the EEST case, [we have a script](https://github.com/eth-act/zkevm-benchmark-workload/blob/master/scripts/download-and-extract-fixtures.sh) that automatically downloads the latest release. (After this issue is resolved, we have to tune a bit some stuff there)
- After the release asset is downloaded (i.e., test fixtures), then we:
  - Use [reth to deserialize each fixture into blockchain state tests](https://github.com/eth-act/zkevm-benchmark-workload/blob/master/crates/witness-generator/src/generate_stateless_witness.rs#L55-L62).
  - That `BlockchainTestCase::load(...) ` is basically a JSON deserialization of [this model](https://github.com/paradigmxyz/reth/blob/ea5ffa51fc11021289ecf656f77001d95eb9b53e/testing/ef-tests/src/models.rs#L15-L36).
  - After we have this deserialized, we [run the test with a fork of reth](https://github.com/kevaundray/reth/blob/kw/zkevm-benchmark-workload-repo/testing/ef-tests/src/cases/blockchain_test.rs#L174-L188), which also returns the `ExecutionWitness` which is required for proving.

After this, we have a `Block` and the corresponding `ExecutionWitness`, which contains all the pre-state, bytecodes, and MPT proof. With this information, we can statelessly run the block in a RISCV reth program that zkEVM proves.

The relevant part is that we try to leverage existing code in reth to deserialize blockchain tests, so we can generate the execution witness for proving without reinventing the wheel. The blockchain test format is very nice since each benchmark fixture is self-contained.

chfast:
I'd very much appreciate this so they can be used e.g. as `evm statetest --bench test.json`.

danceratopz:
- #1855
[...]

PR #1804, opened 2025-06-25
refactor(benchmark): rename `zkevm` tests to `benchmark`

spencer-tb:
## 🗒️ Description
Renames `tests/zkevm` to `tests/benchmark`. Just to get the process started.

Additionally renames the releases & markers to benchmark.

## 🔗 Related Issues
This change will affect all open zkevm/benchmark test PRs.

## ✅ Checklist


.github/configs/evm.yaml
```diff
@@ -16,6 +16,11 @@ zkevm:
   repo: ethereum/evmone
   ref: master
   targets: ["evmone-t8n", "evmone-eofparse"]
+benchmark:
+  impl: evmone
+  repo: ethereum/evmone
+  ref: master
+  targets: ["evmone-t8n", "evmone-eofparse"]
```
spencer-tb, line 23: We currently have multiple features that use the zkevm evm definition:
```
zkevm_36M:
  evm-type: zkevm
  fill-params: --from=Cancun --until=Prague --block-gas-limit 36000000 -m zkevm ./tests
  solc: 0.8.21
  feature_only: true
...
zkevm_120M:
  evm-type: zkevm
  fill-params: --from=Cancun --until=Prague --block-gas-limit 120000000 -m zkevm ./tests
  solc: 0.8.21
  feature_only: true
 ```
 
 Note these only create releases for tests that are marked with the `zkevm` marker.
 
 1) Are all tests within `tests/zkevm` (now `tests/benchmark`), marked with `zkevm` marker? @jsign 
 2) Should we add a list of benchmark release features like the above with `fill ... ./tests/benchmark` that contains all the benchmark tests? This allows us to have specific releases for `zkevm` but also benchmark. @marioevz
 3) Should we then replace the existing tox runs for `zkevm` marked tests with all of the benchmark tests?

spencer-tb: After some thought. I think we should do 3. And only run tox for all benchmark tests IFF those tests have changed. It feels ineffiecient and slow to run the existing zkevm tox for every PR in CI.

marioevz: I agree, we can blanket-rename all `zkevm` to `benchmark`, including markers. This would require a very small change in `tests/benchmark/conftest.py` but I think that would be it.

tests/benchmark/conftest.py
marioevz, line 1: We should rename the markers in this file.

marioevz (commented):
Looks good to me, I think we should go ahead and do the rename.

The fixture issue has an easy fix IMO: https://github.com/ethereum/execution-spec-tests/issues/1778#issuecomment-3005926242

We should also modify https://github.com/ethereum/execution-spec-tests/blob/main/.github/configs/feature.yaml in this same PR and rename all zkevm references to benchmark.

danceratopz (commented):
Thanks @spencer-tb. Would you mind taking care of the release naming (`zkevm` -> `benchmark` in `.github/configs/evm.yaml` and `.github/configs/evm.yaml` if appropriate?

spencer-tb:
> I feel like there's no way the [Evmone Coverage Report / evmone-coverage-diff (pull_request)](https://github.com/ethereum/execution-spec-tests/actions/runs/16005214478/job/45150116441?pr=1804) is going to complete successfully due to the amount of cpu intensive tests that are being touched.
> 
> The only way the coverage would complete is if it could intelligently determine that the fixtures are exactly the same and have just been renamed, and then skip those, but that's not yet a capability of the coverage script sadly.
> 
> If it fails IMO we should still merge.

We should consider adding the hasher logic to the coverage script for this!

(holds urge to YOLO merge :P)

PR #1810, opened 2025-06-27
feat(tooling): log RLP errors to file + shorten RLP errors in console

felix314159:
## 🗒️ Description
This change was originally part of #1614 but was split out into a separate PR. Rationale: When using real blobs logging the entire RLP on error into console is not feasible, so this PR shortens the logged error to 50 chars in these cases. Also logs entire RLP to file (only when error is encountered).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/src/ethereum_test_rpc/rpc.py
+++ b/src/ethereum_test_rpc/rpc.py
@@ -1,7 +1,9 @@
 """JSON-RPC methods and helper functions for EEST consume based hive simulators."""
 
+import logging
 import time
 from itertools import count
+from pathlib import Path
 from pprint import pprint
 from typing import Any, ClassVar, Dict, List, Literal, Union
 
@@ -11,6 +13,7 @@ from pydantic import ValidationError
 
 from ethereum_test_base_types import Address, Bytes, Hash, to_json
 from ethereum_test_types import Transaction
+from pytest_plugins.logging import get_logger
 
 from .types import (
     ForkchoiceState,
@@ -24,6 +27,7 @@ from .types import (
 )
 
 BlockNumberType = Union[int, Literal["latest", "earliest", "pending"]]
+logger = get_logger(__name__)
 
 
 class SendTransactionExceptionError(Exception):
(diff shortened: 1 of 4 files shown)


src/ethereum_test_rpc/rpc.py
```diff
@@ -43,7 +47,28 @@ def __str__(self):
         if self.tx is not None:
             f"{super().__str__()} Transaction={self.tx.model_dump_json()}"
         elif self.tx_rlp is not None:
-            return f"{super().__str__()} Transaction RLP={self.tx_rlp.hex()}"
+            tx_rlp_hex_full = self.tx_rlp.hex()
```
marioevz, line 50: We really should constrain this to only run when debugging is enabled. This would be an overhead that's dumping information to file system without the user having an option to disable it.

src/ethereum_test_rpc/rpc.py
```diff
@@ -43,7 +47,28 @@ def __str__(self):
...
+            # always log shortened errors to console (even when debugging read full from log file)
+            tx_rlp_hex = tx_rlp_hex_full[:50]
+
+            # create ./logs/rlp folder if it does not exist already
+            rlp_logs_folder = Path(".") / "logs" / "rlp"
```
marioevz, line 55: We have `AppConfig().DEFAULT_EVM_LOGS_DIR` instead of `Path(".")` but I'm not sure if that's the correct source for this location, so I'll defer to @danceratopz who is more familiar with the consume command logging.

marioevz (commented):
Some comments, but I think @danceratopz should give this one a review since he's more familiar with the logging mechanism of consume. Thanks!

src/ethereum_test_rpc/rpc.py
felix314159: I agree, but with Dan's custom logger I do not know how to retrieve the user-provided eest-log-level at runtime.

felix314159:
@danceratopz With your custom logger wrapper, is it possible to temporarily specify to log to file only (should take output filepath as argument)? I feel like this PR might introduces bad/unclean code because the custom logger does not provide the functionality we need here (or maybe it does and i just do not know how to use it). It should also be easy to at-runtime access the user provided value of the flag `eest-log-level`

Issue #1532, opened 2025-05-05
feat(consume): enable retries for RPC requests to reduce flakiness


danceratopz:
Use [tenacity](https://github.com/jd/tenacity) with an [exponential back-off to retry](https://tenacity.readthedocs.io/en/latest/api.html?highlight=exponential#tenacity.wait.wait_exponential) all RPC requests made to clients in EEST simulators (any retries should at least be logged; better tracked). This should reduce flakiness (false positives) in tests and greatly improve the experience with `--sim.parallelism` in Hive (`-n` in pytest).

Regarding `--sim.parallelism`, @somnathb1 writes:
> 1. I was able to run the entire consume-engine suite in 40 minutes with --sim.parallelism=64. This was mostly disk io limited, and i did it on a decent nvme. On a cheaper nvme, around ~2-5% fail on beyond --sim.parallelism=8.
>
> 2. Most of those errors with the cheaper SSD on erigon were around the runner/MockCL expecting "VALID" status on FCU and getting "SYNCING" status. In these cases the runner/MockCL should just re-try, as "SYNCING" is a spec-defined.

felix314159:
IMO we should do it like spencer and make our own simple retry mechanism instead of adding yet another dependency. As of now `tenacity` seems to be in our pyproject toml but is not used anywhere, so I think we should remove it if that is the case

danceratopz:
> IMO we should do it like spencer and make our own simple retry mechanism instead of adding yet another dependency.

This is a question of philosophy, here's my take.

https://github.com/jd/tenacity has become the de-factor library (7.9k stars) for known flaky code (e.g., network requests).

The current solution is inline in the simulator and not re-usable. Instead of implementing a library, I would rather outsource the effort (unit tests, documentation, useful extended features like exponential back-off, etc) to a well-maintained and understood project than make this effort in-house. Experienced python devs will know know this library and immediately be able use it/understand the context.

> As of now tenacity seems to be in our pyproject toml but is not used anywhere, so I think we should remove it if that is the case
Yes, unacceptable to have unused deps, thanks for taking care of this in #2019.

PR #1815, opened 2025-06-27
chore(consume): retry logic for initial consume engine fcu

spencer-tb:
## 🗒️ Description

Adds a simple exponential retry for the initial fcu within consume engine.

Using generic retry logic on all RPC calls were considered via #1532, however, this case specifically requires us keep calling the initial fcu until the client is no longer syncing. There may be cases in the future where we want to check syncing in other scenarios.

Following a comment from @danceratopz:
> Hive allows the simulator to define a "check live port" which is used to verify that the client has finished starting up: 
> https://github.com/ethereum/execution-spec-tests/blob/851f9f5bc88d25f869a7fcd7798b10b9c5bceb7e/src/pytest_plugins/consume/simulators/single_test_client.py#L34-L47
> For the engine simulator it's 8551, for example:
> https://github.com/ethereum/execution-spec-tests/blob/851f9f5bc88d25f869a7fcd7798b10b9c5bceb7e/src/pytest_plugins/consume/simulators/base.py#L24-L33

A clients engine port can be live after start up, but does not necessarily mean they are ready to accept all engine requests "properly". In the case for Erigon who prompted the latter, they have an occasional flakey test where their engine port is live, but the client needs some more time to finish syncing.

This retry logic should fix the issue with the Erigon flakey test issues, as now we wait for the client to finish syncing after start up. Note this won't be an issue in consume enginex or consume rlp.

## 🔗 Related Issues or PRs

#1532

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -45,6 +45,7 @@ Users can select any of the artifacts depending on their testing needs for their
 
 - 🔀 `consume` now automatically avoids GitHub API calls when using direct release URLs (better for CI environments), while release specifiers like `stable@latest` continue to use the API for version resolution ([#1788](https://github.com/ethereum/execution-spec-tests/pull/1788)).
 - 🔀 Refactor consume simulator architecture to use explicit pytest plugin structure with forward-looking architecture ([#1801](https://github.com/ethereum/execution-spec-tests/pull/1801)).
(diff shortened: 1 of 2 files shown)


marioevz (approved):
LGTM, thanks!

PR #1820, opened 2025-06-30
feat(benchmark): add empty payload test

LouisTsai-Csie:
## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```


marioevz:
#1804 got merged! Please rebase the branch and move the tests to the appropriate folder.

tests/benchmark/test_worst_blocks.py
```diff
@@ -156,3 +157,35 @@ def test_block_full_of_ether_transfers(
         blocks=[Block(txs=txs)],
         exclude_full_post_state_in_output=True,
     )
+
+
+def calldata_cost_per_byte(byte_data: bytes):
+    """Calculate the cost of calldata per byte."""
+    return 4 if byte_data == b"\x00" else 16
```
jsign, line 164: I think this isn't https://eips.ethereum.org/EIPS/eip-7623 aware shipped in Pectra? Prob the method should be renamed to `heavy_data_tx_calldata_cost_per_byte` or similarish.

jsign (changes requested):

tests/benchmark/test_worst_blocks.py
LouisTsai-Csie: Oh, I didn’t realize this was included in Pectra, let me check on that. I just ported the test case from Nethermind. Also, I noticed that the information on [evm.codes](https://www.evm.codes/) is outdated; I’d like to help update it.

LouisTsai-Csie:
I read the [proposal](https://eips.ethereum.org/EIPS/eip-7623) as well as some [analysis](https://hackernoon.com/eip-7623-the-proposal-that-will-re-price-calldata-for-ethereum-transactions) (this is a good one), and refactored the benchmark accordingly. Please let me know if my understanding is incorrect:

According to EIP-7623, the gas cost is calculated as follows:
```py
tx.gasUsed = 21000 + max(
    STANDARD_TOKEN_COST * tokens_in_calldata
    + execution_gas_used
    + isContractCreation * (32000 + INITCODE_WORD_COST * words(calldata)),
    TOTAL_COST_FLOOR_PER_TOKEN * tokens_in_calldata,
)
```

In this test case, since the transaction only includes a payload and does not execute any bytecode or create a contract, both `execution_gas_used` and `isContractCreation` are zero. This simplifies the formula to:

```py
tx.gasUsed = 21000 + max(
    STANDARD_TOKEN_COST * tokens_in_calldata,
    TOTAL_COST_FLOOR_PER_TOKEN * tokens_in_calldata,
)
```

Given that `STANDARD_TOKEN_COST` = 4 and `TOTAL_COST_FLOOR_PER_TOKEN` = 10, the max(...) term always evaluates to the latter. So the formula can be further simplified to:

```py
tx.gasUsed = 21000 + 10 * tokens_in_calldata
```

By definition:
```py
tokens_in_calldata = zero_bytes_in_calldata + 4 * nonzero_bytes_in_calldata
```

We use this relationship to compute the total number of bytes that can fit in the payload under the available gas limit.

tests/benchmark/test_worst_blocks.py
```diff
@@ -156,3 +157,69 @@ def test_block_full_of_ether_transfers(
...
+    #
+    # So we calculate how many bytes we can fit into calldata based on available gas.
+
+    gas_available = attack_gas_limit - intrinsic_cost
+
+    # Calculate the token_in_calldata
+    gas_available //= total_cost_floor_per_token
+
+    # Calculate the number of bytes that can be stored in the calldata
+    num_of_bytes = gas_available // 4 if zero_byte else attack_gas_limit
```
jsign, line 210:
```suggestion
    # Calculate the token_in_calldata
    max_tokens_in_calldata = gas_available // total_cost_floor_per_token

    # Calculate the number of bytes that can be stored in the calldata
    num_of_bytes = max_tokens_in_calldata if zero_byte else max_tokens_in_calldata // 4
```
Suggest two changes:
- Use a `max_tokens_in_calldata` instead of `gas_limit` when `//4` since after doing that division, "gas limit" semantics seems to be lost since we aren't talking about gas anymore but tokens.
- I think your previous formula (i.e., `num_of_bytes = gas_available // 4 if zero_byte else attack_gas_limit`) was maybe incorrect?. We should `//4` if `not zero_byte`, and also under no case we should us `attack_gas_limit` as the number of bytes?

tests/benchmark/test_worst_blocks.py
```diff
@@ -156,3 +157,68 @@ def test_block_full_of_ether_transfers(
...
+
+@pytest.fixture
+def total_cost_floor_per_token():
+    """Total cost floor per token."""
+    return 10
+
+
+@pytest.mark.valid_from("Prague")
+@pytest.mark.parametrize("zero_byte", [True, False])
+def test_block_full_of_empty_payload(
```
jochem-brouwer, line 170: I think we need to change the name of this test, because it is not empty (its a tx full of data). Do we also want to test this sending value? And maybe sending value to self or some other account? :thinking:

jochem-brouwer (commented):
One question/remark :smile: :+1:

tests/benchmark/test_worst_blocks.py
jochem-brouwer: Meh sending value is actually another situation and it would mix two things (sending data / sending value) so I guess it is not very effective to do so here.

LouisTsai-Csie: maybe we should limit this test to pure calldata only. 

There’s another test case specifically for plain ether transfers, in case that’s what you’re looking for. - [Link](https://github.com/ethereum/execution-spec-tests/blob/0f77f07a8537c447e1113547975fe9b9b4378baf/tests/benchmark/test_worst_blocks.py#L109)

LouisTsai-Csie: > I think we need to change the name of this test, because it is not empty (its a tx full of data)

I agree on this, it is more clear. I change the PR title also

spencer-tb (approved):
LGTM! Thanks

PR #1831, opened 2025-07-01
Clz initcode context

LouisTsai-Csie:
## 🗒️ Description

WIP

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```


LouisTsai-Csie:
This is my understanding of contract creation: during deployment, the contract’s creation code is executed. This code can store the final contract bytecode in memory, as shown in the `test_clz_initcode_context` in this PR

I’ve split the creation code into two parts. The first part is the `initcode`, which stores the result of the CLZ operation into storage slot 0 during contract creation for testing purpose. The second part uses `EXTCODECOPY` to load the deployed contract bytecode into memory.

Based on this structure, I want to extend the test case to cover `CREATE` and `CREATE2`, using the same technique of separating the `initcode`. 

This is my ideal test case, but I’m unsure why the `initcode` isn’t executing during contract creation as expected.

```python
@pytest.mark.valid_from("Osaka")
@pytest.mark.parametrize("bits", [0, 1, 4, 8, 16])
@pytest.mark.parametrize("opcode", [Op.CREATE])
def test_clz_initcode_create(state_test: StateTestFiller, pre: Alloc, bits: int, opcode: Op):
    """Test CLZ opcode behavior when creating a contract."""
    init_code = Op.SSTORE(0, Op.CLZ(1 << bits)) + Op.RETURN(0, Op.MSIZE)

    # init code is executed first, so the storage is set to the CLZ result
    creation_code = Op.SSTORE(1, Op.CLZ(1 << bits)) + Op.MSTORE(0, init_code.hex())

    sender_address = pre.fund_eoa()

    # factory contract will create a new contract
    if opcode == Op.CREATE:
        factory_code = creation_code + Op.CREATE(offset=0, size=len(init_code))

    factory_address = pre.deploy_contract(code=factory_code)

    # the new contract address is computed from the factory contract address and the nonce
    if opcode == Op.CREATE:
        contract_address = compute_create_address(address=factory_address, nonce=1)

    tx = Transaction(
        to=factory_address,
        gas_limit=6_000_000,
        data=creation_code,
        sender=sender_address,
    )

    expected_clz = 255 - bits

    # During the creation, the storage is set to the CLZ result based on the init code
    post = {
        contract_address: Account(storage={"0x00": expected_clz}),
    }

    state_test(pre=pre, post=post, tx=tx)
```

I am not sure if I got something wrong with create/create2, any recommendation is appreciated!

tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py
```diff
@@ -276,3 +278,95 @@ def test_clz_jump_operation(
         post[callee_address] = Account(storage={"0x00": expected_clz})
 
     state_test(pre=pre, post=post, tx=tx)
+
+
+@pytest.mark.valid_from("Osaka")
+@pytest.mark.parametrize("bits", [0, 16, 64, 128, 255])
```
marioevz, line 284: Same comment as in the other EIP, I think we should have these bits inside of the same test case just to reduce the number of generated tests.

tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py
```diff
@@ -276,3 +278,95 @@ def test_clz_jump_operation(
...
+
+    creation_code = (
+        init_code
+        + Op.EXTCODECOPY(
+            address=code_address,
+            dest_offset=0,
+            offset=0,
+            size=Op.EXTCODESIZE(code_address),
+        )
+        + Op.RETURN(0, Op.MSIZE)
```
marioevz, line 301: We could simplify this part by simply returning `Op.RETURN(0, 1)`, since we are not really calling the deployed contract, it does not matter what ends up deployed.

With this we can also remove `code` and its deployment from the test.

tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py
```diff
@@ -276,3 +278,95 @@ def test_clz_jump_operation(
...
+
+    post = {
+        contract_address: Account(storage={"0x00": expected_clz}),
+    }
+
+    state_test(pre=pre, post=post, tx=tx)
+
+
+@pytest.mark.valid_from("Osaka")
+@pytest.mark.parametrize("bits", [0, 1, 4, 8, 16])
```
marioevz, line 323: Same as previous comment.

tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py
```diff
@@ -276,3 +278,95 @@ def test_clz_jump_operation(
...
+    sender_address = pre.fund_eoa()
+
+    if opcode == Op.CREATE:
+        create_contract = creation_code + opcode(offset=0, size=len(ext_code))
+    elif opcode == Op.CREATE2:
+        create_contract = creation_code + opcode(
+            offset=0,
+            size=len(ext_code),
+            salt="00" * 32,
+        )
```
marioevz, line 348: If we are not using the salt we can simplify this a bit by not having to provide it as a kw-argument:
```suggestion
    create_contract = creation_code + opcode(offset=0, size=len(ext_code))
```
And then it's assumed to be zero during the `compute_create_address` call.

tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py
```diff
@@ -276,3 +278,95 @@ def test_clz_jump_operation(
...
+
+    create_contract_address = pre.deploy_contract(code=create_contract)
+
+    if opcode == Op.CREATE:
+        contract_address = compute_create_address(address=create_contract_address, nonce=1)
+    elif opcode == Op.CREATE2:
+        contract_address = compute_create2_address(
+            address=create_contract_address, salt="00" * 32, initcode=ext_code
+        )
```
marioevz, line 357: Another small simplification here since `compute_create_address` can handle both computations internally:
```suggestion
    contract_address = compute_create_address(address=create_contract_address, nonce=1, initcode=ext_code, opcode=opcode)
```

marioevz: Also would change `create_contract_address` to `factory_contract_address`, and `contract_address` to `created_contract_address`. It makes the code easier to follow.

tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py
```diff
@@ -276,3 +278,95 @@ def test_clz_jump_operation(
...
+@pytest.mark.parametrize("bits", [0, 1, 4, 8, 16])
+@pytest.mark.parametrize("opcode", [Op.CREATE, Op.CREATE2])
+def test_clz_initcode_create(state_test: StateTestFiller, pre: Alloc, bits: int, opcode: Op):
+    """Test CLZ opcode behavior when creating a contract."""
+    init_code = Op.SSTORE(0, Op.CLZ(1 << bits))
+
+    ext_code = Op.SSTORE(1, Op.CLZ(1 << bits))
+    ext_code_address = pre.deploy_contract(code=ext_code)
+
+    creation_code = init_code + Op.EXTCODECOPY(
```
marioevz, line 332: In this case we could pass `ext_code` as the data of the transaction (see line 362), and use `CALLDATACOPY` instead of `EXTCODECOPY` instead, just to save the extra contract deployment that the tests currently requires.

I think we can also remove `init_code` entirely.

marioevz (commented):
Looks mostly good, just some optimization comments.
Also a rebase is needed, thanks!

LouisTsai-Csie:
@marioevz Thanks for your review, I've updated this PR, but I am not sure if the change is what you want, let me know if there is anything wrong.

I was confused by the `<init code><runtime code><constructor params>` pattern in Solidity, so I try to create `initcode` and creation code in each test.

After reviewing this blog, everything becomes more clear now.

Resource: https://rareskills.io/post/ethereum-contract-creation-code

marioevz (approved):
LGTM, thanks for the changes!

Issue #1795, opened 2025-06-24
feat(tests): additional test cases for eip-7939


spencer-tb:
#### Count leading zeros (CLZ) opcode

**Initcode Context**
- [x] From `CODECOPY` / `EXTCODECOPY` - https://github.com/ethereum/execution-spec-tests/pull/1858
- [x] Contract creation transaction - https://github.com/ethereum/execution-spec-tests/pull/1831
- [x] Contract creation via `CREATE` & `CREATE2` - https://github.com/ethereum/execution-spec-tests/pull/1831

**Execution Context**
- [x] `CALL`, `STATICCALL`, `DELEGATECALL`, `CALLCODE` - https://github.com/ethereum/execution-spec-tests/pull/1829
- [x] setCode transaction (eip-7702) - https://github.com/ethereum/execution-spec-tests/pull/1853

**Gas Cost**
- [x] Not enough gas - https://github.com/ethereum/execution-spec-tests/pull/1859
- [x] Exact gas cost (5 unit) - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/df014bf76be563b157a665abafb4885bea99071b/tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py#L111)
- [x] Extra gas - https://github.com/ethereum/execution-spec-tests/pull/1859
- [x] Short by N, N = 3 - https://github.com/ethereum/execution-spec-tests/pull/1859

**Benchmark**
- [x]  ZK- proving time - https://github.com/ethereum/execution-spec-tests/pull/1845
- [x] Execution time for client - https://github.com/ethereum/execution-spec-tests/pull/1845

**New Opcode**
- [x] Fork transition test - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/df014bf76be563b157a665abafb4885bea99071b/tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py#L159)
- [x] Execution succeed on Osaka but fail on Prague - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/df014bf76be563b157a665abafb4885bea99071b/tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py#L159)

**0x1e Opcode**
- [x] jump / jumpi into opcode should fail - https://github.com/ethereum/execution-spec-tests/pull/1824
[...]

PR #1845, opened 2025-07-02
feat(benchmark): add clz opcode test case

LouisTsai-Csie:
## 🗒️ Description

Add benchmark test for CLZ opcode.

## 🔗 Related Issues or PRs

Issue https://github.com/ethereum/execution-spec-tests/issues/1795

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/tests/benchmark/test_worst_compute.py
+++ b/tests/benchmark/test_worst_compute.py
@@ -22,6 +22,7 @@ from ethereum_test_tools import (
     Block,
     BlockchainTestFiller,
     Bytecode,
+    Environment,
     StateTestFiller,
     Transaction,
     add_kzg_version,
@@ -2727,12 +2728,14 @@ def test_worst_return_revert(
 
 @pytest.mark.valid_from("Osaka")
 def test_worst_clz_same_input(
-    state_test: StateTestFiller, blockchain_test: BlockchainTestFiller, pre: Alloc, fork: Fork
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+    env: Environment,
 ):
     """Test running a block with as many CLZ with same input as possible."""
-    env = Environment()
-    attack_gas_limit = env.gas_limit
-    tx_gas_limit = fork.transaction_gas_limit_cap()
+    tx_gas_limit = fork.transaction_gas_limit_cap() or env.gas_limit
 
     magic_value = 248  # CLZ(248) = 248
 
@@ -2743,42 +2746,38 @@ def test_worst_clz_same_input(
 
     code_address = pre.deploy_contract(code=code)
 
-    tx = Transaction(
-        to=code_address,
-        gas_limit=tx_gas_limit if tx_gas_limit else attack_gas_limit,
-        sender=pre.fund_eoa(),
+    sender = pre.fund_eoa()
+    tx_count = gas_benchmark_value // tx_gas_limit
+    remainder_gas = gas_benchmark_value % tx_gas_limit
+
+    txs = [
+        Transaction(
+            to=code_address,
+            gas_limit=tx_gas_limit if i < tx_count else remainder_gas,
+            nonce=i,
+            sender=sender,
+        )
+        for i in range(tx_count + 1)
+    ]
+
+    blockchain_test(
+        genesis_environment=env,
+        pre=pre,
+        post={},
+        blocks=[Block(txs=txs)],
     )
-
-    if (tx_gas_limit is None) or (tx_gas_limit > attack_gas_limit):
-        state_test(
-            env=env,
-            pre=pre,
-            post={},
-            tx=tx,
-        )
-    else:
-        tx_count = attack_gas_limit // tx_gas_limit
-        txs = [tx for _ in range(tx_count)]
-
-        blockchain_test(
-            genesis_environment=Environment(),
-            pre=pre,
-            post={},
-            blocks=[Block(txs=txs)],
-        )
 
 
 @pytest.mark.valid_from("Osaka")
 def test_worst_clz_diff_input(
-    state_test: StateTestFiller,
     blockchain_test: BlockchainTestFiller,
     pre: Alloc,
     fork: Fork,
+    gas_benchmark_value: int,
+    env: Environment,
 ):
     """Test running a block with as many CLZ with different input as possible."""
-    env = Environment()
-    attack_gas_limit = env.gas_limit
-    tx_gas_limit = fork.transaction_gas_limit_cap()
+    tx_gas_limit = fork.transaction_gas_limit_cap() or env.gas_limit
     max_code_size = fork.max_code_size()
 
     code_prefix = Op.JUMPDEST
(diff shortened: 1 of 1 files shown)


tests/benchmark/test_worst_compute.py
```diff
@@ -2255,3 +2255,95 @@ def test_worst_push(
...
+@pytest.mark.valid_from("Osaka")
+def test_worst_clz_same_input(
+    state_test: StateTestFiller, blockchain_test: BlockchainTestFiller, pre: Alloc, fork: Fork
+):
+    """Test running a block with as many CLZ with same input as possible."""
+    env = Environment()
+    attack_gas_limit = env.gas_limit
+    tx_gas_limit = fork.transaction_gas_limit_cap()
+
+    magic_value = 248  # CLZ(248) = 248
```
LouisTsai-Csie, line 2269: In this test case, each `CLZ` operation uses the same input. I discovered that` CLZ(248) == 248`, which makes it more efficient than the approach where we first push a value and then `DUP` it before calling `CLZ`.

To verify, you could try this script:

```py
# This function is from https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7939.md
def clz(x):
    """Returns the number of zeros preceding the most significant one bit."""
    if x < 0:
        raise ValueError("clz is undefined for negative numbers")
    if x > 2**256 - 1:
        raise ValueError("clz is undefined for numbers larger than 2**256 - 1")
    if x == 0:
        return 256
    # Convert to binary string and remove any '0b' prefix.
    bin_str = bin(x).replace('0b', '')
    return 256 - len(bin_str)

bits = 0
magic_value = 248
stack = [magic_value]

for i in range(100):
    x = stack.pop()
    assert clz(x) == magic_value
    stack.append(clz(x))
```

tests/benchmark/test_worst_compute.py
```diff
@@ -2255,3 +2255,95 @@ def test_worst_push(
...
+    code = code_loop_precompile_call(calldata, attack_block, fork)
+    code_address = pre.deploy_contract(code=code)
+
+    tx = Transaction(
+        to=code_address,
+        gas_limit=tx_gas_limit if tx_gas_limit else attack_gas_limit,
+        sender=pre.fund_eoa(),
+    )
+
+    if (tx_gas_limit is None) or (tx_gas_limit > attack_gas_limit):
```
LouisTsai-Csie, line 2282: This if-statement is for EIP-7852 for transaction limit cap, more details could be found in this [issue](https://github.com/ethereum/execution-spec-tests/issues/1770).

tests/benchmark/test_worst_compute.py
```diff
@@ -2255,3 +2255,95 @@ def test_worst_push(
...
+
+    code_prefix = Op.JUMPDEST
+    code_suffix = Op.PUSH0 + Op.JUMP
+
+    iteration = (max_code_size - len(code_prefix) - len(code_suffix)) // 2
+
+    code_seq = Bytecode()
+
+    for i in range(2 * iteration):
+        value = i if i % 2 else 2**256 - 1 - i
```
LouisTsai-Csie, line 2322: In the second test case, I aim to add some randomness. Using a simple range from 1 to iteration results in many values being too similar.

tests/benchmark/test_worst_compute.py
```diff
@@ -2317,3 +2317,95 @@ def test_worst_return_revert(
...
+        code_seq += Op.CLZ(value) + Op.POP
+
+    code_address = pre.deploy_contract(code=code_prefix + code_seq + code_suffix)
```
jochem-brouwer, line 2387: This will deploy a contract too large. Could you add the sanity check?

```suggestion
    code_address = pre.deploy_contract(code=code_prefix + code_seq + code_suffix)
    if len(attack_code) > max_contract_size:
        raise ValueError(
            f"Code size {len(attack_code)} exceeds maximum code size {max_contract_size}"
        )
```

tests/benchmark/test_worst_compute.py
jochem-brouwer: I get the idea but I would set this up differently. We have `code_prefix` and `code_suffix` so we know that we can add at most `max_code_size - len(code_prefix) - len(code_suffix)` bytes. I'd keep incrementing some number here and then let CLZ work on `2^256 - 1 >> i % 256`. This `i` increments each round and will thus right shift the values such that CLZ yields a different value. Note that for each CLZ you add you should check `Op.CLZ(value) + Op.POP` if this length does not exceed the max size (else `break`). This length is not constant because it will depend on how many bytes we are pushing (to push `2^256-1` need `PUSH32`, and for `1` need `PUSH1`)

jochem-brouwer (changes requested):
Some comments :smile: :+1: 

Need to also think on how to use these benchmarks, i.e. how to test these against clients and how to interpret the results. The end goal could be to determine the gas price of the opcode. For that we also need a baseline in order to interpret the results and thus also to price the opcode. This is non-trivial because these benchmarks use other opcodes as well, and we thus have to find a way to measure the costs for CLZ only (and not the cost of the other opcodes). Food for thought!

Left a comment on one of the tests. The 248 trick is beautiful :smile: :+1:

tests/benchmark/test_worst_compute.py
LouisTsai-Csie: I might go with the following approach for consistency now, since the remaining benchmarks use it as well. 

However, we could add this refactoring for all the assertions later in this [issue](https://github.com/ethereum/execution-spec-tests/issues/1704)

```py
 assert len(attack_code) <= max_code_size
```

LouisTsai-Csie:
@jochem-brouwer Thanks for your review, I've updated accordingly! Please let me know if the change is good to you.

tests/benchmark/test_worst_compute.py
```diff
@@ -2317,3 +2317,103 @@ def test_worst_return_revert(
...
+
+    for i in range(available_code_size):
+        value = (2**256 - 1) >> (i % 256)
+        clz_op = Op.CLZ(value) + Op.POP
+        if len(code_seq) + len(clz_op) > available_code_size:
+            break
+        code_seq += clz_op
+
+    attack_code = code_prefix + code_seq + code_suffix
+    assert len(attack_code) <= max_code_size
```
MariusVanDerWijden, line 2393: I think this whole logic is very generalizable, you set up the stack, setup a jumpdest then do `24k - prefix - suffix times / inner code` include the operations then add the suffix.
I think this could be refactored out into a helper cc @marioevz e.g.

```
func createBenchLoop(prefix, code, suffix []byte) []byte{
    result := []byte
    result := append(result, prefix)
    dest := len(result)
    result := append(result, JUMPDEST)
    suffix := append(suffix, PUSH32)
    suffix := append(suffix, extend32(dest)
    suffixLen := len(suffix)
    
    available_code_size := max_contract_size - len(result) - suffixLen
    
    for i := range available_code_size / len(code)
         result := append(result, code)
    
    result := append(result, suffix)
    return result
```

LouisTsai-Csie: This is part of our refactoring plan, we will create new test types: benchmark_test, and benchmark_state_test to handle the repetitive code, the transaction gas limit cap, and more. You could check more information in this [issue](https://github.com/ethereum/execution-spec-tests/issues/1896).

marioevz (approved):
Amazing implementation, thanks!

marioevz:
I think this needs a careful rebase since I tried rebasing locally and the merge messed up the test, @LouisTsai-Csie could you try rebasing to latest main and see that it works? thanks!

LouisTsai-Csie:
@marioevz I did not see any conflict from my side, but I still rebase to the latest changes. Let's see if this works.

PR #1889, opened 2025-07-09
fix(benchmark): Fix CALLDATACOPY filling for low gas limits

jsign:
## 🗒️ Description
I performed a pre-filling of current tests in `main` for 1M and 5M, and I noticed that some cases failed.

The reason for failure is that the intrinsic cost is higher than the gas limit; thus the transaction execution failed.

Since these 1M and 5M cases are primarily toy-like cases for zkVM CIs and quick experiments, I addressed this by calculating the intrinsic cost and skipping the case if the gas limit is too low.

This is something that would have surfaced whenever we tagged the next release (see linked PR below).

## 🔗 Related Issues or PRs
https://github.com/ethereum/execution-spec-tests/pull/1876

### DIFF
--- a/tests/benchmark/test_worst_memory.py
+++ b/tests/benchmark/test_worst_memory.py
@@ -77,6 +77,15 @@ def test_worst_calldatacopy(
     if size == 0 and non_zero_data:
         pytest.skip("Non-zero data with size 0 is not applicable.")
 
+    # If `non_zero_data` is True, we fill the calldata with deterministic random data.
+    # Note that if `size == 0` and `non_zero_data` is a skipped case.
+    data = Bytes([i % 256 for i in range(size)]) if non_zero_data else Bytes()
+
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+    min_gas = intrinsic_gas_calculator(calldata=data)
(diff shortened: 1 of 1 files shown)


tests/benchmark/test_worst_memory.py
```diff
@@ -77,6 +77,15 @@ def test_worst_calldatacopy(
...
         pytest.skip("Non-zero data with size 0 is not applicable.")
 
+    # If `non_zero_data` is True, we fill the calldata with deterministic random data.
+    # Note that if `size == 0` and `non_zero_data` is a skipped case.
+    data = Bytes([i % 256 for i in range(size)]) if non_zero_data else Bytes()
+
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+    min_gas = intrinsic_gas_calculator(calldata=data)
+    if min_gas > env.gas_limit:
+        pytest.skip("Minimum gas required for calldata ({min_gas}) is greater than the gas limit")
```
jsign, line 87: Theoretically, there could be other gas limits where this check isn't enough. But practically speaking, the low targets are 1M and 5M, so IMO this is good enough without getting into more complex calculations.

Issue #1891, opened 2025-07-10
feat(benchmark,fill): generate all gas value benchmarks in one `fill` execution


danceratopz:
The current status of benchmark fixture releases should be improved to help use EEST benchmarking test fixtures with Nethermind's benchmarking infra.

More background from Kamil:

> For a "proper" benchmarking what we need to do is to have a solid stable SINGLE genesis file on which we can execute various-sized scenarios and grasp the data from them.
> 
> So if we have 5 different sized tests - they all should be executable on single genesis file so we can execute all of them one after another, get the data and analyze.
> 
> Current approach would require us to stop node multiple times and restart which makes it difficult from a warmup reasons.

## Status quo

1. To generate benchmark tests at a specific gas value/limit `fill` is called with the `--block-gas-limit` flag specifying the limit.
2. `fill` must be executed once per gas value to be benchmarked, with different values to `--block-gas-limit`.
3. As a consequence of 2., a benchmark release consists of multiple tarballs, one per gas value, e.g. [benchmark@v0.0.2](https://github.com/ethereum/execution-spec-tests/releases/tag/benchmark%40v0.0.2).
4. Each tarball in the release corresponds requires a different client genesis configuration as the `gasLimit` specified via `--block-gas-limit` is set as the default gas limit. This value lands in the block header, of course, this also influences the block `hash`.

## Goal

1. Tests for all gas values get created in one `fill` execution, as a result, a benchmark release will now be contained in a single tarball.
2. All tests, even for different gas benchmark values, from one pre-alloc group can use the same genesis configuration.

## Approach

1. Add a new command-line flag for `fill`: Suggestion --gas-benchmark-values. This takes a comma-separated list of the gas benchmark values.
2. The default value of the `Environment`'s gas is set to a large enough value that it is irrelevant for the benchmarking: This will ensure that the genesis configuration is the same for all tests generated from the single `fill` execution.
3. Parametrize the tests with the values from `--gas-benchmark-values`. This should be done via `fill`s `pytest_generate_tests()` (parameter name suggestion: `gas_benchmark_value`, but only if the flag from 1. has been set, otherwise we can take the default value from the flag? (TBD). We need to be careful with non-benchmark tests, if a test is not a benchmark test it should be unaffected.(and this should not show up in its test id). See example test IDs below. 
[...]

PR #1895, opened 2025-07-11
feat(fill): add `--gas-benchmark-values` command to support single genesis file

LouisTsai-Csie:
## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```


src/pytest_plugins/filler/filler.py
```diff
@@ -1038,6 +1063,22 @@ def pytest_generate_tests(metafunc: pytest.Metafunc):
                 indirect=True,
             )
 
+            # Handle gas benchmark values parametrization
+            if "gas_benchmark_value" in metafunc.fixturenames:
+                gas_benchmark_values = metafunc.config.getoption("gas_benchmark_value")
+                if gas_benchmark_values:
+                    gas_values = [int(x) for x in gas_benchmark_values.replace(" ", "").split(",")]
```
marioevz, line 1070:
```suggestion
                    gas_values = [int(x.strip()) for x in gas_benchmark_values.split(",")]
```
Not sure if necessary but maybe in case there's a white space between commas?

src/pytest_plugins/filler/filler.py
```diff
@@ -326,6 +337,12 @@ def pytest_configure(config):
     if config.getoption("block_gas_limit"):
         EnvironmentDefaults.gas_limit = config.getoption("block_gas_limit")
 
+    if config.getoption("gas_benchmark_value"):
+        benchmark_values = [
+            int(x) for x in config.getoption("gas_benchmark_value").replace(" ", "").split(",")
+        ]
+        EnvironmentDefaults.gas_limit = max(benchmark_values) * 1_000_000
+
```
marioevz, line 345: I would propose that we leave this out and instead create a fixture in the [benchmark folder's conftest.py](https://github.com/ethereum/execution-spec-tests/blob/main/tests/benchmark/conftest.py), that generates a default gas value for the setups:
```python
@pytest.fixture
def env() -> Environment:  # noqa: D103
    return Environment(gas_limit=1_000_000_000)
```
The 1 billion default seems reasonable for most benchmark setups that deploy a ton of contracts for the actual benchmark test.

The tests would now change from `Environment().gas_limit` to adding the `env` parameter to the test function and then `env.gas_limit` to determine the max gas to use in the setup block, while respecting `gas_benchmark_value` for the attack block.

src/pytest_plugins/filler/filler.py
```diff
@@ -265,6 +265,17 @@ def pytest_addoption(parser: pytest.Parser):
         default=False,
         help="Fill tests using existing pre-allocation groups (phase 2 only).",
     )
+    test_group.addoption(
+        "--gas-benchmark-values",
+        action="store",
+        dest="gas_benchmark_value",
+        type=str,
+        default=None,
```
marioevz, line 273:
```suggestion
        default='5',
```
This would default to run the benchmark tests with 5 million gas for sanity checking for example, what do you think?

marioevz (commented):
Looks great! I think this a great change for maintainability and make it easier for us to generate more vectors as they are required.

Small downside is that we have to remove `Environment().gas_limit` from most of tests, but I would say we do it the earlier the better.

cc @jsign for some feedback on my comments.

Thanks!

jsign:
Nice! @LouisTsai-Csie @marioevz, is this compatible with supporting all the test formats too? (i.e. https://github.com/ethereum/execution-spec-tests/issues/1778).

Mostly asking since I think this is coming from the fact of simplifying the single genesis for perfnets, but wondering if it should still be fine for the other formats that we need for zkVMs.

src/pytest_plugins/filler/filler.py
jsign: Nice, sounds good to me.

marioevz:
> Nice! @LouisTsai-Csie @marioevz, is this compatible with supporting all the test formats too? (i.e. #1778).
> 
> Mostly asking since I think this is coming from the fact of simplifying the single genesis for perfnets, but wondering if it should still be fine for the other formats that we need for zkVMs.

Should be compatible out of the box, but I'll give that a look again and raise if the there's any concerns.

src/pytest_plugins/filler/filler.py
LouisTsai-Csie: It is necessary to handle cases like `5, 10,   20,30`, string with white spaces between. But the current `.replace(" ", "")` method already removes the white space

src/pytest_plugins/filler/filler.py
LouisTsai-Csie: > We need to be careful with non-benchmark tests, if a test is not a benchmark test it should be unaffected.
According to the issue description, non-benchmark tests shouldn’t be impacted, since this would change the gas limit.

If we set the default value of `--gas-benchmark-values` to "5", the non-benchmark ones would automatically apply a benchmark fixture with a 5M gas setting.

Based on my experiment, setting the default to "5" doesn’t apply when the flag is explicitly included without a value, like this:
```bash
uv run fill -v tests/benchmark/test_worst_blocks.py::test_block_full_data \
  --fork Prague \
  --gas-benchmark-values\
  --generate-pre-alloc-groups \
  --clean
# This will lead to error
```

However, when the flag is ignored the default value is applied, as seen in this command:
```bash
uv run fill -v tests/benchmark/test_worst_blocks.py::test_block_full_data \
  --fork Prague \
  --generate-pre-alloc-groups \
  --clean
# It will have fork_Prague-blockchain_test_engine_x_from_state_test-benchmark-gas-value_5M-zero_byte_False
```

src/pytest_plugins/filler/filler.py
LouisTsai-Csie: Looks good, but this looks to be part of the general refactoring, should I do it in this PR, or I could add it in this [issue](https://github.com/ethereum/execution-spec-tests/issues/1704) and do it later?

danceratopz (commented):
Thanks, this looks great @LouisTsai-Csie!

Shame, that this didn't occur to me up front in #1891, but I'd suggest that we move this codeto a new plugin that gets activated with fill by default. This should work well due to the composability of pytest plugins. 

This means, we:
1. Add these changes (and other benchmarking related pytest config, if any) to a separate pytest plugin, I'd suggest `src/pytest_plugins/filling/benchmarking.py`.
2. Enable this plugin using `-p` via the `fill` command's pytest ini: 
    https://github.com/ethereum/execution-spec-tests/blob/0f7c73a7e177abf58cdba116a3b4275e9be9a61b/src/cli/pytest_commands/pytest_ini_files/pytest-fill.ini#L11-L22

All benchmarking-related plugin customizations (e.g. `pytest_addoption`, `pytest_generate_tests`, etc.) currently in `filler/filler.py` can be moved directly to `filler/benchmarking.py`. This keeps the benchmarking logic self-contained. Pytest hooks from both modules should compose as expected.

To cleanly handle options/values that are specific to benchmarking, I'd suggestion the following approach, if you agree/like it feel free to go for it!

#### 1. Define a filling mode enum in `filler/filler.py`:
```python
from enum import StrEnum, unique

@unique
class FillMode(StrEnum):
    CONSENSUS = "consensus"
    BENCHMARKING = "benchmarking"
```

---

#### 2. In the filler plugin (`filler.py`), set the default:
```python
from _pytest.config import Config
from .filler import FillMode

def pytest_configure(config: Config) -> None:
    if not hasattr(config, "fill_mode"):
        config.fill_mode = FillMode.CONSENSUS
```

---

#### 3. In the benchmarking plugin (`filler/benchmarking.py`), override only if `--benchmark-gas-values` is set:
```python
from _pytest.config import Config
from .filler import FillMode

def pytest_configure(config: Config) -> None:
    if config.getoption("--benchmark-gas-values") is not None:
        config.fill_mode = FillMode.BENCHMARKING
```

---

#### 4. Example usage in filler logic, wrapped in a fixture:
```python
import pytest
from ,filler import FillMode

GIGA_GAS = 1_000_000_000

@pytest.fixture
def env() -> Environment:  # noqa: D103
    return 1_000_000_000)
    if config.fill_mode == FillMode.BENCHMARKING:
        return Environment(gas_limit=GIGA_GAS)
    else:
        return Environment()
```

src/pytest_plugins/filler/filler.py
```diff
@@ -335,6 +343,9 @@ def pytest_configure(config):
         called before the pytest-html plugin's pytest_configure to ensure that
         it uses the modified `htmlpath` option.
     """
+    if not hasattr(config, "fill_mode"):
+        config.fill_mode = FillMode.CONSENSUS
```
LouisTsai-Csie, line 347: There is a `mypy` linting issue because `fill_code` is not defined in the config type, but we assign it here. This affects other parts of `conftest.py`. For now, I only ignore the warning.

tests/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+from ethereum_test_tools import Environment
+from pytest_plugins.filler.filler import FillMode
+
+GIGA_GAS = 1_000_000_000
+
+
+@pytest.fixture
+def env(request: pytest.FixtureRequest) -> Environment:  # noqa: D103
+    """Return an Environment instance with appropriate gas limit based on fill mode."""
+    if request.config.fill_mode == FillMode.BENCHMARKING:  # type: ignore[attr-defined]
```
LouisTsai-Csie, line 14: Same linting issue as described above

danceratopz (commented):
Thanks! This looks great to me! 

One comment below.

danceratopz:
> @danceratopz Thank you for review, but I am wondering the following:
> 
>     * Should I add test cases under `src/cli/tests/`? I noticed your recent [PR](https://github.com/ethereum/execution-spec-tests/pull/1855/files#diff-5c3633f8cbee135e20eb35f9537277edaf7ff69714db9f5c0993431a312ca5f5) included tests there.

I don't think it's strictly necessary for the PR, but some sanity check that the flag works is nice, of course. Recently, I've been pointing Claude at unit testing tasks.

>     * Should I add documentation for the new flag? I’m happy to do that, but I ran into some [issues](https://github.com/ethereum/execution-spec-tests/issues/1908) building the docs locally with `mkdocs` (related to cairosvg and missing `libcairo` on `macOS`). Let me know if there’s a preferred workaround or if I should just update the markdown and let CI verify the build (not a good idea).

Does this work?
```
uvx --with=tox-uv tox -e mkdocs
```
If so, its' because of the macOS trick found in these lines (you can then set the env var locally):
https://github.com/ethereum/execution-spec-tests/blob/dfdd43369928253bd862405886d028c65e2a4cc3/tox.ini#L56-L58

src/pytest_plugins/filler/filler.py
```diff
@@ -335,6 +344,9 @@ def pytest_configure(config):
...
```
marioevz, line 348: This should be as an `else` branch in `pytest_configure` in `src/pytest_plugins/filler/benchmarking.py`, because otherwise we require to import this `filler.py` during execute.

src/pytest_plugins/filler/filler.py
```diff
@@ -557,6 +569,11 @@ def pytest_html_report_title(report):
     report.title = "Fill Test Report"
 
 
+@pytest.fixture
```
marioevz, line 572: Can be moved to `src/pytest_plugins/shared/execute_fill.py` in order for this all to work with execute too.

src/pytest_plugins/filler/benchmarking.py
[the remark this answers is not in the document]
danceratopz: This occurred to me in the call today :slightly_smiling_face:

.github/configs/feature.yaml
```diff
@@ -11,34 +11,9 @@ static:
...
-  feature_only: true
-benchmark_90M:
-  evm-type: benchmark
-  fill-params: --from=Cancun --until=Prague --block-gas-limit 90000000 -m benchmark ./tests
-  solc: 0.8.21
-  feature_only: true
-benchmark_120M:
-  evm-type: benchmark
-  fill-params: --from=Cancun --until=Prague --block-gas-limit 120000000 -m benchmark ./tests
+  fill-params: --from=Cancun --until=Prague --gas-benchmark-values 1,10,30,60,90,120 -m benchmark ./tests
```
marioevz, line 16:
```suggestion
  fill-params: --from=Cancun --until=Prague --gas-benchmark-values 1,10,30,60,90,120 -m benchmark --generate-all-formats ./tests
```

src/pytest_plugins/filler/tests/test_benchmarking.py
```diff
@@ -0,0 +1,118 @@
...
+
+    return test_module
+
+
+def test_gas_benchmark_option_added(pytester: pytest.Pytester):
+    """Test that the --gas-benchmark-values option is properly added."""
+    pytester.copy_example(name="src/cli/pytest_commands/pytest_ini_files/pytest-fill.ini")
+
+    # Command: pytest -p pytest_plugins.filler.benchmarking --help
+    result = pytester.runpytest("-p", "pytest_plugins.filler.benchmarking", "--help")
```
marioevz, line 66:
```suggestion
    result = pytester.runpytest("-c", "pytest-fill.ini", "--help")
```
Fixes the last remaining failing unit test.

marioevz (commented):
Awesome work, I would like to see if we could rebase and use `gas_benchmark_value` in all benchmark tests so we can prepare for the next benchmark release if possible.

PR #1911, opened 2025-07-15
Fix CONTRIBUTING.md and SECURITY.md copy paths in docs script

zeevick10:
**Description:**  
Updated the documentation script to copy `CONTRIBUTING.md` and `SECURITY.md` to the project root instead of the `getting_started` directory, ensuring correct file placement for documentation builds.


danceratopz:
I'm sorry @zeevick10 but the script works as it is. Your change would break the script.

PR #1914, opened 2025-07-17
fix(fixtures,specs): Fix Engine X Env->Genesis Conversion

marioevz:
### Fix failing tests with modified pre during grouping

Fixes tests that removed contracts from the default fork pre-allocation (required system contracts).

The issue was a `model_post_init` hook that re-applied the fork default pre-allocation when the pre-alloc group was parsed from file: https://github.com/ethereum/execution-spec-tests/blob/9b30723e8af0f966f4de4e617e7f69b7549b6ccd/src/ethereum_test_fixtures/pre_alloc_groups.py#L33-L40

This hook is now removed and we instead make sure to apply this fork default pre-alloc when initializing the group for the first time and never do it again.

### State tests and Blockchain tests with the default `Environment` now fall into the same pre-alloc group

There were a couple of small issues but the main was that the default block number in `Environment` is 1, but the blockchain test in reality uses a hard-coded zero to generate the genesis header. This on top of the fact that the number did not propagate correctly to the method that requested the genesis environment to calculate the pre-alloc group hash.

Now if two tests, one state test and the other blockchain test, use the same default environment (`Environment()`), and they also don't make modifications to the pre-alloc (via account removal or similar), they will land in the same pre-allocation group.

### Removal of unnecessary custom environments in some tests

There were (and still some remain) that use a custom environment without any real reason, these must be removed in order to minimize the number of pre-allocation groups, and the environment should only be touched if it's absolutely necessary.

### Fix mismatches between `env` and `genesis` in pre-alloc group fixtures

Some fields were not exported correctly to the pre-alloc group fixture, namely the `currentCoinbase`, due to the use of defaults and different code paths to generate the genesis environments in state vs blockchain tests.

This is now fixed and unit tests were added to ensure that the values match (see `src/pytest_plugins/filler/tests/test_prealloc_group.py::test_pre_alloc_grouping_by_test_type`).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist
    ```console
    ```


src/ethereum_test_specs/blockchain.py
```diff
@@ -692,7 +710,10 @@ def make_hive_fixture(
         """Create a hive fixture from the blocktest definition."""
         fixture_payloads: List[FixtureEngineNewPayload] = []
 
-        pre, genesis = BlockchainTest.make_genesis(self.genesis_environment, self.pre, fork)
+        pre, genesis = self.make_genesis(
+            fork=fork,
+            apply_pre_allocation_blockchain=fixture_format != BlockchainEngineXFixture,
```
marioevz, line 715: The reason behind this is that, `BlockchainEngineFixture` receives the environment directly from the `genesis_environment` field as it comes from the test, and the tests do not set the fork pre-required system contracts.
On the other hand, `BlockchainEngineXFixture` receives the environment from the pre-allocation group, where the `fork.pre_allocation_blockchain` has already been applied before saving to file (and loading again during this step).

src/pytest_plugins/filler/tests/test_prealloc_group.py
```diff
@@ -203,125 +204,216 @@ def test_pre_alloc_group_with_reason():
...
+            [
+                StateTest(env="Environment(extra_data='0x01')"),
+                BlockchainTest(env="Environment(extra_data='0x02')"),
+            ],
+            2,
+            id="different_extra_data_different_types",
+            marks=pytest.mark.xfail(
+                reason="Extra data is excluded=True in the Environment model, so it does not "
+                "propagate correctly to the genesis header without a lot of code changes.",
+            ),
```
spencer-tb, line 369: Do you think its worth creating an issue for this? I wondered if we could improve the changes in `src/ethereum_test_fixtures/blockchain.py` too surrounding the `extra_data` field.

src/ethereum_test_specs/blockchain.py
```diff
@@ -432,25 +444,31 @@ def discard_fixture_format_by_marks(
             return fixture_format != BlockchainEngineFixture
         return False
 
-    @staticmethod
+    def get_genesis_environment(self, fork: Fork) -> Environment:
+        """Get the genesis environment for pre-allocation groups."""
+        modified_values = self.genesis_environment.set_fork_requirements(fork).model_dump(
+            exclude_unset=True
+        )
+        return Environment(**(GENESIS_ENVIRONMENT_DEFAULTS | modified_values))
```
spencer-tb, line 452: It might be nice to add some basic tests for this in `src/pytest_plugins/filler/tests/test_prealloc_group.py`.

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -219,32 +221,16 @@ def block_hash(self) -> Hash:
...
-        )
+        environment_values = env.model_dump(exclude_none=True, exclude={"withdrawals"})
+        if env.withdrawals is not None:
+            environment_values["withdrawals_root"] = Withdrawal.list_root(env.withdrawals)
+        environment_values["extra_data"] = env.extra_data
+        extras = {
+            "state_root": state_root,
+            "requests_hash": Requests() if fork.header_requests_required(0, 0) else None,
+            "fork": fork,
+        }
```
spencer-tb, line 232: Why do we not add `parent_hash`, `transactions_trie`, `receipts_root`, `logs_bloom`, `gas_used` here - since the env doesn't provide these?

spencer-tb (approved):
LGTM! A couple of comments nonetheless!

src/pytest_plugins/filler/tests/test_prealloc_group.py
marioevz: I think it's not necessary `genesis_environment` is very much useless so unless we find some specific use at some point, I would refrain from spending any time on this.

src/ethereum_test_fixtures/blockchain.py
marioevz: Thanks, this is due to defaults being added to the fields (see `FixtureHeader` field changes in this file).

I removed them to make it more clear what are we adding that is not zeros or defaults.

src/ethereum_test_specs/blockchain.py
marioevz: I feel that's the purpose of `test_pre_alloc_grouping_by_test_type` in that test file, maybe implicitly perhaps, just to verify that they generate the same genesis environment only when needed.

Issue #1919, opened 2025-07-17
chore(benchmark): don't fill benchmark tests by default


spencer-tb:
## Description

EELS is the default t8n tool used to fill tests in EEST. Currently when filling all tests:
```
uv run fill --until Prague -n 10 --clean
```
benchmark tests are included. I propose that we make benchmark tests require the `-m benchmark` marker to be filled.

1) EELS is non-perfomant focused. It takes a long time to fill these tests.
2) Currently we use `-m "not benchmark"` in CI and during localy runs as a work around.

## Proposed Solution

The change in EEST for this should be rather straight forward. Adding the following to the `pyproject.toml`:
```
[tool.pytest.ini_options]
addopts = ["-m", "not benchmark"]
```

## Blockers

Confirmation with the zkevm team (and others using benchmark tests) must be clarified (cc @jsign).

spencer-tb:
Perfect! Thanks for the fast reply :)

PR #1920, opened 2025-07-17
chore(benchmark): don't fill benchmark tests by default

spencer-tb:
## 🗒️ Description

Updates EEST to not fill benchmark tests by default.

## 🔗 Related Issues or PRs

#1919 

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/tests/benchmark/conftest.py
+++ b/tests/benchmark/conftest.py
@@ -12,22 +12,18 @@ def pytest_collection_modifyitems(config, items):
     run_benchmarks = marker_expr and (
         "benchmark" in marker_expr and "not benchmark" not in marker_expr
     )
-
     if gas_benchmark_values:
         run_benchmarks = True
-
     items_for_removal = []
     for i, item in enumerate(items):
-        is_benchmark_test = Path(__file__).parent in Path(item.fspath).parents
-
+        is_in_benchmark_dir = Path(__file__).parent in Path(item.fspath).parents
+        has_benchmark_marker = item.get_closest_marker("benchmark") is not None
+        is_benchmark_test = is_in_benchmark_dir or has_benchmark_marker
         if is_benchmark_test:
-            benchmark_marker = pytest.mark.benchmark
+            if is_in_benchmark_dir and not has_benchmark_marker:
+                benchmark_marker = pytest.mark.benchmark
 
-            if gas_benchmark_values:
-                gas_values = [int(v.strip()) for v in gas_benchmark_values.split(",")]
-                benchmark_marker = pytest.mark.benchmark(gas_values=gas_values)
-
-            item.add_marker(benchmark_marker)
+                item.add_marker(benchmark_marker)
             if not run_benchmarks:
                 items_for_removal.append(i)
 
--- a/tests/benchmark/conftest.py
+++ b/tests/benchmark/conftest.py
@@ -7,6 +7,17 @@ import pytest
 
 def pytest_collection_modifyitems(config, items):
     """Add the `benchmark` marker to all tests under `./tests/benchmark`."""
+    marker_expr = config.getoption("-m", default="")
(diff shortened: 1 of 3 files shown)


spencer-tb:
I'm considering doing the same for `-m slow`...

spencer-tb:
I added a better solution by modifying the benchmark conftest!

tests/benchmark/conftest.py
```diff
@@ -7,6 +7,17 @@
 
 def pytest_collection_modifyitems(config, items):
     """Add the `benchmark` marker to all tests under `./tests/benchmark`."""
+    marker_expr = config.getoption("-m", default="")
```
marioevz, line 10: I think we could even use `--gas-benchmark-values` introduced in #1895, I'm not sure if we can access that option from here but that would be ideal IMO.

tests/benchmark/conftest.py
```diff
@@ -7,6 +7,17 @@
...
+    run_benchmarks = marker_expr and (
+        "benchmark" in marker_expr and "not benchmark" not in marker_expr
+    )
+
+    items_to_remove = []
     for item in items:
```
marioevz, line 16: Let's use the trick from #1936 and use indexes when removing the items, it's significantly faster.

tests/benchmark/conftest.py
```diff
@@ -7,6 +7,17 @@
...
         if Path(__file__).parent in Path(item.fspath).parents:
             item.add_marker(pytest.mark.benchmark)
+            if not run_benchmarks:
```
marioevz, line 19: We could even also do the opposite on top of this: I.e. if we are running in benchmark, remove the tests that are not marked as benchmark tests.

marioevz (commented):
Really good idea! Just a couple of suggestions that would make it even easier to use IMO.

tests/benchmark/conftest.py
```diff
@@ -7,6 +7,32 @@
...
+    items_for_removal = []
+    for i, item in enumerate(items):
+        is_benchmark_test = Path(__file__).parent in Path(item.fspath).parents
+
+        if is_benchmark_test:
+            benchmark_marker = pytest.mark.benchmark
+
+            if gas_benchmark_values:
+                gas_values = [int(v.strip()) for v in gas_benchmark_values.split(",")]
+                benchmark_marker = pytest.mark.benchmark(gas_values=gas_values)
```
marioevz, line 28:
```suggestion
```
I'd lean towards not doing this here because the gas-benchmark-values are a properties of the current test filling execution, rather than the particular test itself.

Plus if we change this parameter's input format, we have to come to this file too to match it and we might miss it.

tests/benchmark/conftest.py
```diff
@@ -7,6 +7,32 @@
...
+    run_benchmarks = marker_expr and (
+        "benchmark" in marker_expr and "not benchmark" not in marker_expr
+    )
+
+    if gas_benchmark_values:
+        run_benchmarks = True
+
+    items_for_removal = []
+    for i, item in enumerate(items):
+        is_benchmark_test = Path(__file__).parent in Path(item.fspath).parents
```
marioevz, line 21: The test could also be a benchmark test if it already had the `pytest.mark.benchmark` (e.g. a benchmark-marked test outside of the `benchmark` folder).

This is not super well defined, but until we do I'd check for the marker too.

marioevz (commented):
Some more comments!

spencer-tb:
I think I addressed your other comments.

Nice way to test:
```
uv run fill --fork Prague -m benchmark --clean --co | grep Module
uv run fill --fork Prague --gas-benchmark-values 1 --clean --co | grep Module
uv run fill --fork Prague --clean --co | grep Module
```

Issue #1791, opened 2025-06-24
feat(tests): additional test cases for eip-7883


spencer-tb:
#### ModExp Gas Cost Increase

- [ ] Add here!

Please comment any additional test cases or edit the issue and add more.

PR #1929, opened 2025-07-21
refactor(eip7883): update vector input structure

LouisTsai-Csie:
## 🗒️ Description

WIP

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```


tests/osaka/eip7883_modexp_gas_increase/conftest.py
```diff
@@ -8,108 +8,191 @@
...
+        code += Op.SSTORE(call_contract_post_storage.store_next(precompile_gas), gas_calculation)
+        # + Op.RETURNDATACOPY(dest_offset=0, offset=0, size=Op.RETURNDATASIZE())
+        # + Op.SSTORE(call_contract_post_storage.store_next(
+        #     keccak256(Bytes(vector.expected))), Op.SHA3(0, Op.RETURNDATASIZE()))
+
+        for i in range(len(modexp_expected) // 32):
+            code += Op.RETURNDATACOPY(0, i * 32, 32)
+            code += Op.SSTORE(
+                call_contract_post_storage.store_next(modexp_expected[i * 32 : (i + 1) * 32]),
+                Op.MLOAD(0),
```
LouisTsai-Csie, line 120: I tried to use a hash-based method (see the comment) here for simplicity but failed, so I simply store all the data in storage for comparison. I still prefer the hash-based method as it reduces the `SSTORE` operation count

tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds_transition.py
```diff
@@ -0,0 +1,86 @@
...
+REFERENCE_SPEC_GIT_PATH = ref_spec_7883.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7883.version
+
+pytestmark = pytest.mark.valid_at_transition_to("Osaka", subsequent_forks=True)
+
+
+@pytest.mark.parametrize(
+    "modexp_input,modexp_expected,gas_old,gas_new",
+    [
+        pytest.param(Spec.modexp_input, Spec.modexp_expected, 200, 500),  # Should be 1200
```
LouisTsai-Csie, line 20: For this test case, it should take 1200 gas cost after Osaka is activated, but in this test case, it only takes 500. Still looking into the root cause here.

This is one is the "", you can find it in the `vectors.json` file

tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
```diff
@@ -36,3 +42,271 @@ def test_vectors_from_file(
...
+            exponent="FF" * (Spec.MAX_LENGTH_BYTES + 1),
+            modulus="FF" * (Spec.MAX_LENGTH_BYTES + 1),
+            case_id="exponent-modulus-too-long",
+        ),
+        create_boundary_modexp_case(
+            base="FF" * (Spec.MAX_LENGTH_BYTES + 1),
+            exponent="FF" * (Spec.MAX_LENGTH_BYTES + 1),
+            modulus="FF" * (Spec.MAX_LENGTH_BYTES + 1),
+            case_id="all-too-long",
+        ),
```
LouisTsai-Csie, line 183: For these test cases, the input exceeds the boundary for the ModExp precompile. However, it’s unclear whether the failure is due to exceeding the transaction gas limit or violating the input boundary constraints.

marioevz: We should double check this and do the actual calculation of the gas required to do this operation, It might even be higher than the tx gas limit cap introduced in Osaka, and in that case the test might be unnecessary.

tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
```diff
@@ -36,3 +42,340 @@ def test_vectors_from_file(
...
+    # │ S4  │  S   │  =  │  D   │ False │  2048   │ Small, equal, large exp+non-zero low256     │
+    # │ S5  │  S   │  >  │  A   │ True  │   500   │ Small, base>mod, zero exp, clamped          │
+    # │ S6  │  S   │  <  │  B   │ True  │   500   │ Small, base<mod, small exp, clamped         │
+    # │ L0  │  L   │  =  │  A   │ True  │   500   │ Large, equal, zero exp, clamped             │
+    # │ L1  │  L   │  =  │  B   │ False │ 12750   │ Large, equal, large exp, unclamped          │
+    # │ L2  │  L   │  =  │  C   │ False │  6350   │ Large, equal, large exp+zero low256         │
+    # │ L3  │  L   │  =  │  D   │ False │  6400   │ Large, equal, large exp+non-zero low256     │
+    # │ L4  │  L   │  >  │  B   │ True  │   500   │ Large, base>mod, small exp, clamped         │
+    # │ L5  │  L   │  <  │  C   │ False │  9144   │ Large, base<mod, large exp+zero low256      │
+    # └─────┴──────┴─────┴──────┴───────┴─────────┴─────────────────────────────────────────────┘
```
LouisTsai-Csie, line 360: To verify the table, you can check this script, compare to eip-7883, and use it ot generate the test case as well as the labels: https://gist.github.com/1c8fd82ac1e75e9cfd1c79e5a2f5fbe6.git

LouisTsai-Csie:
Some ideas from the EIPs:

> If any of these inputs are larger than the limit, the precompile execution stops, returns an error, and consumes all gas.
What would happen if the input is invalid? Does it consume all the gas as well?

tests/byzantium/eip198_modexp_precompile/helpers.py
```diff
@@ -73,6 +73,8 @@ def from_bytes(cls, input_data: Bytes | str) -> "ModExpInput":
 
         modulus = input_data[current_index : current_index + modulus_length]
 
+        modulus = modulus.ljust(min(1024, modulus_length), b"\x00")
```
LouisTsai-Csie, line 76: This change is necessary but requires additional refactoring.

For case 4 in `modexpFiller.json`, the test description is:

> 4 - Would also parse as a base of 3, an exponent of 65535, and a modulus of 2**255. It attempts to read 32 bytes for the modulus starting from 0x80, but since there’s no further data, it right-pads the result with 31 zero bytes.

Although the operation modulus = `input_data[current_index : current_index + modulus_length]` is technically valid, as Python silently pads with null bytes when slicing beyond the end, we actually need to left-pad with `\x00` instead.

Example: https://onecompiler.com/python/43smfe6dm

Regarding the 1024-byte length restriction, in case 2, the exponent length is `2**256 - 1`, which can cause `current_index + modulus_length` to overflow. Therefore, I added this condition to prevent overflow. It aligns with the logic in [EIP-7883](https://eips.ethereum.org/EIPS/eip-7883), although such a restriction does not exist in [EIP-198](https://eips.ethereum.org/EIPS/eip-198).

Some idea for refactoring in `ModexpInput`: We should consider not passing `ModexpInput` data type to the test, but `bytes` data instead, this would be much more flexible for strange testing scenario.

tests/osaka/eip7883_modexp_gas_increase/conftest.py
spencer-tb: I tried to do this too but had issues with `guido-4` and `geth-fail` vectors. I then realised we have a bug in the chunking method for cases where `len(modexp_expected) // 32 = 0`. For `guido-4` this equates to `8 // 32 = 0`. So then we get `for i in range(0)`, meaning we don't store the result for `guido-4`.

We should try and get the hash method working!

tests/osaka/eip7883_modexp_gas_increase/spec.py
```diff
@@ -32,11 +34,23 @@ class Spec:
     LARGE_BASE_MODULUS_MULTIPLIER = 1
     MAX_LENGTH_THRESHOLD = 32
     EXPONENT_BYTE_MULTIPLIER = 8
+    MAX_LENGTH_BYTES = 1024
 
     WORD_SIZE = 8
     EXPONENT_THRESHOLD = 32
     GAS_DIVISOR = 3
 
+    # Test Constants
```
spencer-tb, line 43:
```suggestion
    # Arbitrary Test Constants
```

tests/osaka/eip7883_modexp_gas_increase/conftest.py
spencer-tb: > I update the implementation and then run consume engine command on the server, it now works fine.
> Screenshot 2025-08-12 at 4 30 39 PM

Nice! Thanks for taking care of this

LouisTsai-Csie: I've already updated the test vector. My current understanding also, is that we never really test the actual data for the return data! For the `geth-fail-length`, it would be critical as the output provided is even incorrect.

spencer-tb:
> I didn’t create a separate PR for [issue #1971](https://github.com/ethereum/execution-spec-tests/issues/1971), as it depends on the infrastructure introduced in this PR.
> 
> There are 37 test cases in the legacy `modexpFiller.json` test. Most of them have been ported, but the following cases are still pending and require further analysis:
> 
>     * Case 2: Would parse a base length of 0, a modulus length of 32, and an exponent length of 2**256 - 1, where the base is empty, the modulus is 2**256 - 2 and the exponent is (2**256 - 3) * 256**(2**256 - 33) (yes, that's a really big number). It would then immediately fail, as it's not possible to provide enough gas to make that computation.
> 
>     * Case 28: base length 4TiB
> 
>     * Case 29: exp length 4TiB; returns 0 because mod is zero
> 
>     * Case 30: base and mod have zero-length.  exp's length is 2^255.  Since mod is zero, the result should be zero.
> 
>     * [ ]  Case 36: the input found on 10 Oct. 2017 that overflows the gas calculation
> 
>     * [ ]  Case 37: input found in July 2022, overflows the gas calculation
> 
> 
>     1. Based on the description of case 2, 28, 29 and 30, they will fail as eip7623 introduces the upper bound for each field, we already have similar test in `test_modexp_invalid_inputs`.
> 
>     2. For case 36, 37, I am checking if it is valid after Fusaka.

Added this here: https://github.com/ethereum/execution-spec-tests/issues/1791

Can follow up in a future PR!

tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
```diff
@@ -36,3 +38,322 @@ def test_vectors_from_file(
...
+        ),
+        create_boundary_modexp_case(
+            base="FF" * (Spec.MAX_LENGTH_BYTES + 1),
+            exponent="FF",
+            modulus="FF" * (Spec.MAX_LENGTH_BYTES + 1),
+            case_id="base-modulus-too-long",
+        ),
+    ],
+)
+def test_modexp_boundary_inputs(
```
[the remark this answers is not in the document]
LouisTsai-Csie: I initially created test cases for two scenarios:
1. base/exponent is invalid while modulus is valid.
2. All components are invalid.

However, I later realized that these tests fail not due to the input upper bound, but because they exceed the transaction gas limit cap (see explanation [here](https://github.com/ethereum/execution-spec-tests/pull/1929#discussion_r2241517616)).

While filling each component with `0xFF` bytes makes such scenarios impossible in practice, I believe it may still be possible to construct cases that trigger this condition. I explored how to generate these cases but have not much idea now

tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
LouisTsai-Csie: This is sooooo great! Do you think we've covered all of the edge cases?

tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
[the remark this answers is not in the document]
spencer-tb: Nice! I'll add it as a future addition to the meta issue :)

tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
```diff
@@ -7,24 +7,26 @@
...
-from .helpers import Vector
-from .spec import ref_spec_7883
+from ...byzantium.eip198_modexp_precompile.helpers import ModExpInput
+from .helpers import vectors_from_file
+from .spec import Spec, ref_spec_7883
 
 REFERENCE_SPEC_GIT_PATH = ref_spec_7883.git_path
 REFERENCE_SPEC_VERSION = ref_spec_7883.version
 
-pytestmark = pytest.mark.valid_from("Prague")
```
spencer-tb, line 22: Can we make this `valid_from` Berlin? This means we can run these tests for EIP-2565.

And then add `@pytest.mark.valid_from("Osaka")` to the boundary test (due to limits from 7823).

spencer-tb:
Validated coverage script in #2071.

PR #1936, opened 2025-07-23
fix(plugins): Don't copy test list during collection

marioevz:
## 🗒️ Description
Minor optimization during test collection to skip copying the entire list of tests to modify it.

Required to fill benchmark tests.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    ```
- [x] ~~All: Considered adding an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~

### DIFF
--- a/src/pytest_plugins/execute/execute.py
+++ b/src/pytest_plugins/execute/execute.py
@@ -349,25 +349,30 @@ def pytest_generate_tests(metafunc: pytest.Metafunc):
 
 def pytest_collection_modifyitems(config: pytest.Config, items: List[pytest.Item]):
     """Remove transition tests and add the appropriate execute markers to the test."""
-    for item in items[:]:  # use a copy of the list, as we'll be modifying it
+    items_for_removal = []
+    for i, item in enumerate(items):
         if isinstance(item, EIPSpecTestItem):
             continue
         params: Dict[str, Any] = item.callspec.params  # type: ignore
         if "fork" not in params or params["fork"] is None:
-            items.remove(item)
+            items_for_removal.append(i)
             continue
         fork: Fork = params["fork"]
         spec_type, execute_format = get_spec_format_for_item(params)
         assert issubclass(execute_format, BaseExecute)
         markers = list(item.iter_markers())
         if spec_type.discard_execute_format_by_marks(execute_format, fork, markers):
-            items.remove(item)
+            items_for_removal.append(i)
             continue
         for marker in markers:
             if marker.name == "execute":
                 for mark in marker.args:
                     item.add_marker(mark)
             elif marker.name == "valid_at_transition_to":
-                items.remove(item)
+                items_for_removal.append(i)
+                continue
         if "yul" in item.fixturenames:  # type: ignore
             item.add_marker(pytest.mark.yul_test)
+
+    for i in reversed(items_for_removal):
+        items.pop(i)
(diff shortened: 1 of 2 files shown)


src/pytest_plugins/execute/execute.py
```diff
@@ -349,25 +349,28 @@ def pytest_generate_tests(metafunc: pytest.Metafunc):
 
 def pytest_collection_modifyitems(config: pytest.Config, items: List[pytest.Item]):
     """Remove transition tests and add the appropriate execute markers to the test."""
-    for item in items[:]:  # use a copy of the list, as we'll be modifying it
+    i = 0
+    while i < len(items):
+        item = items[i]
         if isinstance(item, EIPSpecTestItem):
             continue
```
felix314159, line 356: Have you ensured that we can't get stuck in an infinite loop? If you only +1 at the very end but there are cases where it does `continue` earlier..

I don't like both functions this PR touches:
* don't modify a list while iterating over it (if there is no other way at least use deepcopy for safety)
* the logic is too complicated. instead of removing from the list (dangerous) why not inverse the logic and make a new list where we only keep what would not have been deleted in our current logic. we should avoid manual index management

marioevz: I like the second idea, it's much easier to just keep appending, I can try that and see if that approach still let's me fill the benchmark tests.
I would not do deepcopy because the items are pytest test objects and I'm not really sure how they would behave on deepcopy.

marioevz: It seems like we cannot make a copy of the list because pytest really wants that same instance that it passes as a parameter to be the final list of test it will execute.

The culprit of the slowness is probably not the copy but rather the usage of `.remove` because of the way that it goes through every item and compares until it finds it to do the removal.

marioevz:
@felix314159 I reworked the function to make it safer while still skipping copying the list.

I timed the benchmark test collection before and after this commit:
### Before
```
(ethereum-execution-spec-tests) ➜  execution-spec-tests git:(update-all-benchmark-tests) time uv run fill --clean --fork=Prague --evm-bin=evmone-t8n --gas-benchmark-values 1,10,30,45,60,90,120 -m benchmark --collect-only > /dev/null
uv run fill --clean --fork=Prague --evm-bin=evmone-t8n --gas-benchmark-values  185.64s user 0.77s system 99% cpu 3:06.42 total
```

Time: 3:06.42

### After
```
(ethereum-execution-spec-tests) ➜  execution-spec-tests git:(update-all-benchmark-tests) time uv run fill --clean --fork=Prague --evm-bin=evmone-t8n --gas-benchmark-values 1,10,30,45,60,90,120 -m benchmark --collect-only > /dev/null
uv run fill --clean --fork=Prague --evm-bin=evmone-t8n --gas-benchmark-values  27.25s user 0.70s system 99% cpu 27.953 total
```
Time: 27.953

felix314159 (commented):
Nice, gj

src/pytest_plugins/filler/filler.py
```diff
@@ -1102,27 +1102,29 @@ def pytest_collection_modifyitems(
     These can't be handled in this plugins pytest_generate_tests() as the fork
     parametrization occurs in the forks plugin.
     """
```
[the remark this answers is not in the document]
marioevz: Absolutely agree, `FillerTestItem` was just a bad choice, I think we can rework it a bit once we have more time after Fusaka passes.

PR #1941, opened 2025-07-23
feat(tests): remove Blake2b max rounds static test

chfast:
## 🗒️ Description
This test brings no additional coverage and with 10G gas limit is very annoying. It is also included in the regular releases so on the Client side you have to filter it out because it runs to long.

There are also benchmarks for Blake2b, so if you think a case with very high number of rounds is desirable, I can add one more case to https://github.com/ethereum/execution-spec-tests/blob/main/tests/benchmark/test_worst_compute.py#L995.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```


marioevz (approved):
Yes, I agree, this case would make sense if we didn't constrain the transaction gas limit on Osaka, but now it's kinda meaningless.

We should add a test consumes the full block gas limit using many transactions of which each uses the most amount of rounds that fit in the new ~16M gas limit of the transaction.

marioevz:
I would even see the point of removing the entire `stTimeConsuming` folder now, these tests are impossible to fill in Osaka anyway, so they are pretty much useless now.

I made an issue to track the porting of them to benchmark tests here: #1947

What do you think @chfast ?

chfast:
> I would even see the point of removing the entire `stTimeConsuming` folder now, these tests are impossible to fill in Osaka anyway, so they are pretty much useless now.
> 
> I made an issue to track the porting of them to benchmark tests here: #1947
> 
> What do you think @chfast ?

I think the 50000x SHA256 test can also be removed easily because it brings not much new coverage.
The benchmarks for blake2b and sha256 are not perfect at the moment, but this is a separate concern.

Finally, the "sstore combinations" as the exhaustive test for one of the storage net metering EIPs. I don't think it has good replacement currently, but probably should be ported to pytest.

PR #1979, opened 2025-07-31
feat(clis,filler): Add trace types to allow trace analysis and gas optimizations

marioevz:
## 🗒️ Description
Introduces a new flag to the test filler that allows re-running the specified tests with lower gas limits to find the lowest value which still produces the same test output, result, and the exact same traces.

### - Trace Types (efdf63a4d24731491c90d42000bc622d877cbc9b)

Introduces trace types to use pydantic to parse the traces from the transition tool.

The types accept the formats used in evmone and execution-specs, and account for their slight differences.

They also implement "equivalence" check methods that compare two different traces:
- The trace lines are compared and always expected to be equal.
- For each trace line, values including the program counter, the gas cost, the memory size, the stack, depth, refund, opcode name and errors, are compared and expected to be exactly equal.
- The remaining gas value is ignored with the reasoning that the remaining gas depends on the gas limit, and therefore changing it in the slightest would result in two traces that are impossible to compare.
- A "post-processing" mode can be included where the top stack element is ignored if the previous opcode is the `Op.GAS`, which allows more flexibility because most tests only use `Op.GAS` to use it for a `Op.CALL*`.

Downsides:
- If the gas stack value produced by the `Op.GAS` opcode is not consumed in the very next operation, the value propagates and the post processing is ineffective, but this should not be a big issue since most tests that use the gas opcode will consume it immediately on the next CALL* opcode.

### - Gas optimization in state tests (7b1d0bdc6125af6838e46b9a7cc4b3c00679d14e)

Implements a procedure to find the minimum gas-limit value where the test still works and yields the same result and traces.

The algorithm works as follows:
- Execute the test with the current gas limit specified in the test.
- Reduce the gas limit by one, re-execute the test: if the runs are still equivalent proceed, otherwise abort and deem the test impossible to lower its gas.
- Perform a binary search by lowering the gas by half, re-executing the test, if the test fails set the new minimum to the last gas limit value plus 1, if the test passes set the new maximum to the last gas limit value.
- Iterate until the difference between the minimum and maximum is zero, or the minimum is higher than 16,777,216

### - Fill command's flags to enable and configure optimize-gas feature (d178bdbe67d49666574df5e54bf2779a2e8437bb)

Flags `--optimize-gas`, `--optimize-gas-output` and `--optimize-gas-post-processing` are added to the fill command to enable and configure this new feature.

### - modify_static_test_gas_limits.py command (dc4c5fca9659f8ecf80cf34e339e01c6cd651bce)

It's a simple command to take the output from the fill's optimize-gas mode and apply it to static tests.

The command has a dry-run mode and aborts if the static test contains more than one gas value, in order to not override incorrect values.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    ```

### DIFF
--- a/docs/writing_tests/gas_optimization.md
+++ b/docs/writing_tests/gas_optimization.md
@@ -45,6 +45,7 @@ When `enable_post_processing=True` is passed to the `verify_modified_gas_limit` 
 ### When to Use Post-Processing
 
 Use `--optimize-gas-post-processing` when your tests:
+
 - Use the `GAS` opcode to read current gas
 - Have contracts that push gas values onto the stack
 - Would otherwise fail gas optimization due to gas-dependent stack operations


src/ethereum_test_specs/state.py
```diff
@@ -213,6 +280,58 @@ def make_state_test_fixture(
...
+                        base_tool_output=base_tool_output,
+                        fork=fork,
+                        current_gas_limit=current_gas_limit,
+                        pre_alloc=pre_alloc,
+                        env=env,
+                        enable_post_processing=enable_post_processing,
+                    ):
+                        maximum_gas_limit = current_gas_limit
+                    else:
+                        if current_gas_limit > 16_777_216:
```
felix314159, line 318: this is a sneaky constant, it should be defined as capslock constant somewhere where we would search for it and then access it, if we ever change the constant (not that unlikely) it would be much easier if we always use references to the constant defined once

src/ethereum_test_specs/state.py
```diff
@@ -213,6 +280,58 @@ def make_state_test_fixture(
...
+                            raise Exception("Requires more than the minimum 16_777_216 wanted.")
```
felix314159, line 319: the value itself should not be hard-coded as text but instead by dynamically read from the referenced constant

docs/writing_tests/gas_optimization.md
```diff
@@ -0,0 +1,93 @@
...
+
+Specify a custom output file for gas optimization results:
+
+```bash
+uv run fill --optimize-gas --optimize-gas-output=my_gas_results.json path/to/some/test/to/optimize
+```
+
+## Post-Processing Mode
+
+Enable post-processing to handle opcodes that put the current gas in the stack (like `GAS` opcode):
```
felix314159, line 23: > "opcodes that put the current gas in the stack (like `GAS` opcode)"

it would be better if you would put the full list of opcodes that are relevant here. it is also unclear to me what exactly this flag does (it "handles opcodes", but what does that mean). it's also unclear whether this is mandatory or not, because in a later paragraph you refer to it as "optional post-processing". is it even optional when `GAS` opcode is used? i feel like this flag maybe does not have to exist, can you not scan the code for opcodes and then dynamically toggle this when certain opcodes are found? less flags make this gas optimization feature easier to use

docs/writing_tests/gas_optimization.md
```diff
@@ -0,0 +1,93 @@
...
+
+- **Test Efficiency**: Create tests with minimal gas requirements
+- **Gas Analysis**: Understand exact gas costs for specific operations
+- **Regression Testing**: Ensure gas optimizations don't break test correctness
+- **Performance Testing**: Benchmark gas usage across different scenarios
+
+## Limitations
+
+- Only works with state tests (not blockchain tests)
+- Requires trace collection to be enabled
```
felix314159, line 59: When reading this I thought that the command would fail if I don't provide `--trace` but that does not seem to be the case, so it seems to add that automatically?

felix314159:
When I run `uv run fill --optimize-gas ./tests/static/state_tests/stRandom/randomStatetest135Filler.json --clean --fill-static-tests --evm-bin=evmone-t8n` I get 2 failed with `Exception: Impossible to compare.` But from the error message alone it not clear what caused this problem. We should adjust the exception message so that it is clear that the error only occurred because `--optimize-gas` was toggled

PR #1989, opened 2025-08-04
fix(tests): Issues with rlp block limit tests and `--generate-pre-alloc-groups`

fselmo:
## 🗒️ Description

Previously, I was using a random `sender` to fill the "fake" block when estimating the block rlp size. This PR fixes these tests to use the `sender` fixture instead for all transactions, instead of swapping with the real `sender` on the final block.

The issue here is that a signed transaction's signature fields, if they have leading zeros, these will be stripped in the rlp encoding. So we have variability across different senders with different signature fields on the final block rlp.

Not only is it not necessary to use a fake `sender` for the block filling and then swap it, it also leads to issues like these. This is fixed here, as well as always using the block gas limit from the `env`, rather than hard-coding as I originally did during Berlinterop when this test was written (refactor).

## 🔗 Related Issues or PRs

#1975 

## ✅ Checklist

    ```console
    ```


marioevz (approved):
LGTM, thanks!

PR #2007, opened 2025-08-07
feat(consume): Add `consume sync` tests to sync clients after engine tests

fselmo:
## 🗒️ Description

Add a new test type that will send `consume engine` tests to client under test and then have a separate client sync to it. Parametrizes both client under test and sync client using hive client config file.

- Fill e.g. `uv run fill --fork=Osaka -k rlp_limit --clean` (or narrow down to one or two tests)
- Consume e.g. `uv run consume sync -v -s` (`-s` for good logs, `-v` for good test ids)

---

## TODO:

- [x] Add documentation once we agree on the code after reviews
- [x] Add entry to CHANGELOG
- [x] Add some appropriate tests for the new command

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/src/ethereum_test_fixtures/blockchain.py
+++ b/src/ethereum_test_fixtures/blockchain.py
@@ -543,7 +543,6 @@ class BlockchainEngineFixture(BlockchainEngineFixtureCommon):
     genesis: FixtureHeader = Field(..., alias="genesisBlockHeader")
     post_state: Alloc | None = Field(None)
     payloads: List[FixtureEngineNewPayload] = Field(..., alias="engineNewPayloads")
-    sync_payload: FixtureEngineNewPayload | None = None
 
 
 @post_state_validator(alternate_field="post_state_diff")
@@ -571,5 +570,19 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
     payloads: List[FixtureEngineNewPayload] = Field(..., alias="engineNewPayloads")
     """Engine API payloads for blockchain execution."""
 
+
+class BlockchainEngineSyncFixture(BlockchainEngineFixture):
+    """
+    Engine Sync specific test fixture information.
+
+    This fixture format is specifically designed for sync testing where:
+    - The client under test receives all payloads
+    - A sync client attempts to sync from the client under test
+    - Both client types are parametrized from hive client config
+    """
+
+    format_name: ClassVar[str] = "blockchain_test_sync"
+    description: ClassVar[str] = (
+        "Tests that generate a blockchain test fixture for Engine API testing with client sync."
+    )
     sync_payload: FixtureEngineNewPayload | None = None
-    """Optional sync payload for blockchain synchronization."""
--- a/src/ethereum_test_fixtures/blockchain.py
+++ b/src/ethereum_test_fixtures/blockchain.py
@@ -16,6 +16,7 @@ from typing import (
 )
 
 import ethereum_rlp as eth_rlp
+import pytest
 from ethereum_types.numeric import Uint
 from pydantic import AliasChoices, Field, PlainSerializer, computed_field, model_validator
 
@@ -586,3 +587,13 @@ class BlockchainEngineSyncFixture(BlockchainEngineFixture):
         "Tests that generate a blockchain test fixture for Engine API testing with client sync."
     )
     sync_payload: FixtureEngineNewPayload | None = None
+
+    @classmethod
+    def discard_fixture_format_by_marks(
+        cls,
+        fork: Fork,
+        markers: List[pytest.Mark],
+    ) -> bool:
+        """Discard the fixture format based on the provided markers."""
+        marker_names = [m.name for m in markers]
+        return "verify_sync" not in marker_names
--- a/src/ethereum_test_specs/blockchain.py
+++ b/src/ethereum_test_specs/blockchain.py
@@ -446,8 +446,6 @@ class BlockchainTest(BaseTest):
             and "blockchain_test_engine_only" in marker_names
         ):
             return True
-        if fixture_format == BlockchainEngineSyncFixture and "verify_sync" not in marker_names:
-            return True
         return False
 
     def get_genesis_environment(self, fork: Fork) -> Environment:
--- a/src/ethereum_test_specs/blockchain.py
+++ b/src/ethereum_test_specs/blockchain.py
@@ -1,7 +1,7 @@
 """Ethereum blockchain test spec definition and filler."""
 
 from pprint import pprint
-from typing import Any, Callable, ClassVar, Dict, Generator, List, Optional, Sequence, Tuple, Type
+from typing import Any, Callable, ClassVar, Dict, Generator, List, Sequence, Tuple, Type
 
 import pytest
 from pydantic import ConfigDict, Field, field_validator
@@ -33,6 +33,7 @@ from ethereum_test_execution import (
 from ethereum_test_fixtures import (
     BaseFixture,
     BlockchainEngineFixture,
+    BlockchainEngineSyncFixture,
     BlockchainEngineXFixture,
     BlockchainFixture,
     FixtureFormat,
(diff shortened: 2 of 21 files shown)


fselmo:
The logic for checking the `verify_sync` flag is a little tricky. We check if the flag exists at all on the fixture and then later, inside `generate()`, when we can actually get the value, we either skip the test if it's set to `False`, as in this case with invalid payload, or we generate it.

I couldn't figure out a good way to completely skip test filling, instead of `pytest.skip()`, if the value is `False` because by the time we are checking the value we have the fixture already. I think this is fine but lmk if you see it another way!

src/ethereum_test_specs/blockchain.py
```diff
@@ -437,10 +442,17 @@ def discard_fixture_format_by_marks(
...
+        marker_names = [m.name for m in markers]
+
+        if "blockchain_test_only" in marker_names:
             return fixture_format != BlockchainFixture
-        if "blockchain_test_engine_only" in [m.name for m in markers]:
+        if "blockchain_test_engine_only" in marker_names:
             return fixture_format != BlockchainEngineFixture
+
+        # Exception tests cannot be used for sync testing
+        if "exception_test" in marker_names and fixture_format == BlockchainEngineSyncFixture:
```
fselmo, line 453: This logic deviates a bit from the existing logic. Perhaps this belongs somewhere else? The idea here is to entirely skip filling for sync tests that are marked as `exception_test` instead of having a `pytest.mark.skip()` because it is also marked as `verify_sync=False`- e.g. We mark this test as having an exception with this parameter [here](https://github.com/fselmo/execution-spec-tests/blob/7041b1d66a638c9efc18d030cc51c07702735627/tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py#L389-L390) but also have to set `False` [here](https://github.com/fselmo/execution-spec-tests/blob/7041b1d66a638c9efc18d030cc51c07702735627/tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py#L440-L441). This at least gets us to entirely skip filling because we can look for the `exception_test` marker even though we can't look at the `verify_sync` value. Does this make sense?

The unfortunate part of this flow is that we don't have access to the value of `verify_sync` until the fixture is filled so we sort of have to dance around it and mark it as a `pytest.mark.skip()` (e.g. [here](https://github.com/fselmo/execution-spec-tests/blob/7041b1d66a638c9efc18d030cc51c07702735627/src/ethereum_test_specs/blockchain.py#L839-L840)).

src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
```diff
@@ -0,0 +1,540 @@
...
+                            )
+                        last_valid_block_hash = payload.params[0].block_hash
+
+        logger.info("All payloads processed successfully on client under test.")
+
+    # sync_payload creates the final block that the sync client will sync to
+    if not fixture.sync_payload:
+        pytest.fail("Sync tests require a syncPayload that is not present in this test.")
+
+    with timing_data.time("Send sync payload to client under test"):
```
marioevz, line 271: I think this could be reworded because "client under test" I thought at first that was the client that performs the sync.

src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
```diff
@@ -0,0 +1,540 @@
...
+            if time.time() - last_forkchoice_time >= forkchoice_interval:
+                try:
+                    # Send forkchoice update to sync client to trigger/maintain sync
+                    assert sync_engine_rpc is not None, "sync_engine_rpc is required"
+                    sync_fc_response = sync_engine_rpc.forkchoice_updated(
+                        forkchoice_state=ForkchoiceState(
+                            head_block_hash=last_valid_block_hash,
+                            safe_block_hash=last_valid_block_hash,
+                            finalized_block_hash=fixture.genesis.block_hash,
+                        ),
```
marioevz, line 457: Tiny optimization is that this can be saved in a variable and reused.

src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
```diff
@@ -0,0 +1,540 @@
...
+                        payload_attributes=None,
+                        version=fixture.sync_payload.forkchoice_updated_version
+                        if fixture.sync_payload
+                        else fixture.payloads[-1].forkchoice_updated_version,
+                    )
+                    status = sync_fc_response.payload_status.status
```
marioevz, line 463: If this is "VALID" we can break the loop, it means it's synced.

src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
```diff
@@ -0,0 +1,540 @@
...
+                    logger.debug(f"Periodic forkchoice update status: {status}")
+                    last_forkchoice_time = time.time()
+                except Exception as fc_err:
+                    logger.debug(f"Periodic forkchoice update failed: {fc_err}")
+
+            # Check sync progress periodically
+            if time.time() - last_check_time >= sync_check_interval:
```
marioevz, line 470: This second block feels to me that should be outside of the loop once we got "VALID" from the forkchoice updated (no need to check multiple times if we are still getting "SYNCING", once we get "VALID" it's only a one-time check I think).

marioevz (commented):
Looks fantastic! Just a couple of comments, main one being that we should make the refactor of the `verify_sync` field in the `BlockchainTest` spec.

fselmo:
> main one being that we should make the refactor of the `verify_sync` field in the `BlockchainTest` spec.

Ok yeah, this is the best way I think as well. I'm glad you agree. I was wondering if we wanted to try to keep it this way for backward compatibility but since no tests were turned on yet for this, I think this is the way to go. Will go this route 👍🏼

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -568,3 +568,24 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
 
     sync_payload: FixtureEngineNewPayload | None = None
     """Optional sync payload for blockchain synchronization."""
+
+
+class BlockchainEngineSyncFixture(BlockchainEngineFixtureCommon):
```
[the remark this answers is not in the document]
fselmo: Yep! This works well 👍🏼

src/ethereum_test_specs/state.py
```diff
@@ -61,6 +61,11 @@ class StateTest(BaseTest):
             f"A {fixture_format.format_name} generated from a state_test",
         )
         for fixture_format in BlockchainTest.supported_fixture_formats
+        # Exclude sync fixtures from state tests - they don't make sense for state tests
+        if not (
+            (hasattr(fixture_format, "__name__") and "Sync" in fixture_format.__name__)
+            or (hasattr(fixture_format, "format") and "Sync" in fixture_format.format.__name__)
+        )
```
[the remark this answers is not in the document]
fselmo: I didn't see a way to remove this 👀 . Fails for `pytest -k fixture_output_based`.

src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
fselmo: Hmm I've documented it throughout the file and logging as "client under test" (receiving the test payloads) and the "sync client". Let me know if you prefer something else here.

fselmo:
@marioevz Updated this to reflect your comments. I had some unideal paths to get to where we got and rebasing `main` with those changes didn't really add a whole lot to the git history so I squashed this all as a single commit. Hopefully that didn't make it too hard to track the changes here.

src/ethereum_test_specs/state.py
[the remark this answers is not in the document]
fselmo: Ah nice addition! Yeah that makes sense 👍🏼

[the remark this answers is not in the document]
fselmo: Wait but this still only checks the marks on `BlockchainEngineSyncFixture` which if it's being generated _from_ a state test will still yield `True`. I don't think there's a solution inside of `BlockchainEngineSyncFixture.discard_fixture_format_by_marks` that takes into account the fact that it's being generated from a state test.

Unless I'm missing something.

marioevz (approved):
Awesome! Thanks so much

Issue #1794, opened 2025-06-24
feat(tests): additional test cases for eip-7951


spencer-tb:
#### Precompile for secp256r1 Curve Support

**Initcode Context**
- [ ] From CODECOPY / EXTCODECOPY - Does not apply? -kc
- [x] Contract creation transaction - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/main/tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py#L366-L409)
- [x] Contract creation via CREATE & CREATE2 - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/main/tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py#L429-L487)

**Execution Context**
- [x] CALL, STATICCALL, DELEGATECALL, CALLCODE - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/0f77f07a8537c447e1113547975fe9b9b4378baf/tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py#L158)
- [x] setCode transaction (eip-7702) - [implementation](https://github.com/ethereum/execution-spec-tests/blob/0f77f07a8537c447e1113547975fe9b9b4378baf/tests/prague/eip7702_set_code_tx/test_set_code_txs.py#L2545)
- [x] Should never revert - Implicitly checked

**Gas Cost**
- [x] Not enough gas - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/0f77f07a8537c447e1113547975fe9b9b4378baf/tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py#L119)
- [x] Exact gas cost (6900 unit) - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/0f77f07a8537c447e1113547975fe9b9b4378baf/tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py#L36)
- [x] Extra gas - [implementation](https://github.com/ethereum/execution-spec-tests/blob/0f77f07a8537c447e1113547975fe9b9b4378baf/tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py#L112)
- [ ] 63/64th check
- [x] zero gas - [Implementation](https://github.com/ethereum/execution-spec-tests/blob/main/tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py#L155-L161)
[...]

PR #2023, opened 2025-08-11
feat(tests): Add more cases for EIP-7951

kclowes:
## 🗒️ Description

Adds a couple more test cases for EIP-7951, mostly from the hardening call. Commit messages should be modular, but specifically adds more gas testing, more invalid input testing, precompile warming, and adds a check for the return status. 

## 🔗 Related Issues or PRs
#1794 

## ✅ Checklist

    ```console
    ```


tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py
```diff
@@ -123,13 +152,39 @@ def test_invalid(state_test: StateTestFiller, pre: Alloc, post: dict, tx: Transa
...
+    tx: Transaction,
+    call_contract_post_storage: Storage,
+    call_succeeds: bool,
+):
     """Test P256Verify precompile gas requirements."""
+    # Assert that the return status is zero if the call runs out of gas, but doesn't revert
+    if call_succeeds:
+        assert call_contract_post_storage[0] == 1
+    else:
+        assert call_contract_post_storage[0] == 0
```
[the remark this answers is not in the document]
marioevz: The fixture flow (in `tests/osaka/eip7951_p256verify_precompiles/conftest.py`) is that:
- `call_contract_post_storage` is instantiated empty in `call_contract_post_storage()`
- `call_contract_post_storage` and `call_succeeds` are used as inputs in `call_contract_code()` to set the appropriate expectation in the storage.
- `post` is built with `call_contract_address` and `call_contract_post_storage` as inputs

This all happens in `conftest.py` auto-magically 🪄.

I would lean towards removing this specific new assertion in this function though because is disjointed from what's happening inside of `call_contract_code`, so if we modify that fixture and add some other verification to the storage, and say, the call return code is no longer at position zero of the storage for example, this assertion would fail and could be a bit annoying to debug.

tests/berlin/eip2929_gas_cost_increases/test_precompile_warming.py
```diff
@@ -40,15 +40,27 @@ def precompile_addresses_in_predecessor_successor(
             boolean indicating whether the address has existed in the predecessor.
 
     """
+    precompile_range = range(0x01, 0x100)
```
marioevz, line 43: I don't quite understand this constrained range, I think this breaks if for some reason a precompile gets added to `0x101` and beyond?

marioevz (commented):
Added cases look very good to me! Just a couple of comments!

tests/berlin/eip2929_gas_cost_increases/test_precompile_warming.py
kclowes: I added this because I don't think the original Ethereum reserved precompile range has changed (but might be wrong on that - please correct me if I'm wrong). AIUI, `0x01-0xff` are still reserved for Ethereum precompiles, and then `0x100-0x1ff` are reserved for L2 precompiles. 

I wanted to make sure that the original precompile address range are warm and cold appropriately, like it was testing before. Since we added the precompile at 0x100, this test was testing the extra range of `0x100 + 32`, rather than the originally intended reserved precompile addresses from `0x11 + 32`. Now this tests the highest from the original range (as of now: `0x11 + 32`), and `0x100 + 32` as well.  I don't think it'll break because we are still getting the highest overall precompile [here](https://github.com/ethereum/execution-spec-tests/pull/2023/files#diff-c198330bea835ddf3b7ef5a8c7fa2cfe80773f23e55ea2d3495c13db49478323R55). 

If you feel like that's overkill, I'm happy to remove :) And let me know if this needs more clarification too!

PR #2040, opened 2025-08-14
feat(tests): add first few single-opcode test for state access in BloatNet

gballet:
## 🗒️ Description

Add a test required as part of the [BloatNet](https://bloatnet.info) effort. This is the 

## 🔗 Related Issues or PRs

Not an issue, but a test plan is described [here](https://hackmd.io/9icZeLN7R0Sk5mIjKlZAHQ?both#Single-opcode-tests-for-state-access-opcodes).

## ✅ Checklist

    ```console
    ```
- [ ] ~~Tests: For PRs implementing a missed test case, update the [post-mortem document](/ethereum/execution-spec-tests/blob/main/docs/writing_tests/post_mortems.md) to add an entry the list.~~
- [ ] ~~Ported Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) or [tests/static](/ethereum/execution-spec-tests/blob/main/tests/static) have been assigned `@ported_from` marker.~~


tests/prague/eip8047_bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,75 @@
...
+            sstore_code = sstore_code + Op.PUSH1(i)
+        else:
+            sstore_code = sstore_code + Op.PUSH2(i)
+
+        sstore_code = sstore_code + Op.SSTORE(unchecked=True)
+
+        storage[storage_slot] = 0x02 << 248
+        storage_slot += 1
+        i += 1
+    sstore_code = sstore_code + Op.POP # Drop last value on the stack
```
LouisTsai-Csie, line 49: My understanding is that here we want to update storage. Maybe we can try something like:

```python
setup = Op.CALLDATALOAD(0)

available_gas = (
    fork.transaction_gas_limit_cap()
    # intrinsic gas cost
    - fork.transaction_intrinsic_cost_calculator()(
        calldata=code,
    )
    # gas cost for push0 and calldataload
    - fork.gas_costs().G_BASE * 2
)

# gas cost for each storage operation
iteration_count = available_gas // (
    fork.gas_costs().G_VERY_LOW * 2
    + fork.gas_costs().G_STORAGE_SET
    + fork.gas_costs().G_COLD_SLOAD
)

code = setup + sum(Op.SSTORE(i, Op.DUP1) for i in range(iteration_count))
assert(len(code) - fork.max_code_size)
```
For `SSTORE` operation, it will automatically adjust to different push variants based on your data size

I am not sure if this applies to this case, but we use [hash](https://github.com/ethereum/execution-spec-tests/blob/5220d21ba64a9a4cbdc7e4025a7f1a032a957c78/tests/benchmark/test_worst_blocks.py#L265) function for storage key to make it not consecutive in benchmarking access list tx.

tests/prague/eip8047_bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,75 @@
...
+
+import pytest
+
+from ethereum_test_forks import Fork
+from ethereum_test_tools import Account, Alloc, Block, BlockchainTestFiller, Transaction
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "DUMMY/eip-DUMMY.md"
+REFERENCE_SPEC_VERSION = "0.1"
+GAS_LIMIT = 30_000_000   # Default gas limit seems to be >90M in this env
```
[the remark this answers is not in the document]
LouisTsai-Csie: I guess what you found is the following fixture:

```python
BENCHMARKING_MAX_GAS = 1_000_000_000_000

@pytest.fixture
def env(request: pytest.FixtureRequest) -> Environment:  # noqa: D103
    """Return an Environment instance with appropriate gas limit based on test type."""
    if request.node.get_closest_marker("benchmark") is not None:
        return Environment(gas_limit=BENCHMARKING_MAX_GAS)
    return Environment()
```

The value is not the actual benchmarking value we are using, currently we are testing under `1,10,30,45,60,100,150M` gas limit, in the past, we create different genesis file with these configuration (so for each config, we need a corresponding genesis file.) but now we set the gas limit to a extremely high value and then restrict the block gas limit usage to these configuration, so that we only need one genesis file.

LouisTsai-Csie:
Our documentation is incomplete (I will fix them ASAP), for running the test, you will need to add a flag `-m benchmark` to run the test under the `benchmark/` folder. By default, these tests are ignored to avoid some overhead in the CI/release process

tests/prague/eip8047_bloatnet/test_bloatnet.py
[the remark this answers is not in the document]
gballet: nvm, found it

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,85 @@
...
+    Alloc,
+    Block,
+    BlockchainTestFiller,
+    Environment,
+    Storage,
+    Transaction,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "DUMMY/eip-DUMMY.md"
```
gballet, line 20: @fselmo @LouisTsai-Csie this is not related to an EIP, do I need to keep it?

tests/prague/eip8047_bloatnet/test_bloatnet.py
gballet: The `sum() for i in range()` won't work, because we also collect the locations that are touched, although since it grows linearly, we could simply do that in a separate loop. It feels like it'll be slower, but you might know better.
  
 Question: is it really correct to write `Op.SSTORE(i, Op.DUP1)`? sounds too good to be true, so I want to double-check.

gballet: found other instances, ok, that's amazing :heart:

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,168 @@
...
+    # Get gas costs for the current fork
+    gas_costs = fork.gas_costs()
+
+    # this is only used for computing the intinsic gas
+    data = final_storage_value.to_bytes(32, "big").rstrip(b"\x00")
+
+    storage = Storage()
+
+    # Initial gas for PUSH0 + CALLDATALOAD + POP (at the end)
+    totalgas = gas_costs.G_BASE * 2 + gas_costs.G_VERY_LOW
```
fselmo, line 45: nit: we can define this once and still keep the comment if we need it:

```py
totalgas = (
    # Initial gas for PUSH0 + CALLDATALOAD + POP (at the end)
    gas_costs.G_BASE * 2
    + gas_costs.G_VERY_LOW
    # intrinsic cost
    + fork.transaction_intrinsic_cost_calculator()(calldata=data)
)
```

fselmo:
I just wanted to add a bit more context on the `gas_benchmark_value`. This allows us to run something like: 

`uv run fill --fork=Prague -m benchmark --gas-benchmark-values 1,10,30,45,100,150 --clean -k bloatnet`

This allows us to test against the different gas limit values specified. I am looking a bit deeper into the PR next but wanted to provide some better context.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,168 @@
...
+
+@pytest.mark.valid_from("Prague")
+@pytest.mark.parametrize("final_storage_value", [0x02 << 248, 0x02])
+def test_bloatnet(
+    blockchain_test: BlockchainTestFiller, pre: Alloc, fork: Fork, final_storage_value: int
+):
+    """
+    A test that calls a contract with many SSTOREs.
+
+    The first block will have many SSTORES that go from 0 -> 1
```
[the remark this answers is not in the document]
fselmo: @gballet I iterated on this quite a bit. I wanted to get familiar with the benchmark tests. If this isn't the route, no worries but I but up a patch for what I was thinking against your branch [here](https://github.com/gballet/execution-spec-tests/pull/1). Let me know, I only touched the `sstore` tests but I also realized this wasn't very straightforward to set up with tx gas cap.

We can run those tests with `uv run fill --fork=Osaka -m benchmark --gas-benchmark-values 1,10,30,45,60,100,150 --clean -k bloatnet_sstore` for example, though. Which I think is the preferred way to run benchmark tests but @LouisTsai-Csie would know best and I believe he will be taking a look here soon.

Hope I helped in some way 😅 .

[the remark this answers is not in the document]
fselmo: Oh, also... If you do think that patch makes sense and you merge it, you can ignore quite a few of the comments I made here as I implemented those and got the CI passing as well.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,80 @@
...
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "DUMMY/eip-DUMMY.md"
+REFERENCE_SPEC_VERSION = "0.1"
+
+
+@pytest.mark.valid_from("Prague")
+@pytest.mark.parametrize("final_storage_value", [0x02 << 248, 0x02])
+def test_bloatnet(
+    blockchain_test: BlockchainTestFiller, pre: Alloc, fork: Fork, final_storage_value: int
```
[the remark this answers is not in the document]
gballet: The spec says "fill the block", so this is what I'm doing. It's not clear to me what that value is?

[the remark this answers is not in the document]
LouisTsai-Csie: Reposted from mattermost.

There are two variables/terms that can be confusing in the context of benchmarking:

- `Environment().gas_limit`
- `gas_benchmark_value`

In `blockchain_test` benchmarking, our original idea is to fill a block up to the block gas limit using a specific opcode/precompile/operation. But the question is: what should we use as the gas limit in our framework?

You might think it’s `Environment().gas_limit`, but that’s not correct. We use a small workaround here instead of consuming all the gas in the block, we define a value called `gas_benchmark_value` and only consume gas up to this amount for benchmarking purposes.

When running tests with `-m benchmark`, you’ll see two fixtures:

- [genesis_environment](https://github.com/ethereum/execution-spec-tests/blob/1b7092cfa2c6856e879cf12b60bc684462a1a2dc/src/pytest_plugins/shared/benchmarking.py#L57)
- [env](https://github.com/ethereum/execution-spec-tests/blob/1b7092cfa2c6856e879cf12b60bc684462a1a2dc/src/pytest_plugins/shared/benchmarking.py#L65)
In both, you can see the gas_limit is set to 1_000_000_000_000. This is a fixed value and not our benchmark target (e.g., 45M, 60M, 90M). The actual benchmark value is `gas_benchmark_value`, which we pass via the flag: `--gas-benchmark-values 10M,45M,60M`. In this case, it will create 3 different tests with respective `gas_benchmark_value`. And this value should then be used as the effective gas limit in your test.

Again, we’re not literally consuming all the gas in the block. This is a workaround for how the genesis file is generated:

- The genesis file always sets `gasLimit` to `1_000_000_000_000` during benchmarking.
- The `gasUsed` field is set to the chosen `gas_benchmark_value`.
- 
This approach lets us use a single genesis file for the Nethermind team. To change the benchmark gas value, we only need to update gasUsed rather than regenerate a new genesis file for every configuration.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,165 @@
...
+    post = {contract_address: Account(storage=storage)}
+
+    blockchain_test(pre=pre, blocks=[Block(txs=[tx_0_1]), Block(txs=[tx_1_2])], post=post)
+
+
+# Warm reads are very cheap, which means you can really fill a block
+# with them. Only fill the block by a factor of SPEEDUP.
+SPEEDUP: int = 100
+
+
```
jsign, line 86: We already have covered cold and warm SSTORE/SLOAD with same and different values here https://github.com/ethereum/execution-spec-tests/blob/main/tests/benchmark/test_worst_stateful_opcodes.py#L393

Are these tests different in some way?

jsign: @gballet, the tests in `benchmark` are not specific to zkvms. (They are even used for PerfNet).

The test I linked are executing blocks where the full gas limit is used to do cold or warm reads, or to write slots to existent or non-existent storage slots. There's nothing specific to zkvms there, thus why I ask how these tests are different.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,382 @@
...
+    """Test that loads many different cold storage locations."""
+    gas_costs = fork.gas_costs()
+    intrinsic_gas_calc = fork.transaction_intrinsic_cost_calculator()
+    tx_gas_cap = fork.transaction_gas_limit_cap() or gas_benchmark_value
+
+    # PUSH + Cold SLOAD + POP
+    cold_sload_cost = gas_costs.G_VERY_LOW + gas_costs.G_COLD_SLOAD + gas_costs.G_BASE
+
+    # Calculate max slots that fit in a single transaction
+    intrinsic_gas = intrinsic_gas_calc(calldata=b"")
```
LouisTsai-Csie, line 326:
```suggestion
    intrinsic_gas = intrinsic_gas_calc()
```

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,382 @@
...
+            break
+
+        # Calculate how many warm loads we can actually do
+        remaining_budget = actual_opcode_budget - warmup_gas
+        actual_warm_loads = min(num_warm_loads_per_tx, remaining_budget // warm_sload_cost)
+
+        tx = Transaction(
+            to=contract_address,
+            gas_limit=tx_gas_limit,
+            data=b"",
```
LouisTsai-Csie, line 296:
```suggestion
```

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,382 @@
...
+    # Calculate total slots needed to fill the benchmark
+    total_max_slots = gas_benchmark_value // cold_sload_cost
+
+    # Build storage for all slots we'll potentially access
+    storage = Storage({HashInt(i): HashInt(0xC0FFEE + i) for i in range(total_max_slots)})
+
+    # Build code that loads max_slots_per_tx cold slots
+    sload_code = Bytecode()
+    for i in range(max_slots_per_tx):
+        sload_code = sload_code + Op.SLOAD(i) + Op.POP
```
LouisTsai-Csie, line 339:
```suggestion
    sload_code = sum(Op.POP(Op.SLOAD(i)) for i in range(max_slots_per_tx))
```

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,382 @@
...
+    # Build the code that does warm loads
+    # Since storage slots are already warmed within each transaction,
+    # we just need to access them repeatedly
+    sload_code = Bytecode()
+    # First warm up all slots (these will be cold on first access in each tx)
+    for i in range(num_slots):
+        sload_code = sload_code + Op.SLOAD(i) + Op.POP
+    # Then do warm loads
+    for i in range(num_warm_loads_per_tx):
+        sload_code = sload_code + Op.SLOAD(i % num_slots) + Op.POP
```
LouisTsai-Csie, line 262: We could refactor to the following:

```suggestion
sload_code = Bytecode()
sload_code += sum(Op.POP(Op.SLOAD(i)) for i in range(num_slots))
sload_code += sum(Op.POP(Op.SLOAD(i % num_slots)) for i in range(num_warm_loads_per_tx))
```

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,382 @@
...
+        actual_opcode_budget = tx_gas_limit - intrinsic_gas
+        actual_slots = min(max_slots_per_tx, actual_opcode_budget // cold_sload_cost)
+
+        if actual_slots == 0:
+            break
+
+        tx = Transaction(
+            to=contract_address,
+            gas_limit=tx_gas_limit,
+            data=b"",
```
LouisTsai-Csie, line 368:
```suggestion
```

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,382 @@
...
+    # Build code that loads max_slots_per_tx cold slots
+    sload_code = Bytecode()
+    for i in range(max_slots_per_tx):
+        sload_code = sload_code + Op.SLOAD(i) + Op.POP
+
+    # Deploy the contract once
+    contract_address = pre.deploy_contract(
+        code=sload_code,
+        storage=storage,
+    )
```
LouisTsai-Csie, line 345: @marioevz I read your [comment](https://github.com/ethereum/execution-spec-tests/pull/2090#discussion_r2316976271) regarding the execute mode. I wonder if we should also avoid this way to configure the storage?

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,382 @@
...
+
+        # Pre-fill storage with initial values
+        initial_storage = {
+            slot: initial_value for slot in range(total_sstores, total_sstores + tx_sstores_count)
+        }
+
+        contract_address = pre.deploy_contract(
+            code=tx_contract_code,
+            storage=initial_storage,  # type: ignore
+        )
```
LouisTsai-Csie, line 186: I have the same question here - can we configure storage in this way for `execute` command? cc @marioevz

LouisTsai-Csie (changes requested):
There is some refactoring for the code, please help update them, thanks!

Let's wait for the answer of the `execute` mode for bloatnet scenario.

tests/benchmark/test_bloatnet.py
gballet: I consider it has already been addressed by my first comment: it's not the same bytecode, not the same objectives (worst case vs perf), and not the same constraints. Ignacio's tests might not be specific to zkvms, but we do need something specific for us.

fselmo (commented):
I left some things to think about.

For Osaka, when the `tx_gas_cap` is present, I don't think the `SLOAD` tests are doing what, as I read the code, is the intention of the test. We build up enough transactions to max out the `SLOAD`s per transaction - but when we reach the last tx, all the contract knows to do is the same static amount of SLOADs and it can't do any less. I think these could benefit from the similar flow of the `STORE` tests where we can actually control the slots as separate contracts per transaction and we can actually get to some higher numbers with these tests.

Let me know if I missed something and let me know your thoughts on this.

LouisTsai-Csie:
I’ll close this PR for now.
Feel free to open a new one, if you plan to include these changes in a superset later.
Thanks @CPerezz for the update!

PR #2041, opened 2025-08-14
fix(filler,tests): Fix failing static tests

marioevz:
## 🗒️ Description

Fixes all remaining failing static tests for v5.0.0 with the following changes:

### Handle `pre_alloc_group` mark in static tests

Adds a special case for handling the `pre_alloc_group` mark when included in static tests (the marker requires a positional parameter).

In the future we could support markers with parameters in static tests but it's not urgent and could be overkill:

```json
        "_info" : {
            "comment": "Requires a separate pre-alloc group due to time required to fill when grouped with other tests.",
            "pytest_marks": ["pre_alloc_group(\"parameter 1\", \"parameter 2\")"]
        },
```

### tests/static/state_tests/stStaticCall/static_LoopCallsThenRevertFiller.json

This test was taking more than 10 minutes to fill with only 10M gas limit in the transaction when grouped with more tests (big pre-alloc).

The test is now marked as `pytest.mark.pre_alloc_group("separate")` to avoid this.

### tests/static/state_tests/stZeroCallsTest/ZeroValue_SUICIDE_ToEmpty_ParisFiller.json

The test checked a tagged address in the post and since this was an address that was determined randomly, there was 1 in 16 chance that the first byte is zero and therefore the lllc code would use PUSH19 instead of PUSH20.

A couple of things:
- The test was only failing in Prague and with `BlockchainTest` format.
- Also was only failing when the pre-alloc groups were enabled.

The reason is that, when the pre-alloc groups are enabled, the `pre.deploy_contract` function switches from assigning address as 0x1000, 0x1100, ..., to using a hash of the test ID to have deterministic addresses during the pre-alloc group generation that (1) match when we run the test again to fill it and (2) the account addresses don't collide between tests (hence the usage of the test id as entropy source).

Since the test id contains `blockchain_test` and `Prague`, this is the reason only this instance of the test was failing because that's the only address of which its hash had a zero at the beginning.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    ```
- [ ] ~~All: Considered adding an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~

### DIFF
--- a/src/pytest_plugins/filler/pre_alloc.py
+++ b/src/pytest_plugins/filler/pre_alloc.py
@@ -338,7 +338,9 @@ def node_id_for_entropy(request: pytest.FixtureRequest, fork: Fork | None) -> st
         fork = request.node.fork
     for fixture_format_name in ALL_FIXTURE_FORMAT_NAMES:
         if fixture_format_name in node_id:
-            test_file_path, test_name = request.node.nodeid.split("::")
+            parts = request.node.nodeid.split("::")
+            test_file_path = parts[0]
+            test_name = "::".join(parts[1:])
             stripped_test_name = test_name.replace(fixture_format_name, "").replace(
                 fork.name(), ""
             )


src/pytest_plugins/filler/pre_alloc.py
```diff
@@ -298,18 +301,66 @@ def sha256_from_string(s: str) -> int:
...
+    test as source of entropy to get a deterministic address when generating the pre-alloc
+    grouping.
+
+    Removing the fixture format and the fork name from the node id before hashing results in the
+    contracts and senders addresses being the same across fixture types and forks for the same
+    test.
+    """
+    node_id: str = request.node.nodeid
+    if fork is None:
+        # FIXME: Static tests don't have a fork, so we need to get it from the node.
```
spencer-tb, line 336:
```suggestion
        # FIXME: Static tests don't have a fork, so we need to get it from the node.
```
Should we create an issue?

src/pytest_plugins/filler/pre_alloc.py
```diff
@@ -298,18 +301,66 @@ def sha256_from_string(s: str) -> int:
...
+        assert hasattr(request.node, "fork")
+        fork = request.node.fork
+    for fixture_format_name in ALL_FIXTURE_FORMAT_NAMES:
+        if fixture_format_name in node_id:
+            test_file_path, test_name = request.node.nodeid.split("::")
```
spencer-tb, line 341: This seems to be the cause of the CI failure. Taking a look.

src/pytest_plugins/filler/pre_alloc.py
marioevz: I'm not sure if it's solvable tbh, this is a quirk of how we hack pytest to do static tests, it's not priority IMO but we can keep it in mind.

Issue #2049, opened 2025-08-19
feat(tests): additional test cases for eip-7910


LouisTsai-Csie:
## Summary

EIP: https://eips.ethereum.org/EIPS/eip-7910

## Implementation Progress

There are three different status:

| Client             | Branch                               | Progress | Note |
|--------------|--------------------------|----------|------|
| Nethermind   | fusaka-devnet-3, main      |          |  testing new release    |
| Besu               | fusaka-devnet-3, main    | Done     |      |
| Reth                | fusaka-devnet-3              |          |      |
| Geth                | fusaka-devnet-3, [branch](https://github.com/lightclient/go-ethereum/tree/ethapi-add-config) |          |  PR created, pending review    |
| Erigon             | fusaka-devnet-3, main    | Done     |      |
|Nimbus-el       | fusaka-devnet-3, main                                   |               |      |

---
## Verification Guideline

For each execution client, we either fetch the docker image from docker Hub or build it from the release files. We then run the client using the Mainnet configuration, with JSON-RPC enabled, but without syncing the state.

**Step 1**: Start the client docker container

Each client is started in a Docker container using `--p2p-enabled=false` or related flag ensuring it does not sync with the network. The JSON-RPC (`eth_`) endpoints are enabled and exposed for local testing.

**Step 2**: Verify `eth_config` RPC Implementation
After the client is running, use the following command to check whether `eth_config` is implemented:
```
curl -X POST -H "Content-Type: application/json"  --data '{"jsonrpc":"2.0","method":"eth_config","id":1}' http://localhost:8545
```
You should receive a JSON response describing the client’s configuration parameters.

Step 3: Validate the RPC response via `execute`
```
uv run execute eth-config --network Mainnet --rpc-endpoint http://localhost:8545
```

[Reference](https://eest.ethereum.org/main/running_tests/execute/eth_config/?h=eth_config#eth_config-expected-response-details)

---

### Nethermind

**Command**

```
docker build https://github.com/nethermindeth/nethermind.git#1.32.4 -t nethermind:1.32.4

docker run --rm -it -p 8545:8545 nethermind:1.32.4 \
  --config mainnet \
  --JsonRpc.Enabled true --JsonRpc.Host 0.0.0.0 --JsonRpc.Port 8545 \
  --init-discoveryenabled false --network-maxactivepeers 0
```

**Information**
Nethermind Release: [Link](https://github.com/NethermindEth/nethermind/releases)
Version: v1.32.4

---

[...]

PR #2054, opened 2025-08-20
src(eth_config): added eth_config simulator prototype

felix314159:
## 🗒️ Description
You can run it via `uv run execute eth-config --network=Mainnet --rpc-endpoint=https://<replace>@rpc.nimbus-besu-1.fusaka-devnet-3.ethpandaops.io --clients="besu,erigon,nethermind,nimbusel,reth" --network-config-file=./src/pytest_plugins/execute/eth_config/networks.yml -vv -s`

## What happens
In the `test_eth_config_majority()` it requests and compares the `eth_config` responses of all provided `--clients`. Certain cl+el combinations might not be available so it picks the first working one for each exec client. The test is only passed if each execution client from the clients flag was successfully queried and all responses are identical. Otherwise an overview of which endpoints provided which response (which with hash) is given, so that client teams can be contacted about misconfiguration.

All other tests in `execute_eth_config` only contact the `--rpc-endpoint`, so here no additional combinations are derived and you only end up requesting info from one execution client.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/src/ethereum_test_rpc/rpc.py
+++ b/src/ethereum_test_rpc/rpc.py
@@ -19,6 +19,7 @@ from .types import (
     ForkchoiceUpdateResponse,
     GetBlobsResponse,
     GetPayloadResponse,
+    JSONRPCError,
     PayloadAttributes,
     PayloadStatus,
     TransactionByHashResponse,
@@ -96,7 +97,7 @@ class BaseRPC:
         if request_id is None:
             request_id = next_request_id_counter
 
-        json = {
+        payload = {
             "jsonrpc": "2.0",
             "method": f"{self.namespace}_{method}",
             "params": params,
@@ -108,19 +109,12 @@ class BaseRPC:
         headers = base_header | extra_headers
 
         logger.debug(f"Sending RPC request, timeout is set to {timeout}...")
-        response = requests.post(self.url, json=json, headers=headers, timeout=timeout)
+        response = requests.post(self.url, json=payload, headers=headers, timeout=timeout)
+        response.raise_for_status()
+        response_json = response.json()
 
-        try:
-            response_json = response.json()
-        except Exception as e:
-            logger.debug(f"Failed to deserialize response: {e}")
-            return None
-
-        # response.raise_for_status()
         if "error" in response_json:
-            # raise JSONRPCError(**response_json["error"])
-            logger.debug(f"Got response with error: {response_json}")
-            return None
+            raise JSONRPCError(**response_json["error"])
 
         assert "result" in response_json, "RPC response didn't contain a result field"
         result = response_json["result"]
--- a/src/ethereum_test_rpc/rpc.py
+++ b/src/ethereum_test_rpc/rpc.py
@@ -19,7 +19,6 @@ from .types import (
     ForkchoiceUpdateResponse,
     GetBlobsResponse,
     GetPayloadResponse,
-    JSONRPCError,
     PayloadAttributes,
     PayloadStatus,
     TransactionByHashResponse,
@@ -110,11 +109,18 @@ class BaseRPC:
 
         logger.debug(f"Sending RPC request, timeout is set to {timeout}...")
         response = requests.post(self.url, json=json, headers=headers, timeout=timeout)
-        response.raise_for_status()
-        response_json = response.json()
 
+        try:
+            response_json = response.json()
+        except Exception as e:
+            logger.debug(f"Failed to deserialize response: {e}")
+            return None
+
+        # response.raise_for_status()
         if "error" in response_json:
-            raise JSONRPCError(**response_json["error"])
+            # raise JSONRPCError(**response_json["error"])
+            logger.debug(f"Got response with error: {response_json}")
+            return None
 
         assert "result" in response_json, "RPC response didn't contain a result field"
         result = response_json["result"]
@@ -155,7 +161,7 @@ class EthRPC(BaseRPC):
             pprint(e.errors())
             raise e
         except Exception as e:
-            logger.error(f"exception occurred when sending JSON-RPC request: {e}")
+            logger.debug(f"exception occurred when sending JSON-RPC request: {e}")
             raise e
 
     def chain_id(self) -> int:
--- a/src/ethereum_test_rpc/rpc.py
+++ b/src/ethereum_test_rpc/rpc.py
@@ -430,9 +430,9 @@ class EngineRPC(BaseRPC):
         method = f"forkchoiceUpdatedV{version}"
 
         if payload_attributes is None:
-            params = [to_json(forkchoice_state)]
-        else:
             params = [to_json(forkchoice_state), None]
+        else:
+            params = [to_json(forkchoice_state), to_json(payload_attributes)]
 
         return ForkchoiceUpdateResponse.model_validate(
             self.post_request(
--- a/src/ethereum_test_rpc/rpc.py
+++ b/src/ethereum_test_rpc/rpc.py
@@ -79,10 +79,12 @@ class BaseRPC:
 
     def post_request(
         self,
+        *,
         method: str,
-        *params: Any,
+        params: Any | None = None,
         extra_headers: Dict | None = None,
         request_id: int | str | None = None,
+        timeout: int | None = None,
     ) -> Any:
         """Send JSON-RPC POST request to the client RPC server at port defined in the url."""
         if extra_headers is None:
@@ -92,7 +94,8 @@ class BaseRPC:
         next_request_id_counter = next(self.request_id_counter)
         if request_id is None:
             request_id = next_request_id_counter
-        payload = {
+
+        json = {
             "jsonrpc": "2.0",
             "method": f"{self.namespace}_{method}",
             "params": params,
@@ -103,7 +106,9 @@ class BaseRPC:
         }
         headers = base_header | extra_headers
 
-        response = requests.post(self.url, json=payload, headers=headers)
+        # print(f"Sending RPC request to {self.url}, timeout is set to {timeout}...")
+        print(f"Sending RPC request, timeout is set to {timeout}...")  # don't leak url in logs
+        response = requests.post(self.url, json=json, headers=headers, timeout=timeout)
         response.raise_for_status()
         response_json = response.json()
 
@@ -135,11 +140,12 @@ class EthRPC(BaseRPC):
         super().__init__(*args, **kwargs)
         self.transaction_wait_timeout = transaction_wait_timeout
 
-    def config(self):
+    def config(self, timeout: int | None = None):
         """`eth_config`: Returns information about a fork configuration of the client."""
         try:
-            response = self.post_request("config")
+            response = self.post_request(method="config", timeout=timeout)
             if response is None:
+                print("eth_config request: failed to get response")
                 return None
             return EthConfigResponse.model_validate(
                 response, context=self.response_validation_context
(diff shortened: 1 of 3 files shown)


LouisTsai-Csie:
Thanks @felix314159 ! i create a specific [issue](https://github.com/ethereum/execution-spec-tests/issues/2049) for eip-7910, what about link this PR to the issue also?

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -54,21 +77,38 @@ def pytest_configure(config: pytest.Config) -> None:
...
+                f'Network "{network_name}" could not be found in file "{network_configs_path}".'
+            )
+        config.network = network_configs.root[network_name]  # type: ignore
+    else:  # case 2: activate majority test mode
+        #   parse clients list
+        clients.replace(" ", "")
+        clients = clients.split(",")
+        for c in clients:
+            if c not in EXECUTION_CLIENTS:
+                pytest.exit(f"Unsupported client was passed: {c}")
```
LouisTsai-Csie, line 106: small suggestion, based on my previous pr [review](https://github.com/ethereum/execution-spec-tests/pull/1895#discussion_r2201179217):
```py
for c in clients.split(","):
    if c.strip() not in EXECUTION_CLIENTS:
        pytest.exit(f"Unsupported client was passed: {c.strip()}")
```

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -97,3 +136,101 @@ def rpc_endpoint(request) -> str:
...
+        exec_client: [
+            re.sub(
+                pattern,
+                f"\\g<1>{consensus}-{exec_client}\\g<4>",
+                rpc_endpoint,
+            )
+            for consensus in CONSENSUS_CLIENTS
+        ]
+        for exec_client in exec_clients
+    }
```
marioevz, line 162: I think this entire post-processing to the RPC endpoint url does not belong in this script, because we are constraining the commands usage to a single case: Testing clients that ethpandaops already deployed and nothing else.

If I'm a client dev that wants to spawn a client locally to test `eth_config` I can no longer do that because `--rpc-endpoint` is unavoidably modified.

There's several of options though:

### Clients Flag is a dict
```bash
uv run execute eth_config --clients '{"go-ethereum": "http://...",  "nethermind": "http://...", ...'
```

We parametrize the tests where the ID is the name of the client and RPC-endpoint is the value of the dictionary.

This would also mean that `--clients` and `--rpc-endpoint` are mutually exclusive, and that's ok, and the `--rpc-endpoint` remains the same as before this PR.

### RPC endpoint is a list
```bash
uv run execute eth_config --rpc-endpoint 'https://geth.endpoint.url/,https://nethermind.endpoint.url/'
```

We parametrize the tests with each element of this comma-separated string. This gives the command more flexibility because we can keep using it in more situations than only testing for ethpandaops-deployed clients.

If we want to keep the ability to include in the test ID the name of the client we could just check if the name of the client is in the URL and use that client's name as test ID, but otherwise gracefully continue if we don't recognize the URL.

marioevz (commented):
Thanks for the changes! I think we may need to adjust the approach here though, since it looks like some existing functionality would be lost with this implementation. Left some comments with suggestions on how to proceed.

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -97,3 +136,101 @@ def rpc_endpoint(request) -> str:
...
     return EthRPC(rpc_endpoint)
+
+
+def majority_eth_config_test(exec_clients: List[str], rpc_endpoint: str):  # noqa: D103
+    # sanity checks
+    assert ".ethpandaops.io" in rpc_endpoint
+    assert len(exec_clients) > 0
+    if "geth" in exec_clients and "devnet-3" in rpc_endpoint:
+        print("Devnet-3 geth does not support eth_config")
+        return
```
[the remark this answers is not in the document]
felix314159: > This is too specific IMO, if we run this test in Amsterdam devnet-3 it's going to skip geth for no reason.

you are actually saying it is not specific enough (you also want me to check for 'fusaka'). and yes, i agree that we should remove this when fusaka-devnet-3 has ended

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -54,21 +76,38 @@ def pytest_configure(config: pytest.Config) -> None:
...
+        clients.replace(" ", "")
+        clients = clients.split(",")
+        for c in clients:
+            if c not in EXECUTION_CLIENTS:
+                pytest.exit(f"Unsupported client was passed: {c}")
+        print(f"Activating majority mode\nProvided client list: {clients}")
+
+        # request and store and compare all eth_config responses, then terminate
+        majority_eth_config_test(exec_clients=clients, rpc_endpoint=rpc_endpoint)
+        return  # TODO: how to not run the other tests, exit(1)?
```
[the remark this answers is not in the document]
marioevz: The linked function is just an example, we need to create a completely new `pytest_generate_tests` that parametrizes `rpc_endpoint` in `src/pytest_plugins/execute/eth_config/eth_config.py`.

felix314159:
Implemented your feedback, let me know if we still need to change sth

src/ethereum_test_rpc/rpc.py
```diff
@@ -402,10 +460,13 @@ def get_blobs(
...
         """`engine_getBlobsVX`: Retrieves blobs from an execution layers tx pool."""
+        method = f"getBlobsV{version}"
+        params = [f"{h}" for h in versioned_hashes]
+
         return GetBlobsResponse.model_validate(
             self.post_request(
-                f"getBlobsV{version}",
-                [f"{h}" for h in versioned_hashes],
+                method=method,
+                params=params,
```
marioevz, line 469:
```suggestion
                params=[params],
```
The parameter is a list of lists, `tests/osaka/eip7594_peerdas/test_get_blobs.py`  stopped working because of this.

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -54,37 +73,62 @@ def pytest_configure(config: pytest.Config) -> None:
...
     config.network = network_configs.root[network_name]  # type: ignore
 
+    # prepare majority test mode
+    #   parse clients list
+    clients.replace(" ", "")
+    clients = clients.split(",")
+    for c in clients:
+        if c not in EXECUTION_CLIENTS:
+            pytest.exit(f"Unsupported client was passed: {c}")
+    print(f"Provided client list: {clients}")
```
marioevz, line 106: I think this printed info is ok since it lets the user know right away of a misconfiguration.

Only question is whether they should be `logger.info` just for consistency of trying not to use bare `print`.

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -93,7 +137,71 @@ def rpc_endpoint(request) -> str:
...
 
-@pytest.fixture(autouse=True, scope="session")
-def eth_rpc(rpc_endpoint: str) -> EthRPC:
-    """Initialize ethereum RPC client for the execution client under test."""
-    return EthRPC(rpc_endpoint)
+# @pytest.fixture(autouse=True, scope="session")
+# def eth_rpc(rpc_endpoint: str) -> EthRPC:
+#     """Initialize ethereum RPC client for the execution client under test."""
+#     return EthRPC(rpc_endpoint)
+
```
marioevz, line 144:
```suggestion
```

Can be removed IMO.

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -54,37 +73,62 @@ def pytest_configure(config: pytest.Config) -> None:
...
+    #   parse clients list
+    clients.replace(" ", "")
+    clients = clients.split(",")
+    for c in clients:
+        if c not in EXECUTION_CLIENTS:
+            pytest.exit(f"Unsupported client was passed: {c}")
+    print(f"Provided client list: {clients}")
+
+    # store majority mode configuration
+    if ".ethpandaops.io" in rpc_endpoint:
```
marioevz, line 109: This check to enable majority mode still feels a bit off for me, because what if I want to check a single client that is running in an ethpandaops rpc.

My suggestion would be to for `--clients` to work like this:
- The default is `None`, and if so, the majority functionality is off.
- If a comma separated value like `"besu,nethermind,..."` is passed, we enable majority and only check those specific clients.
- If the string `"all"` is passed, we enable majority and test all clients in `EXECUTION_CLIENTS`.

Wdyt?

marioevz (commented):
Looks much better now IMO.

There's a lot of comments but mostly because the change in `src/ethereum_test_rpc/rpc.py` broke `consume` and `execute` sadly. It's a small change but we need to be careful with the RPC package because of the implications it can have.

src/ethereum_test_rpc/rpc.py
```diff
@@ -368,11 +418,17 @@ def forkchoice_updated(
         version: int,
     ) -> ForkchoiceUpdateResponse:
         """`engine_forkchoiceUpdatedVX`: Updates the forkchoice state of the execution client."""
+        method = f"forkchoiceUpdatedV{version}"
+
+        if payload_attributes is None:
+            params = [to_json(forkchoice_state)]
```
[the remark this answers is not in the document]
felix314159: Ty for this, I remember while typing this I was very unsure about this part

src/pytest_plugins/execute/eth_config/eth_config.py
felix314159: i forgor

src/pytest_plugins/execute/eth_config/eth_config.py
felix314159: I agree with everything except the `all` thing, we should avoid special cases when possible. Instead I improved the help description for the flag so that you can see possible values there.

felix314159:
Implemented your feedback and rebased on main. The only thing still unclear to me is the `timeout` comment

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -142,7 +186,67 @@ def rpc_endpoint(request) -> str:
...
+
+    if metafunc.definition.name == "test_eth_config_majority":
+        if len(all_rpc_endpoints_dict) < 2:
+            # The test function is not run because we only have a single client, so no majority comparison  # noqa: E501
+            logger.info(
+                "Skipping eth_config majority because less than 2 exec clients were passed"
+            )
+            metafunc.parametrize(
+                ["all_rpc_endpoints"],
+                [],
```
marioevz, line 233:
```suggestion
                [
                    pytest.param(
                        all_rpc_endpoints_dict,
                        id=metafunc.definition.name,
                        marks=pytest.mark.skip("Only one client"),
                    )
                ],
```
Small fix so the test appears as "SKIPPED", along with the skip reason.

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -142,7 +186,67 @@ def rpc_endpoint(request) -> str:
...
+            )
+        else:
+            metafunc.parametrize(
+                ["all_rpc_endpoints"],
+                [[all_rpc_endpoints_dict]],  # interpret it as a single argument dict
```
marioevz, line 238:
```suggestion
                [
                    pytest.param(
                        all_rpc_endpoints_dict,
                        id=metafunc.definition.name,
                    )
                ],  # interpret it as a single argument dict
```
Another fix to the ids.

src/pytest_plugins/execute/eth_config/execute_eth_config.py
```diff
@@ -171,3 +178,64 @@ def test_eth_config_last_fork_id(
...
+        for eth_rpc_target in all_rpc_endpoints[exec_client]:
+            response = eth_rpc_target.config(timeout=10)
+            if response is None:
+                # safely split url to not leak rpc_endpoint in logs
+                logger.warning(
+                    f"When trying to get eth_config from {eth_rpc_target} a problem occurred"  # problem itself is logged by .config() call # noqa: E501
+                )
+                continue
+
+            response_str = json.dumps(response.model_dump(mode="json"))
```
marioevz, line 201:
```suggestion
            response_str = json.dumps(response.model_dump(mode="json"), sort_keys=True)
```
See comment below.

src/pytest_plugins/execute/eth_config/execute_eth_config.py
```diff
@@ -171,3 +178,64 @@ def test_eth_config_last_fork_id(
...
+        f" from each specified execution client. Full list of execution clients is "
+        f"{all_rpc_endpoints.keys()} but we were only able to gather eth_config responses "
+        f"from: {responses.keys()}\n"
+        "Will try again with a different consensus-execution client combination for "
+        "this execution client"
+    )
+    # determine hashes of client responses
+    client_to_hash_dict = dict()  # Dict[exec_client : response hash] # noqa: C408
+    for client in responses.keys():
+        response_bytes = json.dumps(responses[client], sort_keys=True).encode("utf-8")
```
marioevz, line 221: `responses[client]` is already the dumped json object as a string, so no need to re-dump it here.
```suggestion
        response_bytes = responses[client].encode("utf-8")
```
Test was failing for me locally in nethermind vs besu because the `sort_keys` had no affect on the string, but this fixes it.

marioevz (commented):
Looks very good! I just tried this locally with my suggestions added and both modes are working great.

I came around and I now think `--clients` it's a very convenient addition, thanks!

src/pytest_plugins/execute/eth_config/eth_config.py
```diff
@@ -142,7 +186,67 @@ def rpc_endpoint(request) -> str:
...
+                scope="function",
+            )
+    else:
+        metafunc.parametrize(
+            ["eth_rpc"],
+            [
+                pytest.param(
+                    rpc_endpoint,
+                    id=endpoint_name,
```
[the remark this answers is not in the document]
felix314159: Thanks, I noticed that before this change it was impossible to tell which test even is being run

felix314159:
Thanks for all the improvements. The only thing I left out was the `**kwargs` thing, I find explicit code easier to understand. Would probably rather remove timeout completely or read it from some reasonable constant than have this

src/ethereum_test_rpc/rpc.py
```diff
@@ -103,12 +107,20 @@ def post_request(
...
+        except Exception as e:
+            logger.debug(f"Failed to deserialize response: {e}")
+            return None
 
+        # response.raise_for_status()
         if "error" in response_json:
-            raise JSONRPCError(**response_json["error"])
+            # raise JSONRPCError(**response_json["error"])
+            logger.debug(f"Got response with error: {response_json}")
+            return None
```
marioevz, line 123: This change seems to be causing issues with execute (try `uv run execute hive --fork Osaka ./tests/osaka/eip7594_peerdas/test_get_blobs.py`), since there’s an important distinction between an error and a None response.

For example, `get_transaction_by_hash` can return `None` when the client successfully processes the request but doesn’t find the transaction (I’m not sure if that’s only in the mempool or also before it gets included in a block).

I would lean towards rolling back this change but if you could you share the motivation behind this change please? I’d like to understand the reasoning.

marioevz (commented):
Hi, just a couple more comments, sadly the changes to `src/ethereum_test_rpc/rpc.py` are breaking `uv run execute` so it needs to be fixed.

src/ethereum_test_rpc/rpc.py
felix314159: Yes let's roll it back. While writing my code a specific cl+el combination was not reachable at that time and it kept giving a timeout error (that's why I also added the shorter timeout, but that is not relevant in this context). Since the actual test does sth like:
```
    for eth_rpc_target in all_rpc_endpoints[exec_client]:
            response = eth_rpc_target.config(timeout=10)
            if response is None:
```
Raising an error in .config() would have failed the test, even when it could have easily been passed by trying a different cl client for this el (e.g. when lighthouse+reth does not respond try teku+reth instead of failing the test). So my fault was that instead of making it a `try .. except` in the test itself I modified rpc.py, so let's roll it back and instead adjust the test. I'll do this tomorrow

marioevz (approved):
LGTM!

Let's get this merged! Thanks!

PR #2057, opened 2025-08-20
feat(tests): add test for large args offset with size zero on CALL opcode

JereSalo:
## 🗒️ Description

Hi! I believe a test like this exists for other opcodes but not for CALL, it is one of those that have a very large offset but size zero and therefore execution shouldn't revert because of the offset. In ethrex we had a bug that was fixed in [this PR](https://github.com/lambdaclass/ethrex/pull/3950). I created this test and tried it out before and after the fix and it detects the bug.
I am making these tests just for the CALL opcode but they apply to all xCALL opcodes, I don't know what the nicest way of making this more generic is. But generally this would be enough because these opcodes tend to have pretty similar logic inside of them.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/tests/frontier/opcodes/test_call.py
+++ b/tests/frontier/opcodes/test_call.py
@@ -151,14 +151,16 @@ def test_call_memory_expands_on_early_revert(
 
 # TODO: There's an issue with gas definitions on forks previous to Berlin, remove this when fixed.
 # https://github.com/ethereum/execution-spec-tests/pull/1952#discussion_r2237634275
+@pytest.mark.with_all_call_opcodes
 @pytest.mark.valid_from("Berlin")
 def test_call_large_args_offset_size_zero(
     state_test: StateTestFiller,
     pre: Alloc,
     fork: Fork,
+    call_opcode: Op,
 ):
     """
-    Test CALL with an extremely large args_offset and args_size set to zero.
+    Test xCALL with an extremely large args_offset and args_size set to zero.
     Since the size is zero, the large offset should not cause a revert.
     """
     sender = pre.fund_eoa()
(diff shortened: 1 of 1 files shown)


tests/frontier/opcodes/test_call.py
```diff
@@ -147,3 +147,54 @@ def test_call_memory_expands_on_early_revert(
...
+
+
+# TODO: There's an issue with gas definitions on forks previous to Berlin, remove this when fixed.
+# https://github.com/ethereum/execution-spec-tests/pull/1952#discussion_r2237634275
+@pytest.mark.valid_from("Berlin")
+def test_call_large_args_offset_size_zero(
+    state_test: StateTestFiller,
+    pre: Alloc,
+    fork: Fork,
+):
```
LouisTsai-Csie, line 159: Thanks for this PR! If you want to expand the test to all the call related operation, you can parametrize these ops like:

```py
@pytest.mark.parametrize(
    "call_opcode",
    [
        Op.STATICCALL,
        Op.DELEGATECALL,
        Op.CALLCODE,
    ],
)
```

And then pass these opcode in the function, and replace the `Op.CALL` with `call_opcode` in your test.
```diff
@pytest.mark.valid_from("Berlin")
def test_call_large_args_offset_size_zero(
    state_test: StateTestFiller,
    pre: Alloc,
    fork: Fork,
+  call_opcode: Op
):
```

Some reference implementation [here](https://github.com/ethereum/execution-spec-tests/blob/6cf066e06b61d2a5ddd05104ba1274cabd5b6e5d/tests/osaka/eip7951_p256verify_precompiles/test_p256verify.py#L201) and [here](https://github.com/ethereum/execution-spec-tests/blob/6cf066e06b61d2a5ddd05104ba1274cabd5b6e5d/tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py#L622).

marioevz: I agree with this comment, and ideally we can use [`@pytest.mark.with_all_call_opcodes`](https://eest.ethereum.org/main/writing_tests/test_markers/#pytestmarkwith_all_call_opcodes) in this case.

JereSalo: You never cease to amaze me guys. Thank you very much.
Implemented changes [here](https://github.com/ethereum/execution-spec-tests/pull/2057/commits/656a47f01ce9e4401445ca47d4279cd073427fc3). Maybe in a follow-up I'll make changes to the other CALL tests in this file.

LouisTsai-Csie (commented):
LGTM, let's wait for other members to review again and merge!

PR #2058, opened 2025-08-20
chore(tests): add modexp length mismatch functionality

### LINKED ISSUE #2043: `ModExpInput` prevents creating trimmed inputs
The `ModExpInput` creates integer representations of base, exp and mod. This is a mistake which prevents creating valid and cheap inputs.
E.g. you can create an input of base 0, mod 0 and non-zero exp of length `2**64` as:
```
0000000000000000000000000000000000000000000000000000000000000000 0000000000000000000000000000000000000000000000010000000000000000 0000000000000000000000000000000000000000000000000000000000000000 80
```

Similarly, the `Spec.calculate_gas_cost()` is impractical because it takes the `exponent : bytes` parameter.

spencer-tb:
## 🗒️ Description

Aims to update the modexp helpers to allow for length mismatch tests.

Requires an update to https://github.com/ethereum/execution-spec-tests/blob/main/tests/osaka/eip7883_modexp_gas_increase/conftest.py#L60-L78 once #1929 is merged.

Length mismatch tests can be added with the following, copied over and refactored from #2044.
```
pytest.param(
    ModExpInput.create_mismatch(
        exponent="80",
        declared_exponent_length=2**64,
    ),
    id="exp_2_pow_64_base_0_mod_0",
),
```

## 🔗 Related Issues or PRs

Fixes #2043 (attemps to).

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/tests/byzantium/eip198_modexp_precompile/helpers.py
+++ b/tests/byzantium/eip198_modexp_precompile/helpers.py
@@ -25,20 +25,62 @@ class ModExpInput(TestParameterGroup):
     extra_data: Bytes = Field(default_factory=Bytes)
     raw_input: Bytes | None = None
 
+    override_base_length: int | None = None
+    override_exponent_length: int | None = None
+    override_modulus_length: int | None = None
+
     @property
     def length_base(self) -> Bytes:
         """Return the length of the base."""
-        return Bytes(len(self.base).to_bytes(32, "big"))
+        length = (
+            self.override_base_length if self.override_base_length is not None else len(self.base)
+        )
+        return Bytes(length.to_bytes(32, "big"))
 
     @property
     def length_exponent(self) -> Bytes:
         """Return the length of the exponent."""
-        return Bytes(len(self.exponent).to_bytes(32, "big"))
+        length = (
+            self.override_exponent_length
+            if self.override_exponent_length is not None
+            else len(self.exponent)
+        )
+        return Bytes(length.to_bytes(32, "big"))
 
     @property
     def length_modulus(self) -> Bytes:
         """Return the length of the modulus."""
-        return Bytes(len(self.modulus).to_bytes(32, "big"))
+        length = (
+            self.override_modulus_length
+            if self.override_modulus_length is not None
+            else len(self.modulus)
+        )
+        return Bytes(length.to_bytes(32, "big"))
+
+    @property
+    def declared_base_length(self) -> int:
+        """Return the declared base length as int."""
+        return (
+            self.override_base_length if self.override_base_length is not None else len(self.base)
+        )
+
+    @property
+    def declared_exponent_length(self) -> int:
+        """Return the declared exponent length as int."""
+        return (
+            self.override_exponent_length
+            if self.override_exponent_length is not None
+            else len(self.exponent)
+        )
+
+    @property
+    def declared_modulus_length(self) -> int:
+        """Return the declared modulus length as int."""
+        return (
+            self.override_modulus_length
+            if self.override_modulus_length is not None
+            else len(self.modulus)
+        )
 
     def __bytes__(self):
         """Generate input for the MODEXP precompile."""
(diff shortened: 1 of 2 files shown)


tests/osaka/eip7823_modexp_upper_bounds/test_modexp_upper_bounds.py
```diff
@@ -24,9 +24,9 @@ def precompile_gas(fork: Fork, mod_exp_input: ModExpInput) -> int:
     """Calculate gas cost for the ModExp precompile and verify it matches expected gas."""
     spec = Spec if fork < Osaka else Spec7883
     calculated_gas = spec.calculate_gas_cost(
-        len(mod_exp_input.base),
-        len(mod_exp_input.modulus),
-        len(mod_exp_input.exponent),
+        mod_exp_input.declared_base_length,
+        mod_exp_input.declared_modulus_length,
+        mod_exp_input.declared_exponent_length,
```
chfast, line 29: This doesn't fully solve the problem because you pass `exponent` which may mean this is only the part of the exponent but the function will treat it as full. So this computation may give incorrect results.

tests/byzantium/eip198_modexp_precompile/helpers.py
```diff
@@ -86,6 +128,26 @@ def from_bytes(cls, input_data: Bytes | str) -> "ModExpInput":
 
         return cls(base=base, exponent=exponent, modulus=modulus, raw_input=input_data)
 
+    @classmethod
+    def create_mismatch(
```
chfast, line 132: This feels a bit too much for me. I generally struggle with some data types in EEST which have a lot of interconnected options and flags.

This could be a normal constructor. And you can fix the properties directly in the constructor like:

```python
length_base = declared_base_length if  declared_base_length not None else len(base)
```

tests/osaka/eip7823_modexp_upper_bounds/test_modexp_upper_bounds.py
LouisTsai-Csie: We can take inspiration from the [EELS](https://github.com/ethereum/execution-specs/blob/forks/osaka/src/ethereum/osaka/vm/precompiled_contracts/modexp.py) implementation, which resolve this problem in an elegant way.

According to [eip-7883](https://eips.ethereum.org/EIPS/eip-7883), the iteration count is derived from the exponent. However, in practice we only need at most 32 bytes of the exponent.

```py
def calculate_iteration_count(exponent_length, exponent):
    iteration_count = 0
    if exponent_length <= 32 and exponent == 0:
        iteration_count = 0
    elif exponent_length <= 32:
        iteration_count = exponent.bit_length() - 1
    elif exponent_length > 32:
        iteration_count = (16 * (exponent_length - 32)) + ((exponent & (2**256 - 1)).bit_length() - 1)
    return max(iteration_count, 1)
```

- When `exponent_length` <= 32, we can compute the bit length of the full exponent directly.
- When `exponent_length` > 32, we mask the exponent to its 32 bytes and calculate the bit length of that portion: `(exponent & (2**256 - 1)).bit_length() - 1`

As a result, we only need to fetch the first 32-byte word of the exponent input (`exponent_head`) to perform this calculation, instead of processing the entire exponent.

spencer-tb:
Updated regarding the comments using the EELS like approach..

tests/osaka/eip7883_modexp_gas_increase/spec.py
```diff
@@ -79,14 +79,29 @@ def calculate_iteration_count(cls, exponent_length: int, exponent: bytes) -> int
...
+        if exponent_length <= cls.EXPONENT_THRESHOLD and exponent_head == 0:
+            iteration_count = 0
+        elif exponent_length <= cls.EXPONENT_THRESHOLD:
+            iteration_count = exponent_head.bit_length() - 1 if exponent_head > 0 else 0
+        else:
+            # For large exponents: length_part + bits from first 32 bytes
+            length_part = cls.EXPONENT_BYTE_MULTIPLIER * (exponent_length - 32)
+            bits_part = exponent_head.bit_length() - 1 if exponent_head > 0 else 0
+            iteration_count = length_part + bits_part
+        iteration_count = max(iteration_count, 1)
```
LouisTsai-Csie, line 104: I do not notice this before, it seems like we already have the same functionality in `calculate_iteration_count`, should avoid duplicate here

spencer-tb: Thanks! Nice catch - updated in latest commit.

PR #2066, opened 2025-08-21
feat(fill): add a plugin for optional execution witness generation

danceratopz:
## 🗒️ Description
1. Adds an optional `executionWitness` field to the blockchain fixture base class.
2. Adds a new pytest plugin that:
    - `pytest_addoption()`: add a boolean `--witness` flag.
    - `pytest_configure()`: ~Compiles~ checks whether the `witness-filler` tool is in PATH if `--witness` is enabled; hard error if not.
    - Provides a fixture `generate_witness` that generates a fixture's execution witness using the `witness-filler` tool.
3.  Calls the `generate_witness` fixture in `BaseTestParametrizer`. If `--witness` is not enabled, there is not change in `fill` behavior.

#### Usage

```
uv run fill tests/istanbul/eip1344_chainid/test_chainid.py --output=fixtures-witness --witness --clean
```

image

## 🔗 Related Issues or PRs
#2031

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/src/ethereum_test_fixtures/blockchain.py
+++ b/src/ethereum_test_fixtures/blockchain.py
@@ -1,5 +1,7 @@
 """BlockchainTest types."""
 
+import json
+from dataclasses import dataclass
 from functools import cached_property
 from typing import (
     Annotated,
@@ -401,6 +403,21 @@ class FixtureWithdrawal(WithdrawalGeneric[ZeroPaddedHexNumber]):
         return cls(**w.model_dump())
 
 
+@dataclass(slots=True)
+class WitnessChunk:
+    """Represents execution witness data for a block."""
+
+    state: List[str]
+    codes: List[str]
+    keys: List[str]
+    headers: List[str]
+
+    @classmethod
+    def from_json(cls, s: str) -> List["WitnessChunk"]:
+        """Parse witness chunks from JSON string."""
+        return [cls(**obj) for obj in json.loads(s)]
+
+
 class FixtureBlockBase(CamelModel):
     """Representation of an Ethereum block within a test Fixture without RLP bytes."""
 
@@ -408,6 +425,7 @@ class FixtureBlockBase(CamelModel):
     txs: List[FixtureTransaction] = Field(default_factory=list, alias="transactions")
     ommers: List[FixtureHeader] = Field(default_factory=list, alias="uncleHeaders")
     withdrawals: List[FixtureWithdrawal] | None = None
+    execution_witness: WitnessChunk | None = Field(None, alias="executionWitness")
 
     @computed_field(alias="blocknumber")  # type: ignore[misc]
     @cached_property
--- a/src/ethereum_test_fixtures/blockchain.py
+++ b/src/ethereum_test_fixtures/blockchain.py
@@ -427,7 +427,7 @@ class FixtureBlockBase(CamelModel):
     txs: List[FixtureTransaction] = Field(default_factory=list, alias="transactions")
     ommers: List[FixtureHeader] = Field(default_factory=list, alias="uncleHeaders")
     withdrawals: List[FixtureWithdrawal] | None = None
-    execution_witness: WitnessChunk | None = Field(None, alias="executionWitness")
+    execution_witness: WitnessChunk | None = None
 
     @computed_field(alias="blocknumber")  # type: ignore[misc]
     @cached_property
--- a/src/ethereum_test_fixtures/blockchain.py
+++ b/src/ethereum_test_fixtures/blockchain.py
@@ -411,8 +411,12 @@ class WitnessChunk(CamelModel):
     headers: List[str]
 
     @classmethod
-    def from_json(cls, s: str) -> List["WitnessChunk"]:
-        """Parse witness chunks from JSON string."""
+    def parse_witness_chunks(cls, s: str) -> List["WitnessChunk"]:
+        """
+        Parse multiple witness chunks from JSON string.
+
+        Returns a list of WitnessChunk instances parsed from the JSON array.
+        """
         return [cls(**obj) for obj in json.loads(s)]
 
 
--- a/src/ethereum_test_fixtures/blockchain.py
+++ b/src/ethereum_test_fixtures/blockchain.py
@@ -1,7 +1,6 @@
 """BlockchainTest types."""
 
 import json
-from dataclasses import dataclass
 from functools import cached_property
 from typing import (
     Annotated,
@@ -403,8 +402,7 @@ class FixtureWithdrawal(WithdrawalGeneric[ZeroPaddedHexNumber]):
         return cls(**w.model_dump())
 
 
-@dataclass(slots=True)
-class WitnessChunk:
+class WitnessChunk(CamelModel):
     """Represents execution witness data for a block."""
 
     state: List[str]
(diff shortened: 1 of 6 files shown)


src/pytest_plugins/filler/witness.py
```diff
@@ -0,0 +1,119 @@
...
+
+from ethereum_test_fixtures.blockchain import FixtureBlock, WitnessChunk
+
+
+def pytest_addoption(parser: pytest.Parser):
+    """Add witness command-line options to pytest."""
+    witness_group = parser.getgroup("witness", "Arguments for witness functionality")
+    witness_group.addoption(
+        "--witness",
+        "--witness-the-fitness",
```
jsign, line 21: :eyes: :sweat_smile:

jsign (approved):
Thank you very much sir!

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -401,13 +403,29 @@ def from_withdrawal(cls, w: WithdrawalGeneric) -> "FixtureWithdrawal":
...
 
 
+@dataclass(slots=True)
+class WitnessChunk:
+    """Represents execution witness data for a block."""
+
+    state: List[str]
+    codes: List[str]
+    keys: List[str]
+    headers: List[str]
```
marioevz, line 413: We cannot add anything to our fixtures that is not a proper pydantic model.

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -401,13 +403,29 @@ def from_withdrawal(cls, w: WithdrawalGeneric) -> "FixtureWithdrawal":
...
+
+    @classmethod
+    def from_json(cls, s: str) -> List["WitnessChunk"]:
+        """Parse witness chunks from JSON string."""
+        return [cls(**obj) for obj in json.loads(s)]
```
marioevz, line 418: This function is a bit misleading since it returns a list of instances of the object instead of single instance.

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -401,13 +403,29 @@ def from_withdrawal(cls, w: WithdrawalGeneric) -> "FixtureWithdrawal":
...
+
+
 class FixtureBlockBase(CamelModel):
     """Representation of an Ethereum block within a test Fixture without RLP bytes."""
 
     header: FixtureHeader = Field(..., alias="blockHeader")
     txs: List[FixtureTransaction] = Field(default_factory=list, alias="transactions")
     ommers: List[FixtureHeader] = Field(default_factory=list, alias="uncleHeaders")
     withdrawals: List[FixtureWithdrawal] | None = None
+    execution_witness: WitnessChunk | None = Field(None, alias="executionWitness")
```
marioevz, line 428: I don't know how is this not failing since `WitnessChunk` is not a pydantic model.

marioevz, line 428:
```suggestion
    execution_witness: WitnessChunk | None = None
```
Also the alias is automatically converted to camel case since we are using `CamelModel`.

src/pytest_plugins/filler/witness.py
```diff
@@ -0,0 +1,119 @@
...
+            [
+                "cargo",
+                "install",
+                "--git",
+                "https://github.com/kevaundray/reth.git",
+                "--rev",
+                "8016a8a5736e4427b3d285c82cd39c4ece70f8c4",
+                "witness-filler",
+            ],
+        )
```
marioevz, line 53: This doesn't make any sense to me, this is a big side-effect of running this command, it should be a requirement rather than an automatic install.

src/pytest_plugins/filler/witness.py
```diff
@@ -0,0 +1,119 @@
...
+        Callable that generates witness data for a fixture if enabled.
+
+    """
+    if not request.config.getoption("witness"):
+        return None
+
+    def generate_witness(fixture: Any) -> None:
+        """Generate witness data for a fixture using the witness-filler tool."""
+        if not hasattr(fixture, "blocks") or not fixture.blocks:
+            return
```
marioevz, line 82: We should instead constrain the type of parameter `fixture` to be a specific fixture type, because:
- Using `hasattr` too many times
- `witness-filler`, whatever it does, it's probably only going to support one of our fixture types
- The caller of the function is the one that should use `isinstance` before calling this function instead of trying to guess what were we passed as a parameter.

src/pytest_plugins/filler/witness.py
```diff
@@ -0,0 +1,119 @@
...
+        try:
+            witnesses = WitnessChunk.from_json(result.stdout)
+            for i, witness in enumerate(witnesses):
+                if i < len(fixture.blocks) and isinstance(fixture.blocks[i], FixtureBlock):
+                    fixture.blocks[i].execution_witness = witness
+        except (ValueError, IndexError, AttributeError) as e:
+            raise RuntimeError(
+                f"Failed to parse witness data from witness-filler tool. "
+                f"Output was: {result.stdout[:500]}{'...' if len(result.stdout) > 500 else ''}"
+            ) from e
```
marioevz, line 117: Just use a `TypeAdapter(List[WitnessChunk])` and it will give you a proper and detailed validation error.

marioevz (changes requested):
Not a huge fan of how hacky this is tbh, but if it's a single use case I can live with it once the changes are applied.

src/ethereum_test_fixtures/blockchain.py
danceratopz: I guess it's only because pydantic is an incredibly flexible library that correctly guesses the correct thing to do.

src/pytest_plugins/filler/witness.py
```diff
@@ -0,0 +1,119 @@
...
+        print("   This may take several minutes for first-time compilation...")
+
+        result = subprocess.run(
+            [
+                "cargo",
+                "install",
+                "--git",
+                "https://github.com/kevaundray/reth.git",
+                "--rev",
+                "8016a8a5736e4427b3d285c82cd39c4ece70f8c4",
```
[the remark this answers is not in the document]
danceratopz: Done.

src/pytest_plugins/filler/witness.py
marioevz: > We won't be running with `--witness` typically. We can always revisit later.

Famous last words before CI breaks because we forgot and tried to install something as non-root.

Also, what happens when users start to deviate from this? Can we guarantee that zkevm and other stateless efforts are never going to use different versions?

Also, if this external tool needs update at any point, it now requires a PR to EEST.

marioevz (commented):
LGTM, other than the cargo install.

src/pytest_plugins/filler/witness.py
marioevz: > I think we can point to a branch instead of a commit maybe? --branch branch-name instead of --rev <commit>?
It's a similar problem IMO. `witness-filler` has to be already in path for this to work, we could at most check that it's indeed there and fail if not.

The CI workflow (if any) is the one responsible to install the correct version before running the `fill` command.

danceratopz: Ok, ok, fair enough, I know it's not very clean. Let me take it out.

marioevz (approved):
LGTM, thanks!

jsign:
@danceratopz, @marioevz, thanks again for this. :pray:

PR #2067, opened 2025-08-21
feat(fork,tests): Adds amsterdam fork and block access list tests

fselmo:
## 🗒️ Description

This builds upon #1823 and #1866. I worked on this side by side with #ethereum/execution-specs#1381 to get BAL support and tests working with each other.

## 🔗 Related Issues or PRs

#ethereum/execution-specs#1381

## ✅ Checklist

    ```console
    ```


src/ethereum_test_types/block_access_list.py
```diff
@@ -0,0 +1,337 @@
...
+    rlp_fields: ClassVar[List[str]] = ["slot", "slot_changes"]
+
+
+class BalAccountChange(CamelModel, RLPSerializable):
+    """Represents all changes to a specific account in a block."""
+
+    address: Address = Field(..., description="Account address")
+    nonce_changes: Optional[List[BalNonceChange]] = Field(
+        None, description="List of nonce changes"
+    )
```
marioevz, line 76:
```suggestion
    nonce_changes: List[BalNonceChange] = Field(
        default_factory=list, description="List of nonce changes"
    )
```
Should remove the need for the `to_list` changes.

If you need to know whether a field was set or the default was used we can check `model_fields_set` in the object's instance.

src/ethereum_test_types/block_access_list.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def to_list(self, signing: bool = False) -> List[Any]:
+        """
+        Override to_list to return the account changes list directly.
+
+        The BlockAccessList IS the list of account changes, not a container
+        that contains a list, per EIP-7928.
+        """
+        # Return the list of accounts directly, not wrapped in another list
+        return to_serializable_element(self.account_changes)
```
marioevz, line 167: I think what we need here is then a `EthereumTestRootModel`:
```python
class BlockAccessList(EthereumTestRootModel, RLPSerializable):
    ...
    root: List[BalAccountChange] = Field(
        default_factory=list, description="List of account changes in the block"
    )
```

It might be that `RLPSerializable` root types, but it's something we could fix in that class perhaps.

marioevz (commented):
Some initial comments!

scripts/convert_addresses.py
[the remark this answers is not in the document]
fselmo: ah yeah... running lint without `src tests` at the end. I do this all the time and finally committed it :(

src/ethereum_test_types/block_access_list.py
fselmo: Ah, yeah! I think this is exactly the route. I will explore making `RLPSerializable` work with a root type. That would be ideal and we may run into this again in the future.

fselmo: @marioevz Hmm actually, now that I'm looking at the code more and played around with a `RLPSerializableRootModel` and a 2-model approach, I think the current approach may be best but with slight modifications. Hear me out and then let's compare it against what you're thinking because I may have missed something.

The current `BlockAccessList` works well in that it validates JSON from t8n... and since only its elements need to be `RLPSerializable` (not the container itself), we can add very basic functionality via `rlp()` and `to_list()` methods to handle the proper serialization. The `verify_against()` already uses `model_fields_set()` for partial validation as you mentioned.

Making this a `RootModel` would sacrifice the JSON validation part and add complexity imo. A 2-model approach (`RootModel` + `CamelModel`) introduces maintenance overhead + devex takes a hit in knowing which to use and where + doesn't seem to provide very clear benefits.

Am I missing any advantages to the `RootModel` approach beyond the semantics of "`BlockAccessList` IS a list so it should be one"? (which I agree makes a whole lot of sense but seems to add complexity everywhere else). Happy to reconsider if I'm missing something!

fselmo: I pushed some minor changes to simplify it, rather than trying to override `RLPSerializable` methods directly.

fselmo: Right. If we change what we pass on the specs side. At that point we would be restricting the model to be more correct as a root model but giving up possible properties on the class itself (since only `root` can be defined as a property), like `version`, etc. It could lose some flexibility but I'd be OK with either route. Thoughts?

fselmo: > allows us to better describe what we are trying to verify in the actual BlockAccessList object.

Curious what this looks like. The current flow specifically only compares what is in the `model_fields_set` between one and the other. Is there anything else we should take into account?

marioevz: So the two classes in my mind should accomplish these two different purposes:
`BlockAccessList`:
- As close to the EIP description as possible
- Should be easily transformed to RLP because it adheres to the RLP description in the EIP
- It's used for communication in the t8n interface.
- Should be also what ends up in fixtures (as JSON or as RLP)

`BlockAccessListCheck` (Name tbd):
- Allows easier writing of expectation in tests
- Allows to easily describe the absence-of expectation (e.g. we are testing that an address is _not_ present in the block level access list, and the test should fail if it's present in the t8n response).
- Contains the `verify_bal` method which is applied to the response from the t8n tool
- Can be skipped by a test if the test is not BAL specific, and in this case no verification is done to the t8n response and that is passed verbatim to the fixture (we trust t8n)

fselmo: Yeah the 2-class approach makes sense here for this separation of concerns. I'll keep these questions in mind when building out the `BlockAccessListExpectation` (name, maybe?) class.

src/ethereum_test_types/block_access_list/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+        # Only check fields that were explicitly set in the expected model
+        for field_name in expected.model_fields_set:
+            if field_name == "address":
+                continue  # Already matched by account lookup
+
+            expected_value = getattr(expected, field_name)
+            actual_value = getattr(actual, field_name)
+
+            # If we explicitly set a field to None, verify it's None/empty
+            if expected_value is None:
```
marioevz, line 277: I think `BalAccountChange` as is right now does not allow `None` in any of its fields, so it could be a bit annoying because type checker is going to complain a lot, but I think the tester should anyway use `[]` to explicitly state that the fields needs to be an empty list instead of `None`.

marioevz (commented):
Looks amazing so far! Some comments and I think we can push to make the merge and make a release.

src/ethereum_test_types/block_access_list/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+            ]
+        )
+
+    """
+
+    account_changes: List[BalAccountChange] = Field(
+        default_factory=list, description="Expected account changes to verify"
+    )
+
+    modifier: Optional[Callable[["BlockAccessList"], "BlockAccessList"]] = Field(
```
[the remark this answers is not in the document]
fselmo: Ah you mean `Optional`. Yeah, bad habit from previous repos 👍🏼

src/ethereum_test_types/block_access_list/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+            actual_accounts_by_addr.keys()
+        )
+        if missing_accounts:
+            raise Exception(
+                "Expected accounts not found in actual BAL: "
+                f"{', '.join(str(a) for a in missing_accounts)}"
+            )
+
+        # Verify each expected account
+        for address, expected_account in expected_accounts_by_addr.items():
```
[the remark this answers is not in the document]
fselmo: No, I like it. I think this is consistent with the rest of the codebase too. Great idea 👍🏼

src/ethereum_test_types/block_access_list/__init__.py
fselmo: > I think the tester should anyway use [] to explicitly state that the fields needs to be an empty list instead of None.

I think this is the right approach. We can relax it to accept both `None` or `[]` but I'm not sure that's great either. I went with raising with a good message to use empty lists on `None`. Thanks for catching this, lot's of changes to keep track. This was a `None` check at one point 😅. I will write some basic unit tests for this.

fselmo:
@marioevz this needed a rebase to fix some conflicts, hence the force push. For ease of a re-review, only the last 4 commits should have any changes since you last reviewed and I believe I addressed all your comments. Thanks for the thorough review 🙏🏼

src/ethereum_test_specs/blockchain.py
```diff
@@ -585,12 +611,31 @@ def generate_block_data(
...
             header = block.rlp_modifier.apply(header)
             header.fork = fork  # Deleted during `apply` because `exclude=True`
 
+        # Process block access list - apply transformer if present for invalid tests
+        bal = transition_tool_output.result.block_access_list
+        if block.expected_block_access_list is not None and bal is not None:
+            # Use to_fixture_bal to validate and potentially transform the BAL
+            bal = block.expected_block_access_list.to_fixture_bal(bal)
+            # Don't update the header hash - leave it as the hash of the correct BAL
+            # This creates a mismatch that should cause the block to be rejected
```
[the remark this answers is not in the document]
fselmo: Hmm yeah, good point. I'm just now starting to explore this for BAL. Not sure if we want to tackle this further down the line. Should we maybe put together some items to think about for BAL as an issue so we can track them while we iterate and try to get this consolidated for a merge? Let me know.

src/pytest_plugins/eels_resolutions.json
```diff
@@ -41,20 +41,20 @@
...
-        "same_as": "EELSMaster"
+      "same_as": "EELSMaster"
     },
     "Amsterdam": {
-        "git_url": "https://github.com/fselmo/execution-specs.git",
-        "branch": "feat/amsterdam-fork-and-block-access-lists",
-        "commit": "59ff2946ca30879079b2ef2b601ae09106ca08f7"
+      "git_url": "https://github.com/fselmo/execution-specs.git",
+      "branch": "feat/amsterdam-fork-and-block-access-lists",
+      "commit": "a5c7b29a658320c2432de78883d350e9f4444d14"
```
spencer-tb, line 58: Could we add this here too: https://github.com/ethereum/execution-spec-tests/blob/main/.github/configs/eels_resolutions.json

spencer-tb (commented):
For an initial release it might be worth adding following [to](https://github.com/ethereum/execution-spec-tests/blob/main/.github/configs/feature.yaml) in the PR:

```yaml
glamsterdam-alpha:
  evm-type: develop
  fill-params: --fork=Amsterdam ./tests/amsterdam
  feature_only: true
```

Note evm-type can be stable/develop as both point to EELS. `glamsterdam-alpha` must be the name of the head of the release tag, I used that as an example. So with the following you should be able to run the following once the PR is merged to main, and you local git is up to date with the latter:
```
git tag glamsterdam-alpha@v1.0.0
git push upstream glamsterdam-alpha@v1.0.0
```

src/ethereum_test_fixtures/blockchain.py
```diff
@@ -428,6 +443,9 @@ class FixtureBlockBase(CamelModel):
     ommers: List[FixtureHeader] = Field(default_factory=list, alias="uncleHeaders")
     withdrawals: List[FixtureWithdrawal] | None = None
     execution_witness: WitnessChunk | None = None
+    block_access_list: Bytes | None = Field(
+        None, description="Serialized EIP-7928 Block Access List"
+    )
```
marioevz, line 448: I think we need to add the block level access list to `with_rlp` in order for it to be included in the full block body RLP in the fixture, which is basically what ends up being used by consume.

The thing is that I think we have to do it using the approach like in:
```python
        if self.withdrawals is not None:
            block.append([w.to_serializable_list() for w in self.withdrawals])
```
And the key is the `to_serializable_list`, because if we use the block access list in an already-rlp-encoded-bytes it's going to be encoded as a bytes string, but I don't think that's what the EIP means (and I also don't think that should be the approach).

One step towards the solution could be to change the type of this field to be the actual `BlockAccessList` type (perhaps we need a `FixtureBlockAccessList` but not sure) so we can encode it properly, with the caveat that the full block access list is going to appear as human-readable in the blockchain test fixture JSON, which I think it's another plus.

marioevz (approved):
LGTM, thanks!

PR #2084, opened 2025-08-28
feat(types,plugins): Add `ChainConfig` to make chain configuration accessible to tests

### LINKED ISSUE #2081: bug(execute): some tests still have hard coded chain-id, use chain-id set via `--rpc-chain-id`
Hi 👋 we are trying to port and use execution-spec-tests on Abritrum. We are using the `uv run execute remote` functionality which requires a `--rpc-endpoint` and `--rpc-chain-id` https://eest.ethereum.org/v4.1.0/executing_tests/#running-test-on-a-live-remote-network Quite a few of the tests will fail, because they hardcode Ethereum mainnet's chain-id.

It would be nice if all tests used `--rpc-chain-id` going forward

anyways let me know what you guys think 👍

marioevz:
### feat(types): Add `ChainConfig` type

Adds `ChainConfig` and `ChainConfigDefaults` to `ethereum_test_types` that contain `chain_id` and should be modified by pytest plugins fill or execute according to the current environment's chain configuration.

### feat(execute): Set `ChainConfig` according to `--chain-id` or `--rpc-chain-id`

Adds a shared (between fill and execute) fixture called `chain_config` that is available session-wide to all tests.

In `fill` there is no way to modify the contents of the chain configuration since the T8N interface does not provide a field to customize the chain-id during the state transition execution.

In `execute` we read the `--chain-id` or `--rpc-chain-id` command line arguments and modify `ChainConfigDefaults.chain_id` so when the `chain_config` fixture is instantiated, the `chain_id` value is already updated.

### fix(static): Add `fix(static): Add `pre_alloc_modify` mark to not-fully-tagged tests` mark to not-fully-tagged tests

Adds `pre_alloc_modify` to all static tests that are not fully tagged, which allows to skip static tests that require direct modifications to the pre (less execute-command failures).

### fix(tests): Chain ID usages in transactions

Fixes chain ID usages accross all tests (I could find) by:
- Removing the `chain_id` parameter from transactions where it was not necessary.
- Using `chain_config.chain_id` value when the test was directly related to the chain ID, so the test now passes in execute-mode.

### fix(plugins): Failing unit tests after test_chainid.py update

Lastly, since we updated and increased coverage of `test_chainid.py`, the file was modified, and this lead to some unit tests which were using this test to verify `fill` command's behavior starting to fail.

Solution was to instead hardcode a minimal state test for these unit tests to use so we can keep updating `test_chainid.py` in the future.

## 🔗 Related Issues or PRs
Fixes #2081

## ✅ Checklist
    ```console
    ```


KolbyML (approved):
:shipit: looks good. I tested the PR on my branch and it works as expected for `execute remote` :fire:

fselmo:
This might be overboard but do you think it would be useful to internally call `eth_chainId` on the rpc endpoint as a sanity check?

spencer-tb:
> This might be overboard but do you think it would be useful to internally call `eth_chainId` via rpc as a sanity check when executing against testnets or standalone clients?

This also sounds like a really nice addition! It could be added in another PR or this PR imo

marioevz:
Found another potential improvement:
If the Engine API's JWT secret is incorrect, `execute remote` starts failing all tests one by one, instead of failing early in `pytest_config`. We could do an early call to the Engine endpoint to check if the provided JWT secret is incorrect and exit early if that's the case.

We only need to implement some endpoint that does not have any effect in the client for this to work (perhaps this? https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md#engine_getclientversionv1).

I'll create an issue for this.

fselmo (approved):
lgtm!

PR #2090, opened 2025-09-01
feat(bloatnet): Add first multi-opcode benchmarks for Bloatnet

gballet:
## 🗒️ Description

- Add CREATE2 deterministic address calculation to overcome 24KB bytecode limit
- Fix While loop condition to properly iterate through contracts
- Account for memory expansion costs in gas calculations
- Add safety margins (50k gas reserve, 98% utilization) for stability
- Tests now scale to any gas limit without bytecode constraints
- Achieve 98% gas utilization with 10M and 20M gas limits

## 🔗 Related Issues or PRs

https://github.com/ethereum/execution-specs/issues/1578

## ✅ Checklist

    ```console
    ```


LouisTsai-Csie (changes requested):

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,588 @@
...
+    factory_address = pre.deploy_contract(code=Op.STOP)  # Simple factory placeholder
+
+    # Pre-deploy all contracts at CREATE2 addresses using sequential salts
+    deployed_contracts = []  # Track for post-state validation
+    for salt in range(num_contracts):
+        # Calculate the CREATE2 address
+        create2_addr = calculate_create2_address(factory_address, salt, bytes(init_code))
+        # Deploy at the calculated address
+        pre[create2_addr] = Account(code=deploy_bytecode)
+        deployed_contracts.append(create2_addr)
```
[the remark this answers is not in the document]
marioevz: We try to never do `pre[address] = account` approach because it basically breaks our assumptions and we cannot run it in `execute` mode.

By the name of the file and test function, I assume running in bloatnet with execute is the goal, so there's a couple of questions:
- Should the factory contract be already in the state of bloatnet?
- Should the contracts deployed by the factory contract also already be deployed in bloatnet before the test starts?

Both cases I feel should be [stubbed contracts](https://eest.ethereum.org/main/running_tests/execute/remote/#address-stubs-for-pre-deployed-contracts).

For the factory contract it's already doable and we simply need to add the appropriate label.

For the contracts deployed by the factory contract it's a bit tricky because we: (a) label them all and pre-deploy them to the bloatnet, or (b) find a more automatic way of pre-deploy if these are not already there.

I think this is all pretty new and the discussion is still open on how to handle this!

CPerezz:
Hey @LouisTsai-Csie  thanks for the comments. 

> In step 2, the `CREATE2` address is calculated in the attack contract, but since these address are calculated in Step 1 also, we could hardcode these value in the Step 2, thus reducing the cost per iteration.

As per this, notice that there's a constraint to what you propose which is that a contract is capped at 24kB atm. Thus, we can only fit ~945 addresses in it. Therefore, unless I'm missing something, this should be the best way to do this (otherwise we can have a chain of contracts that calls each one another as a linkedlist, but it's quite complex for the benefit (arround 6%).

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+        addr_hex = factory_address[2:] if factory_address.startswith("0x") else factory_address
+        factory_bytes = bytes.fromhex(addr_hex)
+    else:
+        # Assume it's an Address object with bytes representation
+        factory_bytes = bytes(factory_address)
+    code += Op.PUSH20(int.from_bytes(factory_bytes, "big")) + Op.PUSH2(base + 1) + Op.MSTORE
+
+    # Store salt at memory[base+21:base+53] (base+0x15)
+    # Assumes salt is already on stack
+    code += Op.PUSH2(base + 0x15) + Op.MSTORE
```
jochem-brouwer, line 58: I would suggest to add this stack item is part of the method arguments, since this is not directly clear when using this method (a side effect or a requirement someone is not aware of)

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+
+    # Store init code hash at memory[base+53:base+85] (base+0x35)
+    code += Op.PUSH32(int.from_bytes(init_code_hash, "big")) + Op.PUSH2(base + 0x35) + Op.MSTORE
+
+    # Calculate keccak256 of 85 bytes starting at memory[base]
+    code += Op.PUSH1(0x55) + Op.PUSH2(base) + Op.SHA3
+
+    # The address is the last 20 bytes (already on stack)
```
jochem-brouwer, line 66: Side note: in EVM you do not have to trim this value to the 20 bytes (not sure if this is being done here), opcodes which use the address (`*CALL`, `BALANCE`, etc.) will trim this value for you.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+    # this is only used for computing the intinsic gas
+    data = final_storage_value.to_bytes(32, "big").rstrip(b"\x00")
+
+    storage = Storage()
+
+    # Initial gas for PUSH0 + CALLDATALOAD + POP (at the end)
+    totalgas = gas_costs.G_BASE * 2 + gas_costs.G_VERY_LOW
+    totalgas = totalgas + fork.transaction_intrinsic_cost_calculator()(calldata=data)
+    gas_increment = gas_costs.G_VERY_LOW * 2 + gas_costs.G_STORAGE_SET + gas_costs.G_COLD_SLOAD
+    sstore_code = Op.PUSH0 + Op.CALLDATALOAD
```
jochem-brouwer, line 97: PUSH0 is 2 gas and CALLDATALOAD is 3 gas, so I believe `gas_increment` is not entirely correct here (but the size of `G_STORAGE_SET` and `G_COLD_SLOAD` are so large that this one-off likely does not yield a bug)

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+    storage_slot: int = 0
+    while totalgas + gas_increment < gas_benchmark_value:
+        totalgas += gas_increment
+        sstore_code = sstore_code + Op.SSTORE(storage_slot, Op.DUP1)
+        storage[storage_slot] = final_storage_value
+        storage_slot += 1
+
+    sstore_code = sstore_code + Op.POP  # Drop last value on the stack
```
jochem-brouwer, line 105: `POP` is not necessary if this is the final opcode in this code.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+
+    sender = pre.fund_eoa()
+    print(sender)
+    contract_address = pre.deploy_contract(
+        code=sstore_code,
```
jochem-brouwer, line 110: Note: keep in mind there is a 24 KiB code size limit.

For the first tx, we assume ~20k gas per SSTORE and 3-4 bytes (DUP1+PUSH1 (or PUSH2)+SSTORE). We would fit approx. 24*1024/4 = 6144 storage stores here so multiplying this with 20_000 yields 122.88M gas until we hit this limit.

However the second tx will hit this limit much sooner since storage changes from 1 to 2 is now much cheaper (2900 gas for the SSTORE (?)). So this is already at ~17M gas limit.

`tx_1_2` will thus consume much less gas than the gas benchmark value and **can thus change more storage values than the ones initialized in `tx_0_1`** (given the same tx gas limit)

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+    pre: Alloc,
+    fork: Fork,
+    final_storage_value: int,
+    gas_benchmark_value: int,
+):
+    """
+    A test that calls a contract with many SSTOREs.
+
+    The first block will have many SSTORES that go from 0 -> 1
+    and the 2nd block will have many SSTORES that go from 1 -> 2
```
jochem-brouwer, line 83: Note: test currently runs both txs in same block

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+    return code
+
+
+@pytest.mark.valid_from("Prague")
+@pytest.mark.parametrize("final_storage_value", [0x02 << 248, 0x02])
+def test_bloatnet(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    final_storage_value: int,
```
jochem-brouwer, line 76: Test description assumes `final_storage_value` is 2 here.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+
+    # Calculate keccak256 of 85 bytes starting at memory[base]
+    code += Op.PUSH1(0x55) + Op.PUSH2(base) + Op.SHA3
+
+    # The address is the last 20 bytes (already on stack)
+    return code
+
+
+@pytest.mark.valid_from("Prague")
+@pytest.mark.parametrize("final_storage_value", [0x02 << 248, 0x02])
```
jochem-brouwer, line 71: I believe both these params will yield the same test. If you want to test storing 0x0200..00 and also storing 0x0000..02 then you likely want to use `Hash(0x02)`, this will left-pad the data (so you get `0x00..02` (32 bytes)) and this will then yield the 0x02 on stack and not the 0x02..00.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+        to=contract_address,
+        gas_limit=gas_benchmark_value,
+        data=(final_storage_value // 2).to_bytes(32, "big").rstrip(b"\x00"),
+        value=0,
+        sender=sender,
+    )
+    tx_1_2 = Transaction(
+        to=contract_address,
+        gas_limit=gas_benchmark_value,
+        data=final_storage_value.to_bytes(32, "big").rstrip(b"\x00"),
```
jochem-brouwer, line 124: As mentioned this transaction will a lot less gas than the benchmark value.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+
+# Warm reads are very cheap, which means you can really fill a block
+# with them. Only fill the block by a factor of SPEEDUP.
+SPEEDUP: int = 100
+
+
+@pytest.mark.valid_from("Prague")
+def test_bloatnet_sload_warm(
+    blockchain_test: BlockchainTestFiller, pre: Alloc, fork: Fork, gas_benchmark_value: int
+):
```
jochem-brouwer, line 142: These benchmark tests are somewhat similar:

https://github.com/ethereum/execution-spec-tests/blob/562fde148440cf1a6a80305dd899dd15d91f4c1e/tests/benchmark/test_worst_stateful_opcodes.py#L411

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+
+    # Calculate how many warm loads we can fit
+    gas_increment = gas_costs.G_WARM_SLOAD + gas_costs.G_BASE  # Warm SLOAD + POP
+    remaining_gas = gas_benchmark_value - totalgas
+    num_warm_loads = remaining_gas // (SPEEDUP * gas_increment)
+
+    # Build the complete code: warmup + repeated warm loads
+    sload_code = Op.SLOAD(0) + Op.POP if num_slots > 0 else Op.STOP
+    for i in range(1, num_slots):
+        sload_code = sload_code + Op.SLOAD(i) + Op.POP
```
jochem-brouwer, line 165: Idea: could use the storage value to point to the next storage slot to load. For instance this storage layout (key,value): `(0,1), (1,2), (2,0)` will thus iterate 0->1->2->0. Note: `POP` is thus not used here and neither is the `PUSH`). So this yields more SLOADS.

tests/benchmark/test_bloatnet.py
```diff
@@ -0,0 +1,618 @@
...
+
+    # Calculate expected gas usage precisely
+    expected_benchmark_gas_used = (
+        intrinsic_gas
+        + memory_expansion_cost
+        + (num_contracts * cost_per_iteration)
+        + final_storage_cost
+    )
+
+    blockchain_test(
```
jochem-brouwer, line 402: I believe this test is similar to https://github.com/ethereum/execution-spec-tests/blob/562fde148440cf1a6a80305dd899dd15d91f4c1e/tests/benchmark/test_worst_stateful_opcodes.py#L122 (this test does similar things to BALANCE/EXTCODEHASH and also CALLs them)

jochem-brouwer (commented):
Some comments and gotcha's mainly EVM related - I also think some of these tests are somewhat implemented in the benchmarks currently! Let me know if something is unclear, happy to help or to check if certain things exist elsewhere! 😄 👍

LouisTsai-Csie:
@jochem-brouwer Thanks for your review, but there is more context on this PR:

This PR is based on this [PR](https://github.com/ethereum/execution-spec-tests/pull/2040), it is still pending since we are exploring the best way to maintain both benchmark test and bloatnet cases(see discussion [here](https://github.com/ethereum/execution-spec-tests/pull/2040#discussion_r2312560433)) and also there is some non-resolved comment.

Now there are 6 tests in the PR, you could focus on the latter 2 tests here, or move the discussion to PR #2040, thanks a lot!

CPerezz:
At least for now, this should be the idea. I've tried to make some rough prototype of how the PR aims to look like (there's quite some polish left to do.

@jochem-brouwer @LouisTsai-Csie could you take a quick look and point anything that seems off? I think this is the way that makes the most sense as of now.

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,242 @@
...
+GAS_PER_CONTRACT_BALANCE_EXTCODECOPY = 5007  # BALANCE(cold) + EXTCODECOPY(warm, 24KB)
+
+# Configuration for CREATE2 pre-deployed contracts
+# These values must match what the deployment script generates
+# UPDATE THESE VALUES after running deploy_bloatnet_simple.py
+FACTORY_ADDRESS = Address("0x07EADb2f6b02Bb9fE994c0AeFf106625c0d3C93f")  # UPDATE THIS
+INIT_CODE_HASH = bytes.fromhex(
+    "8eba8b9d04aea4e6f3d6601ad364021e8e37b6282a0f57e40c635ba7f80aa0cb"
+)  # UPDATE THIS
+NUM_DEPLOYED_CONTRACTS = 1838  # UPDATE THIS - current setup for 5M gas
```
CPerezz, line 43: Currently using Geth development account which I've used to test that the overall deployment script + fixture generation + execute mode work as expected.

This needs to be updated with the final values after the deployment in Bloatnet (which I'll perform once we all agree this is the way to go).

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,242 @@
...
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "DUMMY/bloatnet.md"
+REFERENCE_SPEC_VERSION = "1.0"
+
+# Gas cost constants - used to calculate required contracts for any gas limit
+# See README.md for detailed breakdown of these costs
+GAS_PER_CONTRACT_BALANCE_EXTCODESIZE = 2707  # BALANCE(cold) + EXTCODESIZE(warm)
+GAS_PER_CONTRACT_BALANCE_EXTCODECOPY = 5007  # BALANCE(cold) + EXTCODECOPY(warm, 24KB)
```
CPerezz, line 34: These are rough estimates (enough for the user to be able to compute how many contracts are needed to deploy for particular tests.

We can remove the constants if preferred as these values are also found inside the tests.

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,242 @@
...
+    "8eba8b9d04aea4e6f3d6601ad364021e8e37b6282a0f57e40c635ba7f80aa0cb"
+)  # UPDATE THIS
+NUM_DEPLOYED_CONTRACTS = 1838  # UPDATE THIS - current setup for 5M gas
+
+
+def calculate_create2_address(factory: Address, salt: int, init_code_hash: bytes) -> Address:
+    """Calculate CREATE2 address from factory, salt, and init code hash."""
+    create2_input = b"\xff" + bytes(factory) + salt.to_bytes(32, "big") + init_code_hash
+    addr_bytes = keccak256(create2_input)[12:]
+    return Address(addr_bytes)
```
CPerezz, line 50: I should probably reuse the one that comes with this codebase already. And manage a bit the `factory_address` + `salt` mixing. (As IIRC the fn expected initcode directly and salt.

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,242 @@
...
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+):
+    """
+    BloatNet test using BALANCE + EXTCODESIZE pattern.
+
+    This test:
+    1. Uses 100+ pre-deployed 24KB contracts
```
CPerezz, line 64: I realize this is unneded. Will remove

CPerezz:
Left to do in this PR:
- Add a label for Bloatnet benchmarks (so we can trigger as a group with `execute`.
- Update SLOAD case from @gballet 's PR in #2040  and possibly superseed his PR.
- The rest is in the comments I added.

scripts/test_create2.py
```diff
@@ -0,0 +1,91 @@
...
+    "60" + "00"  # PUSH1 0x00 (salt - using 0 for simplicity)
+    "36"  # CALLDATASIZE (size of init code)
+    "60" + "00"  # PUSH1 0x00 (offset in memory)
+    "60" + "00"  # PUSH1 0x00 (value)
+    "f5"  # CREATE2
+
+    "60" + "00"  # PUSH1 0x00
+    "52"  # MSTORE (store address at 0)
+    "60" + "20"  # PUSH1 0x20
+    "60" + "00"  # PUSH1 0x00
```
jochem-brouwer, line 43: PUSH1 0 could also be written as PUSH0 (`0x5f`) - this is cheaper but not supported on chains which do not have PUSH0. See also https://eips.ethereum.org/EIPS/eip-7997

scripts/test_create2.py
```diff
@@ -0,0 +1,91 @@
...
+    'data': '0x' + simple_bytecode
+})
+
+if result:
+    deployed_addr = '0x' + result[-20:].hex()
+    print(f"Contract deployed at: {deployed_addr}")
+
+    # Verify by checking code
+    code = w3.eth.get_code(deployed_addr)
+    print(f"Deployed code length: {len(code)} bytes")
```
jochem-brouwer, line 89: This should store as code: 0x00..0042 (32 bytes)

tests/benchmark/bloatnet/deploy_bloatnet_simple.py
```diff
@@ -0,0 +1,383 @@
...
+}
+
+
+def generate_max_size_bytecode(salt: int = 0) -> Tuple[bytes, bytes]:
+    """Generate 24KB contract bytecode for standard bloatnet tests."""
+    # Init code copies runtime bytecode to memory and returns it
+    init_code = bytearray()
+
+    # Init code: PUSH2 size, PUSH1 offset, PUSH1 dest, CODECOPY, PUSH2 size, PUSH1 0, RETURN
+    bytecode_size = 24576
```
jochem-brouwer, line 43: You can read this constant from `fork: Fork` (provided in tests) and read `fork.max_code_size()` to read the constant from the fork.

tests/benchmark/bloatnet/deploy_create2_factory.py
```diff
@@ -0,0 +1,130 @@
...
+
+    test_account = w3.eth.accounts[0]
+    print(f"Using test account: {test_account}")
+
+    # Minimal CREATE2 factory bytecode
+    # Takes salt (first 32 bytes) and bytecode (rest) from calldata
+    # Returns the deployed address
+    factory_bytecode = bytes(
+        [
+            # Runtime code
```
jochem-brouwer, line 29: I strongly suggest to read a bit about the tooling EEST provides to write these kind of bytecode tests. 

For instance, here is CREATE2 factory code: https://github.com/ethereum/execution-spec-tests/blob/291fe00e65d82e05ca2bd5e7a49e4ac4a8c7a5b9/tests/benchmark/test_worst_bytecode.py#L142-L160 which will factory-deploy this initcode: https://github.com/ethereum/execution-spec-tests/blob/291fe00e65d82e05ca2bd5e7a49e4ac4a8c7a5b9/tests/benchmark/test_worst_bytecode.py#L118-L137 which itself gets called from another contract which will invoke the factory contract N times (calldata input) https://github.com/ethereum/execution-spec-tests/blob/291fe00e65d82e05ca2bd5e7a49e4ac4a8c7a5b9/tests/benchmark/test_worst_bytecode.py#L165-L168.

jochem-brouwer: EEST will else help with things like formatting numbers for you, i.e. 0xff gets put in a PUSH1 but 0xffff will be put in a PUSH2 (so you dont have to do the bitwise logic like the shifting and the AND masks here)

tests/benchmark/bloatnet/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Gas Cost Constants
+
+### BALANCE + EXTCODESIZE Pattern
+**Gas per contract: 2,707**
+- `PUSH20` (address): 3 gas
+- `BALANCE` (cold access): 2,600 gas
+- `POP`: 2 gas
+- `EXTCODESIZE` (warm): 100 gas
+- `POP`: 2 gas
```
jochem-brouwer, line 19: This tests is basically two scenarios in one. It tests both BALANCE (which marks it warm) and EXTCODESIZE.

Note that for accounts in the Merkle Patricia Trie in the state, account are stored as:

`[nonce, balance, storageRoot, codeHash]`

Thus reading balance from MPT will "just" require reading the account. `EXTCODESIZE` however means we have to query `codeHash`, and to get the size we have to lookup **all the code** from the DB in order to determine the size (this assumes that the client has not optimized this some way, for instance via an extra database like a `codeHash => codeSize` lookup which would skip first reading all the code to determine size). 

So, I believe we need scenarios for BALANCE/EXTCODESIZE.

For EXTCODESIZE, I think this benchmark test is what you want: https://github.com/ethereum/execution-spec-tests/blob/291fe00e65d82e05ca2bd5e7a49e4ac4a8c7a5b9/tests/benchmark/test_worst_bytecode.py#L41

jochem-brouwer: For BALANCE (cold) this test: https://github.com/ethereum/execution-spec-tests/blob/291fe00e65d82e05ca2bd5e7a49e4ac4a8c7a5b9/tests/benchmark/test_worst_stateful_opcodes.py#L39

tests/benchmark/bloatnet/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### BALANCE + EXTCODECOPY Pattern
+**Gas per contract: ~5,007**
+- `PUSH20` (address): 3 gas
+- `BALANCE` (cold access): 2,600 gas
+- `POP`: 2 gas
+- `EXTCODECOPY` setup: ~100 gas
+- `EXTCODECOPY` (24KB): ~2,300 gas
```
jochem-brouwer, line 27: (same comment regarding BALANCE + EXTCODECOPY here, should be in different tests so we can investigate the individual behavior of these opcodes instead of mixed)

For EXTCODECOPY, we want to force clients to read all code. But since we expand memory for the copied bytes and pay for those copied bytes also, copying 1 byte instead of the 24 kb should be sufficient (this forces clients to load code from disk since we need to know what that specific byte is). Therefore we should read the final byte of the contract. 

This test is relevant: https://github.com/ethereum/execution-spec-tests/blob/291fe00e65d82e05ca2bd5e7a49e4ac4a8c7a5b9/tests/benchmark/test_worst_bytecode.py#L192

I'm not sure why it copies 1000 bytes, but this should be edited to read 1 byte so we can target more accounts. (1000 byte copy likely from the original idea of these benchmarks (zkEVM) because we want to measure the worst case zk cycles there, not the worst state attack, so slightly different performance perspective regarding worst case scenarios there)

jochem-brouwer: Note: in linked tests there is also a calculation in the pre-setup phase to calculate how much contracts are necessary. This uses an upper bound, it *should* be slightly less in practice, so attack block will always read non-empty accounts

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # Attack transaction
+    tx = Transaction(
+        to=attack_address,
+        gas_limit=gas_benchmark_value,
+        sender=pre.fund_eoa(),
+    )
+
+    # Post-state: just verify attack contract exists
+    # Benchmark tests run out of gas, so no success flag
```
jochem-brouwer, line 130: For these kind of attacks OOG is fine, but if we do edit anything in state, it should not OOG because we have to ensure the writes are flushed to trie and not discarded (just a side note)

jochem-brouwer (commented):
Left some general comments! I think to write these bytecode it would be helpful to take a look at the tooling EEST provides to make writing these bytecode easier!

In general: most of the attack scenarios (in isolated form, so not BALANCE + EXTCODESIZE/EXTCODECOPY but rather BALANCE or EXTCODESIZE or EXTCODECOPY) are written in the benchmarks folder already (I linked those). 

These are from tests related to zkEVM where we want to find the scenarios which yield the most zk cycles to get an idea how "bad" these are. This is a different perspective of "worst case" than we want to investigate here. Some tests will need some slight edits.

Here is what I think we should do:

The current tests (EXTCODESIZE/EXTCODECOPY (and BALANCE but this will not change if we target accounts with or without code)) target a HUGE AMOUNT of contracts with code. What we want to do is **only deploy these accounts on the bloatnet once** and then write our tests in such way that they target these accounts. (Note: the "risk" of this is that these contracts are in the memory cache of the clients, which we don't want, since the worst case scenario means that client has to read from disk and not from memory cache). This is (I think) a good start.

We can therefore take these written tests, edit those, skip the pre-deploy phase and instead run the attacks directly on the bloatnet where we target these predeployed contracts. We can also repurpose those accounts for other tests, like EXTCODEHASH tests. 

For tests specifically executing code (so CALL, CALLCODE, DELEGATECALL, STATICCALL) we have to also take into account the JUMPDEST analysis which will perform good on some clients for specific code, but bad on other clients which will perform better on other formatted code (this depends on what pattern they have optimized for, mostly if they store valid JUMPDEST or if they store intermediate bytes). That test is here: https://github.com/ethereum/execution-spec-tests/blob/291fe00e65d82e05ca2bd5e7a49e4ac4a8c7a5b9/tests/benchmark/test_worst_bytecode.py#L247

It would be helpful to immediately deploy the contracts with one of these patterns for the EXTCODECOPY/EXTCODESIZE tests such that we can repurpose the factory-created contracts for these CALL tests. (To investigate other JUMPDEST patterns and their impact we should thus deploy a new set of contracts, but this is something we will tackle later).

Let me know what you think! 😄 👍

tests/benchmark/bloatnet/README.md
CPerezz: I'm not sure I follow you here. 

If these standalone scenarios already exist as you correctly pointed out, and my PR adds the combination of them into a single test, what is actually needed further from what this PR adds?

tests/benchmark/bloatnet/deploy_bloatnet_simple.py
CPerezz: Modified!

tests/benchmark/bloatnet/test_bloatnet.py
CPerezz: In our cases for SLOAD, EXTCODE_X for now we never write to state. So I'm getting more and more convinced of removing them basically.

WDYT?

tests/benchmark/bloatnet/deploy_create2_factory.py
CPerezz: To be clear, If STEEL team doesn't want to have the deployment helper files within EEST (which I could def understand), then I'd prefer avoiding the dependency on EEST for the standalone files. (As I'll publish them as a gist or similar so that they can be executed after a wget or whatever).

If they remain here, I agree the refactor makes sense. Pushed it for you all to see the diff in: cf2c7c6

CPerezz:
@jochem-brouwer thanks a lot for taking the time to review. A couple points:
> In general: most of the attack scenarios (in isolated form, so not BALANCE + EXTCODESIZE/EXTCODECOPY but rather BALANCE or EXTCODESIZE or EXTCODECOPY) are written in the benchmarks folder already (I linked those).

Thanks for pointing out! Yes, in fact I ignored the single-opcode implementations of these cases because of this exact reason. I think anyways that we will need to "fork them" or similar since we want to also explore the opcodes by themseleves but targeting already-deployed contracts (as our tests are meant to run worst-cases of 150M gas_limit (to account for 20% refund tricks).

So overall, I think this is a good next-task too. And even if it's a bit ugly to duplicate the tests, I think the avoidance of deploying all these contracts is really important. And, at the same time, I don't want to touch zkevm tests so that they have consistent behaviour agains their prior benchmarks against their specific code.
LMK what you think.

> For tests specifically executing code (so CALL, CALLCODE, DELEGATECALL, STATICCALL) we have to also take into account the JUMPDEST analysis which will perform good on some clients for specific code, but bad on other clients which will perform better on other formatted code (this depends on what pattern they have optimized for, mostly if they store valid JUMPDEST or if they store intermediate bytes).

Here with the 63/64 rule will mostly kill any attempt to abuse `CALL`-like opcodes (at least that was my thought). Anyways, there are proposals that suggest removing it (see: https://ethereum-magicians.org/t/eip-7923-linear-page-based-memory-costing/23290/11). 
Nonetheless, this is another of the tasks that needs to be done. I wonder if you have any suggestions on how to exploit the Bloatnet characteristics (big state & high gas limit) to abuse these opcodes as much as possible.

tests/benchmark/bloatnet/test_bloatnet.py
CPerezz: Done in a862f765

tests/benchmark/bloatnet/test_bloatnet.py
CPerezz: Done in a862f765

CPerezz:
The PR is ready for review @LouisTsai-Csie @marioevz 

I think once you approve, I will deploy all the contracts in Bloatnet and make the final commit updating the addresses. Then we can merge.

tests/benchmark/bloatnet/test_bloatnet.py
jochem-brouwer: Removing those read tests? They are definitely necessary. Although you do not write anything to state, it is not known beforehand that you do not write anything to state (besides: you actually *do* write something to state: namely, you deduct the gas*gasPrice from the tx sender and update the nonce, so you still need to know the gas spent on execution) so you have to execute these (disk read) operations anyways. We would still need to know the worst case behavior is also if there are only disk read operations. So I would not recommend to remove those.

jochem-brouwer:
> Here with the 63/64 rule will mostly kill any attempt to abuse CALL-like opcodes (at least that was my thought). Anyways, there are proposals that suggest removing it (see: https://ethereum-magicians.org/t/eip-7923-linear-page-based-memory-costing/23290/11).

Yes this would be the case if we keep CALLing into contracts, but this is not the attack scenario here. The attack is to call from a root contract into the children contract deployed via the CREATE2 factory (these should all have unique code). The call depth is thus max 2. In the children contract we perform a JUMP to the end of the code. This forces clients to perform JUMPDEST analysis, which means to scan (from the start of the code all to the target `pc` we are targeting) all the opcodes to mark invalid bytes (the intermediate bytes of PUSH) as invalid JUMPDESTs in order to verify the JUMP target is indeed a valid JUMPDEST. This would therefore clients to load (1) X unique contracts from disk and (2) perform JUMPDEST analysis on X*24 KiB contracts. Which my feeling would be relatively heavy on both I/O and on execution logic. (This test is written, linked here: https://github.com/ethereum/execution-spec-tests/blob/9f96fadb3836d20e7afbdc58a8678f9c2df02e30/tests/benchmark/test_worst_bytecode.py#L247)

> So overall, I think this is a good next-task too. And even if it's a bit ugly to duplicate the tests, I think the avoidance of deploying all these contracts is really important. And, at the same time, I don't want to touch zkevm tests so that they have consistent behaviour agains their prior benchmarks against their specific code.
LMK what you think.

I think we could duplicate the tests but instead write them in such way, that if two tests target the same type of code deployed from the CREATE2 factory, we assume when executing these tests those child contract are already deployed and then only run the attack phase of the test. This would avoid having to deploy these large amount of contracts and, especially when benching these scenarios, thus makes it much easier to re-bench it. (Without having to wait X time before those thousands of contracts are deployed)

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,246 @@
...
+from ethereum_test_tools import (
+    Account,
+    Address,
+    Alloc,
+    Block,
+    BlockchainTestFiller,
+    Transaction,
+    While,
+)
+from ethereum_test_tools.vm.opcode import Opcodes as Op
```
LouisTsai-Csie, line 20: There are some folder restructuring in EEST, please rebase first and the path could be updated as follows to avoid conflict.

```suggestion
from ethereum_test_vm.opcodes import Opcodes as Op
```

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # Calculate how many contracts to access
+    available_gas = gas_benchmark_value - intrinsic_gas - 1000  # Reserve for cleanup
+    contracts_needed = int(available_gas // cost_per_contract)
+
+    # Limit to actually deployed contracts
+    num_contracts = min(contracts_needed, NUM_DEPLOYED_CONTRACTS)
+
+    if contracts_needed > NUM_DEPLOYED_CONTRACTS:
+        import warnings
```
LouisTsai-Csie, line 75: nit: could we import at the beginning of the file.

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,246 @@
...
+    # Post-state: just verify attack contract exists
+    post = {
+        attack_address: Account(storage={}),
+    }
+
+    blockchain_test(
+        pre=pre,
+        blocks=[Block(txs=[attack_tx])],
+        post=post,
+    )
```
LouisTsai-Csie, line 134: The CI will fail sine the gas usage is mismatched, but we already resolve the issue by PR https://github.com/ethereum/execution-spec-tests/pull/2155:

```suggestion
    blockchain_test(
        pre=pre,
        blocks=[Block(txs=[attack_tx])],
        post=post,
        skip_gas_used_validation=True,
    )
```

tests/benchmark/bloatnet/test_bloatnet.py
```diff
@@ -0,0 +1,246 @@
...
+    # Post-state
+    post = {
+        attack_address: Account(storage={}),
+    }
+
+    blockchain_test(
+        pre=pre,
+        blocks=[Block(txs=[attack_tx])],
+        post=post,
+    )
```
LouisTsai-Csie, line 246:
```suggestion
  blockchain_test(
        pre=pre,
        blocks=[Block(txs=[attack_tx])],
        post=post,
        skip_gas_used_validation=True,
    )
```

LouisTsai-Csie (changes requested):
Hi @CPerezz, thanks for the latest changes. I ran your tests locally and have a few suggestions to help them pass.

Since this PR is based on PR #2040, I believe it can’t be merged until that one is merged (please correct me if I’m wrong). At the moment, we also can’t run the CI. My suggestion would be to open a new PR and migrate the `tests/benchmark/bloatnet/test_bloatnet.py` file there. We do not need to migrate the `deploy_create2_factory_refactored.py` and `deploy_bloatnet_simple.py` for now, as we discussed integration in a separate PR.

We can keep this PR open so I can create follow-up issues from it.

LouisTsai-Csie:
Closes as PR #2186 and PR #2242 completes. Feel free to create issues if I miss any todo in this PR. Thanks everyone for the contribution, review and feedback.

PR #2098, opened 2025-09-03
refactor(tests): add checklist marker for eip7883

LouisTsai-Csie:
## 🗒️ Description

EIP-7883 updates the gas calculation formula. This PR marks the appropriate checklist item for the eip.

This category aligns with the `Gas Cost Changes` section, but since test coverage in the previous fork was incomplete, I’ve included it under `New Precompiles` for review.

**New Precompiles** - [Link](https://eest.ethereum.org/main/tests/osaka/eip7883_modexp_gas_increase/checklist/#new-precompile) 

- [x] Call contexts
    - [x] CALL
    - [x] DELEGATECALL
    - [x] STATICCALL
    - [x] CALLCODE
    - [x] Transaction Entry-point
    - [x] Initcode call
    - [x] Precompile as Set-code Delegated Address -> used in `test_set_code_txs.py`.
- [ ] Inputs
- [x] Value Transfer
    - Minimum Fee Precompile: No value transfer is required
    - [x]No-Fee Precompile
- [ ] Out-of-bounds checks
    - [ ] precompile/test/out_of_bounds/max
    - [x] precompile/test/out_of_bounds/max_plus_one
- [x] Input Lengths
    - [x] Zero-length Input
    - [x] Static Required Input Length
    - [x] Dynamic Required Input Length
- [x] Gas usage
    - Constant Gas Cost: The gas cost is dynamic
    - [x] Variable Gas Cost
    - [x] Excessive Gas
- [x] Fork transition
    - [x] Pre-fork Block Call
    - [x] Cold/Warm Precompile Address State: Note Modexp is not a new one, so it is warm even before fork activation

**Gas Cost Changes** - [Link](https://eest.ethereum.org/main/tests/osaka/eip7883_modexp_gas_increase/checklist/#gas-cost-changes) 
- [x] Gas Usage: 
    - [x] gas_cost_changes/test/gas_updates_measurement
- [ ] Out-of-gas
- [x] Fork transition

Missing Test:

- precompile/test/gas_usage/dynamic/exact
- precompile/test/gas_usage/dynamic/oog
- gas_cost_changes/test/out_of_gas 

## 🔗 Related Issues or PRs

Issue https://github.com/ethereum/execution-spec-tests/issues/1791

## ✅ Checklist

[...]

PR #2111, opened 2025-09-08
chore(tests): fix modexp conftest for tx gas limit cap

spencer-tb:
## 🗒️ Description

Fixes the tx gas limit cap (from EIP-7825) for ModExp (EIP-7883) tests.

## 🔗 Related Issues or PRs

#2108 

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
+++ b/tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
@@ -322,7 +322,6 @@ def create_modexp_variable_gas_test_cases():
         ("00" * 1024, "00" * 32, "01" * 32, "00" * 31 + "01", "Z13"),
         ("01" * 32, "00" * 1024, "00" * 32, "00" * 32, "Z14"),
         ("01" * 32, "00" * 31 + "01", "00" * 1024, "00" * 1024, "Z15"),
-        ("00" * 32, "00" * 1024, "01" * 1024, "00" * 1023 + "01", "Z16"),
         # Maximum value stress tests
         ("FF" * 64, "FF" * 64, "FF" * 64, "00" * 64, "M1"),
         ("FF" * 32, "01", "FF" * 32, "00" * 32, "M2"),
@@ -456,4 +455,10 @@ def test_modexp_variable_gas_cost_exceed_tx_gas_cap(state_test, pre, tx, post):
     Test ModExp variable gas cost.
     Inputs with an expected gas cost over the EIP-7825 tx gas cap.
     """
+    # Test case coverage table (gas cap):
+    # ┌─────┬──────┬─────┬──────┬───────┬─────────┬───────────────────────────────────────────────┐
+    # │ ID  │ Comp │ Rel │ Iter │ Clamp │   Gas   │ Description                                   │
+    # ├─────┼──────┼─────┼──────┼───────┼─────────┼───────────────────────────────────────────────┤
+    # │ Z16 │  L   │  <  │  C   │ False │520060928│ Zero base, zero exp, large modulus (gas cap)  |
+    # └─────┴──────┴─────┴──────┴───────┴─────────┴───────────────────────────────────────────────┘
     state_test(pre=pre, tx=tx, post=post)


tests/osaka/eip7883_modexp_gas_increase/conftest.py
```diff
@@ -40,12 +40,38 @@ def call_contract_post_storage() -> Storage:
...
 @pytest.fixture
-def call_succeeds() -> bool:
+def total_tx_gas_needed(
+    fork: Fork, modexp_expected: bytes, modexp_input: ModExpInput, precompile_gas: int
+) -> int:
+    """Calculate total tx gas needed for the transaction."""
+    intrinsic_gas_cost_calculator = fork.transaction_intrinsic_cost_calculator()
+    memory_expansion_gas_calculator = fork.memory_expansion_gas_calculator()
+    sstore_gas = fork.gas_costs().G_STORAGE_SET * (len(modexp_expected) // 32)
+    extra_gas = 100_000
```
LouisTsai-Csie, line 50: I think we no longer need `sstore_gas`, since we switch to hash-based comparison method. I also do the refactor in my [PR](https://github.com/ethereum/execution-spec-tests/pull/2098). What do you think?

```suggestion
intrinsic_gas_cost_calculator = fork.transaction_intrinsic_cost_calculator()
memory_expansion_gas_calculator = fork.memory_expansion_gas_calculator()
extra_gas = 500_000
```

LouisTsai-Csie: If this works, also remove the `sstore_gas` in the return.

tests/osaka/eip7883_modexp_gas_increase/conftest.py
```diff
@@ -40,12 +40,38 @@ def call_contract_post_storage() -> Storage:
...
+        + precompile_gas
+        + sstore_gas
+    )
+
+
+@pytest.fixture
+def exceeds_tx_gas_cap(total_tx_gas_needed: int, fork: Fork) -> bool:
+    """Determine if total gas requirements exceed transaction gas cap."""
+    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
+    return tx_gas_limit_cap is not None and total_tx_gas_needed > tx_gas_limit_cap
```
LouisTsai-Csie, line 65: Checking the data type for tx_gas_limit_cap is so annoying XD! Do you think it is a good idea to do the following?

```suggestion
@pytest.fixture
def exceeds_tx_gas_cap(total_tx_gas_needed: int, fork: Fork, env: Environment) -> bool:
    """Determine if total gas requirements exceed transaction gas cap."""
    tx_gas_limit_cap = fork.transaction_gas_limit_cap() or env.gas_limit
    return total_tx_gas_needed > tx_gas_limit_cap
```

tests/osaka/eip7883_modexp_gas_increase/conftest.py
```diff
@@ -174,28 +200,12 @@ def tx(
...
-        + sstore_gas
-    )
-
     tx_gas_limit_cap = fork.transaction_gas_limit_cap()
-
     if tx_gas_limit_cap is not None:
-        return min(tx_gas_limit_cap, total_gas)
-    return total_gas
+        return min(tx_gas_limit_cap, total_tx_gas_needed)
+    return total_tx_gas_needed
```
LouisTsai-Csie, line 208: What about:

```python
@pytest.fixture
def tx_gas_limit(total_tx_gas_needed: int, fork: Fork, env: Environment) -> int:
    """Transaction gas limit used for the test (Can be overridden in the test)."""
    tx_gas_limit_cap = fork.transaction_gas_limit_cap() or env.gas_limit
    return min(tx_gas_limit_cap, total_tx_gas_needed)
```

 `tx_gas_limit_cap = fork.transaction_gas_limit_cap() or env.gas_limit` -> This means if you have tx gas limit cap, use it, or we take block gas limit instead.

This could run successfully, and lint & type is passing on my local machine!

LouisTsai-Csie (changes requested):

tests/osaka/eip7883_modexp_gas_increase/conftest.py
spencer-tb: Added :)

tests/osaka/eip7883_modexp_gas_increase/conftest.py
spencer-tb: Added :)

LouisTsai-Csie (approved):
Nice! I will try the remaining refactoring in my PR, thanks for this

tests/osaka/eip7883_modexp_gas_increase/conftest.py
```diff
@@ -40,12 +49,38 @@ def call_contract_post_storage() -> Storage:
...
+
+@pytest.fixture
+def call_succeeds(exceeds_tx_gas_cap: bool) -> bool:
     """
     By default, depending on the expected output, we can deduce if the call is expected to succeed
-    or fail.
+    or fail. Under EIP-7825, transactions requiring more gas than the cap should fail.
     """
-    return True
+    return not exceeds_tx_gas_cap
```
marioevz, line 83: My worry with this change is that if for some reason in the future most of the tests exceed the gas limit, either because we further change the precompile costs or we lower the transaction gas limit cap, we are not going to realize if more tests start exceeding this limit and we lose coverage.

I would lean towards:
- Raise an exception and fail to fill the test if we exceed the gas limit cap in any of the cases.
- Mark this specific case as valid-until-prague.
- Ideally, generate a similar valid-from-osaka case where the gas limit cap is not exceeded but using similar inputs if possible.

marioevz (changes requested):
See my comment, I think current solution could be improved because future changes might break tests if we are not careful.

tests/osaka/eip7883_modexp_gas_increase/conftest.py
spencer-tb: I understand what you mean. I added a reasonable solution I think.

We get this output using this commit: https://github.com/ethereum/execution-spec-tests/pull/2111/commits/c296a9cb2871c7dc608eef1e42f13405e9ac47ae
```
ERROR tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py::test_modexp_variable_gas_cost[fork_Osaka-state_test-Z16] - Failed: Test unexpectedly exceeds tx gas cap. Either mark with `expected_tx_cap_fail=True` or adjust inputs.
```

spencer-tb: The test then passes using this commit: https://github.com/ethereum/execution-spec-tests/pull/2111/commits/c57586ff7f1de533c736b467aad51f9f29ac6241

marioevz: I see the fix, great solution, thanks so much!

PR #2124, opened 2025-09-09
feat(tests): Add flexibility to expected absent scenarios for BALs

fselmo:
## 🗒️ Description

We should have flexibility in defining the absence cases we expect for block-level access lists. This allows us to define absence validators for any case we might want to validate against. I don't expect these to grow very much but this does provide some flexibility.

I would very much welcome something simpler, without having to define validators, but with all the possible conditions for BALs, it feels that defining quick validators would give us the most flexibility for `should_not_exist` checks. This was also similar to the approach for the modifiers so the devex feels fairly similar.

---

Note: This PR is rebased on top of #2138. Once that is merged, I will do a final cleanup.

---

## ✅ Checklist

    ```console
    ```


src/ethereum_test_types/block_access_list/absence_validators.py
```diff
@@ -0,0 +1,97 @@
...
+            if balance_change.tx_index in tx_set:
+                raise AssertionError(
+                    f"Unexpected balance change found at tx {balance_change.tx_index}"
+                )
+
+    return check
+
+
+@validate_call
+def storage_changes_at_slots(*slots: StorageKey) -> AbsenceValidator:
```
raxhvl, line 50: Shouldn't this ask for a transaction index to?

src/ethereum_test_types/block_access_list/absence_validators.py
```diff
@@ -0,0 +1,97 @@
...
+            if storage_slot.slot in slots_set:
+                raise AssertionError(
+                    f"Unexpected storage change found at slot {storage_slot.slot}"
+                )
+
+    return check
+
+
+@validate_call
+def storage_reads_at_slots(*slots: StorageKey) -> AbsenceValidator:
```
raxhvl, line 65: Also transaction index here?

src/ethereum_test_types/tests/test_block_access_lists.py
```diff
@@ -206,3 +214,273 @@ def test_missing_expected_address():
...
     with pytest.raises(Exception, match="Expected address .* not found in actual BAL"):
         expectation.verify_against(actual_bal)
+
+
+@pytest.mark.parametrize("has_change_should_raise", [True, False])
+def test_absence_validator_nonce_changes(has_change_should_raise):
+    """Test nonce_changes_at_tx validator with present/absent changes."""
+    alice = Address("0x000000000000000000000000000000000000000a")
+
+    nonce_changes = [BalNonceChange(tx_index=1, post_nonce=1)]
```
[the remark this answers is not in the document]
fselmo: resolved 👍🏼

fselmo:
Rebased on top of #2138, which is more important to get in first and touches the same files.

fselmo:
There are updates we need to add to the engine payloads, from updates on the comment here: https://discord.com/channels/595666850260713488/1364000387195076608/1417401978413121539. I will put this in a separate PR right now to separate concerns a bit more.

spencer-tb:
> I tried this out, instead of:
> 
> ```python
>         account_expectations={
>             alice: BalAccountExpectation(
>                 nonce_changes=[
>                     BalNonceChange(tx_index=1, post_nonce=1),
>                 ],
>                 should_not_exist=[
>                     no_nonce_changes({2, 4}),
>                 ],
>             ),
>         }
> ```
> 
> if we make the should not exist a dict of {nonce_changes, blance_changes, ...} like account expectation:
> 
> ```python
>         account_expectations={
>             alice: BalAccountExpectation(
>                 nonce_changes=[
>                     BalNonceChange(tx_index=1, post_nonce=1),
>                 ],
>                 should_not_exist={
>                   nonce_changes=[
>                       BalNonceChange(tx_index=2, post_nonce=4),
>                   ],                ],
>             ),
>         }
> ```
> 
> It would make the API consistent. What do you think?

I personally like this better! So in favor from my side.

fselmo:
I like the approach here as well. I think the functional approach makes sense for the modifiers moreso than here. The only question I have is whether we have enough flexibility for testing this but I think so right? Because we would define these as sub-sequences of the expectation, just like we do with the positive cases.

One small syntax correction, that I think opens up another conversation though. This isn't a dictionary you defined here. You opened the brackets and then used `kwargs` syntax. So do we want a `dict` where you have to spell out `"nonce_changes": [...]`, for example... or do we want something like `BalAccountAbsences(nonce_changes=[...]`?

cc: @spencer-tb, @raxhvl

raxhvl:
The class approach is much better.

PR #2138, opened 2025-09-11
feat(tests): Enforce EIP-7928 ordering rules

fselmo:
## 🗒️ Description

To be clear, the current tests are not broken because the spec is implemented correctly. However, we were not enforcing certain validation rules that are required by EIP-7928. These changes introduce an initial up-front check on the entire order of the BAL coming from t8n as defined in EIP-7928:

> Ordering and Determinism
> 
> The following ordering rules MUST apply:
> 
>     - Addresses: lexicographic (bytewise).
>     - Storage keys: lexicographic within each account.
>     - Block access indices: ascending within each change list.

### `BlockAccessList` (t8n model)

We fully validate all ordering upfront, based on the above rules, for the entire BAL coming from t8n. This is crucial that it meets all the requirements for the EIP implementation.

#### Address Ordering

We define the expectation as a `dict` with `address: BalAccountChanges`. We don't need to check that our ordering is correct here because we will validate that the BAL coming from t8n is already in the correct order. So if the addresses we define are present at all, they will be in the correct order already, and we are free to define this as an unordered `dict` in the expectation test class. This is especially "nice" because utilities like `pre.deploy_contract()` and `pre.fund_eoa()` will generate addresses for us and it would be very difficult to put the burden on the test writer to place these within an `OrderedDict`, for example.

### Transaction Index and Slot Ordering

We DO validate that the transaction indexes and slots appear in the correct order, but only as a sub-sequence of the full list - there’s no need to define every element. Since these are defined within a Python `list`, we do impose on the test writer that the sub-sequence they define here should be in order to maintain readability and reduce confusion for a reader of the test. I don't think this is too much to ask nor too big of a burden to place.

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/src/ethereum_test_types/block_access_list/__init__.py
+++ b/src/ethereum_test_types/block_access_list/__init__.py
@@ -6,7 +6,7 @@ these are simple data classes that can be composed together.
 """
 
 from functools import cached_property
-from typing import Any, Callable, ClassVar, Dict, List
+from typing import Any, Callable, ClassVar, Dict, List, Union
 
 import ethereum_rlp as eth_rlp
 from pydantic import Field, PrivateAttr
@@ -256,6 +256,11 @@ class BlockAccessListExpectation(CamelModel):
         """
         Verify that the actual BAL from the client matches this expected BAL.
 
+        Validation steps:
+        1. Validate actual BAL conforms to EIP-7928 ordering requirements
+        2. Verify address expectations - presence or explicit absence
+        3. Verify expected changes within accounts match actual changes
+
         Args:
             actual_bal: The BlockAccessList model from the client
 
@@ -263,15 +268,17 @@ class BlockAccessListExpectation(CamelModel):
             Exception: If verification fails
 
         """
-        actual_accounts_by_addr = {acc.address: acc for acc in actual_bal.root}
+        # validate the actual BAL structure follows EIP-7928 ordering
+        self._validate_bal_ordering(actual_bal)
 
+        actual_accounts_by_addr = {acc.address: acc for acc in actual_bal.root}
         for address, expectation in self.account_expectations.items():
             if expectation is None:
                 # check explicit exclusion of address when set to `None`
                 if address in actual_accounts_by_addr:
                     raise Exception(f"Address {address} should not be in BAL but was found")
             else:
-                # Address should be in BAL with expected values
+                # check address is present and validate changes
                 if address not in actual_accounts_by_addr:
                     raise Exception(f"Expected address {address} not found in actual BAL")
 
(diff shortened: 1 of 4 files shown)


fselmo:
The last commit here attempts to fix issues relate to `consume direct` with geth which we agreed, for now (and for consistency I think it's also good), on Eth R&D would be hex-encoded integers.

Unrelated but maybe relevant, the current state of the tests  on geth branch `jwasinger/bal-execution` (the latest BAL branch):

- Pass all `consume rlp` tests 
- Fail for `consume engine`. But after debugging issues with the current engine flow in geth, pass all tests except for invalid tests - which is just an exception mismatch issue and can be mitigated via exception mapper
- Fail `consume direct` which geth expects as uint16 currently but will be expecting hex-encoded integers which this PR fixes. I tested the current implementation changing those values to `int` and all `consume direct` tests pass so if we agree both on the EEST side and geth side, they should end up passing once the agreement is met (hex-encoded).

src/ethereum_test_types/block_access_list/__init__.py
```diff
@@ -281,6 +299,89 @@ def verify_against(self, actual_bal: "BlockAccessList") -> None:
...
+            )
+
+        # Check ordering within each account
+        for account in bal.root:
+            # Check storage slots are in lexicographic order
+            if account.storage_changes:
+                slots = [s.slot for s in account.storage_changes]
+                sorted_slots = sorted(slots, key=lambda x: bytes(x))
+                if slots != sorted_slots:
+                    raise Exception(
```
spencer-tb, line 337: Should we define python specific BAL exception types? Just an idea but not essential: `BALStorageOrderError` etc

src/ethereum_test_types/block_access_list/__init__.py
```diff
@@ -281,6 +299,89 @@ def verify_against(self, actual_bal: "BlockAccessList") -> None:
...
+                        f"Account {account.address}: Storage slots not in lexicographic order. "
+                        f"Got: {slots}, Expected: {sorted_slots}"
+                    )
+
+                # Check tx indices within each storage slot are in ascending order
+                for slot_change in account.storage_changes:
+                    if slot_change.slot_changes:
+                        tx_indices = [c.tx_index for c in slot_change.slot_changes]
+                        if tx_indices != sorted(tx_indices):
```
spencer-tb, line 346: We use do this a few times I think. It could be a seperate function: `_validate_tx_indices`

src/ethereum_test_types/block_access_list/__init__.py
```diff
@@ -281,6 +299,89 @@ def verify_against(self, actual_bal: "BlockAccessList") -> None:
...
+
+            # Check tx indices in other change lists
+            changes_to_check: List[
+                tuple[
+                    str,
+                    Union[
+                        List[BalNonceChange],
+                        List[BalBalanceChange],
+                        List[BalCodeChange],
+                    ],
```
spencer-tb, line 369: Does it make sense for this to be its own type?
```
TxChangeType = Union[BalNonceChange, BalBalanceChange, BalCodeChange]
```

src/ethereum_test_types/block_access_list/__init__.py
```diff
@@ -304,109 +413,154 @@ def _compare_account_expectations(
...
+                            actual_idx += 1
+                            break
+                        actual_idx += 1
+
+                    if not found:
+                        exp_str = exp_read.hex() if isinstance(exp_read, bytes) else str(exp_read)
+                        actual_str = [
+                            r.hex() if isinstance(r, bytes) else str(r) for r in actual_reads
+                        ]
+                        raise AssertionError(
```
spencer-tb, line 439: Similar here for the BAL specific exceptions. And same for `_validate_change_lists`. I am not too deep though but if you think they would be useful!

spencer-tb (commented):
Some small comments for now! The validation functions look good to me. Will review the tests tmo :D

src/ethereum_test_types/block_access_list/__init__.py
fselmo: Yeah... some are AssertionError and some are Exception. I definitely like this idea. Do you think we need to be as specific as having `StorageOrder` separated from any other order issues? Or can we just benefit from a more general `BlockAccessListValidationError` with good messaging in each raise?

src/ethereum_test_types/block_access_list/__init__.py
fselmo: Yes! It's certainly more readable as its own type I think. Nice abstraction 👌🏼

src/ethereum_test_types/block_access_list/__init__.py
fselmo: I defined this as a more general `BlockAccessListValidationError`, same with the previous comment. I do think that's a really nice addition but if you see reason to break this up even more, let me know!

PR #2155, opened 2025-09-16
refactor(benchmark): update expected gas usage rule for flexibility

LouisTsai-Csie:
## 🗒️ Description

Currently, we enforce verification of `expected_benchmark_gas_used`, which ensures that no invalid operation terminates execution prematurely. This PR relaxes that rule to give benchmark developers more flexibility.

For example, when benchmarks use while-loops or conditional comparisons, calculating the exact gas usage becomes difficult. In such cases, developers may end up spending more effort on tracking precise gas costs than on the test logic itself. Moreover, not all benchmark cases are intended to fully exhaust the block’s gas like it does in gas-limit-testing.

That said, this change does not mean future benchmarks can skip gas accounting altogether. This exception should only apply when gas usage is inherently hard to predict. By default, reviewers must still verify that `expected_benchmark_gas_used` is correctly set.

## 🔗 Related Issues or PRs

PR #2090 

## ✅ Checklist

    ```console
    ```


spencer-tb (approved):
Works from my side. From the call we had a couple weeks ago on the 1st of September with Guilliaume my expectation was that we'd make this change. LGTM!

cc-ing @marioevz for when he's back - happy to merge nonetheless!

spencer-tb:
Tagged @fselmo for another opinion - if all good I say we should merge.

src/ethereum_test_specs/blockchain.py
```diff
@@ -576,15 +576,14 @@ def generate_block_data(
...
-                "expected_benchmark_gas_used is not set"
-            )
             gas_used = int(transition_tool_output.result.gas_used)
-            assert gas_used == expected_benchmark_gas_used, (
-                f"gas_used ({gas_used}) does not match expected_benchmark_gas_used "
-                f"({expected_benchmark_gas_used})"
-                f", difference: {gas_used - expected_benchmark_gas_used}"
-            )
+
+            if expected_benchmark_gas_used is not None:
```
fselmo, line 581: My only thought is if we should use a different marker other than `None` here as this reads like "expected benchmark gas used is none" i.e. "no gas".

If we had something like `skip_gas_used_validation=True` on the `blockchain_test` itself, and we checked `self.skip_gas_used_validation` in this method instead to skip this validation, that could work well and still keeps this as an explicit effort on the tester to set this.

This feels like a total nit, solely for readability's sake, so feel free to ignore. No blockers from me 🙂

spencer-tb: Yeah I like this - its much safer! :)

LouisTsai-Csie: I really like the idea, i was also exploring how to do something like this, and it is exactly what i wanted, thanks for advice!

PR #2157, opened 2025-09-16
feat(fill, execute): track execution & setup testing phase

LouisTsai-Csie:
## 🗒️ Description

Background for reviewers:

In some benchmark tests, there are not only benchmark transactions but also `setup` transactions that create the scenario for the benchmark target. For example, in the `SSTORE` benchmark we first need a `setup` transaction that deploys as many contracts as possible, and then a separate transaction to perform the storage updates. This feature helps gas-limit testing distinguish between the `setup` and `execution` phases.

This PR extends transaction [metadata](https://eest.ethereum.org/main/running_tests/execute/transaction_metadata/?h=), which labeling phases such as `setup`, `testing`, and `cleanup`. Currently, operations like `deploy_contract` or `fund_eoa` are tagged as `setup` phase, while transactions/blocks included in `state_test` or `blockchain_test` are tagged as testing. However, the current labeling is not always precise enough.

Take `test_worst_blockhash` as an example. The first 256 blocks should be classified as `setup`, while only the final block should count as the testing phase. Without this distinction, we might mistakenly include the setup blocks in benchmark accounting.

```python
def test_worst_blockhash(
    pre: Alloc,
    gas_benchmark_value: int,
):
    """Test running a block with as many blockhash accesses to the oldest allowed block as possible."""
    # Create 256 dummy blocks to fill the blockhash window.
    blocks = [Block()] * 256

    # Always ask for the oldest allowed BLOCKHASH block.
    execution_code = Op.PUSH1(1) + While(
        body=Op.POP(Op.BLOCKHASH(Op.DUP1)),
    )
    execution_code_address = pre.deploy_contract(code=execution_code)
    op_tx = Transaction(
        to=execution_code_address,
        gas_limit=gas_benchmark_value,
        sender=pre.fund_eoa(),
    )
    blocks.append(Block(txs=[op_tx]))

    blockchain_test(
        pre=pre,
        post={},
        blocks=blocks,
    )
```

The design introduces a `test_phase` attribute at both the block and transaction level (currently supporting `execution` and `setup` phases). This attribute is used during execution (see `transaction_post.py` for details), and the metadata is updated accordingly.

## 🔗 Related Issues or PRs

This is the follow-up PR for ethereum/execution-spec-tests#1945, more description could be found in issue ethereum/execution-specs#1588.
Related discussion: [1](https://github.com/ethereum/execution-spec-tests/pull/1945#discussion_r2331768720), [2](https://github.com/ethereum/execution-spec-tests/pull/1945#discussion_r2353702048)

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/tests/benchmark/test_worst_stateful_opcodes.py
+++ b/tests/benchmark/test_worst_stateful_opcodes.py
@@ -422,6 +422,7 @@ def test_worst_storage_access_warm(
     with TestPhaseManager.execution():
         op_tx = Transaction(
             to=contract_address,
+            gas_limit=gas_benchmark_value,
             sender=pre.fund_eoa(),
         )
         blocks.append(Block(txs=[op_tx]))


LouisTsai-Csie:
Note: I have not yet tested against `execute` command. I will provide some result later

src/ethereum_test_types/tests/test_phase_manager.py
```diff
@@ -0,0 +1,240 @@
...
+    manager3 = TestPhaseManager(some_kwarg="value")
+    manager4 = TestPhaseManager("arg1", "arg2", kwarg1="val1", kwarg2="val2")
+
+    # All should have the same initialization
+    for manager in [manager1, manager2, manager3, manager4]:
+        assert len(manager.setup_transactions) == 0
+        assert len(manager.setup_blocks) == 0
+        assert len(manager.execution_transactions) == 0
+        assert len(manager.execution_blocks) == 0
+        assert manager.get_current_phase() == TestPhase.EXECUTION
```
fselmo, line 41: Perhaps we should have a test with two managers that changing the phase of one does not impact the others (no context muddying). This will be the case for sure because the phase is set on the instance, but just to sanity check against [point 2 in this comment](https://github.com/ethereum/execution-spec-tests/pull/1945#discussion_r2353702048) in the old PR in case the phase logic ever changes.

It could be simple... something like entering a `setup` context block with `manager1`, checking that its context changed to `setup`, and making sure `manager2.get_current_phase()` is still `execution` (default).

Thoughts?

src/ethereum_test_specs/blockchain.py
```diff
@@ -232,7 +241,13 @@ class Block(Header):
     expected_post_state: Alloc | None = None
     """Post state for verification after block execution in BlockchainTest"""
     block_access_list: Bytes | None = Field(None)
-    """EIP-7928: Block-level access lists (serialized)."""
+    """
+        EIP-7928: Block-level access lists (serialized).
```
fselmo, line 245: nit: Any reason this was changed? There's an empty space here but we can probably put it back to one line? 👀

fselmo (commented):
Hey @LouisTsai-Csie. I don't have much to add on the implementation. This looks great. I'm going to test this early next week but just had some minor observations at first pass and left as comments.

I wanted to just get a bit of clarity on the scope on how the metadata of each phase will be used. I think we're capturing the data in a good way and I will continue my review next week, but I'd like to understand all the ways we currently plan to _use_ this kind of metadata to get a better understanding for the design.

I think we mentioned being able to call `execute` on a network with a `setup` phase first, and later call it with the `execution` phase. Is there any other case I'm missing so I can understand the end goal on capturing this split? Thanks!

cc: @marioevz

src/ethereum_test_types/tests/test_phase_manager.py
LouisTsai-Csie: This is a good idea for testing phase isolation, i add a new commit (b7faeef) that include the change.

src/ethereum_test_specs/blockchain.py
LouisTsai-Csie: This is changed by accident! I change it back to the origin.

fselmo (approved):
lgtm! @marioevz I'll let you merge this in case there is something from the original issue we should cover here but the logic for splitting is sound here imo 👍🏼

LouisTsai-Csie:
> Apologies for not being more clear. I completely understand the need for splitting these contexts up. It's clear from this PR that the separation of concerns is working well. I wanted a bit more context on how it will be used. Defining and splitting them up is one thing, but then using this in practice is another.

No worries! i found i understand the initial question wrong. And yes i think this could be the future goal, but might need a dedicate PR for this feature!

```uv run execute --phase=setup``` + ```uv run execute --phase=execution```

Thanks for your review and suggestion!

File: src/ethereum_test_specs/blockchain.py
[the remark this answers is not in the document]
Comment (fselmo) [reply src/ethereum_test_specs/blockchain.py:1]: I see yeah, this is the kind of scope I was looking for.

**Comment by jochem-brouwer** [comment]
Hiya, could you show how this `test_worst_blockhash` test would be written using this new metadata? I'd assume we have to decorate the blocks with a marker? It would be great to see how this would look when writing the tests :smile: :+1: 

(If this is already somewhere in this PR comments, please link me then I have "read over" that part. Or a link to the changed code is also fine! :smile: :+1: )

File: src/ethereum_test_types/tests/test_types.py [hunk 3]
```diff
@@ -545,6 +545,7 @@ def test_account_merge(
                 "r": "0xcc61d852649c34cc0b71803115f38036ace257d2914f087bf885e6806a664fbd",
                 "s": "0x2020cb35f5d7731ab540d62614503a7f2344301a86342f67daf011c1341551ff",
                 "sender": "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b",
+                "testPhase": "execution",
```
marioevz: We can do `Field(default_factory=..., exclude=True)` so we don't (1) need to change the unit tests, (2) send it to the transition tool because it does not need it.

marioevz (approved):
LGTM, thanks for making the changes!

I left just a few comments, and the branch needs a rebase, but we should be ok to merge after that.

src/ethereum_test_types/tests/test_phase_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+def test_default_phase_is_none() -> None:
+    """Test that default phase is None (no context set)."""
+    assert TestPhaseManager.get_current_phase() is None
+
+
+def test_transaction_auto_detects_execution_phase() -> None:
+    """Test that transactions default to 'execution' when no phase set."""
+    tx = Transaction(to=Address(0x123), value=100, gas_limit=21000)
+    assert tx.test_phase == "execution"
```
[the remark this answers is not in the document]
LouisTsai-Csie: I remember we discussed making `execution` the default option, should we consider doing that here?
If the test phase isn’t configured to run in `execution` mode by default, we’ll need to explicitly specify it in the test definitions.

I will made this change to the `transaction_post.py`, making `execution` the default option, if this is not what we want i could change again!

```python
phase = (
    "testing"
    if (tx.test_phase == "execution" or tx.test_phase is None)
    else "setup"
)
```

PR #2159, opened 2025-09-16
✨ feat(tests): EIP-7928 SELFDESTRUCT test

raxhvl:
## 🗒️ Description
Test case effects of `SELFDESTRUCT` on EIP-7928.

cc: @nerolation `bal_changes` for the self-destructing contract is captured too. 

## Additional changes

- Validate not just ordering but duplicate entries in the BAL coming from EELS / t8n
- Restrict extra keys in the models used for BALs so we fail on unexpected kwargs used

## ✅ Checklist

    ```console
    ```


tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_self_destruct.py
```diff
@@ -0,0 +1,116 @@
...
+                    if self_destruct_in_same_tx
+                    else [],
+                ),
+            }
+        ),
+    )
+
+    post = {
+        alice: Account(nonce=1),
+        bob: Account(balance=100),
```
fselmo, line 104: `kaboom` always exists, it's the create kaboom that self-destructs. Let's make that clear here.

```suggestion
        bob: Account(balance=100),
        kaboom: Account(balance=0, code=selfdestruct_code),
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_self_destruct.py
```diff
@@ -0,0 +1,116 @@
...
+    # A template, self-destructing contract
+    kaboom = pre.deploy_contract(code=Op.SELFDESTRUCT(bob))
+
+    if self_destruct_in_same_tx:
+        # The goal is to create a self-destructing contract in the same
+        # transaction to trigger deletion of code as per EIP-6780.
+        # The factory contract below clones the template `kaboom`
+        # contract and calls it in this transaction.
+
+        template = pre[kaboom]
```
fselmo, line 54:
```suggestion
```

Let's get rid of this and the assert and use `selfdestruct_code` below.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_self_destruct.py
```diff
@@ -0,0 +1,116 @@
...
+        assert template is not None, "Template contract MUST be deployed for cloning"
+
+        bytecode_size = len(template.code)
```
fselmo, line 57:
```suggestion
        bytecode_size = len(selfdestruct_code)
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_self_destruct.py
```diff
@@ -0,0 +1,116 @@
...
+
+    post = {
+        alice: Account(nonce=1),
+        bob: Account(balance=100),
+    }
+
+    # If the account was NOT self-destructed in the same contract,
+    # we expect the account code to be present and its balance to be 0.
+    if not self_destruct_in_same_tx:
+        post[kaboom] = Account(balance=0, code=pre[kaboom].code)  # type: ignore
```
fselmo, line 110: This will always be true, so we don't need to be selective with this in a conditional. What we should do is update this to both check the factory and the `kaboom_same_tx`:

```py
if self_destruct_in_same_tx:
    post.update(
        {
            factory: Account(
                nonce=2,  # incremented after CREATE
                balance=0,  # spent on CREATE
                code=factory_bytecode,
            ),
            kaboom_same_tx: Account.NONEXISTENT,  # type: ignore
        }
    )
```

fselmo (commented):
Lgtm overall, thanks for getting this in 👌🏼. I left some comments to think about but also, I don't expect selfdestruct tests to grow too large. I don't think these deserve their own file for this reason, please correct me if I'm wrong though. Perhaps we should move these into the positive test cases for now instead of their own file. Thoughts?

fselmo:
@raxhvl I just rebased this with main and fixed an import issue as well. Left all other comments unaddressed so we can discuss them.

raxhvl:
re: creating a separate file for this test

This is purely done to avoid conflicts while I work on these PRs in parallel. Once all PRs are merged, we could group them. Let me know if you want me to move them now.

fselmo:
> This is purely done to avoid conflicts while I work on these PRs in parallel (I have done the same in other PRs too). Once all PRs are merged, we could group them. Let me know if you want me to move them now.

Oh I see. I wouldn't think it's difficult to rebase if these are standalone cases. It would be better to PR them to the appropriate place, but if that's messy I suppose that's fine. Don't let this be a blocker at all.

fselmo:
@raxhvl I merged in Toni's changes and we're pointing to that commit in here now. I also addressed #2197 in the last commit. I'm good to merge this when CI passes. Anything else holding this back?

fselmo:
@raxhvl, @nerolation and I worked on self destruct back and forth some at the end of last week to get the specs ironed out  and I wrote out the specific case of the pre-funded self-destruct account where we want to record the balance change as `0` in the end. I also pointed to the latest commit in the spec for the resolver. Those tests are currently filling here now.

I think this is good to go? Let me know if we're missing anything 👀

raxhvl:
Thanks @fselmo I made the following changes

1.  parameterize `pre_funded` separately from `self_destruct_in_same_tx` to get better test coverage for both cases
2. Extend the self destruct bytecode to include storage operation (read and write)


@nerolation Here is the impact on specs

1. If the account has been prefunded before self destruct tx (sending ether to a create 2 tx before deploying it) then self destruct will include a `balance_change -> 0`


Open questions:
 If a contract reads / write to storage before `SELFDESTRUCT`  would `storage_changes` and `storage_reads` be still be empty?

raxhvl:
@fselmo Think this test is getting too complicate. Do you think for the sake of simplicity we should sepearte the tests / create a new file? Another thing Toni I were discussing is clearing of nonce to zero. Which can be triggered by a contract like this:

```js
pragma solidity ^0.8.0;

contract SelfDestructOrDeploy {

    // Function to check balance and take action
    function checkBalanceAndAct() external payable {
        uint balance = address(this).balance;

        // If the balance is greater than 50 (in wei), self-destruct
        if (balance > 50) {
            selfdestruct(payable(msg.sender));
        }
        // Otherwise, deploy a new contract
        else {
            new NewContract();
        }
    }
}

contract NewContract {
    // Simple contract that logs a message
    string public message = "New contract deployed!";
}

```

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -7,12 +7,18 @@
...
-| `test_bal_2930_slot_listed_but_untouched` | Ensure 2930 access list alone doesn’t appear in BAL        | Include `(KV, S=0x01)` in tx’s EIP-2930 access list; tx executes code that does **no** `SLOAD`/`SSTORE` to `S` (e.g., pure arithmetic/log). | BAL **MUST NOT** contain any entry for `(KV, S)` — neither reads nor writes — because the slot wasn’t touched.                                                                                                    | 🟡 Planned |
+| `test_bal_self_destruct` | Ensure BAL captures storage access and balance changes caused by `SELFDESTRUCT` | Parameterized test: Alice interacts with a contract (either existing or created same-tx) that reads from storage slot 0x01, writes to storage slot 0x02, then executes `SELFDESTRUCT` with Bob as recipient. Contract may be pre-funded with 10 wei | BAL MUST include Alice's nonce change (increment) and Bob's balance change (100 or 110 depending on pre-funding). For the self-destructing contract: storage_reads=[0x01], empty storage_changes=[], and if pre-funded, balance_changes with post_balance=0; if not pre-funded, no balance change recorded. MUST NOT have code_changes or nonce_changes entries | ✅ Completed |
```
spencer-tb, line 10: Does the `either existing` from `(either existing or created same-tx)` refer to the different tx parameterization from the test? If so maybe it should just say same-tx here as the expectations are different. Just nit-picking 😆

Issue #1926, opened 2025-07-18
docs(benchmark): unify benchmark-test documentation and discussion forum


LouisTsai-Csie:
## Background

zkEVM development is accelerating, plus the plan to raise the gas limit, and the updated EIP process now mandates benchmarks for every new opcode/precompile. Multiple teams might have begun writing their own benchmark tests, which might lead to duplicated work and incompatible infrastructure.

## Proposal
Move all benchmark tests, and the tooling that runs them, into EEST. This is already a consensus during Berlin interop.
We need to further add a dedicated “Benchmarks” section to the EEST docs that provides the following features and functionality.

This is my personal idea, any feedback and suggestion is welcome:

**Opcode coverage table**
- Lists every opcode alongside the benchmark tests that exercise it.
- Highlights missing coverage (e.g. PC is still un-benchmarked; We missed it even though we'd created #1453 ).
- Example: selfdestruct - [test_worst_selfdestruct_existing](https://github.com/ethereum/execution-spec-tests/blob/9b30723e8af0f966f4de4e617e7f69b7549b6ccd/tests/benchmark/test_worst_stateful_opcodes.py#L497), [test_worst_selfdestruct_initcode](https://github.com/ethereum/execution-spec-tests/blob/9b30723e8af0f966f4de4e617e7f69b7549b6ccd/tests/benchmark/test_worst_stateful_opcodes.py#L687), [test_worst_selfdestruct_created](https://github.com/ethereum/execution-spec-tests/blob/9b30723e8af0f966f4de4e617e7f69b7549b6ccd/tests/benchmark/test_worst_stateful_opcodes.py#L630C5-L630C36)

**Concise test descriptions**
- Short and clear description, so contributors grasp each test’s intent without reading code

**Results & resources hub**
- Links to the latest benchmark dashboards, raw results.

**Collaboration channel**
- A lightweight forum (issue template or discussion board) where teams can propose new benchmarks, coordinate on shared scenarios, and avoid duplicating effort. As the testing scenario will be more and more complicated, such as the one in #1923 

Implementing this structure will give contributors a clear view of existing coverage, streamline future additions, and keep benchmark infrastructure consistent across the ecosystem.

danceratopz:
Here's a consideration regarding naming from @spencer-tb in https://github.com/ethereum/execution-spec-tests/pull/2169#pullrequestreview-3253912379

[...]

PR #2169, opened 2025-09-17
feat: add bloatnet marker support

LouisTsai-Csie:
### Background

Bloatnet benchmark tests were introduced in PR #2040 and #2090. Because of the nature of bloatnet, using `execute` together with a stub address is a more ideal solution compared to the `fill` and `consume` method.

With the `fill` command, the pre-deployment state of contracts is encoded in the genesis file. However, this becomes problematic when the state configuration in bloatnet is too large. By leveraging `execute`, we can directly send transactions to the network to interact with contracts.

Instead of creating a new folder for bloatnet, the test cases are placed under the existing `benchmark/` folder, with a specific marker to distinguish them.

Following the discussion in issue #1926, the `test_bloatnet.py` file is placed under `benchmark/state/bloatnet/`. All tests that require the combination of `execute` and stub addresses are grouped under the `stateful/` folder. PR #2101 would also benefit from this structure.

Based on this design, I did not add a separate `bloatnet` marker. Instead, I extended the marker to `stateful` marker, which is shared across all stateful tests.

### Verification
- Flag: `-m stateful` runs only the bloatnet-related tests (excluding benchmark tests).
- Flag: `-m benchmark` runs only the benchmark tests (excluding bloatnet tests).

To verify:
1. Add a test in the `test_bloatnet.py` file, then run:
- `uv run fill -v tests/benchmark/stateful/bloatnet/test_bloatnet.py -m benchmark` → should not run
- `uv run fill -v tests/benchmark/stateful/bloatnet/test_bloatnet.py -m stateful` → should run
2. Add the `state` marker to another file under `benchmark/` (e.g. `test_block_full_data`), then run:
- `uv run fill -v tests/benchmark/test_worst_blocks.py::test_block_full_data -m benchmark` → should not run
- `uv run fill -v tests/benchmark/test_worst_blocks.py::test_block_full_data -m stateful` → should run

Documentation:
1. For test cases under state/ folder, the command should include the `-m stateful` flag
2. For test cases not under state/ folder but label with `stateful`, the command include the `-m stateful` flag
3. For the remaining cases, it should include `-m benchmark`

### Notes
I am open to discussion, please let me know if you want to rename the marker, or 

## 🔗 Related Issues or PRs

PR #2040 , #2090 and #2101 

## ✅ Checklist

    ```console
    ```


spencer-tb (commented):
Can we change from `-m state` to `-m stateful` - the same for the folder name: `state/` to `stateful`? Since we have `-m state_test` already!

The logic all makes sense to me. Having 2 conftest's was the better option after reviewing this ;)

danceratopz:
> Thanks a lot for tackling this @LouisTsai-Csie! The PR description is really helpful and documents the expectation very well 🙏
> 
> The benchmark paths aren't correct in the following, but it should explain my suggestion.
> 
> With the additional requirement to "filter" `stateful` tests, I'm wondering if we're adding too much complexity to our pytest setup via markers in order to run the correct subset of tests. The logic alone for `tests/benchmark/conftest.py` is quite complex and then we overlay the logic from `tests/benchmark/stateful/conftest.py` on top of this.
> 
> I won't block this, but I think the requirement is actually quite simple and perhaps we can quickly consider an alternative approach before merging. My proposal would be to remove all of this conftest logic and use an approach that modifies pytest's [`testpaths`](https://docs.pytest.org/en/latest/reference/customize.html#pytest-ini) at the CLI logic stage.
> 
> I.e., adding a `mode` flag, `None` by default (=just fill regular tests), but that can be set to "benchmark" or "stateful".
> 
> ```
> uv run fill --mode=benchmark
> ```
> 
> Then, we add (simpler) logic in [`src/cli/pytest_commands`](https://github.com/ethereum/execution-spec-tests/tree/main/src/cli/pytest_commands) to modify fill and execute's ini config from (and others...):
> 
>     * https://github.com/LouisTsai-Csie/execution-spec-tests/blob/7e4d3e214252fa878a7ac5c381fcab0d48d1c40a/src/cli/pytest_commands/pytest_ini_files/pytest-fill.ini#L5
>       https://github.com/LouisTsai-Csie/execution-spec-tests/blob/7e4d3e214252fa878a7ac5c381fcab0d48d1c40a/src/cli/pytest_commands/pytest_ini_files/pytest-execute.ini#L5
> 
> 
> to overwite `testpaths` with the relevant path (needs a well-organized structure underneath `./tests/benchmarks/`). For this, we don't modify or add new ini files, but rather modify it using this functionality:
> 
> ```
> -o testpaths=tests/benchmark/stateful
> ```
> 
> From `pytest --help`:
> 
> ```
>   -o OVERRIDE_INI, --override-ini=OVERRIDE_INI
>                         Override ini option with "option=value" style, e.g. `-o
>                         xfail_strict=True -o cache_dir=cache`.
> ```
> 
> For normal filling mode (no benchmarking), we can append:
> 
> ```
> uv run pytest --help | grep ignore
>   --ignore=path         Ignore path during collection (multi-allowed)
> ```
> 
> to ignore `./tests/benchmarks` in regular filling.

Hey @LouisTsai-Csie, tried to add a clearer formulation here:
https://github.com/ethereum/execution-spec-tests/issues/2208

As we discussed, happy to merge this as-is if there are other priorities and tackle the simpler solution later.

marioevz (approved):
I think this is great, I agree there's a better approach but I would like for this to get merged so it's not lost in the Weld transition.

marioevz:
> to overwite `testpaths` with the relevant path (needs a well-organized structure underneath `./tests/benchmarks/`). For this, we don't modify or add new ini files, but rather modify it using this functionality:

I'm not super sure if this will seamlessly work when we transition to the EELS repo, I think it would work but I would defer implementing this approach until we move the code.

PR #2178, opened 2025-09-19
✨ feat(tests): EIP-7928 No-op and adjacent test case

raxhvl:
## 🗒️ Description

Adds No-op and no-op adjacent test cases for EIP-7928

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
+++ b/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
@@ -1118,3 +1118,42 @@ def test_bal_aborted_account_access(
         blocks=[block],
         post={},
     )
+
+
+def test_bal_fully_unmutated_account(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+):
+    """
+    Test that BAL captures account that has zero net mutations.
+
+    oracle account:
+        1. Storage read and write the same value (no net change).
+        2. Receives `0` value transfer (no net change).
+    """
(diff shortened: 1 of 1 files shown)


raxhvl:
Depends on https://github.com/ethereum/execution-specs/issues/1432

fselmo (approved):
Hey @raxhvl, I think this is good to go but can use your eyes as a sanity check before merging. Some things to note:

- As we talked about, I updated the empty account expectations to use `BalAccountExpectation.empty()` explicitly (f2c40315582cdb9571ec88595d8c3c5ada0c81f0). We raise if ambiguous `BalAccountExpectation()` is used. I think this explicitness is much nicer for a reader / implementer. I added unit tests for these as well.
- ~~I removed a duplicate test. It looks like the fully unmutated test was exactly the same as the `bal_noop_storage_write` test except for checking the storage read, which I added to the latter and removed the former [here](https://github.com/ethereum/execution-spec-tests/pull/2178/commits/e0428974526799052be3a6563dc3014a27607bae).~~
- I also moved the tests into the main file and added a changelog entry.

Let me know if you think this is good to go as well and we can get this merged 😄. Thanks for your work on all these 👌🏼

fselmo:
> the fully unmutated test also includes a zero value eth transfer along with a storage no op. This means it checks two no-op operations on two account fields simultaneously (balance and storage).

But there was no balance expectation in this test. They _were_ exactly the same if we don't check that. I think this makes sense then. We can put it back but I will add the explicit no-balance-changes check to it as well.

Issue #2143, opened 2025-09-12
chore: find technical solution to transfer eest issues to execution-specs


danceratopz:
We should find a solution pre-Weld to move remaining open issues to execution-specs at-Switch. For example, this could be one of:
1. Transferring the issues via some official Github functionality as suggested by @kclowes.
2. Scripting via the [gh cli](https://cli.github.com/). I.e., grab all open issues from EEST and then open them again in EELS (respecting label, author, ... if possible?).

Transferring sounds better :)

Other comments:
- Labels as unified by #2142 should be transferred with the issue and applied in execution-specs.
- We should link back to the original issue from the new one (and vice versa).

kclowes:
Do we also want to move PRs over, or just issues?

kclowes:
All issues are transferred!

Issue #2200, opened 2025-09-24
chore: create new post-weld labels and tag issues with them


danceratopz:
Requires:
- #2141 

After we have a label concept from #2141 we can:
1. Add the new labels to the EEST and EELS repos
2. Add the new labels appropriately to EEST issues.
3. Remove the old labels from EEST issues. Is this required ?

danceratopz:
A concept is ready for the STEEL team to review and a start has been made on the helper scripts in https://github.com/ethereum/execution-spec-tests/pull/2258.

kclowes:
Labels are added, issues are moved!

PR #2204, opened 2025-09-24
Script to move EEST issues to EELS

kclowes:
We shouldn't need to merge this one, but would love any feedback around testing it out. It's pretty sloppy, but I'm less concerned about style and more concerned about the functionality since it's not something we'll maintain. This script will move issues from one repo to another and close the issues that get moved. 

There are flags to do a `--dry-run` and `--limit` so we can test it out ahead of time on the actual repos. 

This will transfer labels if they exist on both repos and can translate one label to another. Also transfers milestones if they exist on both repos. Assignees will transfer, but I'm not sure if they have to have maintainer status in order for them to transfer. See  #2143 for more functionality details around sub-issues.

As seen in the script, I made up a couple test repos to test things out. Feel free to play around with those if you'd like to test!

cc @danceratopz

### DIFF
--- a/scripts/transfer_issues.py
+++ b/scripts/transfer_issues.py
@@ -1,0 +1,461 @@
+#!/usr/bin/env python3
+"""Transfer GitHub issues from one repository to another using the GitHub CLI."""
+
+import argparse
+import json
+import re
+import subprocess
+import sys
+from typing import Dict, List, Optional, Tuple
+
+
+# Configuration
+SOURCE_REPO = "kclowes/transfer-issues-from"
+TARGET_REPO = "kclowes/transfer-issues-to"
+
+# Label translation mapping
+LABEL_MAP = {
+    "custom": "to-custom",
+}
+
+
+def run_command(cmd: List[str], capture_output: bool = True) -> Tuple[int, str, str]:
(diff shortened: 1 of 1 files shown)


scripts/transfer_issues.py
```diff
@@ -0,0 +1,461 @@
...
+
+        # Translate labels
+        if labels:
+            translated_labels = translate_labels(labels)
+            if translated_labels != labels:
+                print(
+                    f"  Translating labels: {', '.join(labels)} -> {', '.join(translated_labels)}"
+                )
+                labels = translated_labels
+
```
danceratopz, line 312: From the description:
> This will transfer labels if they exist on both repos and can translate one label to another. 

I think we should add these labels ahead of time (also to help us to agree on our label set), this is:
- https://github.com/ethereum/execution-spec-tests/issues/2200

My suggestion would be to error out (or verbosely skip) if the label doesn't exist in the target repo and abort the transfer. If successfully transferred issues get closed, it should be trivial to run the script again without any side effects.

danceratopz (commented):
Thanks a lot for getting on this @kclowes, especially the testing!

Small comment to avoid having to any manual patching up if labels don't exist in the target repo.

PR #2243, opened 2025-10-01
feat(consume): `consume direct` using evmone-state and blockchaintest

pdobacz:
## 🗒️ Description

First step towards supporting `consume direct` with `evmone`.

Remarks:
1) Starting off with just `state_test` support to get the ball rolling
2) `evmone` has two things which make things harder. One is that it has two separate binaries: `evmone-statetest` and `evmone-blockchaintest`, rather than a single one with 2 subcommands, as EEST assumes. Decision we need to make where do we adjust - on EEST end or on `evmone` end. Gut tells me it is more elegant to adjust `evmone`, but overall easier to do so in EEST (assuming we can cut a corner and let EEST choose state/blockchain binary on the fly as it picks up fixtures). EDIT: using evmone-blockchaintest is handled in this PR :point_down: 
3) Other is that `evmone` tests on the granularity level of a single `json` file. This means a failed test case fails the entire file, making it a bit harder to figure out where the failure is (failure message provides a good hint, but is pretty messy). Also test filtering is very coarse. Something we need to tackle on `evmone` end as we progress.

Opening as draft - let me know if OK to proceed with such first step into `main`.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -23,6 +23,7 @@ Test fixtures for use by clients are available for each release on the [Github r
 
 - ✨ Add retry logic to RPC requests to fix flaky connection issues in Hive ([#2205](https://github.com/ethereum/execution-spec-tests/pull/2205)).
 - 🛠️ Mark `consume sync` tests as `flaky` with 3 retires due to client sync inconsistencies ([#2252](https://github.com/ethereum/execution-spec-tests/pull/2252)).
+- ✨ Add `consume direct` using `evmone-statetest` and `evmone-blockchaintest` ([#2243](https://github.com/ethereum/execution-spec-tests/pull/2243)).
 
 ### 📋 Misc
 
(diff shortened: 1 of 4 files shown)


marioevz:
> This allows us to define two FixtureConsumer classes for evmone:
> 
>     * `EvmOneStateFixtureConsumer`,
> 
>     * `EvmOneBlockchainFixtureConsumer`.

I agree on this point, I think this should be the way to go IMO.

If we implement that, and then doing `--bin=evmone-statetest --bin=evmone-blockchaintest` still results in fails for BlockchainTest-formatted tests, then there's something we need to fix in EEST.

Thanks for implementing this @pdobacz !

pdobacz:
blockchaintest done with the caveat as mentioned above, ready to review.

marioevz (approved):
LGTM, amazing work, thanks!

PR #2244, opened 2025-10-01
chore(cli): add mapping for TYPE_3_TX_CONTRACT_CREATION error

SDartayet:
## 🗒️ Description
This PR adds a mapping for the TYPE_3_TX_CONTRACT_CREATION for Ethrex. This is necessary to pass the recently added Hive tests that check whether the client rejects a type 3 transaction with an empty "to" field. While we do reject such blocks, the exception mapper maps them to a TYPE_4_TX_CONTRACT_CREATION, so this adds the appropriate mapping for these test cases.


marioevz (approved):
LGTM, thanks!

PR #2256, opened 2025-10-03
feat(benchmark): add SLOAD benchmark test with multi-contract support

CPerezz:
## 🗒️ Description
Add test_sload_empty_erc20_balanceof to benchmark SLOAD operations on non-existing storage slots using ERC20 balanceOf() queries.

The idea of this benchmark is to exploit within a single or series of N contracts calls to non-existing addresses. On this way, we force clients to resolve as many tree branches as possible.

## ✅ Checklist

    ```console
    ```
- [x] Tests: For PRs implementing a missed test case, update the [post-mortem document](/ethereum/execution-spec-tests/blob/main/docs/writing_tests/post_mortems.md) to add an entry the list.
- [x] Ported Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) or [tests/static](/ethereum/execution-spec-tests/blob/main/tests/static) have been assigned `@ported_from` marker.

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -1,0 +1,177 @@
+"""
+abstract: BloatNet single-opcode benchmark cases for state-related operations.
+
+   These tests focus on individual EVM opcodes (SLOAD, SSTORE) to measure
+   their performance when accessing many storage slots across pre-deployed
+   contracts. Unlike multi-opcode tests, these isolate single operations
+   to benchmark specific state-handling bottlenecks.
+"""
+
+import pytest
+
+from ethereum_test_forks import Fork
+from ethereum_test_tools import (
+    Account,
+    Alloc,
+    Block,
+    BlockchainTestFiller,
+    Transaction,
+    While,
+)
+from ethereum_test_vm import Bytecode
+from ethereum_test_vm import Opcodes as Op
+
+REFERENCE_SPEC_GIT_PATH = "DUMMY/bloatnet.md"
+REFERENCE_SPEC_VERSION = "1.0"
+
+# ERC20 function selectors
+BALANCEOF_SELECTOR = 0x70A08231  # balanceOf(address)
+APPROVE_SELECTOR = 0x095EA7B3  # approve(address,uint256)
+ALLOWANCE_SELECTOR = 0xDD62ED3E  # allowance(address,address)
+
+
+# SLOAD BENCHMARK ARCHITECTURE:
+#
+#   [Pre-deployed ERC20 Contract] ──── Storage slots for balances
+#           │
+#           │  balanceOf(address) → SLOAD(keccak256(address || slot))
+#           │
+#   [Attack Contract] ──CALL──► ERC20.balanceOf(random_address)
+#           │
+#           └─► Loop(i=0 to N):
+#                 1. Generate random address from counter
+#                 2. CALL balanceOf(random_address) → forces cold SLOAD
+#                 3. Most addresses have zero balance → empty storage slots
+#
+# WHY IT STRESSES CLIENTS:
+#   - Each balanceOf() call forces a cold SLOAD on a likely-empty slot
+#   - Storage slot = keccak256(address || balances_slot)
+#   - Random addresses ensure maximum cache misses
+#   - Tests client's sparse storage handling efficiency
+
+
+@pytest.mark.valid_from("Prague")
+def test_sload_empty_erc20_balanceof(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+    address_stubs,
+):
+    """
+    BloatNet SLOAD benchmark using ERC20 balanceOf queries on random addresses.
+
+    This test:
+    1. Auto-discovers ERC20 contracts from stubs (pattern: erc20_contract_*)
+    2. Splits gas budget evenly across all discovered contracts
+    3. Queries balanceOf() incrementally starting by 0 and increasing by 1.
+    (thus forcing SLOADs to non-existing addresses.)
+    """
+    gas_costs = fork.gas_costs()
+
+    # Calculate gas costs
+    intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(calldata=b"")
+
+    # Cost per iteration (attack contract overhead + balanceOf call)
+    cost_per_iteration = (
+        # Attack contract loop overhead
+        gas_costs.G_VERY_LOW * 2  # MLOAD counter (3*2)
+        + gas_costs.G_VERY_LOW * 2  # MSTORE selector (3*2)
+        + gas_costs.G_VERY_LOW * 3  # MLOAD + MSTORE address (3*3)
+        + gas_costs.G_BASE  # POP (2)
+        + gas_costs.G_BASE * 3  # SUB + MLOAD + MSTORE for counter decrement (2*3)
+        + gas_costs.G_BASE * 2  # ISZERO * 2 for loop condition (2*2)
+        + gas_costs.G_MID  # JUMPI (8)
+        # CALL to ERC20 contract
+        + gas_costs.G_WARM_ACCOUNT_ACCESS  # Warm CALL to same contract (100)
+        # Inside ERC20 balanceOf
+        + gas_costs.G_VERY_LOW  # PUSH4 selector (3)
+        + gas_costs.G_BASE  # EQ selector match (2)
+        + gas_costs.G_MID  # JUMPI to function (8)
+        + gas_costs.G_JUMPDEST  # JUMPDEST at function start (1)
+        + gas_costs.G_VERY_LOW * 2  # CALLDATALOAD arg (3*2)
+        + gas_costs.G_KECCAK_256  # keccak256 static (30)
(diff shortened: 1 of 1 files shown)


tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -0,0 +1,177 @@
...
+#   - Tests client's sparse storage handling efficiency
+
+
+@pytest.mark.valid_from("Prague")
+def test_sload_empty_erc20_balanceof(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+    address_stubs,
```
marioevz, line 59: I didn't quite expect the fixture to be directly used by the test tbh, it's an interesting workaround!

A couple things:
- It will produce unexpected behavior to someone running `execute` and not knowing the inner workings of this tests because it will try to use all the stubs, including those that are meant to be used by other tests.
- The test will change its behavior depending on the stubs passed to the parameter in `execute`, which is not inherently a bad thing, but I think we should really think about it.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -0,0 +1,177 @@
...
+#           │
+#   [Attack Contract] ──CALL──► ERC20.balanceOf(random_address)
+#           │
+#           └─► Loop(i=0 to N):
+#                 1. Generate random address from counter
+#                 2. CALL balanceOf(random_address) → forces cold SLOAD
+#                 3. Most addresses have zero balance → empty storage slots
+#
+# WHY IT STRESSES CLIENTS:
+#   - Each balanceOf() call forces a cold SLOAD on a likely-empty slot
```
marioevz, line 47: How different would this be from a benchmark test that optimizes for `SLOAD`s instead of trying to mimic the `balanceOf()` behavior?
The things that come to mind are the extra jumps, keccak operations, and the fact that `balanceOf()` (iirc) only does a single `SLOAD` per subcall.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -0,0 +1,177 @@
...
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+    address_stubs,
+):
+    """
+    BloatNet SLOAD benchmark using ERC20 balanceOf queries on random addresses.
+
+    This test:
+    1. Auto-discovers ERC20 contracts from stubs (pattern: erc20_contract_*)
```
marioevz, line 65: Ah, this answers my question on the first comment, I think it's ok then.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -0,0 +1,177 @@
...
+        + gas_costs.G_JUMPDEST  # JUMPDEST at function start (1)
+        + gas_costs.G_VERY_LOW * 2  # CALLDATALOAD arg (3*2)
+        + gas_costs.G_KECCAK_256  # keccak256 static (30)
+        + gas_costs.G_KECCAK_256_WORD * 2  # keccak256 dynamic for 64 bytes (2*6)
+        + gas_costs.G_COLD_SLOAD  # Cold SLOAD (2100)
+        + gas_costs.G_VERY_LOW * 3  # MSTORE result + RETURN setup (3*3)
+        # RETURN costs 0 gas
+    )
+
+    num_contracts = len(address_stubs.root)
```
marioevz, line 100: I think we have to do the pattern (`erc20_contract_*`) discrimination here to have a proper number.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -0,0 +1,177 @@
...
+    # Calculate gas budget per contract and calls per contract
+    available_gas = gas_benchmark_value - intrinsic_gas
+    gas_per_contract = available_gas // num_contracts
+    calls_per_contract = int(gas_per_contract // cost_per_iteration)
+
+    # Deploy all discovered ERC20 contracts using stubs
+    # In execute mode: stubs point to already-deployed contracts on chain
+    # In fill mode: empty bytecode is deployed as placeholder
+    erc20_addresses = []
+    for stub_name in address_stubs.root:
```
marioevz, line 111: This loop also needs to discriminate using the pattern.

marioevz (commented):
Did a quick pass and it looks good to me overall.

I left a couple of questions as comments. Thanks!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
CPerezz: Yes. I basically did that because a lot of the tests we will do will just be the same code, but changing the contract against which we run it.
Thus I thought it would not make sense to duplicate code all the time. And instead, I should just reuse the same code for any number of contracts that share interface.

So it's the stubs what actually determines how the contract behaves.

I spoke with @kamilchodola about this and he's also not sure if this way will be the best for his tool. Nevertheless, IMO it's the best in regards code to maintain and simplicity.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
CPerezz: The problem we have here is the following:
1. The attack wants to stress calling big contracts and doing path resolution on them (at least usually). 
2. The biggest contracts (that share interface) are ERC20. And `balanceOf` doesn't really add a ton of overhead.
3. If we had to deploy contracts that allow us to abuse SLOAD, we would need a ton of time to bloat lots of contracts with the same interface and make them 5-20 GB of storage each.
4. Thus, for this iteration, it just seems significantly easier to go this route.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
CPerezz: I have the same question here. Can you elaborate on this a bit?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -0,0 +1,334 @@
...
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+    address_stubs,
+):
+    """
+    BloatNet SSTORE benchmark using ERC20 approve to write to storage.
+
+    This test:
+    1. Auto-discovers ERC20 contracts from stubs (pattern: erc20_contract_*)
```
CPerezz, line 192: I think we might not need the pattern at all.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: Do you mean that, in your benchmarking process, the pre-deployed contracts already contain randomized storage values (such as balances or approvals), and you’re benchmarking the `SLOAD` operation based on that?

I’ve read the recent state analysis report, which shows that `USDT` is one of the largest contracts in terms of storage state. In your case, would this be similar to benchmarking state operations using such a large contract as a reference?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -0,0 +1,334 @@
...
+                    # Store address at memory[64] (just use counter as address)
+                    + Op.MSTORE(offset=64, value=Op.MLOAD(0))
+                    # Call balanceOf(address) on ERC20 contract
+                    + Op.CALL(
+                        address=erc20_address,
+                        value=0,
+                        args_offset=32,
+                        args_size=36,
+                        ret_offset=96,
+                        ret_size=32,
```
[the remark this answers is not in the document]
CPerezz: No. That's true. But at the same time, we gain nothing out of changing it to 0,0. Very very small savings in mem-expansion. I can def change it if you prefer.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -0,0 +1,334 @@
...
+        attack_code += (
+            # Initialize counter in memory[0] = number of calls
+            Op.MSTORE(offset=0, value=calls_per_contract)
+            # Loop for this specific contract
+            + While(
+                condition=Op.MLOAD(0) + Op.ISZERO + Op.ISZERO,  # Continue while counter > 0
+                body=(
+                    # Use counter directly as address (cheapest option)
+                    # Store function selector at memory[32]
+                    Op.MSTORE(offset=32, value=BALANCEOF_SELECTOR)
```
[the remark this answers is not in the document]
CPerezz: That's an awesome suggestion! Thanks!

LouisTsai-Csie (changes requested):
@CPerezz , I’ve left some suggestions. Please take a look and let me know if they’re unclear or not practical. These changes might not reduce much gas usage, but i wonder if they could help simplify the layout a bit.

I've not yet reviewed `test_multi_opcode.py`, but i believe it would be quick if we have consensus on the other test cases!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
marioevz: For example, if we run execute for all tests, and therefore pass stub contracts that are needed for other tests (like `xen_contract` for example), these other contracts are going to be included in `address_stubs` unconditionally, and we are going to try to send a `balanceOf` to those other contracts that are not ERC20 contracts.

PR #2266, opened 2025-10-04
fix(tests): fix storage slot of actual value for `eip7516_blobgasfee` tests in `execute remote`

### LINKED ISSUE #2255: Using `uv run execute remote` test `eip7516_blobgasfee` for `no_stack_overflow|enough_gas` fail
The following `eip7516_blobgasfee` test using `execute remote` fails with `Storage value` error

```console
$ uv run execute remote $RPC_ARGS tests/cancun/eip7516_blobgasfee -k no_stack_overflow
[...]
FAILED tests/cancun/eip7516_blobgasfee/test_blobgasfee_opcode.py::test_blobbasefee_stack_overflow
[fork_Cancun-state_test-no_stack_overflow] - AssertionError: 
Storage value at 0x01 of 0x53b4811d15b8cb1caad9e652d64d412ee66e7971 is
0x0000000000000000000000000000000000000000000000000000000000000000,expected 0x01.
```

_(the same applies for test with id `enough_gas`)_

The problem seems to be that the result of the `CALL` is stored in `NUMBER`

https://github.com/ethereum/execution-spec-tests/blob/476f3c3e437e6a8048e47786389d55e8c0bf6dcc/tests/cancun/eip7516_blobgasfee/test_blobgasfee_opcode.py#L49-L55

while the storage slot used to check the expected value is `1`

https://github.com/ethereum/execution-spec-tests/blob/476f3c3e437e6a8048e47786389d55e8c0bf6dcc/tests/cancun/eip7516_blobgasfee/test_blobgasfee_opcode.py#L110-L112

Replacing `NUMBER` for `1` in `caller_code` makes the test pass.

acuarica:
## 🗒️ Description

This PR fixes the storage slot of the actual value in the test.

In `execute remote` mode, it used `NUMBER` as the storage slot, but the expected value is stored in slot `1`. This causes problems when running tests with `execute remote` where `NUMBER` cannot be ensured to be `1`.

## 🔗 Related Issues or PRs

Fixes #2255.

## ✅ Checklist

    ```console
    ```


LouisTsai-Csie:
This fix resolves the issue in `execute` mode, but it will fail under the `fill` mode in our framework. Please take a look at the failing case `test_blobbasefee_during_fork`, where the block number is used as the storage key.

```python
 pre-set storage just to make sure we detect the change
        code_caller_post_storage[block_number] = 0 if timestamp < 15_000 else 1
```

That is the reason for the failing CI, you could also run it locally via
```python
uv run fill -v tests/cancun/eip7516_blobgasfee/test_blobgasfee_opcode.py::test_blobbasefee_during_fork --fork Osaka --clean
```

LouisTsai-Csie:
Gas usage validation is intended for benchmark mode (see [docs](https://eest.ethereum.org/main/writing_tests/benchmarks/#expected-gas-usage)).
You can temporarily use `skip_gas_used_validation` to bypass it, we’ll later limit this check to benchmark mode after this wip [issue](https://github.com/ethereum/execution-spec-tests/issues/2208) is done

tests/cancun/eip7516_blobgasfee/test_blobgasfee_opcode.py
```diff
@@ -46,13 +46,20 @@ def callee_address(pre: Alloc, callee_code: Bytecode) -> Address:
     return pre.deploy_contract(callee_code)
 
 
+@pytest.fixture
+def expected_value_slot() -> Bytecode:
+    """Storage slot where to save the expected value, by default 1."""
```
LouisTsai-Csie, line 51: Could we rename the variable to something like `store_slot` or `result_slot`? I find `expected_value_slot` a bit confusing in this context.

LouisTsai-Csie:
For the failing evmone CI issue, I checked the report and found an `LBC`. The root cause is the use of a different storage key. In the previous setup, `Op.NUMBER` was used to manage the storage slot, whereas the latest approach instead uses `slot 1`. As a result, evmone detects the discrepancy and reports it as an `LBC`.

Report data
Screenshot 2025-10-06 at 4 10 24 PM

LBC Found
Screenshot 2025-10-06 at 4 11 06 PM

LouisTsai-Csie:
A recent merged [PR](https://github.com/ethereum/execution-spec-tests/pull/2269) might help you with the gas validation issue. For non benchmark mode, it would not apply the gas validation

acuarica:
Thanks for clarifying https://github.com/ethereum/execution-spec-tests/pull/2266#issuecomment-3370418432, and in general for taking the time to review this.

tests/cancun/eip7516_blobgasfee/test_blobgasfee_opcode.py
```diff
@@ -46,13 +46,20 @@ def callee_address(pre: Alloc, callee_code: Bytecode) -> Address:
...
+
 @pytest.fixture
 def caller_code(
     call_gas: int,
     callee_address: Address,
+    store_slot: Bytecode,
 ) -> Bytecode:
     """Bytecode used to call the bytecode containing the BLOBBASEFEE opcode."""
-    return Op.SSTORE(Op.NUMBER, Op.CALL(gas=call_gas, address=callee_address))
+    return Op.SSTORE(store_slot, Op.CALL(gas=call_gas, address=callee_address))
```
marioevz, line 62: One trick could be to use `Op.SELFBALANCE` instead of `Op.NUMBER`, and then add `value=1` (1 wei) to the transaction, so that:
- Each transaction increases the balance of the contract by 1.
- When the contract uses its own balance as the key, each transaction it will save to the next key incrementally because of the 1 wei included with the transaction.
- This is compatible with fill and execute because we no longer rely on the block number, and the contract balance always starts at zero on both instances.

marioevz (commented):
Thanks for this! Just one suggestion that might simplify and fix the issues found in the previous review.

tests/cancun/eip7516_blobgasfee/test_blobgasfee_opcode.py
acuarica: Thanks for the suggestion. I've applied it and it works for both `fill` and `execute`.

However, AFAIU CI is still failing in the same spot as https://github.com/ethereum/execution-spec-tests/pull/2266#issuecomment-3370418432.

PR #2272, opened 2025-10-06
chore(consume): fix consume sync logger errors

spencer-tb:
## 🗒️ Description

Consume sync can fail with `TypeError: 'NoneType' object is not subscriptable` when attempting to verify block state roots immediately after the sync completes. The `get_block_by_hash()` call returns `None` because the block isn't yet queryable, even though the forkchoice update returned `VALID`.

### Solution To Try
Add explicit `None` checks for retrieved blocks in the verification retry loop.

## 🔗 Related Issues or PRs

Follow up from: #2252 

## ✅ Checklist

    ```console
    ```

### DIFF
--- a/src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
+++ b/src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
@@ -419,10 +419,16 @@ def test_blockchain_via_sync(
         assert eth_rpc is not None, "eth_rpc is required"
         assert sync_eth_rpc is not None, "sync_eth_rpc is required"
 
-        for attempt in range(3):
+        for attempt in range(5):
             try:
                 sync_block = sync_eth_rpc.get_block_by_hash(last_valid_block_hash)
                 client_block = eth_rpc.get_block_by_hash(last_valid_block_hash)
+
+                if sync_block is None or client_block is None:
+                    raise LoggedError(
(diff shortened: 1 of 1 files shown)


fselmo:
How does this break the `try / except` though? This is part of the reason I added the `try / except`. If this happens we should hit the `except` block, sleep, and try again. Are you saying you are getting this `None` after 3 attempts still?

marioevz (approved):
LGTM, thanks!

src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
```diff
@@ -424,6 +424,14 @@ def test_blockchain_via_sync(
                 sync_block = sync_eth_rpc.get_block_by_hash(last_valid_block_hash)
                 client_block = eth_rpc.get_block_by_hash(last_valid_block_hash)
 
+                if sync_block is None or client_block is None:
+                    if attempt < 2:
```
fselmo, line 428: I think this `continue` here is the only thing that is repetitive since if we just go on to raise `LoggedError` below, we already account for the attempts + continue in the `except` block.

spencer-tb: Bumped! Thanks for catching this. Let me know what you think. I also changed the attempts from 3 to 5 for good luck 🙏

